Issue #2, opened 2023-03-07
Remove OCI backend dependency from Controlplane


migmartri:
Currently, the controlplane has a dependency on a OCI backend provider so it can upload/download the generated attestation as well as downloading SBOMs for re-routing to Dependency-Track.

We have a component that it's already handling OCI upload/downloads which is the Artifact CAS, and this is leveraged by the CLI already. 

We should delegate OCI access from the control plane to the artifact CAS and leverage its API instead this means.

- [x] Extract the CAS client code that it's currently in the CLI and move it to a top level package https://github.com/chainloop-dev/chainloop/pull/43
- [x] Implement CAS upload use-case for attestation https://github.com/chainloop-dev/chainloop/pull/45
- [x] Implement CAS download https://github.com/chainloop-dev/chainloop/pull/46
- [x] Download attestation using CAS https://github.com/chainloop-dev/chainloop/pull/46
- [ ] Download SBOMs using CAS
- [ ] Remove credentials.Reader and backend provider use from the control plane

PR #43, opened 2023-03-20
chore: decouple cas-client file upload

migmartri:
The CAS client will be used by the control-plane too and the upload input in such case is an io.Reader not a file.

This patch updates the Uploader interface and implementation accordingly. It also keeps the `UploadFile` method encapsulated in the casclient package since it's being used by two callers, the CLI action package and the attestation materials package.

Refs #2

### DIFF
--- a/internal/casclient/casclient.go
+++ b/internal/casclient/casclient.go
@@ -24,8 +24,8 @@ import (
 )
 
 type UpDownStatus struct {
-	Filepath, Filename, Digest string
-	ProcessedBytes             int64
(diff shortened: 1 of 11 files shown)


internal/casclient/casclient.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 type UpDownStatus struct {
-	Filepath, Filename, Digest string
```
migmartri, line 27: `Filepath` was not actually used and added an unnecessary dependency to a local file

PR #55, opened 2023-03-24
feat: CAS readiness probe support

migmartri:
Update the Control plane readiness probe API endpoint to take into account that the CAS configuration is correct and can in fact reach the CAS gRPC API server.

In other words, now it checks that it can reach the `statusz` endpoint of the configured CAS dependency.

Refs #2

### DIFF
--- a/app/artifact-cas/internal/server/grpc.go
+++ b/app/artifact-cas/internal/server/grpc.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
 
 	v1 "github.com/chainloop-dev/chainloop/app/artifact-cas/api/cas/v1"
 	"github.com/chainloop-dev/chainloop/app/artifact-cas/internal/conf"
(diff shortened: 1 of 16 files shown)


app/artifact-cas/internal/server/grpc.go
```diff
@@ -62,11 +64,14 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
...
 			logging.Server(logger),
 			// NOTE: JWT middleware only works for unary requests
-			// below you can see a reimplementation of the middleware as a stream interceptor
-			jwtMiddleware.Server(
-				loadPublicKey(rawKey),
-				jwtMiddleware.WithSigningMethod(casJWT.SigningMethod),
-				jwtMiddleware.WithClaims(func() jwt.Claims { return &casJWT.Claims{} })),
+			// below you can see a re-implementation of the middleware as a stream interceptor
+			// If we require a logged in user we
+			selector.Server(
```
migmartri, line 69: allow status endpoint without authentication

Issue #108, opened 2023-05-08
CAS: Add support for GCP Secret Manager


gr0:
Right now CAS provides support for AWS secret manager and Hashicorp Vault. It would be awesome to have support for GCP Secret Manager.

migmartri:
hi @gr0 thanks for filling the issue.

I agree, this could be a great addition specially for those running Chainloop in GKE. 

In case it helps. There is some sort of encapsulation already defined to achieve the addition of alternative backends, which long story short should be implementing this interface 
https://github.com/chainloop-dev/chainloop/blob/90c67652f78b231f0f472c01df1626cdb783a27b/internal/credentials/credentials.go#L32-L44

In addition to adding the backend to the `credentials` package, you'll need to a) add support to configuring (providing GCP auth) both at the ControlPlane and CAS i.e https://github.com/chainloop-dev/chainloop/blob/90c67652f78b231f0f472c01df1626cdb783a27b/app/controlplane/internal/conf/conf.proto#L94-L98

and b) bootstrap the new backend if needed as seen here https://github.com/chainloop-dev/chainloop/blob/90c67652f78b231f0f472c01df1626cdb783a27b/app/controlplane/cmd/main.go#L160 

Let me know if this is smth you'd like to work on and I'll be happy to assign the issue to you :)

Thank you!

migmartri:
I am reopening this issue to tackle the deployment part, that way the feature can be used end to end.

https://github.com/chainloop-dev/chainloop/issues/134

PR #130, opened 2023-05-26
chore: remove prefix from CAS

migmartri:
This is a follow up of #108 with some minor changes.

* `secret_prefix` is not required on the CAS since it's only used for writes (control plane).
* Adapted the config sample files indentation and added also an example in the control plane.
* The rest is just indentation that was in general incorrect in those files. [Buf style guide](https://buf.build/docs/format/style#indentation) recommends 2 spaces. I am not sure why `buf lint` doesn't complain about it.

cc/ @gr0 (I am figuring out how I can add you as reviewer :)


danlishka (approved):

gr0 (approved):
The changes look good to me, thanks @migmartri for handling that and sorry I missed that.

PR #133, opened 2023-05-27
refactor(config): extract secret manager configuration

migmartri:
Both the control plane and the CAS use the same way to configure and initialize the `secrets` backend. This code extracts this to the parent module, specifically:

- Move initialization code and proto declaration to parent credentials module
- Adds declarative validations to the config.proto file and runs validations, see attached tests to learn more

Note that this does not have any impact in the external UX of an user.

cc/ @gr0 

Refs #108


internal/credentials/api/credentials/v1/config.go
```diff
@@ -0,0 +1,106 @@
...
+	"fmt"
+
+	"github.com/chainloop-dev/chainloop/internal/credentials"
+	"github.com/chainloop-dev/chainloop/internal/credentials/aws"
+	"github.com/chainloop-dev/chainloop/internal/credentials/gcp"
+	"github.com/chainloop-dev/chainloop/internal/credentials/vault"
+	"github.com/go-kratos/kratos/v2/log"
+)
+
+func NewFromConfig(conf *Credentials, l log.Logger) (credentials.ReaderWriter, error) {
```
migmartri, line 29: Add tests

internal/credentials/api/credentials/v1/config.proto
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package credentials.v1;
+
+option go_package = "github.com/chainloop-dev/chainloop/internal/credentials/api/credentials/v1;v1";
+
+// Where the credentials to access the backends are stored
+message Credentials {
```
migmartri, line 23: add validations

gr0 (approved):
👍 I like the simplification and putting the code responsible for initialization in a single place.

migmartri:
Thanks for the review!

danlishka (approved):

PR #135, opened 2023-05-30
feat: Add support for JUnit XML material type (#120)

### LINKED ISSUE #120: JUnit XML material type
It would be useful to add JUnit XML report format to Chainloop as first class citizen material type.

This will allow developer teams to attest their test results as part of their attestation as well as operators to enable custom destination endpoints and analysis integrations in the future.

https://www.ibm.com/docs/en/developer-for-zos/14.1?topic=formats-junit-xml-format 

This requires

- [x] Addition of new material type in the proto definitions for the crafting tool
- [x] Add handling (upload if applicable + validation) in CLI tool for this new kind of material
- [x] Document it in reference docs https://docs.chainloop.dev/reference/operator/contract#material-schema

danlishka:
This PR aims to add support for a new material type: JUnit XML.

We know that more types will be added in the future, so I propose to extract the upload&craft code and move it to a new reusable private function uploadAndCraft. All similar materials, such as SBOMs (SPDX, CycloneDX), Artifact, and more, may use this function instead of duplicating the code in the Craft method.

Documentation is updated in this PR https://github.com/chainloop-dev/docs/pull/126

closes #120


internal/attestation/crafter/materials/materials.go
```diff
@@ -34,6 +36,32 @@ type crafterCommon struct {
 	input  *schemaapi.CraftingSchema_Material
 }
 
+type crafterUploader struct {
+	*crafterCommon
+	uploader casclient.Uploader
+}
+
+// Craft will calculate the digest of the artifact, simulate an upload and return the material definition
```
migmartri, line 44: "simulate the upload" simulate or actually uploading it?

internal/attestation/crafter/materials/materials.go
```diff
@@ -34,6 +36,32 @@ type crafterCommon struct {
...
+func (i *crafterUploader) Craft(ctx context.Context, artifactPath string) (*api.Attestation_Material, error) {
```
migmartri, line 45: probably to make it clearer I'd not 1) expose this function (a.k.a downcase it) and 2) call it something else, there is no need to be called craft right, maybe upload?

migmartri (commented):
Cool, good start

internal/attestation/crafter/materials/materials.go
danlishka: I would use craft, we upload there and we return attestation material there.

internal/attestation/crafter/materials/junit_xml.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := xml.Unmarshal(bytes, &junit.Suite{}); err != nil {
+		return nil, fmt.Errorf("invalid JUnit XML file: %w", ErrInvalidMaterialType)
+	}
+	_, err = junit.IngestReader(f)
+	if err != nil {
+		i.logger.Debug().Err(err).Msg("error decoding file: " + filePath)
+		return nil, fmt.Errorf("invalid JUnit XML file: %w", ErrInvalidMaterialType)
+	}
+
+	return i.crafterUploader.craft(ctx, filePath)
```
danlishka, line 73: @migmartri should I also update artifact, spdx and cyclonedx crafters to use this approach with crafterUploader?

internal/attestation/crafter/materials/junit_xml.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, fmt.Errorf("material type is not JUnit XML")
+	}
+	craftCommon := &crafterCommon{logger: l, input: schema}
+	return &JUnitXMLCrafter{uploader: uploader, crafterCommon: craftCommon}, nil
+}
+
+func (i *JUnitXMLCrafter) Craft(ctx context.Context, filePath string) (*api.Attestation_Material, error) {
+	if err := i.validate(ctx, filePath); err != nil {
+		return nil, err
+	}
```
migmartri, line 49: nitpick. I tend to add extra lines between code blocks, i.e after the closing `}`

internal/attestation/crafter/materials/junit_xml_test.go
```diff
@@ -0,0 +1,135 @@
...
+		wantErr  string
+	}{
+		{
+			name:     "invalid path",
+			filePath: "./testdata/non-existing.json",
+			wantErr:  "no such file or directory",
+		},
+		{
+			name:     "invalid JUnit XML format",
+			filePath: "./testdata/sbom.spdx.json",
```
migmartri, line 79: what's the difference between this test-case and the next one. 

Shouldn't this one actually test an xml format that happens to not to be valid (which I see you do below)

danlishka: The JUnit lib does not really return an error in case of non-xml files, that is why I had to improve a little bit more the validate function. 

I think we can remove one of tests related to the invalid xml. I was just trying to be on the safe side.

internal/attestation/crafter/materials/materials.go
```diff
@@ -34,6 +36,30 @@ type crafterCommon struct {
 	input  *schemaapi.CraftingSchema_Material
 }
 
+// uploadAndCraft uploads the artifact to CAS and crafts the material
+// this function is used by all the uploadable artifacts crafters (SBOMs, JUnit, and more in the future)
+func uploadAndCraft(ctx context.Context, input *schemaapi.CraftingSchema_Material, uploader casclient.Uploader, artifactPath string) (*api.Attestation_Material, error) {
+	result, err := uploader.UploadFile(ctx, artifactPath)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
danlishka: A very good point! Thanks!

internal/attestation/crafter/materials/junit_xml.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Errorf("can't read the file: %w", err)
+	}
+
+	if err := xml.Unmarshal(bytes, &junit.Suite{}); err != nil {
+		return fmt.Errorf("invalid JUnit XML file: %w", ErrInvalidMaterialType)
+	}
+
+	_, err = junit.IngestReader(f)
+	if err != nil {
+		i.logger.Debug().Err(err).Msg("error decoding file: " + filePath)
```
migmartri, line 72: Nitpick, no need to change though.

There might be a more native way of performing interpolation, maybe smith like MsgF or Sprintf

migmartri (approved):
LGTM!

internal/attestation/crafter/materials/junit_xml.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if err := xml.Unmarshal(bytes, &junit.Suite{}); err != nil {
+		return fmt.Errorf("invalid JUnit XML file: %w", ErrInvalidMaterialType)
+	}
+
+	_, err = junit.IngestReader(f)
+	if err != nil {
+		s := fmt.Sprintf("error decoding file: %s", filePath)
+		i.logger.Debug().Err(err).Msg(s)
```
migmartri, line 73: even better

```suggestion
		i.logger.Debug().Err(err).Msgf("error decoding file: %s", filePath)
```

migmartri (approved):
LGTM! :rocket: it!

PR #141, opened 2023-06-01
chore: refactor actuation

migmartri:
Part two of #38. Actuation abstraction

WIP

Issue #39, opened 2023-03-16
[fan-out integration] add resiliency to firing mechanism


migmartri:
Chainloop, currently has a mechanism to send attestation and artifact metadata to third party-integrations such an OCI registry (attestation) and Dependency-Track (CycloneDX SBOM).

![225039624-72816cc9-8247-426e-8b94-622c469d93d9](https://user-images.githubusercontent.com/24523/225775285-357fb74e-25f8-4305-9a41-3168a077b348.png)

It's implementation today happens in the service layer and has some drawbacks due to it's naive initial implementation using a go-routine https://github.com/chainloop-dev/chainloop/blob/e770faeeed854cfcff45a06100f9a2eef57d779a/app/controlplane/internal/service/attestation.go#L223

* It does not have delivery guarantees. 
* It happens in memory so container restarts or new rollouts could potentially drop the task.
* There is no receipt, ack of this fan-out execution of happening. Only looking at the logs (or sentry alerts) can give us some indication if a third-party integration fire has happened correctly.
* There is no way to handle an attestation that came in the past, a.k.a replay
* The firing logic is coupled to the service layer.

Lucky for us those shortcomings can be easily overcame by using an event bus / queue mechanism such as [nats.io streams](https://docs.nats.io/nats-concepts/jetstream/streams). 

We should design a new architecture that fixes the aforementioned shortcomings and sets the stage to a generic third-party integration framework https://github.com/chainloop-dev/chainloop/issues/38

waveywaves:
can this be assigned to me ?

migmartri:
hi! Feel free to add a proposal, it's not clear to us how to tackle it

Issue #38, opened 2023-03-16
abstract / define third party integration interface


migmartri:
NOTE: We should make sure that whatever design we make here meets also this https://github.com/chainloop-dev/chainloop/issues/28 other effort. 

Currently, adding a third party integration requires changes in many places in the control-plane codebase as described here https://github.com/chainloop-dev/chainloop/issues/37#issuecomment-1471902115. In short it requires changes is pretty much every layer of the application, a) api proto changes, b) biz/use-cases changes and of course c) actual integration code.

c) is the only part that's clearly defined with interfaces https://github.com/chainloop-dev/chainloop/blob/e770faeeed854cfcff45a06100f9a2eef57d779a/app/controlplane/internal/integrations/integration.go#L20-L29

The other two changes are currently custom and per-integration. 

The goal of this effort is to design an architecture where adding a new integration just requires such custom code to implement some interfaces, custom code that should probably live in https://github.com/chainloop-dev/chainloop/tree/e770faeeed854cfcff45a06100f9a2eef57d779a/app/controlplane/internal/integrations

Some goals

* These new interfaces should cover the "registration" and "attachment" use cases
* We should aim to no code changes required for a new attestation in the biz, service or API layer

migmartri:
A WIP code for the dispatcher codo can be found here https://github.com/migmartri/chainloop/pull/1/files

migmartri:
Our latest release contains all the sdk changes and new plugins https://github.com/chainloop-dev/chainloop/releases/tag/v0.11.1

Issue #28, opened 2023-03-14
Merge OCI-repositories and Integration models


migmartri:
Chainloop control plane has a mechanism to forward received DSSE attestation envelopes or SBOMs to different backends, i.e OCI registry or Dependency-Track.    

![overview-2](https://user-images.githubusercontent.com/24523/225039624-72816cc9-8247-426e-8b94-622c469d93d9.png)

This means that conceptually both OCI registries and SBOM processors are at the same level, third party fan out integrations, but reality is that the implementation grew in a different manner.

Because of historical development reasons, currently, an OCI repository is mapped as its own identity in both the data base and business logic.

![image](https://user-images.githubusercontent.com/24523/225041902-703e14dc-7850-4b89-8e76-3480b2a5190c.png)

while dependency track integration makes use of a generic "integration" model meant to encapsulate different processors. 

```golang
type Integration struct {
	ent.Schema
}

func (Integration) Fields() []ent.Field {
	return []ent.Field{
		field.UUID("id", uuid.UUID{}).Default(uuid.New).Unique().Immutable(),
		field.String("kind").Immutable(),
		field.String("secret_name").Immutable(),
		field.Time("created_at").
			Default(time.Now).
			Immutable().
			Annotations(&entsql.Annotation{Default: "CURRENT_TIMESTAMP"}),
		// kind specific config
		field.Bytes("config").GoType(&pb.IntegrationConfig{}),
		field.Time("deleted_at").Optional(),
	}
}

func (Integration) Edges() []ent.Edge {
	return []ent.Edge{
		edge.From("attachments", IntegrationAttachment.Type).Ref("integration"),
		edge.From("organization", Organization.Type).Ref("integrations").Unique().Required(),
	}
}
```  

both of them belongs to an `organization` and get attached to different workflows. The only difference is that **having a valid OCI registry is mandatory**

This issue is about making the OCI repository feature part of the more generic "integrations" feature for consistency.

danlishka:
+1 for moving OCI registries to integrations. 

How we add or update the OCI registries today might be confusing. I will explain why.

There are two related commands in the CLI:

```
  - config       Configure this client
  - integration  Third party integrations
```

The config one is about the CLI setup, so as an end user, **I assume this is only about local changes**.

Now let's take a look at this example:

```
$ chainloop config set-oci-repo --username danlishka --password <secret> --repo ghcr.io/danlishka/chainloop-test-lab
```

[...]

PR #142, opened 2023-06-05
feat(third-party-integrations): generic dispatcher

migmartri:
Continuation of #38

> TLDR; The server side integrations code is now fully generic and decoupled from the dependency-track implementation. It also works for attestation-based integrations which we expect to implement soon https://github.com/chainloop-dev/chainloop/issues/28 

It defines a generic dispatcher (and removes previous dependency-track specific code). 

The fanout dispatcher is called when an attestation is received and performs two phases

### 1 - Integrations configuration preparation (integrationsMap)

- Load the integration attached to the workflow
- Calculate what kind of inputs these want, material type (i.e CDX SBOM) or attestation (thanks to this https://github.com/chainloop-dev/chainloop/pull/141)
- For each one it makes sure to collect all the context required for a plugin to be executed. Registration and attachment configurations, credentials and the actual input)

### 2 - Attestation analysis, materials extraction and integration execution

The next step is for the dispatcher to analyze the kind of materials is contained in the received attestation and match them with the previously computed map.

- For each material type we check if there is any integration waiting for it, if so, we download it (once) and enqueue an execution https://github.com/chainloop-dev/chainloop/pull/141 (this part will eventually be moved to a background processing system https://github.com/chainloop-dev/chainloop/issues/39)
- In addition, integrations can be configured to listen to attestation input types, we sent the attestations to those directly as well.

### DIFF
--- a/app/controlplane/internal/biz/integration_test.go
+++ b/app/controlplane/internal/biz/integration_test.go
@@ -20,7 +20,7 @@ import (
 	"errors"
 	"testing"
 
-	integrationsSDK "github.com/chainloop-dev/chainloop/app/controlplane/integrations/sdk/v1"
+	"github.com/chainloop-dev/chainloop/app/controlplane/integrations/sdk/v1"
 	integrationMocks "github.com/chainloop-dev/chainloop/app/controlplane/integrations/sdk/v1/mocks"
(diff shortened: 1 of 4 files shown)


danlishka (approved):
LGTM!

PR #143, opened 2023-06-06
refactor(integrations sdk): UX improvements

migmartri:
Multiple UX improvements with regards to the integrations SDK are included in this patch

- `FanOut` interface methods are now consistent, `req` and `response` wise.
- Methods down the pipe now receive the results of previous ones. For example `Attach` will now receive the result of `Register`. Execute will receive both the `attachment` and `register` configuration.  The lifecycle of a fan out integration is `Registration` -> `Attachment` -> `Execution`
- The use of `proto.any` has been removed from the SDK. Now it's only used by the API wrapper.  
- the protobuffers definition in each extension is now used only to define the expected `inputs` for both the registration and attachment phases.

Refs #38


app/controlplane/integrations/dependencytrack/cyclonedx/v1/api.proto
```diff
@@ -22,28 +22,19 @@ import "validate/validate.proto";
 
 // RegistrationRequest is what the integration expect as input of the pre-registration step
```
danlishka, line 23: Do we follow any naming convention here dependencytrack/cyclonedx/v1? Is "dependencytrack" the integration name and "cyclonedx" a feature available in the extension? I know this will change but wondering what is the current approach that we follow here.

app/controlplane/integrations/dependencytrack/cyclonedx/v1/api.proto
```diff
@@ -22,28 +22,19 @@ import "validate/validate.proto";
...
 message RegistrationRequest {
-	RegistrationConfig config = 1 [(validate.rules).message.required = true];
+	string domain = 1 [(validate.rules).string.min_len = 1];
```
danlishka, line 25: Should we add proper validations here: 

```
string domain = 1 (validate.rules).string.address = true];
```

 or

```
string domain = 1 (validate.rules).string.uri = true];
```

app/controlplane/integrations/dependencytrack/cyclonedx/v1/api.proto
migmartri: Good question.

I actually followed a similar approach [than the one at Kubeapps](https://github.com/vmware-tanzu/kubeapps/tree/main/cmd/kubeapps-apis/plugins) which seems to be `{solution}/{purpose}/{version}`. I replaced `purpose` for `input type`.

That said, I do not think this pattern works anymore since an extension might be subscribed to more than one `material type` or to both materials and attestations

In fact, after our discussion offline this morning, where we laid of some potential extension `roles`, it might make sense to in fact do something like

`{role}/{provider}/{version}` for example

- `fanout/dependencytrack/v1`
- `fanout/oci/v1`

What do you think? @danlishka ?

app/controlplane/integrations/dependencytrack/cyclonedx/v1/api.proto
migmartri: that probably makes sense, good catch! Will change it!

app/controlplane/integrations/dependencytrack/cyclonedx/v1/integration.go
```diff
@@ -26,56 +26,59 @@ import (
...
-const description = "Dependency Track CycloneDX Software Bill Of Materials Integration"
-
-var _ sdk.FanOut = (*DependencyTrack)(nil)
 
 type DependencyTrack struct {
-	*sdk.BaseIntegration
+	*sdk.FanOutIntegration
+}
+
+type registrationConfig struct {
```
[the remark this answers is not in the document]
migmartri: The idea here is to decouple the `API schema` (currently defined and validated with protobuffers but in the future it might be defined differently) from the `state` that an extension author wants to preserve during the life-cycle of the extension.

It's true that today both the API and the state looks pretty much identical but it does not need to be like that. In fact if you look closely, the `api key` is not in the state. So credentials are handled differently. 

The other reason for maintaining your own structs is to make sure the plugin is decoupled from proto messages, for the most part.

Before this change, both the state and the API was defined in the proto files, what do you think, should we keep it like before, btw. note, nothing prevents an author to define and re-use structs from protos.

app/controlplane/integrations/dependencytrack/cyclonedx/v1/api.proto
danlishka: I would make it flat for now, and I think it makes more sense to rename integrations to extensions.

app/controlplane/integrations/dependencytrack/cyclonedx/v1/integration.go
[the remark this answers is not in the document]
danlishka: Makes sense! Thanks for the detailed explanation.

danlishka (approved):
LGTM! Please consider taking a look at my pending comments.

migmartri:
> LGTM! Please consider taking a look at my pending comments.

Done! ptal!

PR #144, opened 2023-06-07
doc(integrations): create template

migmartri:
- Moved current dependency-track integration to `core` since we expect the number of integrations to grow and the `extensions` directory will get polluted quick.
- The extensions loader has been moved to the extensions directory and now iterates over a provided list of factories instead.
- Added a `template` fanOut integration that includes the minimal, plus some commented out code to get started.
- Initialized some documentation in the form of Readme(s), more to come.

Refs #38


app/cli/internal/action/integration_add_deptrack.go
```diff
@@ -19,8 +19,8 @@ import (
 	"context"
 
 	pb "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1"
-	deptrack "github.com/chainloop-dev/chainloop/app/controlplane/extensions/dependencytrack/v1"
-	cxpb "github.com/chainloop-dev/chainloop/app/controlplane/extensions/dependencytrack/v1/api"
+	deptrack "github.com/chainloop-dev/chainloop/app/controlplane/extensions/core/dependencytrack/v1"
```
danlishka, line 22: Love it!

app/controlplane/extensions/core/template/v1/extension.go
```diff
@@ -0,0 +1,197 @@
...
+		&sdk.NewParams{
+			ID:      "template",
+			Version: "1.0",
+			Logger:  l,
+		},
+		// You can specify the inputs this attestation will be subscribed to, materials and or attestation envelope.
+		// In this case we are subscribing to SBOM_CYCLONEDX_JSON
+		sdk.WithInputMaterial(schemaapi.CraftingSchema_Material_SBOM_CYCLONEDX_JSON),
+		// You can also subscribed to any kind of materials
+		// sdk.WithInputMaterial(schemaapi.CraftingSchema_Material_MATERIAL_TYPE_UNSPECIFIED),
```
danlishka, line 50: What if you want both all materials and attestation?

app/controlplane/extensions/extensions.go
```diff
@@ -0,0 +1,52 @@
...
+// Load the available third party integrations
+// In the future this code will iterate over a dynamic directory of plugins
+// and try to load them one by one
+func Load(l log.Logger) (sdk.Loaded, error) {
+	var res sdk.Loaded
+
+	// Array of integrations that are meant to be loaded
+	// Eventually this will be dynamically loaded from a directory
+	toEnable := []sdk.FanOutFactory{
+		dependencytrack.New,
```
danlishka, line 34: This is the only place where you need to add a new extension. If we have a hello / inspect extension, we could add it here.

We could also move it to extensions_to_enable.go, up to you.

app/controlplane/extensions/core/template/v1/extension.go
migmartri: You can put both. It's commented out.

app/controlplane/extensions/extensions.go
migmartri: > This is the only place where you need to add a new extension. If we have a hello / inspect extension, we could add it here.

That would mean that's available, should we make it available?

> We could also move it to extensions_to_enable.go, up to you.

That sounds to specific to me, this file will contain more methods once we enable plugins.

PR #145, opened 2023-06-07
fix(dispatcher): only download materials that are in the CAS

migmartri:
This fixes an issue where the dispatcher fails if there are attached integrations in a workflow and one of the associated materials was a container image type.

This change makes sure that the dispatcher only downloads the materials that are in fact in the CAS. Before it used to infer a downloadable item if there was a `digest` and no content available. This is wrong, since container images match this criteria and are not in the CAS. 

To solve it I added an additional annotation to the materials that is set once its uploaded to the CAS. To do that I leverage the common download and craft method introduced in #120

image

I also made sure that all the uploadable materials are now using such method and hence the new annotation. 

Refs #38


app/controlplane/internal/dispatcher/dispatcher.go
```diff
@@ -199,7 +199,7 @@ func (d *FanOutDispatcher) Run(ctx context.Context, envelope *dsse.Envelope, org
 		// Retrieve material content
 		content := []byte(material.Value)
 		// It's a downloadable so we retrieve and override the content variable
-		if material.Hash != nil {
+		if material.Hash != nil && material.DownloadableFromCAS {
```
danlishka, line 202: Is this check needed here anymore?

```
material.Hash != nil
```

internal/attestation/crafter/materials/junit_xml_test.go
```diff
@@ -45,7 +46,7 @@ func TestNewJUnitXMLCrafter(t *testing.T) {
 		{
 			name: "wrong type",
 			input: &contractAPI.CraftingSchema_Material{
-				Type: contractAPI.CraftingSchema_Material_CONTAINER_IMAGE,
+				Type: contractAPI.CraftingSchema_Material_SBOM_CYCLONEDX_JSON,
```
danlishka, line 49: Why did we change it here?

internal/attestation/renderer/chainloop/v02.go
```diff
@@ -194,6 +199,10 @@ func normalizeMaterial(material *slsa_v1.ResourceDescriptor) (*NormalizedMateria
 		return nil, fmt.Errorf("material name not found")
 	}
 
+	if v, ok := material.Annotations[AnnotationMaterialCAS]; ok && v.(bool) {
+		m.DownloadableFromCAS = true
```
migmartri, line 203: change to uploaded to CAS

danlishka: Based on our conversation in the morning. Yes, it will be more consistent this way.

danlishka (approved):
LGTM

PR #147, opened 2023-06-12
feat(extension sdk): jsonSchema based input definition

migmartri:
This PR drops the use of protocol buffers to define the registration and attachment request inputs. Instead we now rely on regular golang structs with jsonSchema annotations for validations.

The use of jsonSchema is transparent to the extension author. To them, the procedure looks like creating and optionally annotating the Go structs that then they can use for marshaling the JSON API request.

The SDK in turn will take these structs, generate a JSON schema based on it and use them for server side validations.

Next, the idea would be to expose this schema to the client so the different options can be shown to the user and run client side validations. 

This patch also updates the `dependencytrack` extension and `CLI entrypoints` to leverage the new methods.

Refs #38 
Refs https://github.com/invopop/jsonschema for schema generation and https://github.com/santhosh-tekuri/jsonschema/ for validation


app/cli/internal/action/integration_add_deptrack.go
```diff
@@ -34,21 +33,22 @@ func NewIntegrationAddDeptrack(cfg *ActionsOpts) *IntegrationAddDeptrack {
...
-	}
 
-	anyConfig, err := anypb.New(&cdxRegistrationRequest)
+	config := make(map[string]any)
+	config["instanceURI"] = host
+	config["apiKey"] = apiKey
+	config["allowAutoCreate"] = allowAutoProjectCreation
+
+	// Transform to structpb for transport
+	configRequest, err := structpb.NewStruct(config)
```
danlishka, line 43: configRequest or requestConfig?

PR #148, opened 2023-06-12
feat(cli): generic integration register and attachment endpoints

migmartri:
This patch updates our CLI by adding generic `integration registration` in the existing `integration add` command as well as a generic `integration attach` one 

This patch also

- Maintains current dependency-track registration/attachment UX but replaces the underlying logic to use a generic action handler.
- The dependency-track specific commands has been marked as deprecated so in practice they are hidden now.

New registration

```console
$ chainloop integration add -h               
Register a new instance of an integration

Usage:
  chainloop integration add INTEGRATION_ID --options key=value,key=value [flags]

Examples:
  chainloop integration add dependencytrack --options instance=https://deptrack.company.com,apiKey=1234567890

Flags:
      --description string   integration registration description
  -h, --help                 help for add
      --options strings      integration arguments
```

Attachment 

```console
$ chainloop wf integration attach -h
Attach an existing registered integration to a workflow

Usage:
  chainloop workflow integration attach [flags]

Examples:
  chainloop workflow integration attach --workflow deadbeef --integration beefdoingwell --options projectName=MyProject

Flags:
  -h, --help                 help for attach
      --integration string   ID of the integration already registered in this organization
      --options strings      integration attachment arguments
      --workflow string      ID of the workflow to attach this integration
```

Refs #38


app/cli/cmd/integration_add_deptrack.go
```diff
@@ -29,17 +29,24 @@ func newIntegrationAddDepTrackCmd() *cobra.Command {
 	var allowAutoCreate bool
 
 	cmd := &cobra.Command{
-		Use:     "dependency-track",
-		Aliases: []string{"deptrack"},
-		Short:   "Add Dependency-Track integration ",
+		Use:        "dependency-track",
+		Aliases:    []string{"deptrack"},
+		Short:      "Add Dependency-Track integration ",
+		Deprecated: "use `chainloop integration add` instead",
```
danlishka, line 35: I am confused, so we are deprecating the old method with "dependency-track" and the new method will follow almost the same UX but the key is "dependencytrack"?

app/cli/internal/action/integration_add.go
```diff
@@ -23,30 +23,25 @@ import (
...
 	// Transform to structpb for transport
-	requestConfig, err := structpb.NewStruct(config)
+	requestConfig, err := structpb.NewStruct(options)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse arguments: %w", err)
 	}
 
 	i, err := client.Register(context.Background(), &pb.IntegrationsServiceRegisterRequest{
-		Kind:        "dependencytrack",
+		Kind:        kind,
```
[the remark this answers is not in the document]
migmartri: basically to not to be confused with integrationID which is an UUID used during attachment. 

That said I think we should change it to `integrationID` and just document it.

danlishka (approved):
LGTM. There are some minor comments.

app/cli/cmd/integration_add_deptrack.go
migmartri: > and the new method will follow almost the same UX but the key is "dependencytrack"?

The new command will use as first argument whatever `id` the extension you are invoking has. In our case `dependencytrack` yes. 

What is the source of the confusion? The deprecation? The fact that we are building a new functionality on a parent command?

app/cli/internal/action/integration_add.go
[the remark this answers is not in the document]
danlishka: +1

app/cli/cmd/integration_add_deptrack.go
migmartri: > I would suggest to exit with error and a proper deprecation message in case of dependency-track.

but that will break the current commands in the guides no?

And there is a deprecation message already


```
$ go run main.go --insecure integration add dependency-track 
Command "dependency-track" is deprecated, use `chainloop integration add` instead
```

migmartri: >  Now there are two working versions: dependency-track and dependencytrack. Are we going to maintain both?

my idea was to first, deprecate the old command (as I did) and once we stabilize the new registration/attachment methods and update the docs we can fully remove it. WDYT?

PR #151, opened 2023-06-12
chore(controlplane): rename display-name

migmartri:
For some reason I didn't apply all the suggestions from here https://github.com/chainloop-dev/chainloop/pull/146

I did not rename the server code from `displayName` to `description`. This patch solves that mistake. Sorry about that. 

Refs #38


app/cli/cmd/integration_add.go
```diff
@@ -47,7 +47,6 @@ func newConfigIntegrationAddCmd() *cobra.Command {
 		},
 	}
 
-	cmd.PersistentFlags().StringVar(&integrationDescription, "description", "", "integration registration description")
```
migmartri, line 50: this is an spurious error from a previous diff.

danlishka (approved):

migmartri:
Thanks @danlishka

Issue #157, opened 2023-06-13
The new integration framework does not properly handle integers


danlishka:
If you use an integer in your extension:

```
type registrationRequest struct {
[...]
	Port     int    `json:"port" jsonschema:"required"`
}
```

and then you try to add the registration:

```
$ chainloop integration add <extension> --options port=123

ERR validation error: expected integer, but got string
exit status 1
```

The control plane log:

```
2023-06-13T15:21:58.972+0200	ERROR	{"kind": "server", "component": "grpc", "operation": "/controlplane.v1.IntegrationsService/Register", "args": "***", "code": 400, "reason": "wrong validation", "stack": "error: code = 400 reason = wrong validation message = validation error: expected integer, but got string metadata = map[] cause = <nil>", "latency": 0.7012075}
```

migmartri:
Awesome, thanks for reporting it, I'll assign it to me and mark is as bug

PR #159, opened 2023-06-13
feat: Added SMTP extension

### LINKED ISSUE #158: SMTP extension
Add a new core extension to allow Chainloop users to send email notifications for every workflow run. 
This will also be an excellent opportunity to validate a new experimental extensions API.

danlishka:
This is a new extension for sending emails via SMTP. This is still a very early experimental version focusing on validating the latest cutting-edge integrations framework.

The README file provides more information on how to use this extension.

 closes #158

### DIFF
--- a/app/controlplane/extensions/core/smtp/v1/README.md
+++ b/app/controlplane/extensions/core/smtp/v1/README.md
@@ -1,6 +1,6 @@
 # SMTP Fan-out Extension
 
-With this extension, you can send an email for every workflow run and attestation, ensuring better communication.
+With this extension, you can send an email for every workflow run and attestation.
 
 ## How to use it
 
@@ -8,13 +8,19 @@ In the following example, we will use the AWS SES service.
 
 1. To get started, you need to register the extension in your Chainloop organization.
 ```
-chainloop integration add smtp --options user=AHDHSYEE7e73,password=kjsdfda8asd****,host=email-smtp.us-east-1.amazonaws.com,port=587,to=platform-team@example.com,from=notifier@example.com
+chainloop integration registered add smtp --options user=AHDHSYEE7e73,password=kjsdfda8asd****,host=email-smtp.us-east-1.amazonaws.com,port=587,to=platform-team@example.com,from=notifier@example.com
 ```
 
 2. When attaching the integration to your workflow, you have the option to specify CC:
 
 ```
 chainloop workflow integration attach --workflow $WID --integration $IID --options cc=security@example.com
+```
+
+cc is optional:
+
+```
+chainloop workflow integration attach --workflow $WID --integration $IID
 ```
 
 Starting now, every time a workflow run occurs, an email notification will be sent containing the details of the run and attestation.
--- a/app/controlplane/extensions/core/smtp/v1/README.md
+++ b/app/controlplane/extensions/core/smtp/v1/README.md
@@ -1,0 +1,20 @@
+# SMTP Fan-out Extension
+
+With this extension, you can send an email for every workflow run and attestation, ensuring better communication.
+
+## How to use it
+
+In the following example, we will use the AWS SES service.
+
+1. To get started, you need to register the extension in your Chainloop organization.
+```
+chainloop integration add smtp --options user=AHDHSYEE7e73,password=kjsdfda8asd****,host=email-smtp.us-east-1.amazonaws.com,port=587,to=platform-team@example.com,from=notifier@example.com
+```
+
+2. When attaching the integration to your workflow, you have the option to specify CC:
+
+```
+chainloop workflow integration attach --workflow $WID --integration $IID --options cc=security@example.com
+```
+
+Starting now, every time a workflow run occurs, an email notification will be sent containing the details of the run and attestation.
--- a/app/controlplane/extensions/core/smtp/v1/extension.go
+++ b/app/controlplane/extensions/core/smtp/v1/extension.go
@@ -110,7 +110,10 @@ func (i *Integration) Register(_ context.Context, req *sdk.RegistrationRequest) 
 	- To: %s
 	`
 	body := fmt.Sprintf(tpl, i.Describe().ID, i.Describe().Version, host, port, user, from, to)
-	sendEmail(host, port, user, password, from, to, "", subject, body)
+	err = sendEmail(host, port, user, password, from, to, "", subject, body)
+	if err != nil {
+		return nil, fmt.Errorf("sending an email: %w", err)
+	}
 
 	return response, nil
 }
@@ -187,7 +190,10 @@ func (i *Integration) Execute(_ context.Context, req *sdk.ExecutionRequest) erro
 This email has been delivered via integration %s version %s.
 	`
 	body := fmt.Sprintf(tpl, req.WorkflowID, jsonBytes, i.Describe().ID, i.Describe().Version)
-	sendEmail(host, port, user, password, from, to, ac.Cc, subject, body)
+	err = sendEmail(host, port, user, password, from, to, ac.Cc, subject, body)
+	if err != nil {
+		return fmt.Errorf("sending an email: %w", err)
+	}
 
 	return nil
 }
--- a/app/controlplane/extensions/core/smtp/v1/extension.go
+++ b/app/controlplane/extensions/core/smtp/v1/extension.go
@@ -1,0 +1,230 @@
+//
+// Copyright 2023 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package smtp
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	nsmtp "net/smtp"
+
+	"github.com/chainloop-dev/chainloop/app/controlplane/extensions/sdk/v1"
+	"github.com/go-kratos/kratos/v2/log"
+	"github.com/in-toto/in-toto-golang/in_toto"
+)
+
+type Integration struct {
+	*sdk.FanOutIntegration
+}
+
+type registrationRequest struct {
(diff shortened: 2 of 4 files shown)


app/controlplane/extensions/core/smtp/v1/README.md
```diff
@@ -0,0 +1,20 @@
...
+## How to use it
+
+In the following example, we will use the AWS SES service.
+
+1. To get started, you need to register the extension in your Chainloop organization.
+```
+chainloop integration add smtp --options user=AHDHSYEE7e73,password=kjsdfda8asd****,host=email-smtp.us-east-1.amazonaws.com,port=587,to=platform-team@example.com,from=notifier@example.com
+```
+
+2. When attaching the integration to your workflow, you have the option to specify CC:
```
migmartri, line 14: nitpick. I'd put two commands here, one without the option and one with the option to make clear `cc` is optional.

app/controlplane/extensions/core/smtp/v1/README.md
```diff
@@ -0,0 +1,20 @@
+# SMTP Fan-out Extension
+
+With this extension, you can send an email for every workflow run and attestation, ensuring better communication.
```
migmartri, line 3: > ensuring better communication.

not sure what you mean by this.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+type Integration struct {
+	*sdk.FanOutIntegration
+}
+
+type registrationRequest struct {
+	To       string `json:"to" jsonschema:"format=email,description=The email address to send the email to."`
+	From     string `json:"from" jsonschema:"format=email,description=The email address of the sender."`
+	User     string `json:"user" jsonschema:"minLength=1,description=The username to use for the SMTP authentication."`
+	Password string `json:"password" jsonschema:"description=The password to use for the SMTP authentication."`
+	Host     string `json:"host" jsonschema:"description=The host to use for the SMTP authentication."`
```
migmartri, line 39: Would any of these format help? https://json-schema.org/understanding-json-schema/reference/string.html#hostnames

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	Port     string `json:"port" jsonschema:"description=The port to use for the SMTP authentication"`
```
migmartri, line 40: I am assuming you are using a string here because of this bug https://github.com/chainloop-dev/chainloop/issues/157

Maybe it would make sense to add a comment here so we do not forget to change it?

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+type registrationState struct {
+	To   string `json:"to"`
+	From string `json:"from"`
+	User string `json:"user"`
+	Host string `json:"host"`
+	Port string `json:"port"`
+}
+
+type attachmentRequest struct {
+	Cc string `json:"cc,omitempty" jsonschema:"format=email,description=The email address of the carbon copy recipient."`
```
migmartri, line 52: nitpick, acronyms should be all in capital letters. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+type attachmentState struct {
+	Cc string `json:"cc"`
```
migmartri, line 56: same uppercase comment here.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func New(l log.Logger) (sdk.FanOut, error) {
+	base, err := sdk.NewFanOut(
+		&sdk.NewParams{
+			ID:      "smtp",
+			Version: "0.1",
```
migmartri, line 63: good, we'll bump down the other integrations too.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	// Unmarshal the request
+	var request *registrationRequest
+	if err := sdk.FromConfig(req.Payload, &request); err != nil {
+		return nil, fmt.Errorf("invalid registration request: %w", err)
+	}
+
+	response := &sdk.RegistrationResponse{}
+
+	to, from, user, password, host, port := request.To, request.From, request.User, request.Password, request.Host, request.Port
+	rawConfig, err := sdk.ToConfig(&registrationState{To: to, From: from, User: user, Host: host, Port: port})
```
migmartri, line 93: I'd move the code about generating the response (and state) after the main registration logic

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("marshalling configuration: %w", err)
+	}
+	response.Configuration = rawConfig
+	response.Credentials = &sdk.Credentials{Password: password}
+
+	// validate and notify
+	subject := "[chainloop] New SMTP integration added!"
+	tpl := `
```
migmartri, line 102: It would be ideal to use go-templates instead https://pkg.go.dev/text/template, it can be in another patch.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	to, from, user, password, host, port := request.To, request.From, request.User, request.Password, request.Host, request.Port
+	rawConfig, err := sdk.ToConfig(&registrationState{To: to, From: from, User: user, Host: host, Port: port})
+	if err != nil {
+		return nil, fmt.Errorf("marshalling configuration: %w", err)
+	}
+	response.Configuration = rawConfig
+	response.Credentials = &sdk.Credentials{Password: password}
+
+	// validate and notify
+	subject := "[chainloop] New SMTP integration added!"
```
migmartri, line 101: ok, so this is in fact sending an email during registration. That's an interesting approach. If we just wanted to check that the credentials are correct, is there any other way to check it?

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	body := fmt.Sprintf(tpl, req.WorkflowID, jsonBytes, i.Describe().ID, i.Describe().Version)
+	err = sendEmail(host, port, user, password, from, to, ac.Cc, subject, body)
+	if err != nil {
+		return fmt.Errorf("sending an email: %w", err)
+	}
+
+	return nil
+}
+
+func validateExecuteRequest(req *sdk.ExecutionRequest) error {
```
migmartri, line 201: mental note for myself: this is starting to look more and more like smth that could get automatically validated.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	message := "From: " + from + "\n" +
+		"To: " + to + "\n" +
+		"Cc: " + cc + "\n" +
+		"Subject: " + subject + "\n\n" +
+		body
+
+	auth := nsmtp.PlainAuth("", user, password, host)
+	err := nsmtp.SendMail(host+":"+port, auth, from, []string{to}, []byte(message))
+	if err != nil {
+		fmt.Println("Error sending email1:", err)
```
migmartri, line 231: there is a namespaced logger instance injected (see dependency track example), use it instead.

app/controlplane/extensions/core/smtp/v1/extension_test.go
```diff
@@ -0,0 +1,101 @@
...
+		errMsg string
+	}{
+		{
+			name:   "missing properties",
+			input:  map[string]interface{}{},
+			errMsg: "missing properties: 'to', 'from', 'user'",
+		},
+		{
+			name:  "valid request",
+			input: map[string]interface{}{"from": "test@example.com", "to": "test@example.com", "host": "localhost", "port": "25", "user": "test", "password": "test"},
```
migmartri, line 39: probably test email formats and potential hostnames.

app/controlplane/extensions/core/smtp/v1/extension.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	if req.AttachmentInfo == nil || req.AttachmentInfo.Configuration == nil {
+		return errors.New("missing attachment configuration")
+	}
+
+	return nil
+}
+
+func sendEmail(host string, port string, user, password, from, to, cc, subject, body string) error {
```
migmartri, line 221: when you start having more than 2 arguments, specially if they are from the same type, it's common to change the function signature to accept a struct instead. That makes it more readable and less error prone. I'd suggest doing that here

migmartri (commented):
Awesome, good start! 

I left some comments, most of them suggestions that can be changed in another patch.

For example, by encapsulating the render code in a function + go templates you'll be able to test them.

I'd suggest to also add comments in the code explaining/separating the different blocks. And as a general rule, make sure that initialization of variables/code happens the closest to the code that'll use it as possible (i.e preparing the response in `register`.

Looking good!

app/controlplane/extensions/core/smtp/v1/extension.go
danlishka: This is not supported by our jsonschema library - I can work on the PR though. We would have to use something like: `anyof_format=hostname,ipv4`

to get

```
"anyOf": [
{ "format": "hostname" },
{ "format": "ipv4" },
{ "format": "ipv6" }
]
```

app/controlplane/extensions/core/smtp/v1/extension.go
danlishka: Yes, this way, you really make sure you receive an email. It might still go to spam, or the `to` address might be incorrect, or the `from` address could be rejected by the SMTP service, etc.

app/controlplane/extensions/core/smtp/v1/extension.go
danlishka: Yes, the SMTP logic will also be moved to a separate package.

app/controlplane/extensions/core/smtp/v1/extension.go
danlishka: yes, this will change once we move it out to a separate package

migmartri:
Please link the issue in the description not in the title of the PR.

migmartri (approved):
LGTM!

PR #160, opened 2023-06-13
fix: Choose the random port for HTTP Metrics in the development mode

### LINKED ISSUE #155: Port 5000 is already in use on OS X
When switching from Linux to OSX for development, I run into this issue when launching a control plane in the development mode:

```
panic: listen tcp 0.0.0.0:5000: bind: address already in use
```

We run HTTP metrics on port 5000 as defined in the `app/controlplane/configs/config.devel.yaml` file.

In the latest versions of OSX, the AirPlay service process uses the 5000 port as well.

```
lsof -i :5000
COMMAND    PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
ControlCe 1081 d    7u  IPv4 0x4dfbc9201515d62b      0t0  TCP *:commplex-main (LISTEN)
ControlCe 1081 d    8u  IPv6 0x4dfbc9201439e933      0t0  TCP *:commplex-main (LISTEN)
```

There is a way to change it in `System Preferences/Sharing` by unchecking the AirPlay Receiver to fix the issue, but I think we better change this port or adequately document it.

danlishka:
Choose the random port for HTTP Metrics in the development mode to fix the already-in-use port issues on OSX, closes #155

### DIFF
--- a/app/controlplane/configs/config.devel.yaml
+++ b/app/controlplane/configs/config.devel.yaml
@@ -7,7 +7,7 @@ server:
     addr: 0.0.0.0:8000
     timeout: 1s
   http_metrics:
-    addr: 0.0.0.0:5000
+    addr: 0.0.0.0:0
   grpc:
     addr: 0.0.0.0:9000
     # We have some slow operations such as verifying an OCI registry


app/controlplane/configs/config.devel.yaml
```diff
@@ -7,7 +7,7 @@ server:
     addr: 0.0.0.0:8000
     timeout: 1s
   http_metrics:
-    addr: 0.0.0.0:5000
+    addr: 0.0.0.0:0
```
migmartri, line 10: I'd add a comment explaining what this means.

migmartri (approved):
LGTM!

PR #161, opened 2023-06-13
feat(cli): Validate and show options during registration/attachment

### LINKED ISSUE #157: The new integration framework does not properly handle integers

migmartri:
TODO: Add tests

Improved client side logic related to integrations

- Fixes #157 by transforming the string values into the actual values the JSON schema is expecting
- Runs client side validations on the schema and if it fails, it shows the available options. 
- Stabilize the available options table, making sure that they have the same order every time. 

To achieve that I needed to add an additional API endpoint (`Describe`) to get information about a specific registered integration. This one is used during attachment to retrieve the extension ID so we can then get the JSON schema.

Refs #38 
Fixes #157


app/cli/cmd/registered_integration_add.go
```diff
@@ -66,5 +123,39 @@ func parseKeyValOpts(opts []string) (map[string]any, error) {
...
+	for k, v := range options {
+		prop, ok := propertiesMap[k]
+		if !ok {
+			continue
+		}
+
+		switch prop.Type {
+		case "string":
+			options[k] = v.(string)
+		case "integer":
```
danlishka, line 145: ok, that's what we are missing. I will update smtp/port implementation once you land this diff.

danlishka (approved):
LGTM!

PR #162, opened 2023-06-14
chore(extensions): fail on extension load

migmartri:
Update the loader code to fail instead of skip in the case of extension load failure.

Skipping is useful if/when we load third-party extensions in the form of plugins. But since for now we are implementing the extensions as part of the core, it's best to fail to detect errors during the development cycle. 

Refs #38


migmartri:
Abandoned in favor of https://github.com/chainloop-dev/chainloop/pull/165

PR #163, opened 2023-06-14
chore: update template extension descriptions

migmartri:
Refs #38

### DIFF
--- a/app/controlplane/extensions/core/template/v1/extension.go
+++ b/app/controlplane/extensions/core/template/v1/extension.go
@@ -34,10 +34,12 @@ type Integration struct {
 // see https://github.com/invopop/jsonschema#example for more information
 type registrationRequest struct {
 	// TestURL is a required, valid URL
-	TestURL string `json:"testURL" jsonschema:"format=uri"`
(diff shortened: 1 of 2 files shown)


app/controlplane/extensions/core/template/v1/extension.go
```diff
@@ -178,10 +180,15 @@ func (i *Integration) Execute(_ context.Context, req *sdk.ExecutionRequest) erro
 // In this case we expect to receive a SBOM in CycloneDX format
 // plus registration, attachment state and credentials
 func validateExecuteRequest(req *sdk.ExecutionRequest) error {
-	if req == nil || req.Input == nil || req.Input.Material == nil || req.Input.Material.Content == nil {
-		return errors.New("invalid input")
+	if req == nil || req.Input == nil {
+		return errors.New("execution input not received")
+	}
+
+	if req.Input.Material == nil || req.Input.Material.Content == nil {
```
danlishka, line 187: We could add two use cases here:
- validation if material
- validation if attestation envelope

danlishka (approved):
LGTM. The comment about the validation is not really related to the scope of the PR and we will improve it in another patch.

PR #167, opened 2023-06-14
feat(extensions): validate schema options

migmartri:
Put limitations on the structs that can be used to generate an extension schema.

For example, now

- Nested structs are not allowed
- Array types are not supported either.

To achieve the first point, now we generate the schema flat.

refs #38

### DIFF
--- a/app/controlplane/extensions/sdk/v1/fanout.go
+++ b/app/controlplane/extensions/sdk/v1/fanout.go
@@ -375,8 +375,37 @@ func FromConfig(data Configuration, v any) error {
 	return json.Unmarshal(data, v)
 }
 
-// generate a JSON schema from a struct, see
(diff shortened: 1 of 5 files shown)


app/controlplane/extensions/sdk/v1/fanout.go
```diff
@@ -375,8 +375,37 @@ func FromConfig(data Configuration, v any) error {
 	return json.Unmarshal(data, v)
 }
 
-// generate a JSON schema from a struct, see
+// generate a flat JSON schema from a struct using https://github.com/invopop/jsonschema
+// We've put some limitations on the kind of input structs we support, for example:
```
danlishka, line 379: That is a very good point!

danlishka (approved):

PR #169, opened 2023-06-14
feat(extensions): OCI extension for attestation upload

### LINKED ISSUE #168: [fanout extensions] Add OCI extension
Now that we have a new extensions framework, we could add an OCI extension that can be attached to any workflow and will receive its attestations.

Note that this is not a direct replacement of #28 (yet) but an optional feature

migmartri:
Create a generic OCI registry extension that works with any static credentials keypair. This means that's compatible with most provider except AWS' Elastic Container Registry (ECR)

It supports an optional `prefix` that can be added at attachment time.

Closes #168

### DIFF
--- a/app/cli/cmd/attached_integration_list.go
+++ b/app/cli/cmd/attached_integration_list.go
@@ -57,14 +57,17 @@ func attachedIntegrationListTableOutput(attachments []*action.AttachedIntegratio
 		wf := i.Workflow
 		integration := i.Integration
 
-		maps.Copy(i.Config, integration.Config)
 		var options []string
-		for k, v := range i.Config {
-			if v == "" {
-				continue
+		if i.Config != nil {
+			maps.Copy(i.Config, integration.Config)
+			for k, v := range i.Config {
+				if v == "" {
+					continue
+				}
+				options = append(options, fmt.Sprintf("%s: %v", k, v))
 			}
-			options = append(options, fmt.Sprintf("%s: %v", k, v))
 		}
+
 		t.AppendRow(table.Row{i.ID, integration.Kind, strings.Join(options, "\n"), i.CreatedAt.Format(time.RFC822), wf.NamespacedName()})
 		t.AppendSeparator()
 	}
--- a/app/cli/cmd/registered_integration_add.go
+++ b/app/cli/cmd/registered_integration_add.go
@@ -33,7 +33,7 @@ func newRegisteredIntegrationAddCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "add INTEGRATION_ID --options key=value,key=value",
 		Short:   "Register a new instance of an integration",
-		Example: `  chainloop integration registered add dependencytrack --options instance=https://deptrack.company.com,apiKey=1234567890`,
+		Example: `  chainloop integration registered add dependencytrack --opt instance=https://deptrack.company.com,apiKey=1234567890 --opt username=chainloop`,
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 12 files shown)


app/cli/cmd/attached_integration_list.go
```diff
@@ -57,14 +57,17 @@ func attachedIntegrationListTableOutput(attachments []*action.AttachedIntegratio
 		wf := i.Workflow
 		integration := i.Integration
 
-		maps.Copy(i.Config, integration.Config)
 		var options []string
-		for k, v := range i.Config {
-			if v == "" {
-				continue
+		if i.Config != nil {
```
migmartri, line 61: I noticed that if there is no attachment configuration, like in this case, the rendering could fail.

app/controlplane/extensions/core/template/v1/extension.go
```diff
@@ -29,6 +29,7 @@ type Integration struct {
 	*sdk.FanOutIntegration
 }
 
+// 1 - API schema definitions
```
migmartri, line 32: I made some improvements in the template based on my experience using it now as base.

app/cli/cmd/registered_integration_add.go
```diff
@@ -93,13 +95,14 @@ func parseAndValidateOpts(opts []string, schema *action.JSONSchema) (map[string]
...
+// 1 - Split the options into key/value pairs
+// 2 - Cast the values to the expected type defined in the schema
 func parseKeyValOpts(opts []string, propertiesMap action.SchemaPropertiesMap) (map[string]any, error) {
-	// Two steps process
-
 	// 1 - Split the options into key/value pairs
 	var options = make(map[string]any)
 	for _, opt := range opts {
-		kv := strings.Split(opt, "=")
+		kv := strings.SplitN(opt, "=", 2)
```
migmartri, line 105: A maximum of two parts since complex values such as json p12 keys didn't work

app/cli/cmd/registered_integration_add.go
```diff
@@ -66,7 +66,9 @@ func newRegisteredIntegrationAddCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&integrationDescription, "description", "", "integration registration description")
-	cmd.Flags().StringSliceVar(&options, "options", nil, "integration arguments")
+	// StringSlice seems to struggle with comma-separated values such as p12 jsonKeys provided as passwords
+	// So we need to use StringArrayVar instead
+	cmd.Flags().StringArrayVar(&options, "opt", nil, "integration arguments")
```
migmartri, line 71: this has a consequence, that we no longer can provide things like `--opt foo=bar,bar=baz` but instead `--opt foo=bar --opt bar=baz`

app/controlplane/extensions/core/ociregistry/v1/README.md
```diff
@@ -0,0 +1,4 @@
+# OCI registry extension
+
+Use this extension to send attestations to a compatible OCI registry.
```
danlishka, line 3: only attestations? What about materials? Another ext?

app/controlplane/extensions/extensions.go
```diff
@@ -34,6 +35,7 @@ func Load(l log.Logger) (sdk.AvailableExtensions, error) {
 	toEnable := []sdk.FanOutFactory{
 		dependencytrack.New,
 		smtp.New,
+		ociregistry.New,
```
danlishka, line 38: boom!

danlishka (approved):
Perfect!

app/controlplane/extensions/core/ociregistry/v1/README.md
migmartri: Maybe, as first version this sends attestations.

PR #192, opened 2023-06-20
docs(extensions): improve some wording

migmartri:
Improve some wording and moves the how to build an extension content to this file.

The reason for that move is so the template readme file can become an actual readme file template not a development guide.

refs #38


danlishka (approved):
LGTM

PR #194, opened 2023-06-21
refactor(plugins): s/extension/plugin

migmartri:
Rename source code references to `extensions` in favor of `plugins`. This is a more commonly understand term.

It also adds a disclaimer note in the plugins/Readme.md about the limitations of the current system and a pointer to a task. 

Refs #38


danlishka (approved):
LGTM

PR #196, opened 2023-06-21
docs(extensions): make links absolute

migmartri:
This readme file will be imported from our docs system so we need to fix the links by making them absolute. 

Refs #38


migmartri:
Merged it to unblock the documentation site from pulling from this required change.

ptal in any case

Issue #193, opened 2023-06-21
Extend extensions readme generator


migmartri:
https://github.com/chainloop-dev/chainloop/pull/191 added support for a generator that keeps extension Readme files up to date with their input schema information.

We'd like to extend this generator to inject additional metadata such as `description` `id` subscribed inputs and so on.

@danlishka had some suggestions on the format https://github.com/chainloop-dev/chainloop/pull/191#discussion_r1236600571

danlishka:
Add supported materials section as well. 

https://github.com/chainloop-dev/chainloop/pull/211/files#r1245729281

Issue #60, opened 2023-04-03
in-toto 1.0 support


migmartri:
In-toto 1.0 [has been released](https://github.com/in-toto/attestation/releases/tag/v1.0) and it includes a couple of major enhancements that seem relevant to the Chainloop Project.

### [New upstream predicates](https://github.com/in-toto/attestation/tree/main/spec/predicates#vetted-predicates) and [guidelines](https://github.com/in-toto/attestation/blob/main/docs/new_predicate_guidelines.md) available

Chainloop uses a custom predicate `chainloop.dev/attestation/v0.1`. We should re-evaluate the need of our custom predicate or use an existing one. If we still do, we should make sure it adheres to the guidelines and if it would make sense to try to upstream it.

### Attestation bundle 

The new [attestation bundle](https://github.com/in-toto/attestation/blob/main/spec/v1.0/bundle.md) can be specially handy in the context of https://github.com/chainloop-dev/chainloop/issues/53. It will allow us to bundle not only the Chainloop generated DSSE envelope but any other generated envelope. 

For extensibility in the future, it might be a good idea to default to a bundle for communication with the control-plane  

### Resource descriptor 

Currently Chainloop encodes the materials in the attestation like this. It's basically an extension of a slsa material (with digest and URI) plus some additional metadata used by the Chainloop controlplane for validation and routing (name and type)

The new [resource descriptor](https://github.com/in-toto/attestation/blob/main/spec/v1.0/resource_descriptor.md) might be a good fit to replace that custom type. The `annotations` field could contain the custom chainloop types. 

```json
 "materials": [
      {
        "material": {
          "slsa": {
            "digest": {
              "sha256": "7e5f829611edeedf36a7318f78e6914b85a8e2f90ac8628cc868b2e87204d93b"
            },
            "uri": "ghcr.io/chainloop-dev/frontend"
          }
        },
        "name": "frontend",
        "type": "CONTAINER_IMAGE"
      },
      {
        "material": {
          "slsa": {
            "digest": {
              "sha256": "092252ea069679c658a949748ca0bee9405d36630d6b460964c7225708de0bb5"
            },
            "uri": "sbom.cyclonedx.json"
          }
        },
        "name": "frontend-sbom",
        "type": "SBOM_CYCLONEDX_JSON"
      }
    ],

```

Congrats @adityasaky, @TomHennen and rest of the team for the release! Feel free to drop any thoughts or comments, always appreciated!

migmartri:
[...]

PR #211, opened 2023-06-27
feat(integration): guacsec/guac integration

### LINKED ISSUE #209: feat(integrations) guacsec/guac
Look into how an initial integration with https://github.com/guacsec/guac would look like. It's feasibility and scope.

At first glance, one way to approach this integration would be for Chainloop to populate a network reachable location (OCI, blob storage, ...) that then can be continuously scanned by a guac collector. 

Untitled-2022-12-20-1126

The data source provided by Chainloop will contain different pieces of evidence, for now we could push SBOMs (cyclonedx, spdx), in-toto statements and DSSE envelopes.

migmartri:
Integration that allows Chainloop users to **automatically send attestation** (`DSSE` envelopes / [`in-toto` statements](https://in-toto.io/)) **and CycloneDX/SPDX Software Bill Of Materials (SBOMs) to a staging area** in the form of a cloud storage bucket. **[GUAC](https://github.com/guacsec/guac) then can be configured to continuously monitor and inject the data** in that bucket.

See [this Readme](https://github.com/chainloop-dev/chainloop/blob/f838b1141ba3777b1f4a20b636526ae663746e61/app/controlplane/plugins/core/guac/v1/README.md) for more information.

overview

Currently, `Guac` only support Google Cloud Storage (GCS) as blob storage backend. So this patch only implements this provider although leaves the API open to other providers in the future.  

## Discovery

The plugin is now available in the system

```sh
$ chainloop integration available ls        
┌──────────────────┬─────────┬─────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────┐
│ ID               │ VERSION │ MATERIAL REQUIREMENT                │ DESCRIPTION                                                                                    │
├──────────────────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ dependency-track │ 1.2     │ SBOM_CYCLONEDX_JSON                 │ Send CycloneDX SBOMs to your Dependency-Track instance                                         │
├──────────────────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ smtp             │ 1.0     │                                     │ Send emails with information about a received attestation                                      │
├──────────────────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ oci-registry     │ 1.0     │                                     │ Send attestations to a compatible OCI registry                                                 │
├──────────────────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ discord-webhook  │ 1.1     │                                     │ Send attestations to Discord                                                                   │
├──────────────────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ guac             │ 1.0     │ SBOM_CYCLONEDX_JSON, SBOM_SPDX_JSON │ Export Attestation and SBOMs metadata to a blob storage backend so guacsec/guac can consume it │
└──────────────────┴─────────┴─────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────┘
```

And it's definition can be retrieved showing that today it sends both SBOM types

```sh
$ chainloop integration available describe --id guac
┌──────┬─────────┬─────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────┐
│ ID   │ VERSION │ MATERIAL REQUIREMENT                │ DESCRIPTION                                                                                    │
├──────┼─────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ guac │ 1.0     │ SBOM_CYCLONEDX_JSON, SBOM_SPDX_JSON │ Export Attestation and SBOMs metadata to a blob storage backend so guacsec/guac can consume it │
└──────┴─────────┴─────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────┘
┌────────────────────────────────────────────────────────────────────────────┐
│ Registration inputs                                                        │
├─────────────┬────────┬──────────┬──────────────────────────────────────────┤
│ FIELD       │ TYPE   │ REQUIRED │ DESCRIPTION                              │
├─────────────┼────────┼──────────┼──────────────────────────────────────────┤
│ bucket      │ string │ yes      │ Bucket name where to store the artifacts │
│ credentials │ string │ yes      │ Credentials to access the bucket         │
│ provider    │ string │ no       │ Blob storage provider: default gcs       │
└─────────────┴────────┴──────────┴──────────────────────────────────────────┘
```

## Registration

Registration by providing a bucket and service account credentials, provider can be sent as well although by default the only option we have now is `gcs`.

```sh
$ chainloop integration registered add guac --opt bucket=test-guac --opt credentials="$(cat /service-account-devel.json)"
┌──────────────────────────────────────┬─────────────┬──────┬───────────────────┬─────────────────────┐
│ ID                                   │ DESCRIPTION │ KIND │ CONFIG            │ CREATED AT          │
├──────────────────────────────────────┼─────────────┼──────┼───────────────────┼─────────────────────┤
│ d19dabf0-d6dd-4f00-91c0-eed209454d05 │             │ guac │ bucket: test-guac │ 28 Jun 23 14:23 UTC │
│                                      │             │      │ provider: gcs     │                     │
└──────────────────────────────────────┴─────────────┴──────┴───────────────────┴─────────────────────┘

```

## Attachment

It does not require any options 

```sh
$ chainloop integration attached add --workflow 5e34b34f-882c-48b8-84c0-4f238e15a5fd --integration d19dabf0-d6dd-4f00-91c0-eed209454d05

┌──────────────────────────────────────┬──────┬───────────────────┬─────────────────────┬───────────────┐
│ ID                                   │ KIND │ CONFIG            │ ATTACHED AT         │ WORKFLOW      │
├──────────────────────────────────────┼──────┼───────────────────┼─────────────────────┼───────────────┤
│ 405b3c39-8654-4a3e-b048-d8e618d9c813 │ guac │ bucket: test-guac │ 28 Jun 23 14:24 UTC │ foo/only-sbom │
│                                      │      │ provider: gcs     │                     │               │
└──────────────────────────────────────┴──────┴───────────────────┴─────────────────────┴───────────────┘
```

## Execution

* SBOMs and DSSE envelopes will be uploaded named after their content digest to prevent file collisions under the chainloop path.
* Such files will contain additional metadata for traceability later on.

image

## Demo

guac

Closes #209


app/controlplane/internal/dispatcher/dispatcher.go
```diff
@@ -184,8 +187,20 @@ func (d *FanOutDispatcher) loadInputs(ctx context.Context, queue dispatchQueue,
 		return fmt.Errorf("extracting predicate: %w", err)
 	}
 
+	// Calculate the attestation hash
+	jsonAtt, err := json.Marshal(att)
+	if err != nil {
+		return fmt.Errorf("marshaling attestation: %w", err)
+	}
+
+	h, _, err := crv1.SHA256(bytes.NewBuffer(jsonAtt))
```
danlishka, line 196: why  google/go-containerregistry here not crypto/sha256?

app/controlplane/plugins/core/guac/v1/README.md
```diff
@@ -0,0 +1,96 @@
+# GUAC integration
+
+Graph for Understanding Artifact Composition ([GUAC](https://github.com/guacsec/guac)) aggregates software security metadata into a high fidelity graph database—normalizing entity identities and mapping standard relationships between them. Querying this graph can drive higher-level organizational outcomes such as audit, policy, risk management, and even developer assistance.
+
+This integration allows Chainloop users to automatically send attestation (DSSE envelopes / in-toto statements) and CycloneDX/SPDX Software Bill Of Materials (SBOMs) to a cloud storage bucket staging area. From there, GUAC can be configured to continuously monitor and inject that data.
```
danlishka, line 5: Should we add the supported input formats section? Could our readme generator do it? I think we have a few on the list to add sooner than later in chainloop, like scorecards: https://github.com/guacsec/guac/tree/main/pkg/handler/processor

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+			Description: "Export Attestation and SBOMs metadata to a blob storage backend so guacsec/guac can consume it",
+			Logger:      l,
+			InputSchema: &sdk.InputSchema{
+				Registration: registrationRequest{},
+				Attachment:   attachmentRequest{},
+			},
+		},
+		// This plugin subscribes to SBOMs and attestations (note: attestations come by default)
+		// In the future the kinds will expand to match
+		// https://github.com/guacsec/guac#supported-input-formats
```
danlishka, line 74: What if I push to GCS something what is not supported by guac? Will it be ignored? Will it make it fail?

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+	Bucket   string `json:"bucket"`
+	Provider string `json:"provider"`
+}
+
+func New(l log.Logger) (sdk.FanOut, error) {
+	base, err := sdk.NewFanOut(
+		&sdk.NewParams{
+			ID:          "guac",
+			Version:     "1.0",
+			Description: "Export Attestation and SBOMs metadata to a blob storage backend so guacsec/guac can consume it",
```
danlishka, line 65: Should we add here that gcs is the only one supported today and ask to create an issue if one of unsupported is needed?

app/controlplane/plugins/core/guac/v1/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+![GUAC integration](./img/overview.png)
+
+Currently, the integration only supports [Google Cloud Storage](https://cloud.google.com/storage) (GCS) as a storage provider.
+
+## How to use it
+
+### Requirements
```
danlishka, line 13: - I would make requirements generic, "you need a writable cloud storage bucket"
- I would add another section: supported cloud storage platforms
- I would move GCP specific info (like service account, json creds file) to a separate GCP section

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// 1 - Upload the attestation
+	envelopeJSON, err := json.Marshal(req.Input.Attestation.Envelope)
+	if err != nil {
+		return fmt.Errorf("marshalling attestation: %w", err)
+	}
+
+	// chainloop-deadbeef-attestation.json
+	fileName := fmt.Sprintf("%s-%s-attestation.json", filePrefix, req.Input.Attestation.Hash.Hex)
```
danlishka, line 155: maybe `%s-attestation-%s.json`?

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+	// 2 - Upload all the materials, in our case they are SBOMs
+	for _, sbom := range req.Input.Materials {
+		// Validate that in fact they are SBOMs
+		if err := validateExecuteOpts(sbom, req.RegistrationInfo, req.AttachmentInfo); err != nil {
+			i.Logger.Warnw("msg", "invalid material received", "err", err)
+			continue
+		}
+
+		// chainloop-deadbeef-sbom.json
+		fileName := fmt.Sprintf("%s-%s-%s", filePrefix, sbom.Hash.Hex, sbom.Value)
```
danlishka, line 169: what if we do something like that here:
-  `fmt.Sprintf("%s-%s-%s", filePrefix, sbom.Value, sbom.Hash.Hex)`
- or `fmt.Sprintf("%s-%s-%s.%s", filePrefix, base(sbom.Value), sbom.Hash.Hex, rxt(sbom.Value))`

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Set metadata and content type
+	// application/json can't be detected automatically by https://mimesniff.spec.whatwg.org/#supplied-mime-type-detection-algorithm
+	// If not set, the underlying library will try to sniff it automatically
+	if filepath.Ext(fileName) == ".json" {
+		w.ObjectAttrs.ContentType = "application/json"
+	}
+
+	w.ObjectAttrs.Metadata = map[string]string{
+		"author":          "chainloop",
```
danlishka, line 202: chainloop.dev

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+	return b, nil
+}
+
+// validateExecuteOpts validates that we receive all the required options, from materials to credentials
+func validateExecuteOpts(m *sdk.ExecuteMaterial, regConfig *sdk.RegistrationResponse, _ *sdk.AttachmentResponse) error {
+	if m == nil || m.Content == nil {
+		return errors.New("invalid input")
+	}
+
+	if m.Type != schemaapi.CraftingSchema_Material_SBOM_CYCLONEDX_JSON.String() && m.Type != schemaapi.CraftingSchema_Material_SBOM_SPDX_JSON.String() {
```
danlishka, line 255: Can we somehow automatically generate this validation? We are already providing the list of accepted material types above.

app/controlplane/internal/dispatcher/dispatcher.go
migmartri: Good question. I am using the google container registry one because we use it in many other places and it has this handy helper. 

That said, I might change it since it's true that it can feel out of place. Thanks!

app/controlplane/plugins/core/guac/v1/README.md
migmartri: Good idea! Maybe as part of this? https://github.com/chainloop-dev/chainloop/issues/193

If so. Would you mind adding a comment there?

Thanks!

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+
+	i.Logger.Info("execution finished")
+	return nil
+}
+
+// uploadToBucket uploads the provided content to the bucket under a deterministic, unique name (digest + filename)
+// It also sets the content type and additional metadata
+func uploadToBucket(ctx context.Context, bucket *storage.BucketHandle, fileName string, content []byte, md *sdk.ChainloopMetadata, logger *log.Helper) error {
```
danlishka, line 181: I this enough to create a GCS plugin?

app/controlplane/plugins/core/guac/v1/README.md
migmartri: Not sure, if feels like an early optimization. Note that right above it I explicitly mention that only GCP is supported. 

Let me think about it, it will definitely make sense if/when we add other backend providers.

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: I'd not add it here, the limitation is defined in the readme and in the schema.

> ask to create an issue if one of unsupported is needed?

Create an issue where? In Guac?

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: For what I can see their parser will fail and ignore the file. In fact that's what will happen with our attestation today since we have a custom in-toto predicate.

But I'd say that's not our scope, we are sending content that's technically supported by Guac, and that system should be smart enough to pick up the pieces that understand and ignore the rest. 

Otherwise we'll not be able to cater multiple versions of Guac, i.e Guac that supports newer versions or even new material types.

WDYT?

app/controlplane/plugins/core/guac/v1/guac.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// 1 - Upload the attestation
+	envelopeJSON, err := json.Marshal(req.Input.Attestation.Envelope)
+	if err != nil {
+		return fmt.Errorf("marshalling attestation: %w", err)
+	}
+
+	// chainloop-deadbeef-attestation.json
+	fileName := fmt.Sprintf("%s-%s-attestation.json", filePrefix, req.Input.Attestation.Hash.Hex)
+	if err := uploadToBucket(ctx, bucket, fileName, envelopeJSON, req.ChainloopMetadata, i.Logger); err != nil {
```
danlishka, line 156: Will Guac be able to process our attestations? Is it fully compatible?

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: I was treating the prefix+digest like a prefix but it's true that I don't love it. 

I'll take a look to maybe [prefix]-[filename (without extension)]-[digest].[extension]

I should probably change the prefix and make it a directory, wdyt?

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: > fmt.Sprintf("%s-%s-%s", filePrefix, sbom.Value, sbom.Hash.Hex)

It will make it look like `chainloop-sbom.json-deadbeef)`

migmartri: > or fmt.Sprintf("%s-%s-%s.%s", filePrefix, base(sbom.Value), sbom.Hash.Hex, rxt(sbom.Value))

If I understand correctly, what you are implicitly saying is that you'd prefer a different format on the filename that includes the original filename at the front, followed by its digest correct?

And yes, I'll take a look at it.

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: I though about it and decided not to add the domain on purpose but instead the project name. I could change it if you think it's much better.

migmartri:
Great review @danlishka! I'll tackle most of your comments! Stay tuned!

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: That's a good point, I don't like it either. As you said, this validation is kind of pointless. The execute method will contain only once of those material types.

I am going to remove this code.

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: Pretty much

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: It is not. It will skip them for now. We should work with them and or the in-toto team to figure out if we can either contribute or adopt a standard predicate type. That effort is tracked here https://github.com/chainloop-dev/chainloop/issues/60

app/controlplane/plugins/core/guac/v1/README.md
danlishka: I just do not want people to assume that this is the default option and show that more storage options are coming in the future.

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: > Create an issue where? In Guac?

In chainloop. First, we have to support it

app/controlplane/internal/dispatcher/dispatcher.go
migmartri: Actually, I just realized why I chose it. 

- To be consistent with the format of the Digest that we sent in the materials already and
- To have a decoded way to represent both the Hex and the algorithm part, separately. 

About the later, I've had many issues representing digests as strings, since sometimes they contained the algorithm and some other times they didn't.

I'd prefer then to keep it like this where the consumer of the digest can decide if they want to use one or two parts of the hash. WDYT?

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: I am not sure. Let's keep it as it is for now and then work on PR to have our chainloop in-toto predicate supported in Guac?

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: I see, in that case that should go in the readme IMO. I'll update it.

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: Should we add somewhere the version of Guac supported by this version of our plugin?

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: make sense to make a prefix a directory? We can also keep attestations and materials in separate ones

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: I think this is better and we make sure it is unique.

app/controlplane/internal/dispatcher/dispatcher.go
danlishka: Should we add a comment then so we do not forget why? :)

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: I am not sure anymore :) Any naming conventions people follow?
```
chainloop-deadbadadadasdasdasdadasdeef-sbom.json
chainloop-sbom-deadbadadadasdasdasdadasdeef.json
chainloop/sbom.json/deadbadadadasdasdasdadasdeef
chainloop-sbom.json-deadbadadadasdasdasdadasdeef
```

migmartri: Updated, ptal 

image

app/controlplane/plugins/core/discord-webhook/v1/README.md
```diff
@@ -50,16 +50,4 @@ chainloop integration attached add --workflow $WID --integration $IID
     "webhook"
   ]
 }
-```
-
-## Attachment Input Schema
```
migmartri, line 55: This is because I fixed the generator

app/controlplane/plugins/core/guac/v1/guac.go
migmartri: Done in the readme, v0.1 beta+

app/controlplane/plugins/core/guac/v1/guac.go
danlishka: #212

danlishka (approved):
LGTM!

PR #215, opened 2023-07-03
feat(guac): add video to guide

migmartri:
Updates guac readme guide to include video and follow the same styleguide and content in docs.chainloop.dev

refs #209


danlishka (approved):
perfect!

PR #225, opened 2023-07-08
feat(deployment): Add support for GCP secret manager to Helm charts

gr0:
A small PR adding support for the GCP secret manager to the Helm Charts. 

refs #108


deployment/chainloop/README.md
```diff
@@ -89,6 +89,19 @@ helm install [RELEASE_NAME] oci://ghcr.io/chainloop-dev/charts/chainloop \
     # ...
 ```
 
+Deploy using GCP secret manager instead of Vault
+
+```console
+helm install [RELEASE_NAME] oci://ghcr.io/chainloop-dev/charts/chainloop \
+    # Open ID Connect (OIDC)
+    # ...
+    # Secrets backend
```
migmartri, line 98: This example in-fact is missing an additional flag
```suggestion
    # Secrets backend
    --set secretsBackend.backend=gcpSecretManager \
```

in fact this is a problem in the aws example too, I just recorded it in this issue #226

deployment/chainloop/templates/_helpers.tpl
```diff
@@ -77,6 +77,12 @@ awsSecretManager:
   creds:
     accessKey: {{ required "access key required" .awsSecretManager.accessKey | quote }}
     secretKey: {{ required "secret key required" .awsSecretManager.secretKey | quote }}
+
+{{- else if eq .backend "gcpSecretManager" }}
```
migmartri, line 81: I believe this is missing a parent key, note how the other sections have `vault:` and `awsSecretManager:` in your case you are missing `gcp_secret_manager` or `gcpSecretManager` both would work I think.

note that I noticed this by rendering the chart before and after your change

1 - first I templated the chart with default settings and no changes and stored it in /tmp/before.yaml
2 - I applied your changes and updated a values.local.yaml file to point to gcpSecretManager and run the following command.
```
helm template test ./deployment/chainloop --values /tmp/values.local.yaml > /tmp/after.yaml && diff -u /tmp/before.yaml /tmp/after.yaml
```

```diff
 stringData:
   config.secret.yaml: |
     data:
@@ -128,22 +127,21 @@
         source: postgresql://chainloop:chainlooppwd@test-postgresql:5432/chainloop-cp
 
     credentials_service:      
-      vault:
         secretPrefix: "chainloop"
-        address: "http://test-vault:8200"
-        token: "notapassword"
+        projectId: "projectID"
+        authKey: "authkey"
```

and notice how the config is missing a namespace key.

migmartri (commented):
Thanks for the PR! Looking good.

Summarizing

a) take a look at the possible missing yaml key and 
b) update the example to include the missing `--set` stanza
c) bump the chart version to 1.7.1 https://github.com/chainloop-dev/chainloop/blob/bfd154b36a47ac858d4934bcdffad6f89c245d73/deployment/chainloop/Chart.yaml#L7. That will re-package the chart as a new version once merged.

Thanks!

deployment/chainloop/README.md
gr0: Good point, thanks a lot for noticing that.

gr0:
@migmartri addressed the comments, have a look when you have time, thanks!

PR #227, opened 2023-07-10
fix(cas): return 404 error if artifact does not exist

migmartri:
Handles the fact that an image doesn't exist in the CAS instead of returning a generic error and notifying the exception manager. 

This patch also includes some makefile refactoring to check that some of the new tools required are in place.

### DIFF
--- a/app/artifact-cas/internal/service/resource.go
+++ b/app/artifact-cas/internal/service/resource.go
@@ -21,6 +21,7 @@ import (
 	v1 "github.com/chainloop-dev/chainloop/app/artifact-cas/api/cas/v1"
 	backend "github.com/chainloop-dev/chainloop/internal/blobmanager"
 	sl "github.com/chainloop-dev/chainloop/internal/servicelogger"
+	"github.com/go-openapi/errors"
 )
 
 type ResourceService struct {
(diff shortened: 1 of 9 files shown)


app/artifact-cas/internal/service/resource.go
```diff
@@ -41,13 +42,15 @@ func (s *ResourceService) Describe(ctx context.Context, req *v1.ResourceServiceD
...
-	backend, err := s.backendP.FromCredentials(ctx, info.StoredSecretID)
+	b, err := s.backendP.FromCredentials(ctx, info.StoredSecretID)
 	if err != nil {
 		return nil, sl.LogAndMaskErr(err, s.log)
 	}
 
-	res, err := backend.Describe(ctx, req.Digest)
-	if err != nil {
+	res, err := b.Describe(ctx, req.Digest)
+	if err != nil && backend.IsNotFound(err) {
```
migmartri, line 51: we now check if the error is not found

internal/blobmanager/oci/backend.go
```diff
@@ -185,6 +187,11 @@ func (b *Backend) Describe(_ context.Context, digest string) (*pb.CASResource, e
 
 	img, err := remote.Image(ref, remote.WithAuthFromKeychain(b.keychain))
 	if err != nil {
+		var e *transport.Error
+		if errors.As(err, &e) && e.StatusCode == http.StatusNotFound {
```
migmartri, line 191: this is the way we bubble up the not found error

danlishka (approved):
LGTM!

PR #231, opened 2023-07-10
fix(release): do not fail if we can not attest a material

migmartri:
Fixes https://github.com/chainloop-dev/chainloop/actions/runs/5507743809/jobs/10038247997

The issue was that now we are also building plugin binaries. Then we iterate over them and try to add them as pieces of evidence. Such pieces of evidence are not in the contract (yet).

This patch basically will do a graceful recover and move on. The verification will happen at `att push` where we make sure the required materials are in place.

### DIFF
--- a/.github/workflows/contracts/releases.cue
+++ b/.github/workflows/contracts/releases.cue
@@ -4,6 +4,7 @@ materials: [
 	{type: "ARTIFACT", name: "cli-linux-amd64", output:           true},
 	{type: "ARTIFACT", name: "control-plane-linux-amd64", output: true},
 	{type: "ARTIFACT", name: "artifact-cas-linux-amd64", output:  true},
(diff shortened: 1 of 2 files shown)


.github/workflows/contracts/releases.cue
```diff
@@ -4,6 +4,7 @@ materials: [
 	{type: "ARTIFACT", name: "cli-linux-amd64", output:           true},
 	{type: "ARTIFACT", name: "control-plane-linux-amd64", output: true},
 	{type: "ARTIFACT", name: "artifact-cas-linux-amd64", output:  true},
+	{type: "ARTIFACT", name: "chainloop-plugin-discord-webhook-linux-amd64"},
```
migmartri, line 7: I'll also apply this change in the contract so we also attest this binary.

PR #236, opened 2023-07-11
chore(controlplane): s/oci_repository/cas_backend/g

migmartri:
As part of #223, we want to enable the controlplane to be able to store other types of CAS_Backend.

- adds migrations to rename current oci_repository DB table to cas_backends.
  - it includes now a `provider` property and a `default` property  
  - This default property is now used to load the default cas backend for the org
  - If the instance already has data in the table, it will keep it and mark is as `OCI`  
- Updates the data and biz layer to meet that new DB schema
- Deprecates the oci_repositories API endpoint. It will get replaced by a cas_backends one.

All of that is backwards compatible.

Note for reviewers. This is a big patch but note that most of the code is generated. 

instead ptal at

- `migrations/migrate`
- `biz/cas_backend.go`


app/controlplane/api/buf.gen.yaml
```diff
@@ -16,7 +16,7 @@ plugins:
   - name: validate
     out: .
     opt: paths=source_relative,lang=go
-  - remote: buf.build/stephenh/plugins/ts-proto
+  - plugin: buf.build/community/stephenh-ts-proto:v1.151.1
```
migmartri, line 19: I needed to change this because of https://buf.build/docs/migration-guides/migrate-remote-generation-alpha/

danlishka (approved):
LGTM

PR #237, opened 2023-07-12
feat(controlplane): make cas-backend used during attestation immutable

migmartri:
We are preparing to support multiple cas backends #223. This patch makes sure that an attestation process, recorded as workflow run in the controlplane, maintains a reference to the cas backend(s in the future) through the whole attestation process (init, add and push)

This code maintains backwards compatibility with previous versions of the CLI.

Some highlights

- New many<->many relationship between CAS backends and Workflow Runs
- Association of CAS backend at `init`, and used during the whole att process, for both uploading artifacts and also calling the fanout dispatcher.

refs #223


danlishka (approved):
LGTM!

Issue #201, opened 2023-06-22
Inline CAS backend


migmartri:
Currently, to use Chainloop, OCI registry credentials [must be provided](https://docs.chainloop.dev/getting-started/setup) to setup the [Content Addressable Storage](https://github.com/chainloop-dev/chainloop/tree/main/app/artifact-cas) backend.

The CAS is then used every-time you need to add an artifact type, or uploadable material (SBOM) to an attestation process `chainloop attestation add ...`

This task is about exploring the possibility of of allowing storing the artifact data in the attestation inline by default. This could be easily fit in our current way of [describing materials](https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md). In fact we already store raw content in the string material type.

Having the data inline in the attestations has some drawbacks

- It will make the attestation grow considerably. To control this we could put a sensible limit i.e 10KB, if an user wants more, they can setup an actual CAS backend.    
- The data will be stored in the controlplane DB, again, size consideration is important.

The nice side-effect of this feature is that we will make easier to use chainloop since the setup step will not be required. It will also move us in the direction of #28 so oci-registries today will become optionals and hence fit the extensions framework

PR #240, opened 2023-07-13
feat(controlplane): update cas-backends

migmartri:
Update CAS backend command/API endpoint. 

This command is safer than the previous `upsert` approach in `set-oci-repo`. This update command only allows you to change the `description`, the `default` status and `rotate its credentials`. You can not change the actual location of the repository.

Some features

- if the user wants to remove the `default` status, it will be warned
  
```sh
$ chainloop cas-backend update oci --id 4e015fe8-658b-4773-9275-bd8771927fbf          
This is the default CAS backend and your are setting default=false.
Please confirm to continue y/N: 
```

- You can now technically have no CAS backends enabled, this today will make the controlplane complain about that requirement, but in the future this will not be a requirement at all because of #201 

Refs #223


danlishka (approved):
LGTM

PR #242, opened 2023-07-14
feat(controlplane): soft-delete CAS backends

migmartri:
Soft delete CAS backends with confirmation in the client side if it's the default one.

Refs #223


app/controlplane/internal/service/casbackend.go
```diff
@@ -109,6 +109,25 @@ func (s *CASBackendService) Update(ctx context.Context, req *pb.CASBackendServic
 	return &pb.CASBackendServiceUpdateResponse{Result: bizOCASBackendToPb(res)}, nil
 }
 
+// Delete the CAS backend
+func (s *CASBackendService) Delete(ctx context.Context, req *pb.CASBackendServiceDeleteRequest) (*pb.CASBackendServiceDeleteResponse, error) {
```
danlishka, line 113: Should we then call it soft delete or archive? I am not sure I am following. We give an option to the end user to delete but we soft delete instead. Why don't we call it properly? Am I missing anything?

migmartri: To me, soft-deletion is an implementation detail of the system. The fact that it does not remove it from the DB is not something that the user should know, unless we plan to offer a recovery mechanism.

The reason we do not want to hard delete is because in the future we might need that data to get historical data on where an artifact was placed.

danlishka: ok. It means that once we delete we are not allowing them to download evidences via cli, but attestations visible in workflow runs and maybe you can see references to the deleted backend so you know where to find the a given material?

danlishka (approved):
LGTM for now. Thanks for providing more information

app/controlplane/internal/service/casbackend.go
migmartri: > ok. It means that once we delete we are not allowing them to download evidences via cli, but attestations visible in workflow runs and maybe you can see references to the deleted backend so you know where to find the a given material?

for now yes, but the idea would be that once an user knows an artifact digest, these scenarios could happen:

the artifact with that digest...

1 - is in the default CAS backend
2 - is in a non-default CAS backend
3 - is in a (soft) deleted CAS backend
4 - is in one of the above but credentials are wrong
5 - none of the above

Today, we only support the first option. 2 and 3) will be implemented with a combination of a) digest <-> backend mapping + b) soft-deleted cas backend retrieval.

PR #243, opened 2023-07-14
chore: refactor cas backend providers

migmartri:
Extract validation logic for the CAS providers to the `provider` struct. 

This means that the controlplane code has no specific logic about a given CAS backend (i.e OCI) and adding (and testing) a new backend is easier, similarly decoupled to our plugins system.

refs #223


danlishka (approved):
LGTM

PR #246, opened 2023-07-17
feat: inline CAS backend support in client

migmartri:
Adds support for inline CAS backend in the client. 

This code basically allows passing a `CASBackend` object to the upload logic, this new struct contains the `name` of the backend, the Max upload size and an optional uploader.

If no uploader is provided, the logic will opt-in for a `inline` store of the content. Within the size validation constraints.

This code does not change anything from the current behavior, next, we'll create a patch to enable it based on the settings in the backend.

Refs #201


app/controlplane/internal/service/attestation.go
```diff
@@ -319,18 +324,23 @@ func extractEnvVariables(in map[string]string) []*cpAPI.AttestationItem_EnvVaria
...
+func extractMaterials(in []*chainloop.NormalizedMaterial) ([]*cpAPI.AttestationItem_Material, error) {
 	res := make([]*cpAPI.AttestationItem_Material, 0, len(in))
 	for _, m := range in {
 		// Initialize simply with the value
 		displayValue := m.Value
 		// Override if there is a hash attached
 		if m.Hash != nil {
-			displayValue = fmt.Sprintf("%s@%s", m.Value, m.Hash)
+			name := m.Value
+			if m.EmbeddedInline || m.UploadedToCAS {
```
danlishka, line 335: I think this deserve a separate helper function like is_downloadable.

migmartri: yeah, the problem is that embedded doesn't mean that can be downloaded.

In fact artifact download doesn't work with embedded (inline) artifacts. We'd need to have an artifact-attestation mapping to make it work.

app/cli/internal/action/attestation_add.go
```diff
@@ -81,7 +81,21 @@ func (action *AttestationAdd) Run(k, v string) error {
...
 
-	if err := action.c.AddMaterial(k, v, cc); err != nil {
+	// // TODO: retrieve this information from backend
+	backend := &casclient.CASBackend{
+		Uploader: cc,
+		Name:     "OCI",
+		// 100MB max size
+		MaxSize: 100 * 1024 * 1024,
+	}
+
```
danlishka, line 91: Do we need this comment here?

migmartri: What comment? the 100MB? or the commented out inline backend?

app/cli/internal/action/attestation_add.go
```diff
@@ -81,7 +81,21 @@ func (action *AttestationAdd) Run(k, v string) error {
 
 	cc := casclient.New(artifactCASConn, casclient.WithLogger(action.Logger))
 
-	if err := action.c.AddMaterial(k, v, cc); err != nil {
+	// // TODO: retrieve this information from backend
+	backend := &casclient.CASBackend{
+		Uploader: cc,
+		Name:     "OCI",
+		// 100MB max size
+		MaxSize: 100 * 1024 * 1024,
```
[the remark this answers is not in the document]
migmartri: Agreed, note that this is changing in https://github.com/chainloop-dev/chainloop/pull/247

```
 go run main.go --insecure cas-backend ls                                                                                    
┌──────────────────────────────────────┬─────────────────────────────────────┬──────────┬─────────────────────────────────────┬───────────────┬─────────┐     │ ID                                   │ LOCATION                            │ PROVIDER │ DESCRIPTION                         │ LIMITS        │ DEFAULT │
├──────────────────────────────────────┼─────────────────────────────────────┼──────────┼─────────────────────────────────────┼───────────────┼─────────┤     │ 36c9b3c2-6946-4972-a832-f0137d7646d7 │ us-central1-docker.pkg.dev/axiomati │ OCI      │ fopoobar                            │ MaxSize: 100M │ false   │
│                                      │ c-grove-366622/chainloop-demo-org   │          │                                     │               │         │     ├──────────────────────────────────────┼─────────────────────────────────────┼──────────┼─────────────────────────────────────┼───────────────┼─────────┤
│ f99fbbae-2a7a-4be3-94bf-1850bfa4b8c2 │                                     │ INLINE   │ Embed artifacts content in the atte │ MaxSize: 500K │ true    │     │                                      │                                     │          │ station                             │               │         │
└──────────────────────────────────────┴─────────────────────────────────────┴──────────┴─────────────────────────────────────┴───────────────┴─────────┘     
```

app/controlplane/internal/service/attestation.go
danlishka: I mean, we can call it differently but we already use this "if m.EmbeddedInline || m.UploadedToCAS" in two places.

app/cli/internal/action/attestation_add.go
danlishka: the commented one about 500KB

migmartri: It will be gone soon.

danlishka (commented):
Added a few comments but in general LGTM!

app/cli/internal/action/attestation_add.go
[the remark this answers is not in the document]
migmartri: re: reducing max size. Until we support changing it I'd not go lower than what we have today. We'd need to check the size of all our binaries and such to make sure they fit.

Issue #195, opened 2023-06-21
[plugins] decouple framework from core


migmartri:
Currently we offer a way to extend Chainloop via a [`core plugins` SDK](./app/controlplane/plugins). 

This mechanism has the limitation that it requires the plugin to be added to Chainloop's Control Plane source code and released with it. In other words, you can not develop a plugin on your end, and test it out in an existing instance of Chainloop.

We want to change that as next step, we'd like to

- Allow users to develop plugins on their own outside of Chainloops source code.
- Being able to dynamically load such plugins by dropping them in a configurable directory.

To implement this, there are different options, but the one we've aiming to implement in our original design is the use of https://github.com/hashicorp/go-plugin

```[tasklist]
### Tasks
- [x] Evaluate what changes are required to be done in the SDK and existing plugins
- [x] Move one plugin to the new mode and test it. This will require us to have a separate build-system for the plugin and add them to the final container image
- [x] bundle dependency-track plugin
- [x] bundle discord plugin
- [x] bundle smtp plugin
- [ ] bundle guac plugin
- [ ] bundle oci-registry
- [ ] bundle slack-plugin
```

migmartri:
Helm Chart version 1.8.1 contains Discord enabled as a plugin. For next release we can migrate the rest of plugins.

PR #251, opened 2023-07-19
chore(plugins): enable smtp and dependency-track

migmartri:
Package `dependency-track` and `smtp` plugins in the upstream container image.

Refs #195


danlishka (approved):
LGTM

PR #253, opened 2023-07-20
feat(cli): robot account automatically created during the workflow creation

### LINKED ISSUE #90: Create first robot-account during workflow creation
Usuallym the steps required to create and use a workflow are

Create a workflow `chainloop wf create`

```sh
$ chainloop wf create --name test --project test
┌──────────────────────────────────────┬──────┬─────────┬─────────────────────┬────────┬─────────────────┐
│ ID                                   │ NAME │ PROJECT │ CREATED AT          │ RUNNER │ LAST RUN STATUS │
├──────────────────────────────────────┼──────┼─────────┼─────────────────────┼────────┼─────────────────┤
│ 71de5436-ef9e-4f27-99c6-a67826f7c0a5 │ test │ test    │ 24 Apr 23 14:09 UTC │        │                 │
└──────────────────────────────────────┴──────┴─────────┴─────────────────────┴────────┴─────────────────┘
```
Create a robot account pointing to that workflow ID

```sh
$ chainloop wf robot-account create --workflow 71de5436-ef9e-4f27-99c6-a67826f7c0a5
┌──────────────────────────────────────┬──────┬──────────────────────────────────────┬─────────────────────┬────────────┐
│ ID                                   │ NAME │ WORKFLOW ID                          │ CREATED AT          │ REVOKED AT │
├──────────────────────────────────────┼──────┼──────────────────────────────────────┼─────────────────────┼────────────┤
│ f1f5cc7d-c7eb-4e84-ae45-3b1437d062a6 │      │ 71de5436-ef9e-4f27-99c6-a67826f7c0a5 │ 24 Apr 23 14:09 UTC │            │
└──────────────────────────────────────┴──────┴──────────────────────────────────────┴─────────────────────┴────────────┘

Save the following token since it will not printed again: 

 eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.REDACTED-I
```

This task is about **by default**, create a token as well, unless an optional flag is set that prevents that, i.e `--skip-robot-account-create`

We need to decide where to implement this, server-side or client-side. My gut tells me, based on the limited information we have, that there is no enough data to corroborate that such a change should be implemented by default in the server. So I am leaning towards a client side implementation of it (although automatic contract creation happens server side)

danlishka:
There is an option to skip the creation of the Robot Account `--skip-robot-account-create`
Screenshot 2023-07-20 at 11 31 19

When you create a workflow with the `output` parameter set to `json`. There are new fields: `robotAccountKey` and `robotAccountID`.
Screenshot 2023-07-20 at 11 32 14

This is how the table look like with the default `table` view:
Screenshot 2023-07-20 at 11 32 54

Closes #90


app/cli/internal/action/workflow_list.go
```diff
@@ -28,14 +28,16 @@ type WorkflowList struct {
...
-	LastRun    *WorkflowRunItem `json:"lastRun,omitempty"`
+	Name            string           `json:"name"`
+	ID              string           `json:"id"`
+	Team            string           `json:"team"`
+	Project         string           `json:"project,omitempty"`
+	CreatedAt       *time.Time       `json:"createdAt"`
+	RunsCount       int32            `json:"runsCount"`
+	ContractID      string           `json:"contractID,omitempty"`
+	LastRun         *WorkflowRunItem `json:"lastRun,omitempty"`
+	RobotAccountID  string           `json:"robotAccountID,omitempty"`
```
migmartri, line 39: I am not sure this is the best place to put robot account properties. 

This struct is in some way a translation between the server object and the client one, and there is an issues with this approach now. A workflow returned by the server will never fill those two fields, in fact those fields are set by the client only on creation. 

What about rendering the message outside in the cmd/workflow_create.go instead? See my comment there

app/cli/cmd/workflow_create.go
```diff
@@ -64,6 +65,15 @@ func newWorkflowCreateCmd() *cobra.Command {
...
+			// If the robot account creation is not skipped via --skip-robot-account-create , create a new one
+			if !skipRACreate {
+				resRA, errRA := action.NewWorkflowRobotAccountCreate(actionOpts).Run(res.ID, "default")
+				if errRA != nil {
+					return errRA
+				}
+				res.RobotAccountID = resRA.ID
+				res.RobotAccountKey = resRA.Key
+			}
 			return encodeOutput([]*action.WorkflowItem{res}, WorkflowListTableOutput)
```
migmartri, line 77: See my comment above of why I think adding extra properties to the struct might not be the best way, what about rendering the robot account here instead using the already defined output?

something like

```diff
-                       return encodeOutput([]*action.WorkflowItem{res}, WorkflowListTableOutput)
+                       if err := encodeOutput([]*action.WorkflowItem{res}, WorkflowListTableOutput); err != nil {
+                               return err
+                       }
+
+                       // Check that we are rendering the table so we don't break the json output
+                       if flagOutputFormat == formatTable {
+                               return encodeOutput([]*action.WorkflowRobotAccountItem{robotAccount}, robotAccountListTableOutput)
+                       }
+
+                       return nil
                },

```

migmartri (commented):
Awesome!

I added two comments, mainly about 

* Making sure that we do not pollute the workflow struct for just rendering at creation
* Maybe return the default output for the robot-account

WDYT?

app/cli/internal/action/workflow_list.go
danlishka: OK. I was just trying to reuse as much as I can what we already have, especially that ContractID is there as well. The problem I see now is that all these fields are probably present in the workflow db table and RobotAccount does not really fit there because it is 1-to-many. I will change it.

migmartri: > OK. I was just trying to reuse as much as I can what we already have

Yes, that makes sense. But in fact my proposal requires less changes, you'll just need to just update one file (I think), `cmd/workflow_create.go`

danlishka: But your proposal does not allow me to automate it because in -o json, I will not be able to get the token. Any ideas?

migmartri: I see, in my head the scope of this was to improve the default user console experience.

Two options, either skipping the creation of the token if you are not in table mode, or if we think it's important to also support json output in one go, go back to your approach. Either way, you choose, i'll approve it.

migmartri (approved):

app/cli/internal/action/workflow_list.go
migmartri: > Or you propose to return:
> 
> * a RA json object if we autocreate RA
> * and workflow json for the current behaviour?
> 
> I did not like this approach because this is confusing and not the best UX

No, my proposal would be to not to auto-create robot-account if you are not in `table` mode. Potentially showing a warning with the logger that the ra will not be created.

But your initial proposal is fine if we want to support everything, maybe adding some comments with the caveats of that output (only works on creation and does not represent the actual set of robot-accounts associated)

danlishka: > either skipping the creation of the token if you are not in table mode

This is important, because this is confusing. How will we explain it? We have more parameters: skip-ra-create(true or false) and output (json or table) -> this makes four different scenarios

migmartri: A "notice" would suffice imo. "robot account creation skipped during json output"

danlishka: Then I would prefer skip this for now and just ask the user to do `chainloop workflow robot-account create`. We are going to review the cli UX soon anyway. What do you think?

app/cli/cmd/workflow_create.go
```diff
@@ -79,6 +109,10 @@ func newWorkflowCreateCmd() *cobra.Command {
 	cmd.Flags().StringVar(&team, "team", "", "team name")
 	cmd.Flags().StringVar(&contract, "contract", "", "the ID of an existing contract or the path/URL to a contract file. If not provided an empty one will be created.")
 
+	cmd.Flags().BoolVarP(&skipRACreate, "skip-robot-account-create", "s", false, "Skip creating a Robot Account for this workflow.")
+
+	cmd.Flags().SortFlags = false
```
migmartri, line 114: I didn't know about this option. Why did you disable it?

danlishka: because if you don't, this is sorted and some options are shown below skip-robot...

PR #270, opened 2023-07-25
fix(slack): add empty line before escape lines

migmartri:
We've noticed that some notifications are not rendering properly, and although it's not clear why some other do, we've detected that during the previous refactoring, the new line before the escape ticks were removed.


danlishka (approved):
Thanks

PR #277, opened 2023-07-28
fix(dispatcher): do not fail if credentials are missing

migmartri:
If during the integrations dispatch process one of the integrations can't find the credentials external dependency, it gets skipped instead of failing the execution.


danlishka (approved):
I think this is ok for now but we should build a better defined log + chainloop operator notifications in the future to let the admin or a person who added an integration know

PR #278, opened 2023-07-28
feat: custom annotations

migmartri:
This patch is the first step towards #252, it allows the definition of key/val annotations in the contract for the different materials.

These annotations are then injected in the attestation automatically and shown in the output

For example, having this contract

```json
{
  "schemaVersion":  "v1",
  "materials":  [
    {
      "type":  "SBOM_CYCLONEDX_JSON",
      "name":  "skynet-sbom",
      "annotations": [{ "name": "component", "value": "controlplane" }]
    },
    {
      "type":  "CONTAINER_IMAGE",
      "name":  "image",
      "annotations": [{ "name": "component", "value": "controlplane" }]
    },
    {
      "type":  "STRING",
      "name":  "string"
    }
  ]
}
```

A resulting attestation will look like

```cue
// $ chainloop wf run describe --id 61aba7ed-8430-401d-89a8-39a8ca12744e -o statement | jq .predicate.materials
[
  {
    "annotations": {
      "chainloop.material.name": "image",
      "chainloop.material.type": "CONTAINER_IMAGE",
      "component": "controlplane" // note this custom annotation!
    },
    "digest": {
      "sha256": "97deaf8ad2e31524d6e7eee65468a4330cb55f7dcf4911ebdf2a580a4261c006"
    },
    "name": "ghcr.io/chainloop-dev/chainloop/control-plane"
  },
  {
    "annotations": {
      "chainloop.material.cas": true,
      "chainloop.material.name": "skynet-sbom",
      "chainloop.material.type": "SBOM_CYCLONEDX_JSON",
      "component": "controlplane" 
    },
    "digest": {
      "sha256": "ebe2e6ffab93f4b2e2743d3557bd094427ea059adc78cfe2cd330a72f61731e8"
    },
    "name": "sbom.cp.cyclonedx.json"
  },
  {
    "annotations": {
      "chainloop.material.name": "string",
      "chainloop.material.type": "STRING"
    },
[...]

Issue #252, opened 2023-07-20
Advanced routing context for integrations


migmartri:
Currently, an artifact gets routed to a given attached integration if the artifact is from a known type (i.e CycloneDX SBOM), this rigid mechanism has some limitations.

### Scenario a) Two SBOM materials defined in a contract are sent to the same dependency-track project, 

So for example, if you are attesting two different SBOMs, they are both sent to the project defined during attachment. This means **project override**.

![image](https://github.com/chainloop-dev/chainloop/assets/24523/34f81ad8-35d9-46dd-9e8a-382884379753)

Ideally, we'd like these to be sent to two different projects.

![image](https://github.com/chainloop-dev/chainloop/assets/24523/bb1d155a-b8ee-43c5-9fc5-a79c7e05d3fe)

### Scenario b) Single SBOM material entry, but containing different SBOMs

Let's say for example that you have a single CI job that is syncing a set of upstream assets. This means that you are using the same contract but technically running it multiple times for each asset.

In the example below, we are running the same workflowRun 3 times, but in the context of a different asset. This run is attaching the sbom material and overriding the same dependency-track project, because of the same issue described above.

![image](https://github.com/chainloop-dev/chainloop/assets/24523/f3d9e17e-b820-43e6-88d3-cc2b7f8a9ed5)

Ideally, we should be able to even though we are re-using the same material entry, attach it some context that then can be used for the final routing. In other words, choose the dependency-track projects based on the asset we are processing.

![image](https://github.com/chainloop-dev/chainloop/assets/24523/c565501b-c5f8-420d-81f9-a13c0bbf2e49)

## Solution

There are two three new implementation concepts relevant to this problem `annotations`, `templating` and `filtering`

### Annotations

Annotations are a set of arbitrary key/values associated with a given material. These annotations will

- Be defined in the contract
- Might have a default value
- Could be set during attestation, i.e `attestation add --name ... --value ... --annotations asset=nginx`

### Templating

Currently, when you attach an integration, i.e dependency-track, you can choose the a project name that will be automatically created, for example

```sh
$ chainloop integration attached add ... --opt projectName=bitnami-asset 
``` 

We could leverage the new annotations to in practice expand a dynamic project name, for example

```sh
$ chainloop integration attached add ... --opt projectName="bitnami-asset-{{.Annotations.Asset}}"
``` 

[...]

PR #284, opened 2023-07-31
feat: global annotations support

migmartri:
Global annotations support.

Similarly to the annotations attached to a material #278, now you can a) define a list of optionally set of annotations at the attestation level. These annotations then will be injected into the resulting attestation and in another patch, exposed to plugins such as the notification ones.

- Annotations that don't have a value, will require the user performing the attestation to inject them during `att push` command.

So for example, having the following schema with an annotation called `asset` already pre-populated and another one called `version` without value.

```json
{
        "schemaVersion":  "v1",
        "annotations":  [
                {
                        "name":  "asset",
                        "value":  "chainloop-oss"
                },
                {
                        "name":  "version"
                }
        ],
        "materials":  [
                {
                        "type":  "SBOM_CYCLONEDX_JSON",
                        "name":  "controlplane-sbom",
```

During the attestation process, when we init we'll see

```bash
$ chainloop att init
│ Initialized At    │ 31 Jul 23 09:16 UTC                  │
│ Workflow          │ 3e4ef02c-0f33-4248-84be-8a55ea2e3cb0 │
│ Name              │ only-sbom                            │
│ Project           │ foo                                  │
│ Contract Revision │ 20                                   │
│ Annotations       │ ------                               │
│                   │ asset: chainloop-oss                 │
│                   │ version: [NOT SET]                   │
```

then the annotation can be provided using the `attestation push` command 

```
chainloop att push --annotation version=123
```

Refs #252


danlishka (approved):
Perfect!  Looking forward to seeing annotations in production.

PR #285, opened 2023-07-31
feat(dependency-track): support global annotations in project name

migmartri:
Allows leveraging global annotations in the dependency-track project name.

For example, let's assume we have the following contract with both material (called component) and global annotations (branch)

```json
cat /tmp/schema-old.json                        
{
        "schemaVersion":  "v1",
        "annotations":  [
                {
                        "name":  "branch",
                        "value": "stable"
                }
        ],
        "materials":  [
                {
                        "type":  "SBOM_CYCLONEDX_JSON",
                        "name":  "controlplane-sbom",
                        "annotations":  [
                                {
                                        "name":  "component",
                                        "value":  "controlplane"
                                }
                        ]
                },
                {
                        "type":  "SBOM_CYCLONEDX_JSON",
                        "name":  "cas-sbom",
                        "annotations":  [
                                {
                                        "name":  "component",
                                        "value":  "cas"
                                }
                        ]
                }
        ]
}
```

You can now attach the dependency-track instance to leverage both types of annotations

```bash
$ chainloop integration attached add .... 
--opt projectName="{{.Material.Annotations.component}}-{{ .Attestation.Annotations.branch }}"  
```

This will push the SBOMs to different projects based on those annotations.

image

### DIFF
--- a/app/controlplane/plugins/core/dependency-track/v1/extension.go
+++ b/app/controlplane/plugins/core/dependency-track/v1/extension.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"text/template"
 
 	schemaapi "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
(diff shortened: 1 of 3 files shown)


danlishka (approved):
LGTM

PR #289, opened 2023-07-31
feat: show annotations in notifications

### LINKED ISSUE #252: Advanced routing context for integrations

migmartri:
- Shows annotations (both global and material based) and new table format in the notifications
- Shows global annotations in chainloop describe

Closes #252


danlishka (approved):
LGTM

PR #292, opened 2023-08-01
feat(slack): truncate message

### LINKED ISSUE #272: Slack plugin gets truncated at 4000 chars
When the content goes over that 4k threshold, it gets split into another message which breaks the escape sequence.

During the rest call, we are currently using the `text` property but there seems to be a way to enrich messages via `blocks` (which seems to be the successor of attachments).

We should take a look at if using a block instead of the fallback text would fix the issue and update the plugin accordingly.

migmartri:
Add render option to truncate the output table to a number of runes. A 4000 chars limit is then set to the Slack message to prevent #272 

The truncate is slightly smart since it leaves the footer where the command to get the attestation info stays intact.

image

Note: ideally, eventually, we should do custom markup for Slack using Slack blocks, but this feature at least prevents broken rendering in the case of overflow.

Closes #272

### DIFF
--- a/app/controlplane/plugins/sdk/v1/helpers.go
+++ b/app/controlplane/plugins/sdk/v1/helpers.go
@@ -27,7 +27,6 @@ import (
 
 type renderer struct {
 	render  func(t table.Writer) string
-	format  string
 	maxSize int
 }
 
@@ -52,7 +51,6 @@ func WithFormat(format string) RenderOpt {
 			return fmt.Errorf("unsupported format %s", format)
 		}
 
-		r.format = format
 		return nil
 	}
 }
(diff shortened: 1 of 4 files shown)


app/controlplane/plugins/sdk/v1/helpers.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 type renderer struct {
-	render func(t table.Writer) string
-	format string
+	render  func(t table.Writer) string
+	format  string
```
danlishka, line 30: After offline conversation, should we remove it?

danlishka (approved):
check the `format`, apart from that LGTM

app/controlplane/plugins/sdk/v1/helpers.go
migmartri: Good catch, removed.

Thanks!

PR #294, opened 2023-08-02
feat(cas): download endpoint

### LINKED ISSUE #293: feature: http download endpoint for  items in CAS
Currently, downloads are performed using the CLI (chainloop artifact download) which hits the `Read` RPC endpoint https://github.com/chainloop-dev/chainloop/blob/f6ec84dfc52b82515938ff1a3b51a4f2f32e3f50/app/artifact-cas/internal/service/bytestream.go#L127

This endpoint requires the `digest` of the artifact to download and an authentication with a short lived token with download permissions.

This task is about creating a similar endpoint but in an HTTP endpoint. i.e `download/sha256:deadbeef?t=short_lived_token` 

This will enable download links from notifications, webUIs and so on.

Note that this task is only about implementing the endpoint in the artifact-cas-proxy, not about an additional endpoint that we will need to do in the control-plane which will craft such download token for the user (that will be another task). 
```[tasklist]
### Tasks
- [ ] https://github.com/chainloop-dev/chainloop/pull/294
- [x] feat(deployment): enable http ingress in Helm Chart https://github.com/chainloop-dev/chainloop/pull/305
- [x] Enable new http CAS endpoint in Chainloop cloud
- [x] Download endpoint in control plane https://github.com/chainloop-dev/chainloop/pull/307
- [x] Update Helm Chart to support downloading https://github.com/chainloop-dev/chainloop/pull/310
```

migmartri:
Exposes a new /download/{digest}?t={token} http endpoint that will

- Stream the resulting file to the browser
- Set the size and content-disposition headers

Peek 2023-08-02 02-20

Closes #293


app/artifact-cas/internal/server/grpc.go
```diff
@@ -130,39 +130,48 @@ func jwtAuthFunc(keyFunc jwt.Keyfunc, signingMethod jwt.SigningMethod) grpc_auth
...
-				return nil, jwtMiddleware.ErrTokenExpired
-			}
+		return jwtMiddleware.NewContext(ctx, claims), nil
+	}
+}
 
-			if ve.Errors&(jwt.ValidationErrorNotValidYet) != 0 {
-				return nil, jwtMiddleware.ErrTokenExpired
-			}
+func extractToken(token string, keyFunc jwt.Keyfunc, signingMethod jwt.SigningMethod) (*casJWT.Claims, error) {
```
danlishka, line 142: I would rename this function to something like extractTokenAndVerify - you do way more here than a token extraction.

app/artifact-cas/internal/server/http.go
```diff
@@ -43,8 +51,40 @@ func NewHTTPServer(c *conf.Server, logger log.Logger) *http.Server {
 	if c.Http.Timeout != nil {
 		opts = append(opts, http.Timeout(c.Http.Timeout.AsDuration()))
 	}
+	// Load the key on initialization instead of on every request
+	// TODO: implement jwks endpoint
+	rawKey, err := os.ReadFile(authConf.RobotAccountPublicKeyPath)
```
danlishka, line 56: Please add the comment here about the key pair. This is not meant to be a robot account public key. We use the Robot Account Public Key here for json token verification.

danlishka (approved):
LGTM, please take a look at my comments

Issue #295, opened 2023-08-13
Helm chart deployment using GCP secret manager does not configure authentication properly


zaibon:
The code configuring the authentication to GCP Secret Manager expects the configuration to have a path to a service account file. https://github.com/Archipelo/chainloop-contrib/blob/a039a641688f6a76b1e443751035841b77ac3f13/internal/credentials/gcp/secretmanager.go#L70

But the hem chart does not upload a file, but just the content of secretsBackend.gcpSecretManager.authKey from the `values.yaml `file.
https://github.com/chainloop-dev/chainloop/blob/f6ec84dfc52b82515938ff1a3b51a4f2f32e3f50/deployment/chainloop/templates/_helpers.tpl#L82-L85

This could be fixed by mounting a k8s secret if the `authKey` is set and use the path mounted in the pod.

migmartri:
Hi @zaibon, thanks for reporting the issue. As discussed [here](https://discord.com/channels/1037381970189111326/1037381970637897840/1140352546129580032), what you pointed out is indeed a bug in the Helm Chart. 

We are already looking into it and will target a fix on the next release of the chart. Thanks!

PR #296, opened 2023-08-14
fix(deployment): properly expose service account key for secret manager

zaibon:
We need to mount the service account key as a file into the controlplane container when using GCP secret manager.

related to https://github.com/chainloop-dev/chainloop/issues/295


deployment/chainloop/templates/_helpers.tpl
```diff
@@ -82,7 +82,7 @@ awsSecretManager:
 gcpSecretManager:
   secretPrefix: {{ required "secret prefix required" .secretPrefix | quote }}
   projectId: {{ required "project id required" .gcpSecretManager.projectId | quote }}
-  authKey: {{ required "auth key required" .gcpSecretManager.authKey | quote }}
```
migmartri, line 85: I think we should still keep the `required` stanza to make sure you provide it and we do not create an empty secret no?

in other words, if we have the required here and we do not properly set the key, deployment will fail at template time, not at runtime. 

What do you think?

migmartri (commented):
It looks good!

Just added one comment about keeping the `required` stanza, let me know what you think

deployment/chainloop/templates/_helpers.tpl
gr0: I think that is a good idea - that would allow to fail fast and correct configuration mistakes.

gr0 (approved):
Great stuff @zaibon - thanks for taking care of that 🙇

deployment/chainloop/templates/_helpers.tpl
zaibon: Actually, is it possible to use the `required` function here? The required function accepts an error message and a path to a key. In this case, we just want to check the key has not an empty value, but we want to enforce a static string for the value. Is there a way to do this?

migmartri: You are right, I missed that, `required` does not make sense anymore. 

Maybe we can append something like?

```
{{- if eq .gcpSecretManager.serviceAccountKey "" }}
{{- fail ".Values.gcpSecretManager.serviceAccountKey not set" }}
{{- end }}
```

migmartri (approved):
LGTM! 

See my comment below in case it makes sense.

PR #300, opened 2023-08-15
fix(deployment): include cloud-sql auth proxy in migration job when needed

### LINKED ISSUE #298: The helm chart needs to deploy sql-proxy in migration job when using external database and cloud-sql proxy
When using an external Cloud SQL database and accessing it using [Cloud SQL Auth Proxy](https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine#proxy), the helm chart only deploy the proxy into the pods to the control-plane. 
It is also required in the migration job.

zaibon:
fixes #298


migmartri (approved):
LGTM!

PR #301, opened 2023-08-17
feat(deployment): allow to configure annotations of controlplane service

zaibon:
this is required in GKE to be able to configure custom backend healthcheck on this service.


migmartri:
Thanks for the PR! The CI seems to be complaining, please take a look at the DCO (signed off) requirement, thanks!

migmartri (approved):
LGTM! 

In order to package this change as a new Chart version, please update the `version` in the Chart.yaml. Bumping the patch portion will do the trick :)


https://github.com/chainloop-dev/chainloop/blob/5bde39edbf7d9f678f0344658c984b7ddb1e8e55/deployment/chainloop/Chart.yaml#L7

Thanks!

migmartri (approved):
Lgtm!

PR #303, opened 2023-08-20
feat: allow to configure TLS for gRPC servers

### LINKED ISSUE #302: TLS termination on gRPC server
Currently, TLS termination is happening in the ingress controller, leaving a last networking leg unencrypted (ingress reverse proxy -> k8s pod).

We'd like to add support to configure TLS termination in both control-plane and CAS gRPC servers. As first version the implementation could consist of adding some knobs in the configuration to provide the cert/pk keypair plus updating the gRPC server initialization logic to load these.

Some suggestions and action items that needs to apply for both the controlplane and the CAS

- Update the configuration options to support providing the path for a cert and a pk https://github.com/chainloop-dev/chainloop/blob/5bde39edbf7d9f678f0344658c984b7ddb1e8e55/app/controlplane/internal/conf/conf.proto#L62-L66
- If provided, we should load it potentially as part of this initialization chain https://github.com/chainloop-dev/chainloop/blob/5bde39edbf7d9f678f0344658c984b7ddb1e8e55/app/controlplane/internal/server/grpc.go#L80-L88
- Note that for now I am thinking that this configuration should be optional for compatibility reasons but this might change in the future.

Note: Kudos to @zaibon for raising this issue since it's a blocker to deploy Chainloop in GKE with GKEs built-in ingress controller. In a follow up patch we could discuss how to expose this option in the Helm Chart and for example automatically load the ingress TLS in the pod if needed.

zaibon:
Allow to add path to files containing TLS server certificate and private key for gRPC servers. The files must contain PEM encoded data.

fixes #302


app/artifact-cas/internal/conf/conf.pb.go
```diff
@@ -15,7 +15,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.30.0
+// 	protoc-gen-go v1.31.0
```
zaibon, line 18: Because of the version change of protoc, I have other files that have this kind of modification.
Would you like for me to add all the files that were re-generated with the new version? Or only keep the ones that are relevant to the configuration only?

migmartri: Thanks for asking. I'd keep it like this, meaning just adding to the patch what's relevant, otherwise the review will get too much noise. 

In another patch, we can bump the generated files if we want, wdyt?

app/artifact-cas/internal/conf/conf.proto
```diff
@@ -43,10 +43,16 @@ message Server {
     string addr = 2;
     google.protobuf.Duration timeout = 3;
   }
+  message TLS {
+    // path to certificate and private key
+    string Certificate = 1;
```
migmartri, line 48: this should be lowercase, I believe that the linter didn't pick it up in the CI because there is a bug and `buf lint` doesn't pick the `config` directory.

Note in local how `make lint` or `buf lint app/controlplane/internal/conf` actually should complain. I'll fix the CI issue.

app/artifact-cas/internal/server/grpc.go
```diff
@@ -93,6 +94,18 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
 	if c.Grpc.Timeout != nil {
 		opts = append(opts, grpc.Timeout(c.Grpc.Timeout.AsDuration()))
 	}
+	cert := c.Grpc.TlsConfig.Certificate
```
migmartri, line 97: it might make sense to use the `secure getter` `getTLSConfig.GetCertificate` this will prevent panics if the TLSConfig is not present, which will not in currently deployed instances.

app/artifact-cas/internal/server/grpc.go
```diff
@@ -93,6 +94,18 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
...
+	privKey := c.Grpc.TlsConfig.PrivateKey
+	if cert != "" && privKey != "" {
+		cert, err := tls.LoadX509KeyPair(cert, privKey)
+		if err != nil {
+			return nil, err
```
migmartri, line 102: it might make sense to return wrapped context in the error i.e `fmt.Errorf("loading gRPC server TLS certificate: %w", err)`

app/artifact-cas/internal/server/grpc.go
```diff
@@ -93,6 +94,18 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
...
+		}
+		opts = append(opts, grpc.TLSConfig(&tls.Config{
+			Certificates: []tls.Certificate{cert},
+			MinVersion:   tls.VersionTLS12,
```
migmartri, line 106: out of curiosity, why do you set this? It the default insecure? If so, it might make sense to have a comment.

app/artifact-cas/internal/server/grpc.go
migmartri: or something like `if tlsConf := c.Grpc.GetTLSConfig(); tlsConf != nil` { /load and append of the key / }

app/controlplane/internal/conf/conf.proto
```diff
@@ -59,10 +59,17 @@ message Server {
     google.protobuf.Duration timeout = 5;
   }
 
+  message TLS {
```
migmartri, line 62: It would be great if you could also add an example in the same files (both CAS and CP) https://github.com/chainloop-dev/chainloop/blob/main/app/controlplane/configs/samples/config.yaml

app/controlplane/internal/server/grpc.go
```diff
@@ -71,7 +72,7 @@ type Opts struct {
 }
 
 // NewGRPCServer new a gRPC server.
-func NewGRPCServer(opts *Opts) *grpc.Server {
+func NewGRPCServer(opts *Opts) (*grpc.Server, error) {
```
migmartri, line 75: nice! so you are familiar with wire's dep injection!

migmartri (commented):
Looking good so far! Thanks for the quick turnaround!

app/artifact-cas/internal/conf/conf.pb.go
zaibon: Sounds good 👍

app/artifact-cas/internal/server/grpc.go
zaibon: Yes, when not set, gosec complains about using a insecure minimum version.

app/artifact-cas/internal/conf/conf.proto
zaibon: Thx. I've updated the file.

app/artifact-cas/internal/server/grpc.go
zaibon: Make sense 👍

app/artifact-cas/configs/samples/config.yaml
```diff
@@ -9,6 +9,10 @@ server:
   grpc:
     addr: 0.0.0.0:9001
     timeout: 1s
+    grpc:
```
migmartri, line 12: This line is spurious.

app/artifact-cas/internal/server/grpc.go
```diff
@@ -93,6 +94,21 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
...
+		cert := tlsConf.GetCertificate()
+		privKey := tlsConf.GetPrivateKey()
+		if cert != "" && privKey != "" {
+			cert, err := tls.LoadX509KeyPair(cert, privKey)
+			if err != nil {
+				return nil, fmt.Errorf("loading gRPC server TLS certificate: %w", err)
+			}
+			opts = append(opts, grpc.TLSConfig(&tls.Config{
+				Certificates: []tls.Certificate{cert},
+				MinVersion:   tls.VersionTLS12, // gosec complains about insecure minumum version we use default value
```
migmartri, line 107: ptal at the linter issuer. Thanks!

app/artifact-cas/internal/server/grpc.go
```diff
@@ -93,6 +94,21 @@ func NewGRPCServer(c *conf.Server, authConf *conf.Auth, byteService *service.Byt
 	if c.Grpc.Timeout != nil {
 		opts = append(opts, grpc.Timeout(c.Grpc.Timeout.AsDuration()))
 	}
+	if tlsConf := c.Grpc.GetTlsConfig(); tlsConf != nil {
```
migmartri, line 97: nitpick: I tend to add extra lines between blocks for readability. This would mean to add an extra line before ``if tlsConf := ...`

migmartri (commented):
Awesome. Just fix the lint errors and the example and we are good to go!

app/artifact-cas/internal/server/grpc.go
zaibon: The previous blocks are not spaced though. Do you want to only space this block, or all of them?

migmartri: Good point, either way in that case. Thanks

migmartri (approved):
LGTM!

PR #305, opened 2023-08-21
feat(deployment): expose HTTP server for CAS

migmartri:
Allow to expose the CAS http server via service/ingress. Refs #293


danlishka (approved):
LGTM

PR #307, opened 2023-08-22
feat(controlplane): download endpoint

migmartri:
Refs #293


app/cli/cmd/auth_login.go
```diff
@@ -71,6 +71,7 @@ func interactiveAuth() error {
 	// Append local callback URL
 	q := serverLoginURL.Query()
 	q.Set("callback", callbackURL.String())
+	q.Set("long-lived", "true")
```
migmartri, line 74: note: This change will require users to update their CLIs otherwise their tokens will expire constantly.

I decided to go this way because

- It just affects operators CLI, not the attestation CLI
- Doing the opposite, meaning making long-lived default while short-lived optional is not good practice.
- We are still pre-1.0 so we do not guarantee backwards compatibility.
- I do not think we can detect the version of the CLI yet, so there is no way to show a warning message to the user.

Please let me know if you have any objection on this breaking change.

danlishka, line 74: It would be perfect if we can use queryParamLongLived here and in the service code.

danlishka: Can you please elaborate on "It just affects operators CLI, not the attestation CLI"?
You mean that it does not affect RobotAccount auth based calls?

danlishka (approved):
LGTM

app/cli/cmd/auth_login.go
migmartri: Correct

migmartri: Right, I thought about it, but there is some refactoring that would need to be done. Such as extracting those constants outside of the `internal` package in the controlplane to a shared `parent` module.

Let me revisit it. Thanks

app/controlplane/configs/config.devel.yaml
```diff
@@ -5,7 +5,7 @@
 server:
   http:
     addr: 0.0.0.0:8000
-    timeout: 1s
+    timeout: 10s
```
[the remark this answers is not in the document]
migmartri: We need now a longer timeout since the HTTP endpoint handler now will check if the cas backend is valid in some cases, that takes some time. 

We have currently configured 10 seconds in the similar gRPC endpoints so I thought of using the same.

app/cli/cmd/auth_login.go
migmartri: Updated

PR #319, opened 2023-08-31
feat(deployment): allow to configure TLS certificate for gRPC servers

zaibon:
The helm chart now allows to specify a sercret name where a TLS certificate and key are store. If specified, it mounts the secret in the pods and update the configuration to enable TLS on the servers.

```diff
@@ -180,6 +180,9 @@
         addr: 0.0.0.0:8000
         timeout: 1s
       grpc:
+        tls_config:
+          certificate: /data/server-certs/tls.crt
+          private_key: /data/server-certs/tls.key
         addr: 0.0.0.0:9000
         timeout: 1s
       http_metrics:
@@ -209,6 +212,9 @@
       grpc:
         addr: 0.0.0.0:9000
         timeout: 10s
+        tls_config:
+          certificate: /data/server-certs/tls.crt
+          private_key: /data/server-certs/tls.key
     cas_server:
       grpc:
         addr: foo-chainloop-cas-api:80
@@ -390,6 +396,8 @@
               mountPath: "/tmp"
             - name: gcp-secretmanager-serviceaccountkey
               mountPath: /gcp-secrets
+            - name: server-certs
+              mountPath: /data/server-certs
       volumes:
         - name: config
           projected:
@@ -401,6 +409,9 @@
         - name: jwt-public-key
           secret:
             secretName: foo-chainloop-cas-jwt-public-key
+        - name: server-certs
+          secret:
+            secretName: server_certs
         - name: gcp-secretmanager-serviceaccountkey
           secret:
             secretName: foo-chainloop-controlplane-gcp-secretmanager-serviceaccountkey
@@ -504,6 +515,8 @@
               mountPath: /tmp
             - name: jwt-cas-private-key
               mountPath: /secrets
+            - name: server-certs
+              mountPath: /data/server-certs
             - name: gcp-secretmanager-serviceaccountkey
               mountPath: /gcp-secrets
       volumes:
@@ -520,6 +533,9 @@
         - name: jwt-cas-private-key
           secret:
             secretName: foo-chainloop-controlplane-jwt-cas
+        - name: server-certs
+          secret:
+            secretName: server_certs
         - name: gcp-secretmanager-serviceaccountkey
           secret:
             secretName: foo-chainloop-controlplane-gcp-secretmanager-serviceaccountkey
```

### DIFF
--- a/deployment/chainloop/values.yaml
+++ b/deployment/chainloop/values.yaml
@@ -103,6 +103,12 @@ controlplane:
     # Overrides the image tag whose default is the chart appVersion.
     # tag: latest
 
+  ## @param controlplane.secret.name name of a secret containing TLS certificate to be used by the controlplane grpc server.
+  tlsConfig:
+    secret:
+      # the secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
+      name: ""
+
   ## @param controlplane.pluginsDir Directory where to look for plugins
   pluginsDir: /plugins
 
@@ -442,6 +448,12 @@ cas:
     repository: ghcr.io/chainloop-dev/chainloop/artifact-cas
     # Overrides the image tag whose default is the chart appVersion.
     # tag: latest
+
+  ## @param cas.tlsConfig.secret.name name of a secret containing TLS certificate to be used by the controlplane grpc server.
+  tlsConfig:
+    secret:
+      # the secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
+      name: ""
 
   ## @skip cas.serviceAccount
   serviceAccount:
--- a/deployment/chainloop/templates/cas/config.configmap.yaml
+++ b/deployment/chainloop/templates/cas/config.configmap.yaml
@@ -11,6 +11,11 @@ data:
         addr: 0.0.0.0:8000
         timeout: 1s
       grpc:
+        {{- if .Values.cas.tlsConfig.secret.name  }}
+        tls_config:
+          certificate: /data/server-certs/tls.crt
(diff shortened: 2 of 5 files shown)


deployment/chainloop/values.yaml
```diff
@@ -103,6 +103,10 @@ controlplane:
     # Overrides the image tag whose default is the chart appVersion.
     # tag: latest
 
+  ## @param controlplane.tlsConfig.secret name of a secret containing TLS certificate to be used by the controlplane grpc server.
+  tlsConfig:
+    secret: ""
```
migmartri, line 108: It makes sense to me to first support providing the name of a secret which has the keypair inside. But I have a couple of suggestions, let me know what you think.

### Make the `secret` key in fact an object

so we end up with something like 

```
tlsConfig:
  secret:
    name: ""
```

The reason being two fold, a) we can clearly leave the door open to other implementations of this API such as providing the cert inline and b) we can easily extend the `by secret name` mechanism to include other opts such as keyNames and so on(as you suggested in the past)

## Add more context

Since not any secret will work here but just a secret that has two entries `tls.crt` and `tls.key`, I'd make add a comment pointing it out.

migmartri (commented):
It looks mostly good to me, ptal at my feeback and let me know your thoughts.

Thanks a lot, this is great work! 

Also, note that your PR is still in draft state, just in case you forgot to change its state.

deployment/chainloop/values.yaml
zaibon: Done

migmartri: Great, thanks!

PR #328, opened 2023-09-04
feat(controlplane):  dynamic CAS backend selection during download 

### LINKED ISSUE #268: Being able to download an artifact even if default CAS backend changed
Currently, when `artifact download` is invoked, the system looks for the default CAS backend and tries to find the artifact with that digest there.

The problem is that if the user performed a CAS backend change (which now they can https://github.com/chainloop-dev/chainloop/issues/223), they will not be able to find previous assets, even if they have the backend registered (but not default) there.

One way we could fix this is by maintaining a lookup table between the digest and the CAS backend where the asset is, and this lookup table can be updated once the attestation is for example finished, since at that stage we have the reference of all the assets + the CAS backend used https://github.com/chainloop-dev/chainloop/pull/237

Then, the download command can try to find the asset in the lookup table and load the actual CAS backend that has it, if it can't find it we could fallback to the current backend as we do today.

migmartri:
Enables download endpoints to find the CAS backend that contains a given digest. This solves the limitation we had of downloading from currentOrgs and default backends only. Now an user will be able to download an item from any cas-backend in any of their orgs.

The selection process works as follows:

- Find the casBackends where we have a record that the digest was uploaded, this is across all the organizations
- Then match these results with the organizations our user has access to (memberships)
- if we still have a match, means that we can download the item from one or more cas-backends
- If we have a single cas-backend or a default one we return that
- Otherwise the first one

image

This download selection works for both the new `/download/sha256:[]` endpoint and for the `artifact download` CLI command.

Closes #268


danlishka (approved):
Thank you.

Issue #263, opened 2023-07-20
Enable the public access to materials and attestations in CAS


danlishka:
Today all materials and attestations are available only to members of an organization in Chainloop. You need to first authenticate via `chainloop auth login` to use later `chainloop artifact download ...`.

We want to allow setting the attestation and materials public at the attestation `init` stage:

```
chainloop attestation init --public
```

This way, for example, we will enable users to download and verify all attestations & materials produced during the Chainloop build and release process.

danlishka:
What if we also add support for directly downloading artifacts through an URL? For example:

```
d.chainloop.dev/9be6e32ee1715a759...
```

PR #329, opened 2023-09-04
feat(controlplane): store and show the digest in CAS of the attestation

migmartri:
Soon, we'll be able to share download URLs for attestations #263. This requires finding the digest to share in the first place. 

This patch does store the digest, uses to store the attestation in the CAS, then it exposes it the API and CLI, so it can be easily shared. 

image

The reason to store it at `save time` instead of calculating it is to guarantee that the format is the same one at the time of saving it in the CAS, it also saves computation time in each `listing`


danlishka (approved):
LGTM

PR #331, opened 2023-09-05
feat(controlplane): public attestations/materials

migmartri:
Workflows now can be set as public and what that means is that its associated attestations and materials can be downloaded by any Chainloop User (regardless of their organization) #263

Downloads can be performed both using the CLI using `chainloop artifact download -d [sha256:...]` or http endpoint `/download/[sha256:...]`

The entrypoint to set the workflow as public for now is the API, no CLI handling has been implemented yet.

So for example you can update a workflow visibility by running.

```
grpcurl -H "authorization: Bearer $TOKEN" -d '{"id": "3e4ef02c-0f33-4248-84be-8a55ea2e3cb0", "public": true }' cp.chainloop.dev:443 controlplane.v1.WorkflowService.ChangeVisibility
```

Once that's done, you will see it in both the CLI table or JSON outputs.

```
$ chainloop wf ls -o json | jq '.[] | [.id, .public]'
[
  "807200c6-b872-45db-84fd-0147142e3104",
  true
]
[
  "7e2f7529-1bef-4ba3-b3d2-297f96df7749",
  false
]
[
  "3e4ef02c-0f33-4248-84be-8a55ea2e3cb0",
  true
]
```
image
 

The logic to pick what CAS backend to use has been extended since #328. It now includes an additional branch (in yellow), that starts when no accessible private backends have been found, in that case we check if there is any public location where the asset is located.  

image


danlishka (approved):
Looking forward to giving it a try!

PR #343, opened 2023-09-14
feat: expose inline material content

migmartri:
- exposes inline material content in the API. This will allow us clients to "simulate" downloads
- Decouples the API material type filename/value properties, similarly to what we already did in the normalizedMaterial/integrations. This allow us to be more precise in what to show.
- The CLI now shows the filename if needed properly in its own option.  

refs #275


danlishka (approved):
LGTM

PR #344, opened 2023-09-14
feat(api): get workflowRuns by digest

migmartri:
Updates the controlplane WorkflowRun API with

- Parameter to retrieve the workflowRun by `attestation digest`
- The controller code now takes into account if a workflow is public


danlishka (approved):
LGTM

PR #358, opened 2023-09-25
chore: add OpenSSF scorecard

migmartri:
image

We can improve it later on, it will also get better over time since it does average commits


danlishka (approved):
Awesome

Issue #357, opened 2023-09-25
Azure Blob Storage support in CAS


migmartri:
Add support to Azure Blob Storage as CAS backend.

This feature will also require to prepare the CAS service to support multiple backend types

![image](https://github.com/chainloop-dev/chainloop/assets/24523/8f89161c-b412-411e-8e60-fdcda5cd9a67)

```[tasklist]
### Pre-requisites
- [x] [Proxy] Multiple Backends support https://github.com/chainloop-dev/chainloop/pull/359
- [x] [Proxy] Selection on the fly https://github.com/chainloop-dev/chainloop/pull/359
- [x] [CLI] Send CAS backend type to Proxy for selection https://github.com/chainloop-dev/chainloop/pull/359
```

and once done we can move ahead with the actual upload/download implementation in the proxy

```[tasklist]
### Azure blob manager in CAS proxy
- [x] Look into authentication mechanisms https://github.com/chainloop-dev/chainloop/pull/360
- [x] Implement upload/download methods + metadata retrieval https://github.com/chainloop-dev/chainloop/pull/360
```
 
and the UX in the controlplane-CLI

```[tasklist]
### Azure CAS backend in control plane
- [x] Selection, addition, onboarding https://github.com/chainloop-dev/chainloop/pull/360
```

PR #359, opened 2023-09-25
refactor(CAS proxy): multiple CAS backend providers support 

migmartri:
Adds support to dynamically select a backend provider in the CAS proxy

image

- [CAS] Loads a list of backend providers at boot
- [CAS] Selects the CAS backend dynamically in each request
- [Controlplane] Generate a CAS token containing also the backend type, which can be used for the selection above.

This is a prerequisite of #357, once merged we will be able to implement a new backend in CAS + enable it in the controlplane


.golangci.yml
```diff
@@ -71,4 +71,8 @@ severity:
     - linters:
         - staticcheck
       text: "SA1019:"
-      severity: info
\ No newline at end of file
+      severity: info
+issues:
+  exclude-rules:
+    - path: _test\.go
+      text: "Potential hardcoded credentials"
```
migmartri, line 78: for some reason now it complains about some of the sha256 digests that we are using during testing

app/artifact-cas/api/cas/v1/status.proto
```diff
@@ -17,34 +17,31 @@ syntax = "proto3";
...
+  }
 }
 
-message InfozRequest { }
+message InfozRequest {}
 
 message InfozResponse {
-	string version = 1;
+  string version = 1;
+  repeated string backends = 2;
```
migmartri, line 37: this is in fact the new part. It exposes the enabled backends.

app/artifact-cas/cmd/main.go
```diff
@@ -56,19 +58,36 @@ func init() {
...
-		kratos.Server(
-			gs,
-			hs,
-			ms,
+type app struct {
+	*kratos.App
+	backend.Providers
+}
+
+func loadCASBackendProviders(creader credentials.Reader) backend.Providers {
```
migmartri, line 66: the new dynamic loader

danlishka (approved):
LGTM based on the review via conf call

PR #360, opened 2023-09-26
feat(cas): Azure Blob Storage support

migmartri:
Adds support for Azure Blob Storage as Content Addressable Storage backend.

In practice this means:

Users can now onboard their Azure Storage Account with Chainloop

```sh
$ chainloop cas-backend add azure-blob \
  --client-id [servicePrincipalID]
  --client-secret [servicePrincipalSecret]
  --tenant [Active directory tenant]\
  --storage-account [Storage Account name]
  --container [optional Storage account container]
```
  
once done, it will appear in your list of registered backends

```console
$ chainloop cas-backend ls

──────────────────────────────────────┬─────────────────────────────────────┬───────────┬─────────────────────────────────────┬───────────────┬─────────┐
│ ID                                   │ LOCATION                            │ PROVIDER  │ DESCRIPTION                         │ LIMITS        │ DEFAULT │
├──────────────────────────────────────┼─────────────────────────────────────┼───────────┼─────────────────────────────────────┼───────────────┼─────────┤
│ 43471279-495e-4c85-975f-822e1619e029 │ chainlooptestcas/chainloop          │ AzureBlob │                                     │ MaxSize: 100M │ true    │
├──────────────────────────────────────┼─────────────────────────────────────┼───────────┼─────────────────────────────────────┼───────────────┼─────────┤
```

At this point there is nothing else that an user needs to do, the system automatically will upload attestations and pieces of evidence to the storageAccount/container automatically :)

image

Note that you can use this backend in combination of an OCI one, meaning that files that have been uploaded to an OCI backend are still accessible and routed accordingly

image

Refs #357

### DIFF
--- a/app/artifact-cas/internal/service/bytestream.go
+++ b/app/artifact-cas/internal/service/bytestream.go
@@ -113,7 +113,7 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
 		return sl.LogAndMaskErr(err, s.log)
 	}
 
-	s.log.Infow("msg", "artifact received, uploading now to OCI backend", "name", req.resource.FileName, "digest", req.resource.Digest, "size", buffer.size)
+	s.log.Infow("msg", "artifact received, uploading now to backend", "name", req.resource.FileName, "digest", req.resource.Digest, "size", buffer.size)
 	if err := backend.Upload(ctx, buffer, req.resource); err != nil {
 		return sl.LogAndMaskErr(err, s.log)
 	}
--- a/app/artifact-cas/internal/service/bytestream.go
+++ b/app/artifact-cas/internal/service/bytestream.go
@@ -18,8 +18,11 @@ package service
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/gob"
+	"fmt"
+	"hash"
 	"io"
 
 	"errors"
@@ -151,7 +154,7 @@ func (s *ByteStreamService) Read(req *bytestream.ReadRequest, stream bytestream.
 	}
 
 	// streamwriter will stream chunks of data to the client
-	sw := &streamWriter{stream, s.log, req.ResourceName}
+	sw := &streamWriter{stream, s.log, req.ResourceName, sha256.New()}
 	if err := backend.Download(ctx, sw, req.ResourceName); err != nil {
 		if errors.Is(err, context.Canceled) {
 			s.log.Infow("msg", "download canceled", "digest", req.ResourceName)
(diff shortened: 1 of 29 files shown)


app/artifact-cas/internal/service/bytestream.go
```diff
@@ -161,6 +164,11 @@ func (s *ByteStreamService) Read(req *bytestream.ReadRequest, stream bytestream.
 		return sl.LogAndMaskErr(err, s.log)
 	}
 
+	// check if the file has been tampered with and notify the client
+	if sw.GetChecksum() != req.ResourceName {
```
migmartri, line 168: This is a verification server-side that the file has not been tampered with and we deliver only the file if it matches the digest. Note that we already do this check client-side but now server side makes it more robust.

app/artifact-cas/internal/service/download.go
```diff
@@ -88,20 +91,44 @@ func (s *DownloadService) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
 
 		http.Error(w, sl.LogAndMaskErr(err, s.log).Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// Verify the checksum
+	if got, want := fmt.Sprintf("%x", gotChecksum.Sum(nil)), wantChecksum.Hex; got != want {
+		msg := fmt.Sprintf("checksums mismatch: got: %s, want: %s", got, want)
```
migmartri, line 117: same here, we only push content to the browser if we know it matches the digest.

app/controlplane/internal/biz/casbackend.go
```diff
@@ -33,8 +35,7 @@ import (
 type CASBackendProvider string
 
 const (
-	CASBackendOCI             CASBackendProvider = "OCI"
```
migmartri, line 36: not needed, removed duplication.

app/controlplane/internal/service/casbackend.go
```diff
@@ -86,7 +86,7 @@ func (s *CASBackendService) Create(ctx context.Context, req *pb.CASBackendServic
 	}
 
 	// For now we only support one backend which is set as default
-	res, err := s.uc.Create(ctx, currentOrg.ID, req.Location, req.Description, biz.CASBackendOCI, creds, req.Default)
+	res, err := s.uc.Create(ctx, currentOrg.ID, req.Location, req.Description, biz.CASBackendProvider(req.Provider), creds, req.Default)
```
migmartri, line 89: so now we create other kinds of backends, specifically azureBlob

deployment/chainloop/templates/cas/config.configmap.yaml
```diff
@@ -17,6 +16,6 @@ data:
           private_key: /data/server-certs/tls.key
         {{- end }}
         addr: 0.0.0.0:9000
-        timeout: 1s
+        timeout: 5s
```
migmartri, line 19: needed to make sure we leave room for the azure requests.

migmartri (commented):
ptal @danlishka 

thanks!

danlishka (approved):
Awesome. LGTM. S3 is the next one!

PR #364, opened 2023-09-27
fix(deployment): wrong labeling templating

migmartri:
The previous PR added a regression in the Helm Chart that was preventing its render.

NOTE: the issue was due to editor auto-formatting


migmartri:
TBR!

PR #369, opened 2023-09-28
fix: extend http download timeout

### LINKED ISSUE #366: Timeout during http downloads
The CAS http download endpoint timeouts if the file takes more than 5 seconds to get sent.

This was a regression added [here](https://github.com/chainloop-dev/chainloop/pull/360/files#diff-94734ef6e7101d93d6329ebb39f71b349a22202f89285e34936caf7ed2f4c903R127). The reason for that change was to check the checksum of the retrieved file before sending it to the browser, but that means blocking/longer request cycles.

migmartri:
The issue in fact is a misconfiguration in the helm Chart and only Azure Blob Storage is affected.

The reason Azure is the only affected is because in its implementation we in fact use the parent `context` which by default has a `timeout`. The OCI implementation doesn't use context (it should)

The reason it only affects the HTTP endpoint is because grpc downloads don't have timeouts, http one does.

Fixes #366

### DIFF
--- a/app/artifact-cas/internal/service/download.go
+++ b/app/artifact-cas/internal/service/download.go
@@ -91,6 +91,10 @@ func (s *DownloadService) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if the buffer contains the actual data we expect we proceed with sending it to the browser
+	// Set headers
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", info.FileName))
+	w.Header().Set("Content-Length", strconv.FormatInt(info.Size, 10))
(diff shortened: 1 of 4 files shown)


app/artifact-cas/internal/service/download.go
```diff
@@ -91,6 +91,10 @@ func (s *DownloadService) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if the buffer contains the actual data we expect we proceed with sending it to the browser
+	// Set headers
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", info.FileName))
```
migmartri, line 96: this has been moved to just notify the browser early of what to expect.

danlishka (approved):
LGTM

PR #372, opened 2023-09-28
chore: remove OCI registry plugin

migmartri:
This PR removes the OCI registry fanout plugin, the reason being:

- It overlaps with the concept if CAS backend, in fact now that we support more advanced (and multiple) cas backends, this plugin makes less sense.
- Unlike CAS backends, it only stores attestations, not materials
- It doesn't store anywhere the result of the upload


danlishka (approved):
LGTM. Thanks

PR #378, opened 2023-10-04
fix(cas): handle client cancellation

### LINKED ISSUE #377: upload error during stress testing
```
{"level":"info","ts":1696448527.64371,"kind":"server","component":"grpc","operation":"/cas.v1.StatusService/Statusz","args":"readiness:true","code":0,"reason":"","stack":"","latency":0.00006399}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa265a1]

goroutine 245650 [running]:
github.com/chainloop-dev/chainloop/app/artifact-cas/internal/service.bufferStream({0x12814f0, 0xc0004627b0}, {0x1286c38, 0xc00576c480}, 0xc005a42420, 0xc0001391a0)
        /home/runner/work/chainloop/chainloop/app/artifact-cas/internal/service/bytestream.go:196 +0x141
created by github.com/chainloop-dev/chainloop/app/artifact-cas/internal/service.(*ByteStreamService).Write in goroutine 245588
        /home/runner/work/chainloop/chainloop/app/artifact-cas/internal/service/bytestream.go:100 +0x6b8
```

The error happens in this line https://github.com/chainloop-dev/chainloop/blob/4b8af6ec5f0c00e3c7b86ece449e1176e6d1c5e8/app/artifact-cas/internal/service/bytestream.go#L196

and it's likely to happen because the client timeouts or looses connection. Canceling the connection also causes the pod to fail

migmartri:
Context cancellation during upload chunk fills was not handled properly.

This code 

- Makes sure that having an error always returns regardless of the type
- Handles the grpc cancellation so we don't get cancellation exceptions
 

Fixes #377

### DIFF
--- a/app/artifact-cas/internal/service/bytestream.go
+++ b/app/artifact-cas/internal/service/bytestream.go
@@ -34,6 +34,8 @@ import (
 	kerrors "github.com/go-kratos/kratos/v2/errors"
 	"github.com/go-kratos/kratos/v2/log"
 	"google.golang.org/genproto/googleapis/bytestream"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // Implements the bytestream interface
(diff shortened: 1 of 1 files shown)


danlishka (approved):
LGTM

PR #388, opened 2023-10-12
feat: Azure KeyVault support

### LINKED ISSUE #376: Azure credentials manage service support
Currently, Chainloop requires a secure storage backend to store credentials (i.e integration settings), these backends include 

- Hashicorp Vault
- AWS Secret Manager
- Google Cloud Secret Manager 

This task is about looking into adding support for an equivalent solution in Azure (Azure Key Vault I think?)

https://azure.microsoft.com/en-us/products/key-vault/

migmartri:
Adds support for [Azure KeyVault](https://azure.microsoft.com/en-us/products/key-vault) for credentials handling.

image

image

Refs #376


danlishka (approved):
Awesome. LGTM

app/controlplane/internal/biz/organization.go
```diff
@@ -101,18 +102,14 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
 		}
 	}
 
-	// Delete the associated repository
-	// Currently there is only one repository per organization
```
migmartri, line 105: this is a bug we had that affected user offboarding. We only deleted the main cas backend instead all of them.

gr0 (approved):
Look great! 👍

PR #392, opened 2023-10-17
fix(azure): existing file regression

migmartri:
This change https://github.com/chainloop-dev/chainloop/pull/390/files?diff=split&w=0#r1360718603 introduced a regression in the Azure code.

The problem was that in Azure, we were returning a pointer to a custom error (instead of the error value), confusing `errors.As`.

This change makes sure we return always a concrete type and we leverage the already existing backend.isNotFound method.


migmartri:
@danlishka you pointed out this code change and I was wrong in my answer, sorry about it.

Issue #383, opened 2023-10-11
add annotations to git.head subject 


migmartri:
In #382 we added a git.head subject that contains the sha1 commit digest. It would be useful to also add additional information such as the author or the commit timestamp. 

Since now in-toto subjects are [resource descriptors](https://github.com/in-toto/attestation/blob/main/go/v1/resource_descriptor.pb.go), we could add annotations.

Note that this will require some refactoring from our end #60. This basically means 

- [x] move from slsa.attestation struct definitions to `in-toto/attestation` ones.
- [x] Update our code to leverage resource descriptors as subject array

PR #398, opened 2023-10-23
feat(attestation): add annotations to subject materials

migmartri:
Propagate material annotations to subject structs. This will enable enhanced ingestion upstream. 

In another patch, a similar feature #383 will be handled for our custom git head subject.

### DIFF
--- a/internal/attestation/renderer/chainloop/testdata/attestation.output.v0.2.json
+++ b/internal/attestation/renderer/chainloop/testdata/attestation.output.v0.2.json
@@ -7,11 +7,15 @@
         "sha256": "b403d1bb2b7739495df20f6d3838956333b04a66bc81d43bc7c64cee1433b604"
       }                                                                        
     },                                                                         
-    {                                                                                                                                                         
-      "name": "index.docker.io/bitnami/nginx",                                
-      "digest": {                                                              
+    {
+      "name": "index.docker.io/bitnami/nginx",
+      "digest": {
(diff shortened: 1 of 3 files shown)


internal/attestation/renderer/chainloop/testdata/attestation.output.v0.2.json
```diff
@@ -7,11 +7,15 @@
...
-    {                                                                                                                                                         
-      "name": "index.docker.io/bitnami/nginx",                                
-      "digest": {                                                              
+    {
+      "name": "index.docker.io/bitnami/nginx",
+      "digest": {
         "sha256": "580ac09da7771920dfd0c214964e7bfe4c27903bcbe075769a4044a67c9a390a"
-      }                                                                        
+      },
+      "annotations": {
```
buccarel, line 15: [not blocking / just a consideration]
Is it expected to have a JSON key in plural which value is not an array? In this case, it's `annotations`, meaning that 1 annotation is <key, value> in that hash.

buccarel (approved):

internal/attestation/renderer/chainloop/testdata/attestation.output.v0.2.json
migmartri: Yes, it's part of the [in-toto spec](https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md#schema) upstream.

danlishka (approved):
LGTM

PR #408, opened 2023-10-31
fix(controlplane): bubble up authorization permission

migmartri:
This code swaps the order of the checks done during invitation, from

- Check that the `receiver` doesn't belong to the `org` already
- Check that the `sender` can invite to the `org`

to

- Check that the `sender` can invite to the `org`
- Check that the `receiver` doesn't belong to the `org` already

That way, a sender will not be able to know if a user belongs to an organization by trying emails/uuids. 

This is an unlikely case (you need to know orgs UUDIDS and email address to positively know that it belongs to the org) but in any case this patch solves that problem.


danlishka (approved):
Good catch. Let's meet and review the flow in the invitation system again.

Issue #88, opened 2023-04-24
Jenkins runner support


migmartri:
Chainloop [has support for CI/CD runners](https://docs.chainloop.dev/reference/operator/contract#runner-context), this task is about adding Jenkins (and JenkinsX) to the list of supported ones.

On the technical side it meens implementing this interface https://github.com/chainloop-dev/chainloop/blob/053637d91d5aa728d50ec3740f246fdca8120b87/internal/attestation/crafter/runner.go#L27-L37

which will require you to inspect and figure out 1) how to know that you are in that runner, 2) what env variables you care about for the tool to auto-discover and 3) how to create a `link` to the workflow.

Additionally you'll need to update the `contract` definition to support this new runner type 

https://github.com/chainloop-dev/chainloop/blob/053637d91d5aa728d50ec3740f246fdca8120b87/app/controlplane/api/workflowcontract/v1/crafting_schema.proto#L36-L40

buccarel:
Looking after this.

buccarel:
This is interesting:

During the installation of jenkins (in this particular case, I'm using a docker container), the following setup step happens through the guided procedure:

<img width="971" alt="image" src="https://github.com/chainloop-dev/chainloop/assets/11213785/eb87581f-e31c-4069-aa0c-8e6f774ec059">

I will need to double-check it but as the wizard says this configuration is *required*, most likely the RunURI() method can be implemented using these values. I am assuming this could be retrieved from the job's node.

migmartri:
Correct, I believe that URL matches `JENKINS_URL` env var

https://e.printstacktrace.blog/jenkins-pipeline-environment-variables-the-definitive-guide/

buccarel:
To make sure, I am creating an alias to 127.0.0.1 that is different than localhost. I will use such alias to configure jenkins, to see if it appears in the job's node. That will be the ultimate confirmation.

buccarel:
Just some notes, mostly for myself.

If we SSH into a jenkins agent node, we can see that the info available at system level is not much:

<img width="1207" alt="image" src="https://github.com/chainloop-dev/chainloop/assets/11213785/53a0b557-f38d-4b59-953b-93cb85fed43d">

The env variable `AGENT_WORKDIR` should be guaranteed to be present.

[...]

PR #411, opened 2023-11-02
feat(runner): Integrate runner for Jenkins jobs

buccarel:
Fixes the following issue:
https://github.com/chainloop-dev/chainloop/issues/88

Definition of done:
[x] Implementation of the runner
[x] Unit tests
[x] Local tests

Jenkins output of `att init`

image

Generated chainloop_attestation.tmp.json

image


internal/attestation/crafter/runners/jenkinsjob.go
```diff
@@ -0,0 +1,65 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runners
+
+import "os"
+
+type JenkinsJob struct{}
```
migmartri, line 20: This is not related to your change, since the other runners have the same issue, but it's something that we could tackle in another patch.

It seems that the "factory functions" are not being used outside of tests, when they could also be used in `runner.go` in the switch clause.

app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go
```diff
@@ -43,6 +43,7 @@ const (
 	CraftingSchema_Runner_GITHUB_ACTION           CraftingSchema_Runner_RunnerType = 1
 	CraftingSchema_Runner_GITLAB_PIPELINE         CraftingSchema_Runner_RunnerType = 2
 	CraftingSchema_Runner_AZURE_PIPELINE          CraftingSchema_Runner_RunnerType = 3
+	CraftingSchema_Runner_JENKINS_JOB	      CraftingSchema_Runner_RunnerType = 4
```
migmartri, line 46: It seems that you've manually updated this value when in fact this is the result of modifying crafting_schema.proto

https://github.com/chainloop-dev/chainloop/blob/01432f04375e2d6a25084fe88c29eff1c32f5da4/app/controlplane/api/workflowcontract/v1/crafting_schema.proto#L42-L46

and then running `make api`. 

The good news is that I believe you've found a bug, for some reason our CI has not detected this, and it should have. We have similar checks for other generated code but not this one interestingly.

Just for runs, if you run `make api` in the code of your patch, you'll see how this line gets removed

```diff
diff --git a/app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go b/app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go
index 3de6c8f..0b782cf 100644
--- a/app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go
+++ b/app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go
@@ -43,7 +43,6 @@ const (
        CraftingSchema_Runner_GITHUB_ACTION           CraftingSchema_Runner_RunnerType = 1
        CraftingSchema_Runner_GITLAB_PIPELINE         CraftingSchema_Runner_RunnerType = 2
        CraftingSchema_Runner_AZURE_PIPELINE          CraftingSchema_Runner_RunnerType = 3
-       CraftingSchema_Runner_JENKINS_JOB             CraftingSchema_Runner_RunnerType = 4
 )

```

So as I said, instead, modify the proto definition and run the generation, that should persist it.

Thanks!

buccarel: Ohh interesting!
I bet it's something I know, like some knowledge I am missing, because I didn't realize at all that file was generated.
My bad my bad.
Will fix it!
Thanks

migmartri:
To run this locally, since this is just a client-side update you have two options:

a) run full development environment including CAS, controlplane locally or
b) just run your CLI pointing to the already existing SASS chainloop instance

For the latter, and to perform an attestation, even locally, you need to configure a workflow + robot-account and this is how you do it.


- login to the controlplane `go run main.go auth login`
- Create an empty contract (github.yaml) that sets the new runner requirement

```yaml
schemaVersion: v1

runner:
  type: "JENKINS_JOB"
```

- create a workflow pointing to that empty contract, that should return a robot-account. This is an API token that's used during attestation.

```console
go run main.go wf create --name test --project jktest --contract jenkins.yaml
```

Export the robot-account as an env variable

```sh
export CHAINLOOP_ROBOT_ACCOUNT=REDACTED.REDACTED.tBAf4SzsCWjFzuaS9R6a0PBNSo2_wdPjZKH1x_ogObc
```

and then run an `attestation init`, you should get the expected error

```console
$ go run main.go att init                                                           
ERR The contract expects the attestation to be crafted in a runner of type "jenkins-job" but couldn't be detected. Use --dry-run flag if development
```

after this you know how to proceed, impersonating the env vars or even running the modified version of this CLI (compiled with `make -C app/cli build` in a Jenkins runner.

I hope it helps

app/controlplane/api/workflowcontract/v1/crafting_schema.pb.go
migmartri: as I said, the CI should have warned us, I'll fix that :)

migmartri: @buccarel, this change https://github.com/chainloop-dev/chainloop/pull/412 is what was missing in the CI to detect your use-case.

If we happen to merge it and your code rebased, your PR should then fail telling you that you are not committing all the generated code or that it differs from the baseline.

migmartri:
#412 has been merged, please rebase main now, thanks!

migmartri (approved):
LGTM! 

Next we'll need to update the docs https://docs.chainloop.dev/reference/operator/contract#runner-context but we can wait until they are moved to the main repo. @danlishka is working on that

Thanks!

PR #412, opened 2023-11-02
fix(ci): detect generated API missing

migmartri:
during #411, we realized that out CI was not checking if all the autogenerated files were checked in. Now we do a similar approach we do for both migrations and ent files.


app/artifact-cas/api/cas/v1/status_http.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-http. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-http v2.7.0
```
migmartri, line 3: this is the result of generating the API bindings with the latest version of the tools

common.mk
```diff
@@ -2,16 +2,21 @@ VERSION=$(shell git describe --tags --always)
...
-	go install github.com/envoyproxy/protoc-gen-validate@v1.0.1
-	go install github.com/go-kratos/kratos/cmd/protoc-gen-go-errors/v2@latest
-	go install github.com/go-kratos/kratos/cmd/protoc-gen-go-http/v2@latest
+init: init-api-tools
 	go install github.com/google/wire/cmd/wire@latest
 	go install github.com/vektra/mockery/v2@v2.20.0
 	go install ariga.io/atlas/cmd/atlas@v0.12.0
+
+# initialize API tooling
+.PHONY: init-api-tools
```
migmartri, line 11: I split the tools for API handling to its own target so the impact in the CI during installation is limited.

common.mk
```diff
@@ -2,16 +2,21 @@ VERSION=$(shell git describe --tags --always)
...
+init-api-tools:
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.31.0
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0
 	go install github.com/bufbuild/buf/cmd/buf@v1.10.0
+	go install github.com/envoyproxy/protoc-gen-validate@v1.0.1
+	# Tools fixed to a specific version via its commit since they are not released standalone
+	go install github.com/go-kratos/kratos/cmd/protoc-gen-go-errors/v2@v2.0.0-20231102162905-3fc8fb7a0a0b
```
migmartri, line 18: I also make sure to "fix the version" of some of these deps, before it was latest.

buccarel (approved):

danlishka (approved):
LGTM

PR #421, opened 2023-11-08
chore: display jenkins job type in runner column

buccarel:
The new runner type for jenkins job needs its entry to be displayed well:

image

### DIFF
--- a/app/cli/internal/action/workflow_run_list.go
+++ b/app/cli/internal/action/workflow_run_list.go
@@ -111,15 +111,16 @@ func pbWorkflowRunItemToAction(in *pb.WorkflowRunItem) *WorkflowRunItem {
 
 func humanizedRunnerType(in v1.CraftingSchema_Runner_RunnerType) string {
 	mapping := map[v1.CraftingSchema_Runner_RunnerType]string{
-		*v1.CraftingSchema_Runner_GITHUB_ACTION.Enum():   "GitHub",
-		*v1.CraftingSchema_Runner_GITLAB_PIPELINE.Enum(): "GitLab",
-		*v1.CraftingSchema_Runner_AZURE_PIPELINE.Enum():  "Azure Pipeline",
-		*v1.CraftingSchema_Runner_JENKINS_JOB.Enum():     "Jenkins Job",
+		*v1.CraftingSchema_Runner_RUNNER_TYPE_UNSPECIFIED.Enum(): "Unspecified",
+		*v1.CraftingSchema_Runner_GITHUB_ACTION.Enum():           "GitHub",
+		*v1.CraftingSchema_Runner_GITLAB_PIPELINE.Enum():         "GitLab",
+		*v1.CraftingSchema_Runner_AZURE_PIPELINE.Enum():          "Azure Pipeline",
+		*v1.CraftingSchema_Runner_JENKINS_JOB.Enum():             "Jenkins Job",
 	}
 
 	hrt := mapping[in]
 	if hrt == "" {
-		return "Unspecified"
+		return "Unknown"
 	}
 
 	return hrt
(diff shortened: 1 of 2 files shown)


app/cli/internal/action/workflow_run_list.go
```diff
@@ -110,14 +110,17 @@ func pbWorkflowRunItemToAction(in *pb.WorkflowRunItem) *WorkflowRunItem {
 }
 
 func humanizedRunnerType(in v1.CraftingSchema_Runner_RunnerType) string {
```
migmartri, line 112: Great! Could you add some unit tests for this function?

app/cli/internal/action/workflow_run_list.go
```diff
@@ -110,14 +110,17 @@ func pbWorkflowRunItemToAction(in *pb.WorkflowRunItem) *WorkflowRunItem {
...
-	switch in {
-	case *v1.CraftingSchema_Runner_GITHUB_ACTION.Enum():
-		return "GitHub"
-	case *v1.CraftingSchema_Runner_GITLAB_PIPELINE.Enum():
-		return "GitLab"
-	case *v1.CraftingSchema_Runner_AZURE_PIPELINE.Enum():
-		return "Azure Pipeline"
-	default:
+	mapping := map[v1.CraftingSchema_Runner_RunnerType]string{
```
migmartri, line 113: This code (and the previous one) has a flaw.

It treats the same the fact that a runner is not specified (generic) or it's not known by the CLI because it's a new type (your Jenkins case)

I would update this code to handle both cases, which means 

* adding to the mapping `CraftingSchema_Runner_RUNNER_TYPE_UNSPECIFIED: "Unspecified"`
* the returned value for things that are not in the map should be called something 
 like "Unknown" 


What do you think?

app/cli/internal/action/workflow_run_list.go
```diff
@@ -110,14 +110,17 @@ func pbWorkflowRunItemToAction(in *pb.WorkflowRunItem) *WorkflowRunItem {
...
+		*v1.CraftingSchema_Runner_GITHUB_ACTION.Enum():   "GitHub",
+		*v1.CraftingSchema_Runner_GITLAB_PIPELINE.Enum(): "GitLab",
+		*v1.CraftingSchema_Runner_AZURE_PIPELINE.Enum():  "Azure Pipeline",
+		*v1.CraftingSchema_Runner_JENKINS_JOB.Enum():     "Jenkins Job",
+	}
+
+	hrt := mapping[in]
```
migmartri, line 120:
```suggestion
	hrt, ok := mapping[in]
```

app/cli/internal/action/workflow_run_list.go
```diff
@@ -110,14 +110,17 @@ func pbWorkflowRunItemToAction(in *pb.WorkflowRunItem) *WorkflowRunItem {
...
+	if hrt == "" {
```
migmartri, line 121:
```suggestion
	if !ok {
```

migmartri (commented):
Thanks for the PR!

app/cli/internal/action/workflow_run_list.go
buccarel: I noticed that. I didn't do anything because I thought it was intentional for some reason.
I agree wityh your solution. I'll do that.
Thanks.

app/cli/internal/action/workflow_run_list.go
buccarel: Yes, will do!

app/cli/internal/action/workflow_run_list_test.go
```diff
@@ -0,0 +1,83 @@
...
+			expectedOutput: "Jenkins Job",
+		}, {
+			name:           "unknown runner",
+			testInput:      -34,
+			expectedOutput: "Unknown",
+		},
+	}
+
+	// enforce 1 test case per runner (+ the unknown)
+	nRunnerTypes := len(v1.CraftingSchema_Runner_RunnerType_name)
```
migmartri, line 63: interesting check, I like it!

migmartri (approved):
Perfect! 🚀

Issue #417, opened 2023-11-06
describe wf run includes empty information


migmartri:
`chainloop wf run describe --id [...] -o json` is showing spurious information that no longer exists like `id` and `createdAt`. The reason they don't exist is because attestations are not stored in their own table anymore.

```
   "attestation": {        
      "id": "",                                                                
      "createdAt": "1970-01-01T00:00:00Z",  
```

To fix that, we need to remove the fields from the client-side struct and deprecate/remove the properties in the API proto message as well.

PR #422, opened 2023-11-08
fix(api): remove empty properties description

Javirln:
This patch removes two empty properties exposed by the attestation API.

Fixes https://github.com/chainloop-dev/chainloop/issues/417


migmartri (approved):
Woohoo!! 

Thanks for the contribution, it looks good to me! 🚀

migmartri:
@Javirln, there seems to be an issue with your commit. You might need to update your GPG key with your personal email address too.

image

PR #423, opened 2023-11-08
fix(DB): cleanup items on organization deletion

migmartri:
`chainloop auth delete-account` was failing with a SQL constraint error.

This is due to the fact that the organization we might be deleting is referenced by other tables, cas_backends, org_invitations and so on.

This code adds the proper reference handling so now deleting the org also deletes these rows.


danlishka (approved):
LGTM

Issue #380, opened 2023-10-05
empty files causes client error


migmartri:
Handle this case gracefully

```
INF uploading goob - sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1bd9e8b]

goroutine 1 [running]:
github.com/chainloop-dev/chainloop/app/cli/internal/action.(*ArtifactUpload).Run(0x0?, {0x7ffc0f3d5cd5, 0x4})
        /home/runner/work/chainloop/chainloop/app/cli/internal/action/artifact_upload.go:75 +0x34b
github.com/chainloop-dev/chainloop/app/cli/cmd.newArtifactUploadCmd.func2(0xc000713500?, {0x219cafd?, 0x4?, 0x219ca15?})
        /home/runner/work/chainloop/chainloop/app/cli/cmd/artifact_upload.go:49 +0x5b
github.com/spf13/cobra.(*Command).execute(0xc000713500, {0xc00007fac0, 0x2, 0x2})
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x87c
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003d0900)
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
github.com/spf13/cobra.(*Command).Execute(0x24e7228?)
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992 +0x13
main.main()
        /home/runner/work/chainloop/chainloop/app/cli/main.go:34 +0x19d

```

PR #430, opened 2023-11-09
fix(cli): handle empty files

jiparis:
Gracefully handle empty files uploads.

Closes https://github.com/chainloop-dev/chainloop/issues/380

### DIFF
--- a/app/cli/internal/action/artifact_upload.go
+++ b/app/cli/internal/action/artifact_upload.go
@@ -59,11 +59,6 @@ func (a *ArtifactUpload) Run(filePath string) (*CASArtifact, error) {
 		return nil, fmt.Errorf("retrieving file information: %w", err)
 	}
 
-	if info.Size() == 0 {
-		a.Logger.Info().Msg("the file is empty, nothing to upload")
-		return nil, nil
-	}
-
 	// render progress bar
 	go renderOperationStatus(context.Background(), client.ProgressStatus, info.Size())
 	defer close(client.ProgressStatus)
--- a/app/cli/internal/action/artifact_upload.go
+++ b/app/cli/internal/action/artifact_upload.go
@@ -59,6 +59,11 @@ func (a *ArtifactUpload) Run(filePath string) (*CASArtifact, error) {
 		return nil, fmt.Errorf("retrieving file information: %w", err)
 	}
 
+	if info.Size() == 0 {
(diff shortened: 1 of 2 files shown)


app/cli/internal/action/artifact_upload.go
```diff
@@ -59,6 +59,11 @@ func (a *ArtifactUpload) Run(filePath string) (*CASArtifact, error) {
 		return nil, fmt.Errorf("retrieving file information: %w", err)
 	}
 
+	if info.Size() == 0 {
+		a.Logger.Info().Msg("the file is empty, nothing to upload")
```
buccarel, line 63: I have a context question, most likely for @migmartri, but anybody who knows the answer is welcome to reply ofc.

What is the context for these files?
Would someone find legit, in the generic scenario, to produce and upload an empty file?

migmartri: Interesting point.

I think you are right, it doesn't harm to upload empty files. In fact, artifact material types in attestations can be empty, but this code doesn't affect that (this is only for `chainloop artifact upload` cmd.) 

The change in `uploader.go` is in fact what fixes the original panic #430, so I think it's ok if we remove 

```go
if info.Size() == 0 {
		a.Logger.Info().Msg("the file is empty, nothing to upload")
...
```

and just keep the change at `internal/casclient/uploader.go`.

What do you think?

> Note: This implementation is based on my prescription offline, so any blame is on me :)

migmartri (commented):
Thanks for the contribution! 

Adding some comments to reply to @buccarel

migmartri (approved):

app/cli/internal/action/artifact_upload.go
jiparis: Makes sense. I'll update the PR with your sugested changes. Thanks!

buccarel: Ok perfect. I'll request changes to remember not to merge this until it's fixed. 
Thanks.

buccarel (changes requested):
Waiting for the fix we agreed on

migmartri (approved):
LGTM! Thanks!

migmartri:
> Waiting for the fix we agreed on

@buccarel ptal, it seems that the fix is now up.

PR #431, opened 2023-11-10
feat(referrer): support same digest for two different kinds

migmartri:
This PR 

- Improves the ingestion code to allow storing two referrers with the same digest but different kinds. it includes some refactoring too.
- Updates the retrieval code to return a well-known error that: 
  - makes the API return a validation error instead of a server error
  - shows actionable information to the user of what's available.

For example, now, on retrieval we get.

```
$ chainloop discover -d sha256:ebe2e6ffab93f4b2e2743d3557bd094427ea059adc78cfe2cd330a72f61731e8
ERR validation error: please provide the referrer kind: failed to get referrer: digest sha256:ebe2e6ffab93f4b2e2743d3557bd094427ea059adc78cfe2cd330a72f61731e8
 present in two kinds ["SBOM_CYCLONEDX_JSON" "ARTIFACT"]
```

In a future patch, the user will be able to filter out by kind.

### DIFF
--- a/app/controlplane/internal/biz/referrer.go
+++ b/app/controlplane/internal/biz/referrer.go
@@ -38,7 +38,12 @@ type Referrer struct {
 	// Wether the item is downloadable from CAS or not
 	Downloadable bool
 	// points to other digests
-	References []string
+	References []*ReferrerReference
+}
+
+type ReferrerReference struct {
+	Digest string
+	Kind   string
 }
 
 // Actual referrer stored in the DB which includes a nested list of storedReferences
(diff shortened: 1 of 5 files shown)


app/controlplane/internal/biz/referrer.go
```diff
@@ -37,27 +39,21 @@ type Referrer struct {
...
 // Actual referrer stored in the DB which includes a nested list of storedReferences
 type StoredReferrer struct {
-	ID     uuid.UUID
-	Digest string
-	Kind   string
-	// Wether the item is downloadable from CAS or not
-	Downloadable bool
-	CreatedAt    *time.Time
+	*Referrer
+	ID        uuid.UUID
```
danlishka, line 48: Why do we need uuid?

danlishka (approved):
LGTM

app/controlplane/internal/biz/referrer.go
migmartri: In favor of another format? or having a primary ID at all?

PR #436, opened 2023-11-14
feat: workflow update

### LINKED ISSUE #432: feat: update workflow
We should be able to update a workflow

- name
- project
- team
- visibility (public or not)

We do not have many examples of updating entities in the CLI but we could technically do something similar to what we do with `workflow contract update` where you can pass the properties you want to change as args. The ones omitted are not changed. This would mean having `chainloop workflow update --name ... --public yes, ...`

NOTE: We want to eventually move to a declarative approach where users will provide a declarative statement of what the entity (workflow in this case would look like). Having an update endpoint could evolve ultimately into this.

migmartri:
Support for updating a workflow name, project, team, and visibility.

This code supersedes the previous changeVisibility API action, and hence it's been removed. 

Closes #432


danlishka (approved):
LGTM

Issue #433, opened 2023-11-13
circle CI runner type support


migmartri:
Our list of [supported runner types](https://docs.chainloop.dev/reference/operator/contract#runner-context) is missing [CircleCI](https://circleci.com/). 

This task is about doing something similar we did for Jenkins #88.

buccarel:
Looking after this

buccarel:
https://circleci.com/docs/variables/

PR #439, opened 2023-11-15
fix(runner): remove optional env variable from the requirements

buccarel:
Main issue:
https://github.com/chainloop-dev/chainloop/issues/433

I am removing the environment variable CIRCLE_PR_NUMBER from the required because it's set only in very specific scenarios. Everything else is working. I artificially set this env variable to make the runner work from within CircleCI and it worked just fine:
image


migmartri (approved):
LGTM! 

Nitpick: could you please reference the associated issue in the description of the PR? 

Thanks!

PR #441, opened 2023-11-16
feat(referrers): shared index  API

migmartri:
Exposes a discovery API that returns the referrers from public workflows of allowed organizations.

In practice, an organization that has been added to the allowList can mark a workflow as public, and its referrers will be exposed through a shared public, index.

## Configuration

The list of organizations that can opt to publish referrers in the shared index is defined via a static configuration sntry. This acts as a trusted publisher mechanism.

```yaml
# example of enabling the index and an org
referrer_shared_index:
  enabled: true
  allowed_orgs:
    - 7ae5a35a-9f01-4711-b293-bd1d161c01a9 # practical_mayer
```

## API changes

This public index is available both via gRPC and HTTP and **DOES NOT require authentication**

### Behavior 

For example, let's say we have an artifact associated with five attestations.

By calling the discover endpoint, you'll get all the attestations regardless of whether they belong to a private or a public workflow. They belong to one of your organizations.

```sh
$ chainloop discover -d sha256:ebe2e6ffab93f4b2e2743d3557bd094427ea059adc78cfe2cd330a72f61731e8 -k ARTIFACT   

```

```json
{
   "digest": "sha256:ebe2e6ffab93f4b2e2743d3557bd094427ea059adc78cfe2cd330a72f61731e8",
   "kind": "ARTIFACT",
   "downloadable": true,
   "public": true,
   "createdAt": "2023-11-09T21:44:31.372575Z",
   "references": [
      {
         "digest": "sha256:119bdefae108a32a81e599e5cf6872c4e88c902cf5a55c58e82598a4c7fcef53",
         "kind": "ATTESTATION",
         "downloadable": true,
         "public": false,
         "createdAt": "2023-11-14T22:20:40.567136Z",
         "references": []
      },
      {
         "digest": "sha256:4b83ebba199499a1cbbd58517f5818d8e3fd56cb5cc244c26f4dd0729e2fccb2",
         "kind": "ATTESTATION",
         "downloadable": true,
         "public": true,
         "createdAt": "2023-11-14T22:45:57.238998Z",
         "references": []
      },
      {
         "digest": "sha256:d1b54d55787308f9ac3facde069afc7c00bf6897fb07d6109253546bec44f2b7",
         "kind": "ATTESTATION",
         "downloadable": true,
         "public": true,
         "createdAt": "2023-11-14T23:00:08.998459Z",
         "references": []
      },
      {
[...]

PR #444, opened 2023-11-17
deploy(referrers): add public index configuration

### LINKED ISSUE #435: feat(referrers): public API
We have added a referrer discovery API in our latest release https://github.com/chainloop-dev/chainloop/releases/tag/v0.30.0

This API returns referrer items associated with one of the current user organizations. This task is about 

- allow organizations to expose referrers to a common, index, as an index that's shared across organizations
- allow the chainloop instance to have control on what organizations can actually publish to the index, some sort of "trusted publisher" approach.

We currently have a way to make attestations and artifacts public #263, we should probably piggy back on the same approach. 

This would mean that an organization that a) makes a workflow public and b) and it's in the allowlist for publishers will get the referrers associated with such workflow to the index.

migmartri:
Adds support to configuring, at deployment time, the organizations allowlist for our public index #441

```diff
 helm template foo deployment/chainloop -f ~/Desktop/values.local.yaml > /tmp/after.yaml && diff -u /tmp/before.yaml /tmp/after.yaml
--- /tmp/before.yaml    2023-11-17 10:39:47.695002643 +0100
+++ /tmp/after.yaml     2023-11-17 11:09:51.669551058 +0100
@@ -232,6 +232,9 @@
       insecure: true
       download_url: /download
     plugins_dir: /plugins
+    referrer_shared_index:
+      allowedOrgs: []
+      enabled: false
```

Closes #435

### DIFF
--- a/deployment/chainloop/Chart.yaml
+++ b/deployment/chainloop/Chart.yaml
@@ -4,7 +4,7 @@ description: Chainloop is an open source software supply chain control plane, a 
 
 type: application
 # Bump the patch (not minor, not major) version on each change in the Chart Source code
-version: 1.23.0
+version: 1.23.1
 # Do not update appVersion, this is handled automatically by the release process
 appVersion: v0.30.0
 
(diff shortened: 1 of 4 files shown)


danlishka (approved):
LGTM

PR #447, opened 2023-11-24
fix(referrer): remove sha256 digest validation

migmartri:
digest `alg` can be different than `sha256`, i.e sha1 for a commit.

This removes a check we have in the API, ideally, we should run some validation that includes the sha1 format too.


buccarel (approved):
Nice

PR #449, opened 2023-11-27
feat(controlplane): update org name

### LINKED ISSUE #448: Update organizaton
Being able to update the organization name

migmartri:
Being able to update an existing organization name that the user is a member of.

This patch also adds some logic to perform [the same validation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) Kubernetes has for the namespace names. 

The reason to perform this validation is that later down the road, we'll add additional constraints to make sure the names are unique across the instance, can be used as part of a URL, and it can become an identifier on a declarative API. 

Closes #448


app/controlplane/api/controlplane/v1/context.proto
```diff
@@ -32,7 +32,7 @@ message ContextServiceCurrentResponse {
   
   message Result {
     User current_user = 1;
-    Org current_org = 2;
+    OrgItem current_org = 2;
```
migmartri, line 35: this change it just to be consistent with the name of other resources in the proto defs

app/controlplane/api/controlplane/v1/organization.proto
```diff
@@ -24,6 +24,7 @@ import "validate/validate.proto";
 service OrganizationService {
   // List the organizations this user has access to
 	rpc ListMemberships (OrganizationServiceListMembershipsRequest) returns (OrganizationServiceListMembershipsResponse);
+	rpc Update (OrganizationServiceUpdateRequest) returns (OrganizationServiceUpdateResponse);
```
migmartri, line 27: this is the actual new method

app/controlplane/internal/biz/organization.go
```diff
@@ -41,25 +45,96 @@ type OrganizationUseCase struct {
...
+	}
+
+	if err := validateOrgName(name); err != nil {
+		return nil, NewErrValidation(fmt.Errorf("invalid organization name: %w", err))
 	}
 
 	return uc.orgRepo.Create(ctx, name)
 }
 
+func validateOrgName(name string) error {
```
migmartri, line 75: and the new validation logic performed on creation and update.

Note that selecting the name on creation is not yet supported, but will.

That's why for the auto-generated names we make sure it complies with this new format.

danlishka (approved):
LGTM

PR #451, opened 2023-12-01
fix(cli attestation): handle non-merged branches

migmartri:
Shows an error message instead of blocking the attestation

> WRN failed to list remotes error="branch config: invalid merge"

### DIFF
--- a/internal/attestation/crafter/crafter.go
+++ b/internal/attestation/crafter/crafter.go
@@ -253,9 +253,12 @@ type CommitRemote struct {
 	Name, URL string
 }
 
+// This error is not exposed by go-git
+var errBranchInvalidMerge = errors.New("branch config: invalid merge")
+
 // Returns the current directory git commit hash if possible
 // If we are not in a git repo it will return an empty string
-func (c *Crafter) gracefulGitRepoHead(path string) (*HeadCommit, error) {
+func gracefulGitRepoHead(path string) (*HeadCommit, error) {
 	repo, err := git.PlainOpenWithOptions(path, &git.PlainOpenOptions{
 		// walk up the directory tree until we find a git repo
 		DetectDotGit: true,
@@ -282,7 +285,7 @@ func (c *Crafter) gracefulGitRepoHead(path string) (*HeadCommit, error) {
 		return nil, fmt.Errorf("finding head commit: %w", err)
 	}
 
-	headCommit := &HeadCommit{
+	c := &HeadCommit{
 		Hash:        commit.Hash.String(),
 		AuthorEmail: commit.Author.Email,
 		AuthorName:  commit.Author.Name,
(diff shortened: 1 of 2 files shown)


internal/attestation/crafter/crafter.go
```diff
@@ -293,7 +293,9 @@ func gracefulGitRepoHead(path string) (*HeadCommit, error) {
 
 	remotes, err := repo.Remotes()
 	if err != nil {
-		return nil, fmt.Errorf("getting remotes: %w", err)
+		// go-git does an additional validation that the branch is pushed upstream
+		// we do not care about that use-case, so we ignore the error
+		return c, nil
```
migmartri, line 298: I know this is swallowing other errors, but I decided to go this way since

- That error type is not exposed in the upstream library
- We really shouldn't block on remotes lookups for any reason.

buccarel: I understand the error, but I can see that the function `repo.Remotes()` can return more than one type of error. I think you want to swallow only if err is `ErrRemoteNotFound` or equivalent, am I right?

buccarel (changes requested):
I'm requesting changes just to make sure my question is answered.

internal/attestation/crafter/crafter.go
migmartri: That's right, that's what I tried to explain in my comment. "I know this is swallowing other errors"

I just thought it was not worth inspecting them as long as we log them (see the logger.Warn), and I do not see a reason why we should fail in any circumstance. 

In any case, let me revisit it.

Thanks!

internal/attestation/crafter/crafter.go
```diff
@@ -186,7 +186,7 @@ func LoadSchema(pathOrURI string) (*schemaapi.CraftingSchema, error) {
 // Initialize the temporary file with the content of the schema
 func (c *Crafter) initCraftingStateFile(schema *schemaapi.CraftingSchema, wf *api.WorkflowMetadata, dryRun bool, runnerType schemaapi.CraftingSchema_Runner_RunnerType, jobURL string) error {
 	// Generate Crafting state
-	state, err := initialCraftingState(c.workingDir, schema, wf, dryRun, runnerType, jobURL)
+	state, err := c.initialCraftingState(c.workingDir, schema, wf, dryRun, runnerType, jobURL)
```
buccarel, line 189: Is it expected that the function belongs to the crafter, but also takes the crafter property as argument?
```
c.initialCraftingState(c.workingDir, ...)
```
Isn't workingDir already accessible without being an argument?

migmartri: Good catch, this is poor coding on my side. I'll fix it.

internal/attestation/crafter/crafter.go
buccarel: Ok so you are ubersure that no matter what happens, we can keep going.
Gotcha, I'll approve now that I'm sure <3

migmartri: ptal at the change now.

PR #452, opened 2023-12-01
feat(controlplane): leave organization

### LINKED ISSUE #414: Allow leaving organization
Users can now be added to an organization #401, but we have no option to leave.

This issue is about supporting leaving an org, it could be done in the CLI by running something like `chainloop org leave --id [UUID of the org]`

Note that we might need to control the case where this user is the only member of an ORG. We might not want to leave orgs stranding and instead suggest them to delete their account `chainloop auth delete-account`

migmartri:
Being able to leave an organization

- Remove membership
- Delete the organization if the currentUser is the only member

```
chainloop org leave --id 669dbbda-9bd0-4f44-b970-254287c9e1d8
You are about to leave the organization "shared-org"
To confirm, please type "J9RY-93XU"
INF Membership deleted
```

closes #414

### DIFF
--- a/app/controlplane/internal/biz/user_integration_test.go
+++ b/app/controlplane/internal/biz/user_integration_test.go
@@ -32,30 +32,30 @@ User mapping:
 - userOne, userTwo -> shared org
 */
 func (s *userIntegrationTestSuite) TestDeleteUser() {
-	assert := assert.New(s.T())
 	ctx := context.Background()
 
 	err := s.User.DeleteUser(ctx, s.userOne.ID)
-	assert.NoError(err)
+	s.NoError(err)
 
 	// Organization where the user is the only member got deleted
 	gotOrgOne, err := s.Organization.FindByID(ctx, s.userOneOrg.ID)
-	assert.NoError(err)
-	assert.Nil(gotOrgOne)
+	s.Error(err)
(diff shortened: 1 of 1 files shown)


app/controlplane/internal/biz/membership.go
```diff
@@ -41,21 +42,57 @@ type MembershipRepo interface {
...
+
+	// Check number of members in the org
+	// If it's the only one, delete the org
+	membershipsInOrg, err := uc.repo.FindByOrg(ctx, m.OrganizationID)
+	if err != nil {
+		return fmt.Errorf("failed to find memberships in org: %w", err)
+	}
+
+	if len(membershipsInOrg) == 0 {
+		// Delete the org
```
buccarel, line 88: Just an architectural question, to make sure I understand everything well.

What happens if, in between the membership deletion and the org deletion, some other membership is created? Where are these races being handled?

migmartri: Good question; they are not. I thought about this one, and to solve it, we'll need to move most of this logic to the data layer, where we could leverage `ent transactions.` 

I just thought that the possibility of this race was very, very low and not worth the implementation in this first pass. That said, I'll create an issue for it so we don't forget.

Issue #453, opened 2023-12-04
Controlplane API tokens support


migmartri:
> NOTE: This is different from workflow robot accounts. Workflow robot-accounts are a mechanism used during attestations. These will eventually be deprecated and superseded by ABAC-enabled API tokens or workload identity. Changing how robot-account work today is not in the scope of this task.

Talking to the control-plane API requires the user to execute an OIDC dance where she will receive a token valid for 24 hours.

This is problematic when you want to perform unattended operations. For example, you might want to enable contract updates via gitOps flow in your CI. To do so, you'll need a long-lasting API token.

This task is about allowing the creation of api-tokens associated with an organization that can be used for different control-plane operations.

These tokens, from the get-go, should:

- Have a required expiration time in duration format. Minimum 1d
- Can be revoked at any time
- Are ORG scoped

For this first version, these tokens will have full-org permissions. In future iterations, we'll add an Authorization layer potentially using https://casbin.org, but that'll come later

PR #455, opened 2023-12-04
Api tokens reference persistence

migmartri:
This is the first step towards supporting API tokens for unattended operations in the control plane #453 

This patch adds api-tokens column that stores the reference (not the actual token) and implements the business logic to store, list and revoke API tokens

The next patch will add API and CLI support for this functionality


danlishka (approved):
LGTM

Issue #456, opened 2023-12-05
Handle org departure aftermatch


migmartri:
Once a user leaves an organization today, no org is set as current. This causes the creation of a new org when the user logs in.

This is something we need to handle during membership deletion. Probably either select one of the other orgs that the user might have available.

migmartri:
We might need to brainstorm what to do in this case, tbh I am not sure what's the turn of action here. Let me know what you thoughts are here and I'll do the same and put it here.

Thanks!

migmartri:
One option I am leaning towards is to prevent the user from leaving an organization if that's the only one they are members of. This is ok, because in that case, the user can opt for deleting their account using `auth delete` instead.

To me, this simplifies the current implementation since a user will always have an organization. 

What do you think?

buccarel:
That is an option I also considered. It's very simple and it's also effective.

For example today I tried to design a solution where another organization is set as current upon deletion of the former current. This not only is more complicated, but it has 2 points that do not convince me:
* we don't have a criterion to use to set the other org
* there might just be 1 org in total

So your solution already looked pretty appealing to me as well.

migmartri:
Thanks @buccarel 

The main reason I did not want to rely on `current` for any further logic in the backend is that it's likely that we'll move away from setting current in the backend and instead be the client who sets it #350

That said, you are right in your observation that we don't have defined a `current` fallback mechanism.

So, to me, your proposal of making sure you can't delete `current` is the best one we have now. We can revisit it in #350

WDYT?

buccarel:
Yep ^_^

buccarel:
Given what you said, I think a good idea would be to implement this in the CLI only, maybe?
When one tries to leave the current organization, the command would fail if the org is current.
This should be compatible with what you said about moving the implementation to the client side, am I right?

PR #457, opened 2023-12-06
feat(org): prevent from leaving current organization

buccarel:
This is a proposal to fix the following issue:
https://github.com/chainloop-dev/chainloop/issues/456

The rationale behind this is based on what @migmartri said in the issue, i.e.:

> it's likely that we'll move away from setting current in the backend and instead be the client who sets it

I think this could be a step in the right direction, because the limitation is happening in the CLI command only.

I also noticed that there are no tests for these implementations, and I think I understand why that is. I think for the moment we could rely on end-to-end, before merging this. Test:
https://pastebin.com/HeGXrewd

### DIFF
--- a/app/cli/cmd/organization_leave.go
+++ b/app/cli/cmd/organization_leave.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


migmartri (approved):
LGTM, ptal at my comment about the error message to be returned.

PR #459, opened 2023-12-07
fix: enable JWT audience verification

migmartri:
This patch updates the JWT generation/verification code used today for authentication and attestation (robot-accounts) to generate and verify respective custom `audience` values.

The reason for this change is in preparation for #453, where we'll be creating additional JWT tokens that are meant to be distinct from the existing authentication ones.

The changes are as follows 

## Temporary authentication tokens

The tokens result of `chainloop auth login` now have the audience `user-auth.chainloop`, existing tokens will fail to validate, but that's ok since they have 24 hours max validity.

## Robot-accounts

These tokens now have `aud=attestations.chainloop` but during verification we still support its previous value. This is to maintain compatibility with robot-accounts that are already in place. New ones will have the right `audience`

### DIFF
--- a/app/controlplane/internal/biz/robotaccount.go
+++ b/app/controlplane/internal/biz/robotaccount.go
@@ -89,7 +89,7 @@ func (uc *RobotAccountUseCase) Create(ctx context.Context, name string, orgID, w
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


danlishka (approved):
LGTM

PR #460, opened 2023-12-11
fix(referrer): remove sha256 requirement

migmartri:
Equivalent to https://github.com/chainloop-dev/chainloop/pull/447 but affecting the public referrer endpoint.


danlishka (approved):
LGTM

PR #461, opened 2023-12-12
API tokens API support

### LINKED ISSUE #351: Service account support
Currently, the operator API can only be used by authenticating with the credentials that they will get through an interactive oauth2 flow (`chainloop auth login`), these credentials, by default, last 24 hours and hence are not suitable for long-lasting automation.

To do that, we might need to offer service accounts (potentially attached to organizations). A service account will provide a long-lasting token (with optional expiration time), that can be revoked at any time.

We've a similar mechanism already implemented for workflow robot-accounts. We should think how/if these are related.

### LINKED ISSUE #453: Controlplane API tokens support

migmartri:
This patch enables API-token management though the CLI as well as implements its use in the controlplane.

An API-token is an alternative authentication method compared to the temporary user auth.  

* [API] Load/validate API tokens in middleware alongside regular user auth tokens
* [API] Implement API handlers for the creation, revocation and listing of API tokens in the client
* [CLI] Add support to providing the token in the CLI via flag `--token` or env var `CHAINLOOP_API_TOKEN`
* [CLI] Add support for creation, revocation and listing of tokens in the CLI 

Closes #351
Closes #453


danoo (approved):
Thanks for the call and reviewing it together. LGTM! Please add an issue to make sure we track the user who created it. And also can we extract the revokedAt validation to a separated method?

danlishka (approved):
Thanks for the call and reviewing it together. LGTM! Please add an issue to make sure we track the user who created it. And also can we extract the revokedAt validation to a separated method?

PR #464, opened 2023-12-16
feat: allow upload/downloads using API token

### LINKED ISSUE #463: Upload/Downloads can't be performed using api-tokens
This PR https://github.com/chainloop-dev/chainloop/pull/461 added support for robot-accounts, but it's implementation didn't cover CAS related handlers in the control-plane. In practice this means that `chainloop artifact upload/download` don't work.

The reason that it doesn't work is because those endpoints rely on the `current_user` to find all the orgs available. In the case of the API-token we'll need to change that and make sure it's using the org associated with it instead.

migmartri:
In a previous patch we added support for API tokens #351, but these couldn't be used for download/upload artifacts to the CAS. 

This PR solves that by

- Update the cas-mapping lookup logic to accept the list of organizations directly, not just the user.
- Update the `service` endpoints to use either of those methods to lookup for the mapping.

Closes #463

### DIFF
--- a/app/controlplane/internal/biz/casmapping.go
+++ b/app/controlplane/internal/biz/casmapping.go
@@ -79,12 +79,12 @@ func (uc *CASMappingUseCase) FindByDigest(ctx context.Context, digest string) ([
 	return uc.repo.FindByDigest(ctx, digest)
 }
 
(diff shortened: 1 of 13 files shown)


danlishka (approved):
LGTM

Issue #136, opened 2023-05-31
The contract revision number gets updated even if the new contract does not change.


danlishka:
Today, if you run 

`chainloop workflow contract update --id $CID -f contract.yml`,

and the contract.yml file does not change since the last time the contract got updated in Chainloop, the contract revision number will still get updated. 

### Example
```
$ chainloop workflow contract update --id $CID -f contract.yml
[...]
│ Revision number      │ 8                                    │
[...]

<contract.yml does not change here, and we run the command again>

$ chainloop workflow contract update --id $CID -f contract.yml
[...]
│ Revision number      │ 9                                    │
[...]

```

### Proposal
- Do not update the contract revision number if the contract does not change. 
- Return an error with the appropriate message.

PR #468, opened 2023-12-18
fix(controlplane): bump revision only during workflow contract update

fgallina:
This change makes sure that a workflow contract revision gets created only if there is an actual update in the schema content.

Closes chainloop-dev/chainloop#136


migmartri (approved):
LGTM! 

Thank you!

migmartri:
Awesome, I can see you fixed the `verified` commit issue. Ready to go 🚀

PR #471, opened 2023-12-18
fix: allow downloading artifacts with http clients

### LINKED ISSUE #470: Allow downloading items with curl
To download an artifact from CAS, in addition to being able to do it with the CLI, you can use the `/download` HTTP endpoint in the control-plane. 

If you open the link below in your browser you'll get redirected and the download will start.

https://cp.chainloop.dev/download/sha256:744ffb7e09fca615818a151745ef872903ae79880f966efb805da736369b1d98

but that's not the case if you use `curl` and authenticate with an API token (new feature #461) 

```
curl -L -H "Authorization: Bearer $API_TOKEN" https://cp.chainloop.dev/download/sha256:744ffb7e09fca615818a151745ef872903ae79880f966efb805da736369b1d98
```

The result of running that is a mere `Your download will begin shortly...` but does not redirect and hence the download process never starts.

It's likely that the issue is related to the use of meta-redirect headers, a feature potentially not supported by clients like `curl`.

migmartri:
image

Closes #470

### DIFF
--- a/app/controlplane/internal/service/casredirect.go
+++ b/app/controlplane/internal/service/casredirect.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	pb "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1"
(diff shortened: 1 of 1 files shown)


danlishka (approved):
LGTM

Issue #466, opened 2023-12-17
add authorization via casbin to API tokens


migmartri:
Currently, there is no authorization layer in Chainloop; since a user or API token is authenticated, it has access to most of the operations in its control organization.

This is especially problematic when we talk about API-tokens, which should be crafted with a limited purpose.

My proposal would be to take a look at https://casbin.org/ along with its [ent adapter](https://github.com/casbin/ent-adapter) and focus first on adding authorization to the API-tokens.

The reason to do it first with auth tokens is because in fact there are actions we do not want those tokens to be able to do, for example, we don't want them to add other tokens, we are avoiding this by checking that we are authentication via user

```go
        // This is a API operation that requires actual user to be logged in not API token
        // TODO: replace with authz layer, i.e casbin policies
        _, err = requireCurrentUser(ctx)
        if err != nil {
                return nil, err
        }
```

but this, instead, should be done by asking the casbin enforcer if we can do that operation instead.

```[tasklist]
### Tasks
- [x] Add casbin library + policy backend setup integrated with ent https://github.com/chainloop-dev/chainloop/pull/474
- [ ] Define a set of properties and limit scope of tokens to that, i.e workflow-contracts:create
- [ ] Add these policy definitions to database
- [ ] Start enforcing in each API call
```

PR #474, opened 2023-12-22
feat: authorization backend for API tokens

migmartri:
Adds authorization backend using casbin attached to API-tokens. It roughly has the following pieces:

### Authorization module

It includes an authorization module where 

- We list the resources, actions, and policies that an API token can have access to i.e. `list workflow contracts` or `get upload credentials`
- A factory method configuring a cabin enforcer using a local permissions model and policies stored in the DB through the `ent` ORM adapter.
- A method to associate policies to an API tokens
- Another one to clear them up

In the following demo you can see how the underlying policies are attached/removed during API-token creation/revocation

policy-handling

## API token business logic (biz.APIToken)

- During the API token creation, a set of default policies are associated with it. This will allow new tokens to go through the middleware successfully
- When the token is revoked, we clear their associated policies, too.
- Since we don't have policies management yet, where a user could add/remove permissions to those tokens. I added a background method that will make sure all the existing API tokens are in sync with the default policies. This allows us to support background compatibility for existing tokens when adding new policies.

## Middleware

Each API call goes through a new authorization middleware which

- Contains a list of API operations with their corresponding policies.
- If the API operation is not in the map, we return forbidden. In practice, this means that only the following operations are allowed: a) list, describe, and update workflow contracts, b) get CAS credentials, and c) use the discovery endpoint for control gates  
- If they are in the map, we check that the authenticated API token has permissions (policies) associated with them.

In this demo, you'll see how only the allowed routes go through

middleware-interception

Refs #466


danlishka (approved):
LGTM. Thanks for walking me through the diff.

app/controlplane/internal/authz/authz.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+//go:embed model.conf
+var modelFile []byte
+
+type Enforcer struct {
+	*casbin.Enforcer
+}
+
+func (e *Enforcer) AddPolicies(sub *SubjectAPIToken, policies ...*Policy) error {
```
buccarel, line 75: I am guessing there is no chance for `e` to be nil, right?

buccarel (approved):

app/controlplane/internal/authz/authz_test.go
```diff
@@ -0,0 +1,163 @@
...
+			subject: &SubjectAPIToken{
+				ID: uuid.NewString(),
+			},
+			policies: []*Policy{
+				PolicyWorkflowContractList,
+				PolicyReferrerRead,
+			},
+		},
+		{
+			name: "another happy path",
```
buccarel, line 60: it could be nicer to state what path these are. "happy path" and "another happy path" are not identifying the test cases

app/controlplane/internal/authz/authz_test.go
```diff
@@ -0,0 +1,163 @@
...
+	assert.Len(t, got, 0)
+	// but the other user should still have 2
+	got = enforcer.GetFilteredPolicy(0, sub2.String())
+	assert.Len(t, got, 2)
+}
+
+func testEnforcer(t *testing.T) (*Enforcer, io.Closer) {
+	policyFilepath := filepath.Join(t.TempDir(), "policy.csv")
+	// create the file if it doesn't exist
+	f, err := os.Create(policyFilepath)
```
buccarel, line 155: small note: if the file exists, it will be truncated. If this is intentional, ok. The truncation can happen if you call `testEnforcer` against the same `t` multiple times, as `t.TempDir()` will return the same directory given the same `t`.

app/controlplane/internal/authz/authz_test.go
```diff
@@ -0,0 +1,163 @@
...
+				PolicyWorkflowContractUpdate,
+				PolicyArtifactDownload,
+			},
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			enforcer, closer := testEnforcer(t)
+			defer closer.Close()
```
buccarel, line 76: Small thing: the inner `t` is shadowing the outer `t`. Maybe it's fine, but I'm just pointing it out, just in case you intended something else.

Small thing n2: you're closing many files at the end of the test function, not at the end of each iteration loop. Even this can be super fine, just pointing it out that you're keeping `len(testcases)` files open, one in its own temp directory, until the end of the test.

buccarel (approved):

app/controlplane/internal/authz/authz.go
migmartri: Technically, there is a chance for a nil receiver. I don't handle it because I think it's pretty idiomatic in this specific scenario to let it panic. For example, look at [this method](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/http/client.go;l=589) in the standard library.

So I'd rather keep it like this, but I'd like to know if you think otherwise.

app/controlplane/internal/authz/authz_test.go
migmartri: You are right, this is sloppy. I'll fix it. 

Thanks!

app/controlplane/internal/authz/authz_test.go
migmartri: Thanks, will review this.

I am curious why the linter didn't caught the shadowing though, maybe because it's in a testing file?

app/controlplane/internal/authz/authz_test.go
migmartri: I'll randomize this. Thanks!

migmartri (commented):
Thanks @buccarel. I'll apply some of your suggestions. Stay tuned!

app/controlplane/internal/authz/authz_test.go
buccarel: Ohh yes you're right! I focused on the `for` and I didn't notice there is a `func` there!

migmartri: No worries. Thanks for raising the possibility of mishandling. Usually, I overlook cleanup procedures.

PR #484, opened 2024-01-29
fix(authZ): propagate new policies across replicas

### LINKED ISSUE #483: bug(authz): policies are not propagated across all the controlplane instances
casbin enforcers are initialized in each instance of the controlplane, and these are backed by a common DB table.

Although they share that DB it seems that `enforcer.Enforce` is not being refreshed with the value in the DB. So for example, if some policies are created by the replica `a`, these policies will not be available to replica `b` until a restart.

migmartri:
Adds a watcher that triggers the reload of the policies when any other casbin replica has injected/removed a policy in the DB. 

Fixes #483 

cc/ @fgallina

### DIFF
--- a/app/controlplane/internal/authz/authz_test.go
+++ b/app/controlplane/internal/authz/authz_test.go
@@ -14,14 +14,18 @@
 // limitations under the License.
 
 // Authorization package
-package authz
+package authz_test
 
 import (
 	"fmt"
 	"io"
 	"os"
 	"testing"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
+	"github.com/chainloop-dev/chainloop/app/controlplane/internal/authz"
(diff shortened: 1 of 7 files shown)


app/controlplane/internal/authz/authz_test.go
```diff
@@ -160,13 +164,68 @@ func TestClearPolicies(t *testing.T) {
...
 		require.FailNow(t, err.Error())
 	}
 
-	enforcer, err := NewFiletypeEnforcer(f.Name())
+	enforcer, err := authz.NewFiletypeEnforcer(f.Name())
 	require.NoError(t, err)
 	return enforcer, f
 }
+
+func TestMultiReplicaPropagation(t *testing.T) {
```
migmartri, line 178: This is the test that handles propagation

danlishka (approved):
LGTM

Issue #479, opened 2024-01-27
unexpected material type cyclonedx


migmartri:
I've a contract that expects a cycloneDX SBOM, but after generating it via  `syft 0.101.1` 

```
$ syft scan ghcr.io/chainloop-dev/chainloop/control-plane:v0.55.0 -o cyclonedx-json > ~/Desktop/sbom.cdx.json
```

Chainloop att fails with the following error.

```
$ chainloop att add --name frontend-sbom --value ~/Desktop/sbom.cdx.json 
ERR adding material: crafting material: invalid cyclonedx sbom file: unexpected material type
```

See attached an example of the generated SBOM  [sbom.cdx.json](https://github.com/chainloop-dev/chainloop/files/14073887/sbom.cdx.json)

migmartri:
The issue is likely because it's a cycloneDX 1.5 format. Refs #210 
```
{
  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
```

PR #485, opened 2024-01-29
chore(deps): Bump the anchore/sbom-action version

danlishka:
Bump anchore/sbom-action from v0.14.3 to v0.15.6 to use the latest version of syft.

Refs #479


migmartri (approved):
LGTM! 

Remember, in another patch you should also upgrade Chainloop's CLI that supports the new CycloneDX 1.5 output

PR #487, opened 2024-01-30
feat: release CLI as container image

migmartri:
This patch extends our release process to also contain the CLI encapsulated in a container image. 

The main reason to do this is to have a simple mechanism to explore the integration of Chainloop in a Dagger pipeline without the need for refactoring on our end (extracting attestation libraries outside of `internal` module path).


danlishka (commented):
LGTM!

danlishka (approved):

Issue #494, opened 2024-02-09
feat: remote attestation support


migmartri:
Allow performing attestations without the need to share a local state (/tmp/attestation.tmp.json) between `init,` `add`, and `push` states.

This will make the app work best in stateless envs like Dagger modules. https://github.com/chainloop-dev/labs/issues/3

PR #495, opened 2024-02-09
feat(api): remote attestation state API

migmartri:
Implements a remote attestation state API. The attestation state is associated with a workflow run. 

- Stores the encrypted-at-rest version of the state in the workflow run database table
- It gets cleared on workflowrun expiration, completion and state reset methods

[Screencast from 2024-02-10 17-03-07.webm](https://github.com/chainloop-dev/chainloop/assets/24523/ef7774ab-753c-4df0-bafc-54e40ecf740c)

The next patch will add experimental support to use this API by the CLI.

refs #494


app/controlplane/api/controlplane/v1/attestation_state.proto
```diff
@@ -0,0 +1,70 @@
...
+syntax = "proto3";
+
+package controlplane.v1;
+
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1;v1";
+
+import "validate/validate.proto";
+
+// This service is used by the CLI to generate attestation
+service AttestationStateService {
```
migmartri, line 25: This API endpoint basically implements 

https://github.com/chainloop-dev/chainloop/blob/bec4b2a73a2abd529f8c66f8a26dd8ce94973fb9/internal/attestation/crafter/crafter.go#L43-L54

app/controlplane/internal/biz/attestationstate.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	return run, nil
+}
+
+// The following code is in charge of symmetric encryption and decryption of the attestation state
+// The only purpose is to have encryption at rest in the database
+const (
+	saltSize   = 16
+	iterations = 10000
```
danlishka, line 160: Why 10k? https://en.wikipedia.org/wiki/PBKDF2

I would also prefer to have the encryption/decryption logic in a separate module, if possible, so it can be reused later.

app/controlplane/internal/service/attestationstate.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := s.uc.Reset(ctx, robotAccount.WorkflowID, req.WorkflowRunId); err != nil {
+		return nil, handleUseCaseErr("state", err, s.log)
+	}
+
+	return &cpAPI.AttestationStateServiceResetResponse{}, nil
+}
+
+// In order to encrypt the state at rest, we will encrypt the state using a passphrase
+// which comes in the request and is used to derive an encryption key.
+// The passphrase that we'll use is the robot-account token that only the workload should have.
```
danlishka, line 117: It would be great to use something unique per workload, like some kind of machineguid + token. Should we add a todo here to review it later?

app/controlplane/internal/biz/attestationstate.go
```diff
@@ -0,0 +1,236 @@
...
+	state := &v1.CraftingState{}
+	if err := proto.Unmarshal(decryptedState, state); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal attestation state: %w", err)
+	}
+
+	return &AttestationState{State: state}, nil
+}
+
+func (uc *AttestationStateUseCase) Reset(ctx context.Context, workflowID, runID string) error {
+	run, err := uc.checkWorkflowRunInWorkflow(ctx, workflowID, runID)
```
danlishka, line 116: I am not sure about workflowID in this method. I was confused about why we need it here , but I don't have any better suggestions at the moment :)

Some questions I asked myself?
- If we want to check whether the runID belongs to workflowId, should we do it on the service level?
- The checkWorkflowRunInWorkflow method looks to me more like a helper to find a wfRun object, and workflowId is just something extra there. Should we rename it?
- Wouldn't using an encryption key suffice in preventing any attempts of spoofing? I know we use the robot account today, and it might be shared

app/controlplane/internal/biz/attestationstate.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil, errors.New("value too large")
+	}
+
+	ciphertext := make([]byte, aes.BlockSize+len(plaintextWithMagic))
+	iv := ciphertext[:aes.BlockSize]
+	if _, err := io.ReadFull(rand.Reader, iv); err != nil {
+		return nil, err
+	}
+
+	stream := cipher.NewCFBEncrypter(block, iv)
```
danlishka, line 199: I would document the reasons for choosing CFB and whether any best practices, patterns, or examples for symmetric encryption were followed here.

danlishka (approved):
LGTM to unblock it but please check my comments

app/controlplane/internal/service/attestationstate.go
migmartri: Added

app/controlplane/internal/biz/attestationstate.go
migmartri: > If we want to check whether the runID belongs to workflowId, should we do it on the service level?

we definitely could, but I am trying to make sure this logic is also easily testable, and our `biz` layer is easy to test both unitarily and functionally. Another reason to add it here is to keep the service layer to just do request handling.

> The checkWorkflowRunInWorkflow method looks to me more like a helper to find a wfRun object, and workflowId is just something extra there. Should we rename it?

That's true, the fact that it returns a run is confusing. Will change it. Thanks!

> Wouldn't using an encryption key suffice in preventing any attempts of spoofing? I know we use the robot account today, and it might be shared

Could you elaborate on how this would work?

PR #496, opened 2024-02-11
chore(api): move attestation API to common package

migmartri:
This patch just moves the attestation document definition from the CLI package to the common `internal/attestation/crafter`. 

The reason to move it is that now the controlplane also will use this definition in its state API #494


danlishka (approved):
LGTM

PR #497, opened 2024-02-11
chore(cli): make attestation cmds context aware

migmartri:
Makes the client-side attestation process context aware. This will fit well with the new remote state API #494


danlishka (approved):
LGTM

PR #499, opened 2024-02-12
feat(cli): remote attestation state support

### LINKED ISSUE #494: feat: remote attestation support

migmartri:
This patch updates the CLI to support the use of a remote state manager for the attestation, making the attestation process in that case stateless in the client.

This feature is disabled by default, to enable it you need to pass the `--remote-state` flag, for example.

```
$ chainloop att init --remote-state 
│ Initialized At    │ 12 Feb 24 15:50 UTC                  │
│ Attestation ID    │ c8bac112-508c-4ab8-8ff2-9b7f22714628 │
│ Name              │ test                                 │
```

This instance of the attestation is being stored now in Chainloop, and to proceed with the attestation process, another flag needs to be provided `--attestation-id` pointing to the value returned by the init process.

For example, to add materials, now:

```
$ chainloop att add --name artifact --value ./main.go --remote-state --attestation-id c8bac112-508c-4ab8-8ff2-9b7f22714628
```

same for reset or push commands.

### Programmatic support

One of the benefits of this new method, is that we could enable attestation processes that span over different jobs (which do not share state). The only parameter all these steps will need to share is the `attestation-id`. This value can be programmatically exposed by for example our github action as an output.  

To make that possible, this patch has also improved the json output of the command by adding a top-level attestationID property.

```
$ att init --remote-state -o json
{
   "attestationID": "fbbfdec6-6183-4062-a703-036dfff034c7",
   "initializedAt": "2024-02-12T16:10:35.326295159Z",
   "workflowMeta": {
      "WorkflowID": "7cf5da85-2508-4f6c-bc60-35ca5e65fa85",
      "Name": "test",
      "Team": "",

```

Closes #494


app/cli/internal/action/attestation_status.go
```diff
@@ -83,8 +86,8 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
 	workflowMeta := att.GetWorkflow()
 
 	res := &AttestationStatusResult{
+		AttestationID: workflowMeta.GetWorkflowRunId(),
```
danlishka, line 89: We may want to review this naming at some point. Sometimes it is confusing and I understand workflow run vs attestation

danlishka (approved):
LGTM. Looking forward to enabling a remote state in my workflows!

Issue #501, opened 2024-02-13
feat: Create Dagger Module


migmartri:
Explore the creation of a [Dagger.io](http://Dagger.io) module for Chainloop.

This module should follow the [new version](https://docs.dagger.io/zenith/) of the Dagger engine that seems to improve encapsulation and composition.

As a first step, we could focus on a module written in Go, that could perform [Chainloop attestations](https://docs.chainloop.dev/getting-started/attestation-crafting).

Once we have the module, we should look into publishing it into the [daggerverse](https://daggerverse.dev/)

Some useful links

* [daggerverse](https://daggerverse.dev/)
* [module building docs](https://docs.dagger.io/zenith/developer/go/457202/test-build-publish)
* [Discord Channel](https://discord.com/channels/707636530424053791/1120503349599543376)

migmartri:
Pushed an initial version of the code, ptal

migmartri:
👍

PR #504, opened 2024-02-13
chore: upgrade module to support dagger 0.9.9+

migmartri:
Our module wasn't compatible with Dagger 0.9.9+. Once of the [changed features](https://discord.com/channels/707636530424053791/1120503349599543376/1205234119898828838) was the way `Optional` arguments are set.

This patch both updates the code and syncs the generated code `dagger develop`

refs #501


danlishka (approved):
LGTM

PR #505, opened 2024-02-14
feat(dagger): leverage secret types

migmartri:
Makes improvements in both the dagger module and its release process.

- The module now leverages `Secret` arguments instead of passing its content by value, this will prevent secret leaking in logs and so on. It also gives flexibility to the user to pick to provide for example the cosign key from an env variable or from a file.
- I removed the license header because it seems that's prepended in the [daggerverse entry](https://daggerverse.dev/mod/github.com/chainloop-dev/chainloop/extras/dagger@cfd5569d30e14592cb48dfc19ebb648e79b603d3)
- Improved the linter to also check this module

Refs #501


danlishka (approved):
LGTM

Issue #511, opened 2024-02-14
bug(dagger): attestation-add issue


migmartri:
AttestationAdd now assumes that what you are providing is a file, that's not the case always i.e `CONTAINER_IMAGE`

Another issue we might run into is that for `CONTAINER_IMAGE` we resolve the full image, we might not be able to do it in the dagger case because "in there" there are no docker creds available.

PR #512, opened 2024-02-14
feat(dagger): support non-artifact materials

migmartri:
Allow providing material types of other kinds other than artifacts refs #511


danlishka (approved):
LGTM

PR #513, opened 2024-02-15
feat(attestation): explicit OCI credentials support

### LINKED ISSUE #511: bug(dagger): attestation-add issue

migmartri:
This patch adds support to providing OCI credentials via command-line flags or env variables, that will be used to authenticate during `CONTAINER_IMAGE` attestations.

Before, `chain loop att add` was loading the default docker credentials stored in `~/.docker/config,` and although this is still the default behavior, now, optionally, the credentials can be explicitly set.

This is important because of our push to support stateless, encapsulated, attestation processes #501 

This is also a requirement to fix #511

### DIFF
--- a/extras/dagger/main.go
+++ b/extras/dagger/main.go
@@ -55,7 +55,6 @@ func (m *Chainloop) AttestationStatus(ctx context.Context, attestationID string)
 // The file type is required for materials of kind ARTIFACT that are uploaded to the CAS
 func (m *Chainloop) AttestationAdd(
 	ctx context.Context,
-	attestationID string,
 	// material name
 	name string,
 	// path to the file to be added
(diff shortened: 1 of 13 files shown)


extras/dagger/main.go
```diff
@@ -55,6 +55,7 @@ func (m *Chainloop) AttestationStatus(ctx context.Context, attestationID string)
 // The file type is required for materials of kind ARTIFACT that are uploaded to the CAS
 func (m *Chainloop) AttestationAdd(
```
migmartri, line 56: don't review this file. It will be removed from this patch.

danlishka (approved):
LGTM

PR #514, opened 2024-02-15
feat(dagger): support explicit OCI credentials

### LINKED ISSUE #511: bug(dagger): attestation-add issue

migmartri:
Adds support to explicit OCI credentials during material addition. It leverages the changes performed here https://github.com/chainloop-dev/chainloop/pull/513

NOTE: This change should be held until a new release of the CLI is pushed

Closes #511


danlishka (approved):
LGTM

Issue #509, opened 2024-02-14
feat: dagger module effects on attestation


migmartri:
We have a [preliminary version](https://github.com/chainloop-dev/chainloop/tree/main/extras/dagger) of a Dagger module. There are a couple of things we need to look into to understand its behavior in production.

### Compatibility with current CI runners. 

Let's say that you are running your workflows in GitHub actions with `runnerType: GITHUB`. Would running the same workflow but using the dagger module instead have the same effect. Specifically, would it detect that it's running in GitHub action and expose the same information as if it was run directly with our CLI?

### Git Subject support

Chainloop automatically adds git information to its attestations if it detects that the attestation is happening in a repo. Would this information be available also if it's ran inside the module?

migmartri:
The answer to both questions are no and no. Neither the env variables nor the git repo info is injected

PR #517, opened 2024-02-16
feat(dagger): support attestation of git repositories

migmartri:
The dagger init command now supports receiving a git repository that can be used by chainloop to automatically gather information during the attestation process.

I've also added the `--revision` feature too

Refs #509


danlishka (approved):
LGTM

PR #518, opened 2024-02-16
feat(cli): automatically discover runner context

### LINKED ISSUE #515: feat: automatically discover runner type
Currently, in order to enforce and gather information about a runner type, the operator needs to add it to the contract.

This task is about splitting the context gathering part of the runner type feature from the enforcement. 

In other words, Chainloop should be able to figure out what runner type an attestation is running into, and enforce it if required

migmartri:
Decouple the detection from the enforcement of runner types.

Now that the CLI will

- detect the runner type
- check that's the expected one, if listed in the contract
- load the required information associated with that runner context.

Closes #515

### DIFF
--- a/internal/attestation/crafter/runner.go
+++ b/internal/attestation/crafter/runner.go
@@ -86,7 +86,7 @@ func discoverRunner(logger zerolog.Logger) SupportedRunner {
 			detectedStr = append(detectedStr, d.ID().String())
 		}
 
-		logger.Info().Strs("detected", detectedStr).Msg("multiple runners detected, incongruent environment")
+		logger.Warn().Strs("detected", detectedStr).Msg("multiple runners detected, incongruent environment")
 		return runners.NewGeneric()
 	}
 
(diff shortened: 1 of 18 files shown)


internal/attestation/crafter/runner.go
```diff
@@ -38,23 +40,59 @@ type supportedRunner interface {
...
+	if r, ok := RunnersMap[t]; ok {
+		return r
 	}
+
+	return runners.NewGeneric()
+}
+
+// Discover the runner environment
+func discoverRunner() SupportedRunner {
+	for _, r := range RunnersMap {
```
danlishka, line 70: We may consider this in the future, but it might be worth checking environments like Dagger first.

danlishka (approved):
LGTM

internal/attestation/crafter/runner.go
migmartri: Yes, this is quite naive. For now I've added a warning if we find more than one environment that matches, plus a fallback to generic.

PR #520, opened 2024-02-16
fix(controlplane): detect environment in server

migmartri:
During this task #515, I missed this last bit which is to also update the state on the server on initialization.

Without this patch, the attestation and the information is automatically added but the runner type in the server is not marked as such automatically


danlishka (approved):
thanks

PR #525, opened 2024-02-18
feat(controlplane): make organization names unique

### LINKED ISSUE #523: feat: make organization-name unique
During https://github.com/chainloop-dev/chainloop/pull/449 we made the organization name DNS-compatible. This means that they can be used in URLs. 

What's left to make sure those names are unique is to

- enforce their uniqueness at the database level
- during automatic creation, make sure we pick one that's unique
- validate them during update

migmartri:
- Adds a database constraint to make sure the organization names are unique
- migrates the existing data to make sure it meets such constrant
- the process of automatically generating orgs, now makes sure the org doesn't exist by checking for conflicts and retrying if needed.
- org.Create checks for duplicates too.

Closes #523


danlishka (approved):
Love it, thanks!

PR #529, opened 2024-02-19
fix(controlplane): handle org update duplicates

### LINKED ISSUE #528: handle duplicated org during name update
Gracefully manage the case where the name of the organization is taken, similarly to the way we do it here #523

migmartri:
When updating an org, instead of returning a generic error, now the API returns a validation error if there is an org with the same name to the one we're trying to set

Closes #528


danlishka (approved):
LGTM

Issue #532, opened 2024-02-20
dagger: write example


migmartri:
Write an example in the repository of how to use Chainloop module both from Github Action or as a dagger pipeline.

PR #533, opened 2024-02-21
docs(dagger): update module readme

migmartri:
Adds information in the module about how to use it as part of a pipeline, pointing to our existing example.

Refs #532
Refs #501


danlishka (approved):
LGTM

PR #537, opened 2024-02-22
chore: improve org name error message

### LINKED ISSUE #536: fix: improve organization update error codes
- When the format is invalid, the error message is too long
- When the org exists, it returns a wrong status code

migmartri:
The other problem listed in the issue #536 was already tackled but not released

Closes #536


danlishka (approved):
LGTM

Issue #535, opened 2024-02-22
Design simple read/write roles for users in organization


migmartri:
We added an authorization backend that's currently configuring ACL for `apiTokens`, we want to add some basic authorization to users too. 

In this case though I am thinking that it might make more sense to look into roles as a first step. i.e `admin`, `writer`, `reader`.

This task is about evaluating if/what those roles could be, how they could be implemented with casbin and configured at the membership level. 
```[tasklist]
### Tasks
- [x] Add support for RBAC
- [x] Define viewer role and resource/action policies mapping
- [x] Load those policies into the policies DB
- [x] Define requests mapping
- [x] Apply authz checks in middleware
- [x] role management on membership invite
- [x] limit download permissions
- [x] expose role information in membership API
- [x] limit invitation mechanism
- [x] limit delete mechanism
```

migmartri:
We are going to implement RBAC that will work alongside the current ACL setup used in the api-tokens 

![Image](https://github.com/chainloop-dev/chainloop/assets/24523/38f651f3-8810-408a-ba9c-c8be188e48de)

In practice, this means that user memberships will have associated roles, and for the first version, we are going to have only `admin` and `viewer` since the short-term goal is to have read-only users.

With regards to the resource/action/role mapping, we've a WIP document with the mapping which can be found [here](https://docs.google.com/document/d/1y_dAR40Zo3WSbYbusxrGWkiVDvweIpAZtx3gQofPM8U/edit?usp=sharing)

migmartri:
We can define inheritance by using RBAC grouping policies, so for example

something like 

```go
// role: viewer can read and write workflow contracts
enforcer.AddPolicy("role:viewer", "workflow_contract", "list")
enforcer.AddPolicy("role:viewer", "workflow_contract", "read")
// role:editor can edit the contract
enforcer.AddPolicy("role:editor", "workflow_contract", "edit")
// We indicate that role:editor should inherit all the permissions from viewer
enforcer.AddGroupingPolicy("role:editor", "role:viewer")
// and finally assign the role editor to the user `deadbeef`
enforcer.AddGroupingPolicy("membership:deadbeef", "role:editor")
```

this ends up in the following policy table

```
 id  | ptype |         v0          |        v1         |  v2  | v3 | v4 | v5 
-----+-------+---------------------+-------------------+------+----+----+----
 155 | p     | role:viewer         | workflow_contract | list |    |    | 
[...]

PR #552, opened 2024-02-29
feat(authz): implement read-only viewer role

migmartri:
This patch defines and enforces a read-only viewer role that can be set in the `membership` entity (and manageable in the future).

In practice, this change should have not functional effect in the existing users, it will only take effect when the role value in a membership is changed to `role:org:viewer`, which can only be done directly in the DB for now.

## Resources, Policies and API enforcement

[This document](https://docs.google.com/document/d/1y_dAR40Zo3WSbYbusxrGWkiVDvweIpAZtx3gQofPM8U/edit) describes what action and resources are allowed within any given `role`. And this patch is an implementation of it which specifically:
 
- Defines in `authz` module the resources and policies needed for the role `org:viewer`
- Makes sure the policies are synched in the policies backend.

```
 id  | ptype |       v0        |           v1           |   v2   | v3 | v4 | v5 
-----+-------+-----------------+------------------------+--------+----+----+----
 251 | p     | role:org:viewer | referrer               | read   |    |    | 
 252 | p     | role:org:viewer | cas_artifact           | read   |    |    | 
 253 | p     | role:org:viewer | cas_backend            | list   |    |    | 
 254 | p     | role:org:viewer | integration_available  | list   |    |    | 
 255 | p     | role:org:viewer | integration_available  | read   |    |    | 
 256 | p     | role:org:viewer | integration_registered | list   |    |    | 
 257 | p     | role:org:viewer | integration_attached   | list   |    |    | 
 258 | p     | role:org:viewer | metrics_org            | list   |    |    | 
 259 | p     | role:org:viewer | robot_account          | list   |    |    | 
 260 | p     | role:org:viewer | workflow_contract      | list   |    |    | 
 261 | p     | role:org:viewer | workflow_contract      | read   |    |    | 
 262 | p     | role:org:viewer | workflow_run           | list   |    |    | 
 263 | p     | role:org:viewer | workflow_run           | read   |    |    | 
 264 | p     | role:org:viewer | workflow               | list   |    |    | 
 265 | p     | role:org:viewer | organization           | read   |    |    | 
 266 | p     | role:org:admin  | cas_artifact           | create |    |    | 
 267 | g     | role:org:admin  | role:org:viewer        |        |    |    | 
 268 | g     | role:org:owner  | role:org:admin         |        |    |    | 
```
The stored policies contain all resource/actions associated with the `viewer` role as well as grouping relationships `owner => admin => viewer` 

- Maps API calls and their required policies. 
- Updates the authorization `middleware` to take effect not only `api tokens` (as it was before) but also `users`.

## Membership role management

Now the membership model has a `role` attribute stored in the database. This attribute is set to

- `role:org:admin ` for the existing memberships in the DB. This is for backward compatibility.
- `role:org:owner` when a user creates an org and or for the org associated automatically during the signup process. 
- `role:org:admin` if the user was invited to an org. Note: this will be configurable in another patch.

Refs #535 

cc/ @fgallina


danlishka (approved):
LGTM

PR #553, opened 2024-03-01
feat(controlplane): scope invitation system

migmartri:
Make the invitation system scoped to the `currentOrganization` instead of allowing the user to provide the org ID.

This is so this API endpoint also fits into the new authorization layer #535 and can be properly secured. With the new change, we can now make sure that only admins can invite.


app/controlplane/internal/data/ent/migrate/migrations/20240229202352.sql
```diff
@@ -2,7 +2,7 @@
 ALTER TABLE "memberships" ADD COLUMN "role" character varying;
 
 -- Set the existing memberships to the role "role:admin" to not to break compatibility
-UPDATE "memberships" SET "role" = 'role:admin';
+UPDATE "memberships" SET "role" = 'role:org:admin';
```
migmartri, line 5: note: this role was invalid, sorry I missed it. 

I updated it here since it has not been released yet.

danlishka (approved):
LGTM

PR #555, opened 2024-03-02
chore(controlplane): continue invite system org scope

migmartri:
This is a continuation of #553 but this time applying to `listing` and `revocation`

Now that we are going to have `admin` users in the org, this change will allow us for each admin to be able to see the pending/sent invites in their org, and match it with a list of members in the future.

Refs #535


danlishka (approved):
LGTM

Issue #558, opened 2024-03-03
feat: org-level membership info


migmartri:
Currently, a user can see what organization they belong to but it's impossible to see what other members are in an organization.

This issue is about figuring out an `organization` scoped membership functionality

## List members in current org

- Show members/emails and their roles
- Make sure this is only available for `admin/owners` or the orgs

## Update member role

- Being able to update a member role, note that we can't update our own role and this feature is only available for admin/owners.

PR #559, opened 2024-03-03
refactor: move membership API service to user

migmartri:
As part of #558 we are going to be creating new API endpoints that operate on org-scoped memberships. 

To leave room to those, this patch moves the existing membership-related API operations to an `user` service, since in fact the current membership handling is user-scoped.

This patch just exposes a new gRPC service `user` and mounts the existing endpoints there.


danlishka (approved):
LGTM

PR #560, opened 2024-03-03
feat(controlplane): org membership API

migmartri:
API implementation of method to 

- List members of the current organization
- Delete members from this org (except myself)
- Update members roles (`admin`, `viewer`) of members of my org (except myself)

Refs #558


danlishka (approved):
LGTM!

PR #565, opened 2024-03-04
fix: allow viewers to change or leave the org

migmartri:
#559 introduced a regression where did not allow a `viewer` to change organizations.

This patch fixes this plus enables such user role to also `leave` the org


danlishka (approved):
LGTM

PR #567, opened 2024-03-04
feat(cli): organization membership

### LINKED ISSUE #558: feat: org-level membership info

migmartri:
Organization-scoped membership management

You can now list the members using `chainloop org member list`

image

update a member 

```
chainloop org member update --id [ID] --role viewer|owner|admin 
```

or delete a member

```
chainloop org delete --id [ID] 
```

Closes #558


danlishka (approved):
LGTM

PR #571, opened 2024-03-05
chore(cli): remove invitation from top-level org cmd

migmartri:
This command is now under `org member` subcommand


danlishka (approved):
LGTM

PR #582, opened 2024-03-08
chore(ci): disable go linter in Dagger module

migmartri:
temporarily disable the linter until we stabilize the module development.


migmartri:
wrong PR

PR #585, opened 2024-03-08
chore: rename pagination options

migmartri:
rename current `cursor` pagination to give room to offset pagination.


danlishka (approved):
LGTM

PR #592, opened 2024-03-11
feat(controlplane): support providing connection string

### LINKED ISSUE #562: feat(authz): forward parameters to upstream oauth2 provider
The oauth dance experience can be tweaked in some oauth providers by providing additional parameters.

For example, in auth0, you can provide a [`connection` parameter](https://auth0.com/docs/api/authentication#social) to select the social provider and skip the popup automatically.  

It could be useful for the controlplane auth endpoint to accept these kind of parameters that can be forwarded upstream to the configured oauth 2 idP

migmartri:
Support for providing `connection` string during login that gets forwarded to the upstream oauth2 provider https://auth0.com/docs/api/authentication#login

Closes #562

### DIFF
--- a/app/controlplane/internal/service/auth.go
+++ b/app/controlplane/internal/service/auth.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023 The Chainloop Authors.
+// Copyright 2024 The Chainloop Authors.
 //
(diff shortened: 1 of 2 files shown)


danlishka (approved):
LGTM

Issue #600, opened 2024-03-12
contract names uniqueness/format handling


migmartri:
We want to make sure the contract names are unique in their organizations and follow DNS1123 similar to our organization names.

- Validate on creation and update
- Migrate existing contract names

PR #601, opened 2024-03-12
feat(controlplane): make contract-names unique and DNS1123 compatible

migmartri:
- DB: 
  - requires the workflow contract name to be unique in the context of an organization
  - updates the existing contract entries to be DNS1123 compatible
- biz: 
  - the name must be a valid `DNS1123` value
  - when we auto-create a contract with a name during workflow creation, we make sure it's unique

Closes https://github.com/chainloop-dev/chainloop/issues/600

### DIFF
--- a/app/controlplane/internal/biz/workflow.go
+++ b/app/controlplane/internal/biz/workflow.go
@@ -122,7 +122,12 @@ func (uc *WorkflowUseCase) findOrCreateContract(ctx context.Context, orgID, cont
 	}
 
 	// No contractID has been provided so we create a new one
-	contractName := fmt.Sprintf("%s-%s", project, name)
+	contractName := name
+	// Project might be empty
+	if project != "" {
+		contractName = fmt.Sprintf("%s-%s", project, name)
+	}
+
 	return uc.contractUC.Create(ctx, &WorkflowContractCreateOpts{OrgID: orgID, Name: contractName, AddUniquePrefix: true})
 }
 
--- a/app/controlplane/internal/biz/workflow.go
+++ b/app/controlplane/internal/biz/workflow.go
@@ -122,8 +122,8 @@ func (uc *WorkflowUseCase) findOrCreateContract(ctx context.Context, orgID, cont
 	}
 
 	// No contractID has been provided so we create a new one
-	contractName := fmt.Sprintf("%s/%s", project, name)
-	return uc.contractUC.Create(ctx, orgID, contractName, nil)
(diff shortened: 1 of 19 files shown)


.github/workflows/lint.yml
```diff
@@ -53,7 +53,6 @@ jobs:
           buf_api_token: ${{ secrets.buf_api_token }}
       - uses: bufbuild/buf-lint-action@bd48f53224baaaf0fc55de9a913e7680ca6dbea4 # v1.0.3
 
-  temporarily disabled until we stabilize the module
```
migmartri, line 56: my bad, this was wrong

.github/workflows/lint.yml
```diff
@@ -76,5 +75,5 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          working-directory: extras/dagger/src
+          working-directory: extras/dagger
```
danlishka, line 78: Should it go to a separate dagger-related PR?

danlishka (approved):
LGTM!
Added a few comments, it is more to better understand it. Thank you!

.github/workflows/lint.yml
migmartri: Yes, probably, let me do it. 

The reason I did it here is because I couldn't make this one go green because of the regression.

app/controlplane/internal/biz/workflow.go
```diff
@@ -122,8 +122,13 @@ func (uc *WorkflowUseCase) findOrCreateContract(ctx context.Context, orgID, cont
 	}
 
 	// No contractID has been provided so we create a new one
-	contractName := fmt.Sprintf("%s/%s", project, name)
-	return uc.contractUC.Create(ctx, orgID, contractName, nil)
+	contractName := name
+	// Project might be empty
+	if project != "" {
+		contractName = fmt.Sprintf("%s-%s", project, name)
```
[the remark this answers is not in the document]
migmartri: It will append a suffix, note the `AddUniquePrefix: true` property below.

That enables a logic to retry the creation of a name with a 4 digits suffix, similar to what we do when we generate a workflow

.github/workflows/lint.yml
migmartri: Here it is https://github.com/chainloop-dev/chainloop/pull/604/files

PR #605, opened 2024-03-12
feat(controlplane): unique workflow name and formatted project

### LINKED ISSUE #602: workflow names uniqueness/format handling
We want to make sure the workflow names are unique in their organizations and both names and projects follow DNS1123.

- Validate on creation/update
- Migrate existing workflow names and projects

migmartri:
Similarly to #601 but applied to workflows, it

- Makes both workflow names and project RFC 1123 compliant
- Make sure that the workflow name is unique at the organization level
- migrates the existing workflow names and projects to meet 1

Closes #602


danlishka (approved):
LGTM!

PR #608, opened 2024-03-12
feat(controlplane): workflow contract description support

migmartri:
- Add description property to workflow contract
- Can be set during creation or update
- This patch also includes the CLI implementation


danlishka (approved):
LGTM

PR #613, opened 2024-03-13
fix(controlplane): support soft-deleted items on name constraint

### LINKED ISSUE #612: make sure uniqueness take into account soft-deleted items
We have a DB contraint in workflow and contracts that makes the `name` of those unique. Some of those entities have soft-deletion. We need to make sure soft-deletion doesn't get on the way of adding new items.

migmartri:
Take into account the `deleted_at` column in the uniqueness constraint

Closes #612

### DIFF
--- a/app/controlplane/internal/data/ent/schema/workflowcontract.go
+++ b/app/controlplane/internal/data/ent/schema/workflowcontract.go
@@ -61,7 +61,9 @@ func (WorkflowContract) Edges() []ent.Edge {
 
 func (WorkflowContract) Indexes() []ent.Index {
 	return []ent.Index{
-		// names are unique within an organization
-		index.Fields("name").Edges("organization").Unique(),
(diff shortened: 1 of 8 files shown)


app/controlplane/internal/data/ent/schema/workflowcontract.go
```diff
@@ -61,7 +61,9 @@ func (WorkflowContract) Edges() []ent.Edge {
 
 func (WorkflowContract) Indexes() []ent.Index {
 	return []ent.Index{
-		// names are unique within an organization
-		index.Fields("name").Edges("organization").Unique(),
+		// names are unique within a organization and affects only to non-deleted items
+		index.Fields("name").Edges("organization").Unique().Annotations(
+			entsql.IndexWhere("deleted_at IS NULL"),
```
migmartri, line 66: This is the important bit

PR #616, opened 2024-03-14
docs: add dagger demo

migmartri:
Include the Dagger demo in our guide in the docs.

I've implemented it so it gets injected when you run

```
yarn run docusaurus download-remote-dagger-module
```

image

refs #501


danlishka (approved):
It looks great!

Issue #32, opened 2023-03-15
implement workflow describe


migmartri:
Currently, there is no option to describe a workflow, this means that there is not a consistent way to for example see the contract that's being used. 

In fact there is a way but that requires listing the workflows and option for either a `--full` table output or `json`

```
chainloop wf ls --full
┌──────────────────────────────────────┬───────────────────┬──────────────────┬──────┬─────────────────────┬────────┬─────────────────┬─────────────────────┬──────────────────────────────────────┬──────────────────────────────────────┐
│ ID                                   │ NAME              │ PROJECT          │ TEAM │ CREATED AT          │ RUNNER │ LAST RUN STATUS │ LAST RUN AT         │ LAST RUN ID                          │ CONTRACT ID                          │
├──────────────────────────────────────┼───────────────────┼──────────────────┼──────┼─────────────────────┼────────┼─────────────────┼─────────────────────┼──────────────────────────────────────┼──────────────────────────────────────┤
│ 37022b2f-34c3-4f47-9fd7-514b4a7baaad │ build-and-release │ integration-demo │      │ 13 Jan 23 14:26 UTC │ GitHub │ success         │ 13 Jan 23 14:26 UTC │ 91093f6e-37c0-4e02-af5d-c9859596d125 │ c8f5d813-8f06-4b09-9593-f2122065d7ee │
└──────────────────────────────────────┴───────────────────┴──────────────────┴──────┴─────────────────────┴────────┴─────────────────┴─────────────────────┴──────────────────────────────────────┴──────────────────────────────────────┘
``` 

```json
$ chainloop wf ls -o json
[
   {
      "name": "build-and-release",
      "id": "37022b2f-34c3-4f47-9fd7-514b4a7baaad",
      "team": "",
      "project": "integration-demo",
      "createdAt": "2023-01-13T14:26:34.224693Z",
      "runsCount": 17,
      "contractID": "c8f5d813-8f06-4b09-9593-f2122065d7ee",
      "lastRun": {
         "id": "91093f6e-37c0-4e02-af5d-c9859596d125",
         "state": "success",
         "createdAt": "2023-03-13T23:35:36.958669Z",
         "finishedAt": "2023-03-13T23:36:37.456564Z",
         "runURL": "https://github.com/chainloop-dev/integration-demo/actions/runs/4410543365",
         "runnerType": "GitHub"
      }
   }
]
```

[...]

PR #617, opened 2024-03-14
fix(authz): add read permissions in workflow.View

migmartri:
This patch enables `viewers` to access the new `Workflow.View` API endpoint introduced here #32 

I do not think this patch will require a re-release since we do not use this endpoint yet in the CLI.
 
Before

```
grpcurl --plaintext -H "authorization: Bearer $TOKEN" -d '{"id": "429a9b9e-63cf-44b9-aa8f-27cdd6775e23" }' localhost:9000 controlplane.v1.WorkflowService.View
ERROR:
  Code: PermissionDenied
```

After

```
grpcurl --plaintext -H "authorization: Bearer $TOKEN" -d '{"id": "429a9b9e-63cf-44b9-aa8f-27cdd6775e23" }' localhost:9000 controlplane.v1.WorkflowService.View
ERROR:
  Code: NotFound
```


danlishka (approved):
LGTM!

PR #628, opened 2024-03-21
chore(api): handle context.Canceled error

### LINKED ISSUE #627: handle context.Cancelled exception
we are raising context.Canceled errors to Sentry and API responses. We might want to gracefully handle those ones.

Image

migmartri:
Gracefully handle context.Canceled error. This patch also makes sure all the error in the API are properly handled. NOTE: this will be moved to a middleware later on.

Closes #627


danlishka (approved):
LGTM

PR #638, opened 2024-03-27
fix(controlplane): handle not-found contract on update

migmartri:
Properly return 404 errors when you try to update a contract that doesn't exist. 

This is done by the service/biz layer that will check if the data layer returned any contract object. The previous code didn't do it, that's fixed now.


danlishka (approved):
LGTM

PR #639, opened 2024-03-27
feat(controlplane): always store the digest of the attestation

migmartri:
This patch makes sure that the digest of the attestation is always present. Before, only for the attestations that were stored in the CAS that was the case.


danlishka (approved):
LGTM

PR #640, opened 2024-03-29
feat(api): support set workflow visibility on creation

migmartri:
Setting the workflow visibility wasn't possible during creation through the API. This patch changes that.


danlishka (approved):
LGTM

Issue #645, opened 2024-04-02
Change contract associated with a workflow


migmartri:
Currently, it's not possible to change the contract associated with an existing workflow. Changing the contract requires updating the associated contract, but that one might be also been used somewhere else.

We should support changing the contract altogether.

PR #646, opened 2024-04-02
feat: add contract selection on workflow update

migmartri:
Closes https://github.com/chainloop-dev/chainloop/issues/645


danlishka (approved):
LGTM

PR #648, opened 2024-04-02
chore: upgrade aws-sdk

migmartri:
The latest release had an issue while being used with AWS secret manager.

The reason was a mismatch of the `secretmanager` library and its SSO companion. More info here https://github.com/aws/aws-sdk-go-v2/issues/2370

This patch just makes sure that the SDK libraries are up to date.


danlishka (approved):
LGTM

PR #653, opened 2024-04-03
fix(api): validate material type

migmartri:
The API wasn't validating that the material type provided was one of the list, it was just making sure that wasn't empty


danlishka (approved):
LGTM

Issue #658, opened 2024-04-08
bug: authorization error download file


migmartri:
I have a read-only user and I am having issues downloading files

![Image](https://github.com/chainloop-dev/chainloop/assets/24523/f0cf8376-8a60-452b-b3aa-a7f305b6ee5e)

PR #659, opened 2024-04-08
fix(controlplane): allow generating downloadURL

migmartri:
It was not possible for read-only users to generate a download URL. They could download items via the CLI but not using the redirect service.

This patch adds the missing policy rule

Fixes https://github.com/chainloop-dev/chainloop/issues/658


danlishka (approved):
LGTM

PR #661, opened 2024-04-08
fix(ci): fix dagger module linter

migmartri:
The dagger module linter is failing because a new directory has been added to the `internal` directory (telemetry) during develop.

This patch updates the gitignores, the module itself as well as the linter process to fix the issue.


danlishka (approved):
LGTM

PR #662, opened 2024-04-09
feat: add unique name to cas-backends

migmartri:
Like #523 and #643, this time is turn for the cas-backends. 

This feature adds a new `name` property to our cas-backends entities that needs to be unique and valid in the organization.

This patch also updates existing cas-backends with unique names.


danlishka (approved):
LGTM

Issue #603, opened 2024-03-12
migrate to https://github.com/bufbuild/protovalidate


migmartri:
https://github.com/bufbuild/protovalidate seems to have replaced the library we are currently using github.com/envoyproxy/protoc-gen-validate 

This task is about migrating our current source code. 
```[tasklist]
### Tasks
- [x] configs https://github.com/chainloop-dev/chainloop/pull/667
- [x] crafter https://github.com/chainloop-dev/chainloop/pull/670
- [x] CAS API https://github.com/chainloop-dev/chainloop/pull/668
- [x] ControlPlane API https://github.com/chainloop-dev/chainloop/pull/673
- [x] middleware http https://github.com/chainloop-dev/chainloop/pull/673
```

migmartri:
https://github.com/bufbuild/protovalidate/blob/main/docs/migrate.md

PR #667, opened 2024-04-11
chore: migrate configs to protovalidate

migmartri:
Migrates `config.proto` to protovalidate from `proto-gen-validate`, this means in practice

- update the buf configuration to use the new plugin
- change the annotations in the proto files
- generate the code and remove the validate.gen.go files since the new library doesn't rely on generated code
- because of ^^, now we need to use a different validation method, no more `ValidateAll`  

This is just part of the work, the difficult one will come when we migrate the API protos, since we need to find an alternative to our current middleware

refs #603


danlishka (approved):
LGTM

PR #669, opened 2024-04-11
fix: auto-create contract

migmartri:
When a contract was provided, its name was incorrect, not matching our new DNS1123 requiremnet


danlishka (approved):
LGTM

PR #670, opened 2024-04-11
chore: migrate crafter/attestations to protovalidate

migmartri:
Migrate attestations part to new validations framework

Refs #603


danlishka (approved):
LGTM

PR #673, opened 2024-04-15
chore: migrate controlplane to protovalidate

migmartri:
Migrate controlplane/api to use protovalidate, this includes

- migrate and re-build proto code
- Update validators
- Update middleware 

refs #603


danlishka (approved):
LGTM

Issue #478, opened 2024-01-19
Define declarative, gitops-based contract management


migmartri:
Now that we have api-tokens that can be used in automation to for example update contracts. We can now define gitops based contract updates in a repository.

In addition to that, since we support contracts in cue lang we could implement inheritance for simple contract management.

This task is about exploring how to configure this kind of setup, a repository that contains multiple generated contracts, that can be easily generated and applied

PR #674, opened 2024-04-16
feat(contracts): updates by name

### LINKED ISSUE #665: being able to update contracts by name

migmartri:
Allows updating a contract by referencing its name as identifier instead of the `id`. 

note that we still support `id` but it's marked as deprecated in both the API and the CLI.

Closes #665 
Refs #478


danlishka (approved):
Please update the docs:
- https://docs.chainloop.dev/getting-started/workflow-definition
- `chainloop workflow contract update ...`

Issue #684, opened 2024-04-23
feat: add Helm chart material type


migmartri:
This task is about adding support for Helm Chart material types for attestations. 

- It will be a material type that inherits from the `artifact type` (gets uploaded to CAS)
- We should validate that the received artifact is a packaged Helm chart
- We should document its existence as part of our readme and documentation page

You can see an example of how a new type can be added here https://github.com/chainloop-dev/chainloop/pull/356

Javirln:
One of the important topics here is what _validate a Helm chart_ means in this context.

migmartri:
> One of the important topics here is what _validate a Helm chart_ means in this context.

Indeed, that's the tricky part. The purpose of client validations is to add guard-rails to developers, not really security (yet). So from that POV I'd just add whatever is minimum to differentiate any other file vs a Helm Chart.

Thanks for looking into this!

Javirln:
Oki nice, so for the moment, I've added some really basic validations on the client side which is the following:

It inspects the tarball and specifically looks for to files to be present and syntactically valid which are Charts.yaml and values.yaml. Thinking about adding to the list Chart.lock as well since it's pretty straightforward.

How do you folks see it? cc @jiparis @migmartri

migmartri:
See a follow up task https://github.com/chainloop-dev/chainloop/issues/693

PR #689, opened 2024-04-23
feat(materials): Helm Chart material support

Javirln:
Add support for Helm Charts files as material types. These get validated before they are uploaded to the CAS and attached to the attestation process.

The validations checks for the following:
- Incoming file is a compressed tarball.
- `Charts.yaml` and `values.yaml` files are present on the uncompressed directory and it's a valid YAML file.

Closes https://github.com/chainloop-dev/chainloop/issues/684


migmartri:
Awesome @Javirln 

As you can see, the CI is complaining that the DCO is not completed. This means that you are not committing with the signed-off line `git commit -s ...`

I'd suggest that you squash all your commits and amend it to have the signed off. Happy to help if you have any questions

Thanks

README.md
```diff
@@ -130,6 +130,7 @@ Chainloop supports the collection of the following pieces of evidence types:
 - [OpenVEX](https://github.com/openvex)
 - [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/)
 - [JUnit](https://www.ibm.com/docs/en/developer-for-zos/14.1?topic=formats-junit-xml-format)
+- [Helm Charts](https://helm.sh/docs/topics/charts/)
```
migmartri, line 133: I'd keep it singular for consistency

```suggestion
- [Helm Chart](https://helm.sh/docs/topics/charts/)
```

internal/attestation/crafter/materials/helmchart.go
```diff
@@ -0,0 +1,128 @@
...
+	uncompressedStream, err := gzip.NewReader(f)
+	if err != nil {
+		return nil, fmt.Errorf("can't uncompress file, unexpected material type: %w", err)
+	}
+
+	// Create a tar reader
+	tarReader := tar.NewReader(uncompressedStream)
+
+	// Flags to track whether required files are found
+	chartFileValid, chartValuesValid := false, false
```
migmartri, line 74: nitpick and curious to knwo your take. I think you can technically leverage go default values instead.

```
var chartFileValid, chartValuesValid bool
```

internal/attestation/crafter/materials/helmchart.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Iterate through the files in the tar archive
+	for {
```
migmartri, line 77: does this work with an empty tarball?

README.md
migmartri, line 133: also, if you want to add the documentation in this PR. I am ok if you do it in another one, you'll need to also update https://github.com/chainloop-dev/chainloop/blob/main/docs/docs/reference/operator/contract.mdx

That's the source code of the doc site. So probably let's do it in another PR

migmartri (approved):

internal/attestation/crafter/materials/helmchart.go
jiparis: +1

internal/attestation/crafter/materials/helmchart.go
```diff
@@ -0,0 +1,128 @@
...
+	schemaapi "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
+	api "github.com/chainloop-dev/chainloop/internal/attestation/crafter/api/attestation/v1"
+	"github.com/chainloop-dev/chainloop/internal/casclient"
+	"github.com/rs/zerolog"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	ChartFileName      = "Chart.yaml"
+	ValuesYamlFileName = "values.yaml"
```
jiparis, line 36: Since these are in a shared package, I would suggest to prefix both constants with an agreed pattern to avoid conflicts. Something like `ChartFileName` and `ChartValuesFileName`.

migmartri: true, and in fact they don't need to be exposed right?

internal/attestation/crafter/materials/helmchart.go
```diff
@@ -0,0 +1,128 @@
...
+	if !chartFileValid || !chartValuesValid {
+		return nil, fmt.Errorf("missing required files in the helm chart: Chart.yaml and values.yaml")
+	}
+
+	// Upload and craft the chart
+	return uploadAndCraft(ctx, c.input, c.backend, filepath, c.logger)
+}
+
+// validateYamlFile validates the YAML file just by trying to unmarshal it
+func validateYamlFile(r io.Reader) error {
```
jiparis, line 121: Please add a receiver to this method so that it's not exposed at package level (unless we want to expose it for some reason).

Javirln: Hmm no no, we don't need to expose it. Eventually in the future this can belong to a shared/utils part to be reused in the the code. Right now, there is no need for that neither for it be exposed.

internal/attestation/crafter/materials/helmchart.go
Javirln: Added a common pattern and made them private outside the package.

migmartri:
Looks good to me! Feel free to merge it!

PR #690, opened 2024-04-23
fix(controlplane): return Notfound errror if robot account doesn't exist

### LINKED ISSUE #677: Handle 404 error in robot-accounts revoke
Currently, the notfound error doesn't seem to be properly handled at the service level for robot-accounts, raising a 500.

image

We should make sure the biz layer is returning a known NotFound error and it's handled by the service layer using `handleUseCaseErr`. You can take a look at how APITokens handle this since it's a very similar use-case.

jiparis:
This PR properly translates the `ent.NotFound` error to `biz.ErrNotFound`, ensuring that service layer will return a 404 error to the client:
```
> go run app/cli/main.go --insecure wf robot-account revoke --id a78c1e6f-c246-4862-aff9-b31946025123
ERR robot account not found
exit status 1
```

Fixes #677


jiparis:
Sure, I'll add some tests

migmartri:
same comment about DCO https://github.com/chainloop-dev/chainloop/pull/689#issuecomment-2074378135

app/controlplane/internal/biz/robotaccount_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	org *biz.Organization
+	ra  *biz.RobotAccount
+}
+
+// // Run the tests
+func TestRobotAccountUseCase(t *testing.T) {
+	suite.Run(t, new(robotAccountTestSuite))
+}
+
+func (s *robotAccountTestSuite) SetupTest() {
```
migmartri, line 68: btw, smth that I learnt not long time ago, and I know that's not consistent in other places, but you can instead of using `assert := assert.New(s.T())` and then assert.noError, you can do directly `s.NoError`)

So basically you can remove `assert` variable and instead use `s.NoError(err)`

You can probably also remove `t := s.T()` at that point too.

app/controlplane/internal/biz/robotaccount_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (s *robotAccountTestSuite) TestRevoke() {
+	ctx := context.Background()
+	s.Run("returns an error if org ID format is not valid", func() {
+		err := s.RobotAccount.Revoke(ctx, "not_valid_uuid", "52597CC8-21BF-45C4-B30B-AE2879556CE1")
+		s.ErrorAs(err, &biz.ErrInvalidUUID{})
+	})
+
+	s.Run("returns an error if robot account ID format is not valid", func() {
+		err := s.RobotAccount.Revoke(ctx, "991B156A-7636-44D2-97B0-E97070E626DC", "not_valid_uuid")
```
migmartri, line 36: any reason for hardcoding the uuid instead of using the one stored in the suite struct?

app/controlplane/internal/biz/robotaccount_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+		s.ErrorAs(err, &biz.ErrInvalidUUID{})
+	})
+
+	s.Run("returns a Not Found if robot account cannot be found", func() {
+		err := s.RobotAccount.Revoke(ctx, s.org.ID, "DCF3895A-22BE-49D0-99D4-0A76835D0BD1")
```
migmartri, line 41: in other places, instead of using a hardcoded UUID I use uuid.NewString(). 

I am curious to know why/if you prefer your approach

app/controlplane/internal/biz/robotaccount_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+		s.ErrorAs(err, &biz.ErrNotFound{})
```
migmartri, line 42: great, thanks for testing this

migmartri (approved):
awesome, added some minor comments.

app/controlplane/internal/biz/robotaccount_integration_test.go
jiparis: Thanks, I was just blindly copying from elsewhere. I've simplified it a little bit.

migmartri (approved):
LGTM!

PR #702, opened 2024-04-26
feat(chainloop): add EVIDENCE material type

### LINKED ISSUE #698: add EVIDENCE material type
This task is about adding support for Helm Chart material types for attestations.

- It will be a material type that inherits from the artifact type (gets uploaded to CAS)
- We should document its existence as part of our readme and documentation page

jiparis:
This PR adds a new material type called EVIDENCE which inherits from ARTIFACT, but it's not meant to represent a main software artifact, but a secondary, stored in CAS, piece of material. A test report, or a generated custom bill of materials are good examples for this new type.

Closes #698


javirln:
Since the EVIDENCE type cannot be set or should be set as `output: true` on the contract level, we could inform the operator by raising an error. On the `crafting_schema_validation.go`:

```golang
// ValidateEvidenceNotOutput validates that materials with type EVIDENCE are not marked as output
func (schema *CraftingSchema) ValidateEvidenceNotOutput() error {
	for _, m := range schema.Materials {
		if m.GetType() != CraftingSchema_Material_EVIDENCE {
			continue
		}

		if m.GetOutput() {
			return fmt.Errorf("material with name=%s and type EVIDENCE cannot be marked as output: true", m.Name)
		}
	}

	return nil
}
```

And then run the validation on the `crafter.go`.

Wdyt @jiparis @migmartri ??

migmartri:
> Since the EVIDENCE type cannot be set or should be set as `output: true` on the contract level, we could inform the operator by raising an error. On the `crafting_schema_validation.go`:
> 
> ```go
> // ValidateEvidenceNotOutput validates that materials with type EVIDENCE are not marked as output
> func (schema *CraftingSchema) ValidateEvidenceNotOutput() error {
> 	for _, m := range schema.Materials {
> 		if m.GetType() != CraftingSchema_Material_EVIDENCE {
> 			continue
> 		}
> 
> 		if m.GetOutput() {
> 			return fmt.Errorf("material with name=%s and type EVIDENCE cannot be marked as output: true", m.Name)
> 		}
> 	}
> 
> 	return nil
> }
> ```
> 
> And then run the validation on the `crafter.go`.
> 
> Wdyt @jiparis @migmartri ??

I do not think that in crafter.go would be enough since we might have other clients creating contracts. because of that, it might make sense to implement it server side, either with custom code as you showed or built-in using a custom protovalidate rule.

migmartri:
in any case, I am ok doing the validation in another patch, because note that we can not enable until this is in place https://github.com/chainloop-dev/chainloop/issues/703

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -84,6 +84,10 @@ message CraftingSchema {
       // https://github.com/microsoft/sarif-tutorials/blob/main/docs/1-Introduction.md
       SARIF = 9;
       HELM_CHART = 10;
+
+      // Evidences are materials that are not meant to be subject of attestations, but still uploaded
```
migmartri, line 88: I believe that `evidences` (plural) is not a correct term, I'd mention it here `Pieces Of Evidences`.

Also, I do not think that the subject part is the important attribute of it but actually what's for, so I'd update the comment to something like this

> Pieces of evidences represent generic, additional context that doesn't fit into one of the well known material types. For example, a custom approval report (in json), ...`

What do you think?

javirln (approved):
LGTM!

internal/attestation/crafter/materials/evidence.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/chainloop-dev/chainloop/internal/casclient"
+	"github.com/rs/zerolog"
+)
+
+type EvidenceCrafter struct {
+	*crafterCommon
+	backend *casclient.CASBackend
+}
+
+func NewEvidenceCrafter(schema *schemaapi.CraftingSchema_Material, backend *casclient.CASBackend, l *zerolog.Logger) (*EvidenceCrafter, error) {
```
migmartri, line 33: I'd add the comment about the purpose of this material type (see my other comment above) here too, and how similar it is from artifacts.

internal/attestation/crafter/materials/evidence_test.go
```diff
@@ -0,0 +1,155 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package materials_test
+
+import (
```
migmartri, line 18: re: linter error

Note that in other tests files we are adding an no-lint exception since the test-cases are very, very similar.

internal/attestation/crafter/materials/evidence.go
jiparis: Done

internal/attestation/crafter/materials/evidence_test.go
jiparis: Thanks, fixed.

migmartri (approved):

migmartri (changes requested):
Let's update also the readme file with the info about the new material type, it seems we only added it to the docs page

migmartri:
> Let's update also the readme file with the info about the new material type, it seems we only added it to the docs page

I've updated it :)

PR #711, opened 2024-04-26
feat(ci): Adds chainloop to codeql pipeline

### LINKED ISSUE #707: attests codeql process
see if we can also store the output, i.e sarif file.

Only run Chainloop if the current run is not provoked by a pull request.

javirln:
This PR adds chainloop CLI to codeql workflows to attest SARIF results.

Closes #707

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,6 +20,9 @@ jobs:
       actions: read
       contents: read
       security-events: write
+    env:
+      CHAINLOOP_VERSION: 0.83.0
+      CHAINLOOP_ROBOT_ACCOUNT: ${{ secrets.CHAINLOOP_ROBOT_ACCOUNT_CODEQL }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


migmartri (approved):
LGTM, but let's wait until we stabilize the other two jobs :)

.github/workflows/codeql.yml
```diff
@@ -48,3 +61,28 @@ jobs:
         uses: github/codeql-action/analyze@6a28655e3dcb49cb0840ea372fd6d17733edd8a4 # v2.21.8
         with:
           category: "/language:${{matrix.language}}"
+        outputs:
+          sarif-output: /tmp/codeql-results.sarif
```
migmartri, line 65: have you tried this? 

I am wondering if what the action documents is actually an output of the action, that can be then accessed with something like ` ${{ steps.analyze.outputs.sarif-output}}`

https://github.com/github/codeql-action/blob/4ebadbc7468649cf79b138f45e20d999351f6ed0/analyze/action.yml#L81

migmartri (changes requested):
asking for changes to make sure we double check the output usage.

.github/workflows/codeql.yml
javirln: Added, the only problem is that it does not include the filename. I've added the changes so it uses the default one along with the absolute path take from the action.

javirln:
Results being uploaded:
Screenshot 2024-04-29 at 10 38 20

.github/workflows/codeql.yml
```diff
@@ -45,6 +58,30 @@ jobs:
...
 
       - name: Perform CodeQL Analysis
+        id: codeqlresults
         uses: github/codeql-action/analyze@6a28655e3dcb49cb0840ea372fd6d17733edd8a4 # v2.21.8
         with:
           category: "/language:${{matrix.language}}"
+
+      - name: Add Attestation (Sarif results)
+        run: |
+          chainloop attestation add --name sarif-results --value ${{steps.codeqlresults.outputs.sarif-output}}/go.sarif
```
migmartri, line 68: nice :)

migmartri (approved):

PR #712, opened 2024-04-26
feat(referrer): store backlinks unconditionally

### LINKED ISSUE #703: feat(referrers API): enable backlinks by default
Currently, links between pieces of evidence and attestations are set only if they are part of the attestation subject. This was a decision based on limiting and scoping what elements can be root nodes in the graph. 

This decision has proven to be confusing to users, since in order to create union points, `output: true` was required to be added to the contract. 

I'd suggest that we do not rely on this casuistic and instead enable backlinks from materials to attestations in the referrer API denormalization always.

what do you think?

migmartri:
makes sure that each node in the graph points back to the attestation. Not only the ones that are `subjects` of the attestations, but all materials or subjects.

Closes #703

### DIFF
--- a/app/controlplane/internal/biz/referrer.go
+++ b/app/controlplane/internal/biz/referrer.go
@@ -305,7 +305,10 @@ func extractReferrers(att *dsse.Envelope) ([]*Referrer, error) {
 
 		materialReferrer := referrersMap[materialRef]
 
-		// Add the reference to the attestation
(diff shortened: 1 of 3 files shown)


migmartri:
converting it to draft since I need to fix the integration tests

danlishka (approved):
LGTM

jiparis (approved):
Awesome, thanks!

PR #715, opened 2024-04-26
fix(ci): Use chainloop action on scorecard pipeline

javirln:
Leverages the use of chainloop action rather the manual steps so it complies with ossf/scorecard constraints.


migmartri (commented):
could it be tested on PR?

javirln:
> could it be tested on PR?

We could test that the use of the action is correct but not the production checks. I'll run it.

danlishka (approved):
LGTM

PR #721, opened 2024-04-29
fix(cli): Send pagination message to stderr

### LINKED ISSUE #716: remove pagination options from json output in wf run ls
The json output includes some pagination options that affect its parsing.

Image

We could move this information to the logger instead of stdout so the result could be successfully piped to for example jq

jiparis:
This fix uses the logger to print the pagination message to stderr instead of stdout.

Fixes #716


migmartri (approved):
Thanks!

javirln (approved):
Awesome!

PR #724, opened 2024-04-29
feat(apitoken): Adds workflow run list policy to default apitoken policies

### LINKED ISSUE #717: support listing and describing workflowruns by default in API tokens
API tokens have a default set of permissions that do not include listing nor describing workflow runs https://github.com/chainloop-dev/chainloop/blob/daaa968255412d41b012cdfc9b5a086beaa702bc/app/controlplane/internal/biz/apitoken.go#L74

Listing runs could be useful o detect if a dependency has finished. This task is about adding that the workflowRunList policy to the token by default 

https://github.com/chainloop-dev/chainloop/blob/daaa968255412d41b012cdfc9b5a086beaa702bc/app/controlplane/internal/authz/authz.go#L102

Note: alternatively, we could add support to query attestations by name in the discovery endpoint.

javirln:
This PR adds to the list of available policies of an apitoken the listing of `WorkflowRuns`.

Tests are covering the case where the apitoken should have all default tokens when created.

Closes #717

### DIFF
--- a/app/controlplane/internal/biz/apitoken.go
+++ b/app/controlplane/internal/biz/apitoken.go
@@ -73,7 +73,7 @@ func NewAPITokenUseCase(apiTokenRepo APITokenRepo, conf *conf.Auth, authzE *auth
 		enforcer:     authzE,
 		DefaultAuthzPolicies: []*authz.Policy{
 			// Add permissions to workflow run
-			authz.PolicyWorkflowRunList, authz.PolicyWorkflowRunRead,
+			authz.PolicyWorkflowRunList, authz.PolicyWorkflowRunmimissRead,
 			// Add permissions to workflow contract management
 			authz.PolicyWorkflowContractList, authz.PolicyWorkflowContractRead, authz.PolicyWorkflowContractUpdate,
 			// to download artifacts and list referrers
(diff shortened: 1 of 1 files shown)


app/controlplane/internal/biz/apitoken.go
```diff
@@ -72,6 +72,8 @@ func NewAPITokenUseCase(apiTokenRepo APITokenRepo, conf *conf.Auth, authzE *auth
 		logger:       servicelogger.ScopedHelper(logger, "biz/APITokenUseCase"),
 		enforcer:     authzE,
 		DefaultAuthzPolicies: []*authz.Policy{
+			// Add permissions to workflow run
+			authz.PolicyWorkflowRunList,
```
migmartri, line 76: we are going to need describe too. The one you provided only allow to list runs not describe them.

migmartri (changes requested):

app/controlplane/internal/biz/apitoken.go
javirln: That was going to be my follow up question. Adding it.

app/controlplane/internal/biz/apitoken.go
```diff
@@ -72,6 +72,8 @@ func NewAPITokenUseCase(apiTokenRepo APITokenRepo, conf *conf.Auth, authzE *auth
 		logger:       servicelogger.ScopedHelper(logger, "biz/APITokenUseCase"),
 		enforcer:     authzE,
 		DefaultAuthzPolicies: []*authz.Policy{
+			// Add permissions to workflow run
+			authz.PolicyWorkflowRunList, authz.PolicyWorkflowRunmimissRead,
```
migmartri, line 76: `authz.PolicyWorkflowRunmimissRead,` ?

migmartri (approved):
LGTM! :)

app/controlplane/internal/biz/apitoken.go
javirln: Yes, IDE going Rambo mode.

PR #726, opened 2024-04-29
feat(cli): Allow to pass --output flag to artifact download

### LINKED ISSUE #492: Allow downloading files with specific name
Currently, `chainloop artifact download` only supports providing a path, and the filename is automatically set.

This issue is about adding an option to select a file output. Something like `--file [filePath]`.

javirln:
This PR allows the user to pass a `--output` to `artifact download` command to specify the output.

This work in different modes:
- Passing an actual filename to to the flag to override the custom name:
```bash
$ chainloop artifact download --digest sha256:a614bb8b24252994a821476bdac82bba822461468b092c196e68a3db297bc6bd --path /tmp/artifacts --output myfile.txt

INF API token provided to the command line
INF downloading file name=myfile.txt to=/tmp/artifacts/myfile.txt
99.96% [#################.] [2.80KB in 224.154ms; 12.49KB/s] ...
 ... done! [2.80KB in 251ms; 11.13KB/s]

$ ls -l /tmp/artifacts
total 8
-rw-r--r--  1 javirln  wheel  2800 Apr 29 17:30 myfile.txt
```

- Sending the output to a pipe:
```bash
$ chainloop artifact download --digest sha256:a614bb8b24252994a821476bdac82bba822461468b092c196e68a3db297bc6bd --output - | jq '.'
INF API token provided to the command line
99.96% [#################.] [2.80KB in 208.439ms; 13.43KB/s] ...
 ... done! [2.80KB in 251ms; 11.16KB/s]
{
  "payload": "eyJfdHlwZ....
}
```

- Using the default filename:
```bash
$ chainloop artifact download --digest sha256:a614bb8b24252994a821476bdac82bba822461468b092c196e68a3db297bc6bd
INF API token provided to the command line
INF downloading file name= to=/tmp/artifacts/attestation-2ff0f163-3b9f-48d9-91d3-7b766a458fc1.json
99.96% [#################.] [2.80KB in 37.387ms; 74.86KB/s] ...
 ... done! [2.80KB in 251ms; 11.14KB/s]
```
Closes #492

### DIFF
--- a/app/cli/internal/action/artifact_download.go
+++ b/app/cli/internal/action/artifact_download.go
@@ -61,6 +61,10 @@ func (a *ArtifactDownload) Run(downloadPath, outputFile, digest string) error {
 		return fmt.Errorf("artifact with digest %s not found", h)
 	}
 
+	if downloadPath != "" && outputFile != "" {
+		return errors.New("both downloadPath and outputFile cannot be set at the same time")
+	}
+
 	if downloadPath == "" {
 		downloadPath, err = os.Getwd()
 		if err != nil {
--- a/app/cli/internal/action/artifact_download.go
+++ b/app/cli/internal/action/artifact_download.go
@@ -35,18 +35,20 @@ import (
 type ArtifactDownload struct {
 	*ActionsOpts
 	artifactsCASConn *grpc.ClientConn
+	stdout           io.Writer
 }
 
 type ArtifactDownloadOpts struct {
 	*ActionsOpts
 	ArtifactsCASConn *grpc.ClientConn
+	Stdout           io.Writer
 }
 
 func NewArtifactDownload(opts *ArtifactDownloadOpts) *ArtifactDownload {
-	return &ArtifactDownload{opts.ActionsOpts, opts.ArtifactsCASConn}
+	return &ArtifactDownload{opts.ActionsOpts, opts.ArtifactsCASConn, opts.Stdout}
 }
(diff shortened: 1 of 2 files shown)


app/cli/internal/action/artifact_download.go
```diff
@@ -60,26 +62,36 @@ func (a *ArtifactDownload) Run(downloadPath, digest string) error {
...
 			return fmt.Errorf("getting current dir: %w", err)
 		}
 	}
 
-	downloadPath = path.Join(downloadPath, info.Filename)
-	f, err := os.Create(downloadPath)
-	if err != nil {
-		return fmt.Errorf("creating destination file: %w", err)
+	// Determine output destination
+	outputPath := path.Join(downloadPath, info.Filename)
```
migmartri, line 72: I'd make the use of `downloadPath` incompatible with providing the file. To me, the output is not the name of the file but the full path. 

So basically, if you provide a file, that's what we use, otherwise the system will work as before.

What do you think?

migmartri (changes requested):

app/cli/internal/action/artifact_download.go
javirln: I had mixed feelings when implementing this and felt the same way as you. Going to do the changes, I believe it's more natural that way!

migmartri (approved):
Thanks!

PR #727, opened 2024-04-29
feat(backend): ATTESTATION material type

### LINKED ISSUE #719: add ATTESTATION kind as material type
In some cases, we might want to add additional attestations as pieces of evidence. This could help with storing externally generated attestations, or connecting different workflows in your organization.

To do so, we could add an ATTESTATION material type that will 

- get uploaded to the CAS
- we'll validate that's a valid DSSE envelope.
- we need to make sure that if we take an attestation from chainloop CLI and add it here as a material, it's still the same digest.
- NOTE: before ingesting this attestation type in the index, we might want to check that it is an existing attestation in your org. This is to make sure that nodes in the graph still represent attestations that have been generated with a run. Just some food for thought, I am not sure if we need this kind of optimization at this point, let's discuss it.

jiparis:
This PR adds a new ATTESTATION material type to allow cross-linking different workflow runs using existing attestations.

This is what the logic do:
* check that the attestation is valid (valid DSSE envelope, valid in-toto attestation and chainloop predicate).
* re-marshall the envelope to mitigate possible digest unmatches with existing previously uploaded attestations
* Some refactor to reuse common code in CLI and backend.

References #719


jiparis:
Marking as draft, as I still want to add some tests for the digests.

app/controlplane/internal/biz/attestation.go
```diff
@@ -47,7 +47,7 @@ func NewAttestationUseCase(client CASClient, logger log.Logger) *AttestationUseC
 
 func (uc *AttestationUseCase) UploadToCAS(ctx context.Context, envelope *dsse.Envelope, backend *CASBackend, workflowRunID string) (*cr_v1.Hash, error) {
 	filename := fmt.Sprintf("attestation-%s.json", workflowRunID)
-	jsonContent, h, err := jsonEnvelopeWithDigest(envelope)
+	jsonContent, h, err := materials.JsonEnvelopeWithDigest(envelope)
```
migmartri, line 50: what about adding this helper in a new file under `internal/attestation/attestation.go`? It seems more related to `attestations` in the general sense than to materials. What do you think?

internal/attestation/crafter/materials/attestation.go
```diff
@@ -0,0 +1,69 @@
...
+	data, err := os.ReadFile(artifactPath)
+	if err != nil {
+		return nil, fmt.Errorf("artifact file cannot be read: %w", err)
+	}
+	var dsseEnvelope dsse.Envelope
+	if err := json.Unmarshal(data, &dsseEnvelope); err != nil {
+		return nil, fmt.Errorf("artifact is not a valid DSEE Envelope: %w", err)
+	}
+
+	_, err = chainloop.ExtractPredicate(&dsseEnvelope)
```
migmartri, line 61: It would be great if we could error here with a more descriptive message. Something like "The provided file does not seem to be chainloop-generated attestation"

internal/attestation/crafter/materials/attestation.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("extracting predicate from envelope: %w", err)
+	}
+
+	jsonContent, h, err := JsonEnvelopeWithDigest(&dsseEnvelope)
```
migmartri, line 66: woudl you mind adding a comment explaining why we need this. To know that this is to preserve the digest

internal/attestation/crafter/materials/materials.go
```diff
@@ -43,9 +44,52 @@ type crafterCommon struct {
 	input  *schemaapi.CraftingSchema_Material
 }
 
-// uploadAndCraft uploads the artifact to CAS and crafts the material
+func uploadAndCraftFromBytes(ctx context.Context, input *schemaapi.CraftingSchema_Material, backend *casclient.CASBackend, filename string, artifactData []byte, h cr_v1.Hash, l *zerolog.Logger) (*api.Attestation_Material, error) {
```
migmartri, line 47: I think there too much duplication here.

What about creating function called `uploadAndCraft` that supports a `io.Reader`, then the from bytes will pass a buffer and the fromFile will pass an opened file?

migmartri (changes requested):

migmartri (commented):
looking good! 

Just added some comments so far with some of my concerns, will do a more complete review soon.


Thank you!!

internal/attestation/crafter/materials/materials.go
jiparis: Much better! that was one of the reasons I marked this PR as draft. I want to keep refactoring.

internal/attestation/crafter/materials/attestation.go
jiparis: done!

javirln (approved):

internal/attestation/crafter/materials/csafvex_test.go
```diff
@@ -102,7 +104,8 @@ func TestCSAFVEXCraft(t *testing.T) {
 			// Mock uploader
 			uploader := mUploader.NewUploader(t)
 			if tc.wantErr == "" {
-				uploader.On("UploadFile", context.TODO(), tc.filePath).
+				uploader.On("Upload", context.TODO(), mock.Anything,
+					"sha256:c27087147fa040909e0ef1b522386608af545b0a163c30c9f11c3d753676fa44", filepath.Base(tc.filePath)).
```
migmartri, line 108: why of this specific digest? which one does it correspond?

internal/attestation/crafter/materials/materials.go
```diff
@@ -43,43 +44,56 @@ type crafterCommon struct {
...
 		},
 	}
 
-	// 2 - Upload the file to CAS
+	// Upload the file to CAS
 	if backend.Uploader != nil {
 		l.Debug().Str("backend", backend.Name).Msg("uploading")
 
-		_, err = backend.Uploader.UploadFile(ctx, artifactPath)
+		_, err := backend.Uploader.Upload(ctx, buf, hash, filename)
```
migmartri, line 96: I am not 100% about these changes. Especially the fact that the content and the ahs are now separate. I am wondering if an alternative less optimal, but that will require less changes would have been to store the content in a temporary file. Not sure, I'd love to get your take

migmartri: Taking into account that the grunt of these changes are because of we want to provide some modified versions of a provided file, an alternative, would have been to during the attestation material craft, take the resulted marshalled data and store it in a temporary file and use that one instead.

Then, the rest of the code wouldn't have changed.

internal/attestation/crafter/materials/csafvex_test.go
migmartri: update: I understand now, this is because now we are using `upload` not `upload file` 

For this, since you do not care about the case when the upload fails, you can just add mock.anything in the digest arg.

migmartri (approved):
I have mixed feelings about changing the `upload` calls but there is nothing wrong with it.

I left some comments to get your take, if you want, feel free to merge. Thanks!

migmartri:
btw, have you tested this and see what's the result in the index?

For future PRs, it would be great if you could add some snippets, demos of the functionality when possible.

Thanks!

internal/attestation/crafter/materials/materials.go
jiparis: Yes, cleaner for sure. Let me refactor a little bit the change.

migmartri:
I am having some issues while running it locally

```
cat /tmp/att.json                                                        
{
   "payload": "eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjEiLCJzdWJqZWN0IjpbeyJuYW1lIjoiY2hhaW5sb29wLndvcmtmbG93LnRlc3QyIiwiZGlnZXN0Ijp7InNoYTI1NiI6IjE1MTAwMzA1OWJmOTRiMTViZTMzNTBiZTc5MGM5MGEyYzE5MzBjOWU0Y2ZkYzQwOTEyYmMxY2UzM2Y3YzFmZGQifX0seyJuYW1lIjoiZ2l0LmhlYWQiLCJkaWdlc3QiOnsic2hhMSI6IjM1MTBkZDE1YzJiYTRmMWFkMDk0YjVlNTI3ZjNjYTAxNjVhNjk2MDAifSwiYW5ub3RhdGlvbnMiOnsiYXV0aG9yLmVtYWlsIjoibWlndWVsQGNoYWlubG9vcC5kZXYiLCJhdXRob3IubmFtZSI6Ik1pZ3VlbCBNYXJ0aW5leiBUcml2aW5vIiwiZGF0ZSI6IjIwMjQtMDQtMzBUMDY6Mjg6NDZaIiwibWVzc2FnZSI6ImZlYXQoY2xpKTogYWRkIGpzb24gb3V0cHV0IHRvIGF0dGVzdGF0aW9uIHB1c2hcblxuU2lnbmVkLW9mZi1ieTogTWlndWVsIE1hcnRpbmV6IFRyaXZpbm8gPG1pZ3VlbEBjaGFpbmxvb3AuZGV2PlxuIiwicmVtb3RlcyI6W3sibmFtZSI6Im9yaWdpbiIsInVybCI6ImdpdEBnaXRodWIuY29tOm1pZ21hcnRyaS9jaGFpbmxvb3AuZ2l0In0seyJuYW1lIjoidXBzdHJlYW0iLCJ1cmwiOiJnaXRAZ2l0aHViLmNvbTpjaGFpbmxvb3AtZGV2L2NoYWlubG9vcC5naXQifSx7Im5hbWUiOiJ0ZXN0LXRva2VuIiwidXJsIjoiaHR0cHM6Ly9naXRsYWItY2ktdG9rZW46Z2xjYnQtNjVfNVg2dUR6SlZSeDlyU3pnZFdES1pAZ2l0aHViLmNvbS9jaGFpbmxvb3AtZGV2L2NoYWlubG9vcC5naXQifV19fV0sInByZWRpY2F0ZVR5cGUiOiJjaGFpbmxvb3AuZGV2L2F0dGVzdGF0aW9uL3YwLjIiLCJwcmVkaWNhdGUiOnsiYnVpbGRUeXBlIjoiY2hhaW5sb29wLmRldi93b3JrZmxvd3J1bi92MC4xIiwiYnVpbGRlciI6eyJpZCI6ImNoYWlubG9vcC5kZXYvY2xpL2RldkBzaGEyNTY6MmE0ZGE4MGU5Y2IzZDJkZTU5YmVjZTAwNmEwOWZlOGNlN2RiZmJmYWQ3MDQ2MTcyMDIzYzY5N2VjMTc4MDcyMCJ9LCJtZXRhZGF0YSI6eyJmaW5pc2hlZEF0IjoiMjAyNC0wNC0zMFQwNjozMjoxOC41NjEwOTA2MzhaIiwiaW5pdGlhbGl6ZWRBdCI6IjIwMjQtMDQtMzBUMDY6MzI6MTEuNjkyOTUyOTk0WiIsIm5hbWUiOiJ0ZXN0MiIsIm9yZ2FuaXphdGlvbiI6ImZvbyIsInByb2plY3QiOiJiYXIiLCJ0ZWFtIjoiIiwid29ya2Zsb3dJRCI6IjY1MmRkMmExLTE0NjItNGFhMC05OTVmLTFlMjg2NWFkMjY3NiIsIndvcmtmbG93UnVuSUQiOiJiYzE2NzhlYy1iMGIzLTQxYWYtYmVhZC1lNzViYzBjNTRiMmMifSwicnVubmVyVHlwZSI6IlJVTk5FUl9UWVBFX1VOU1BFQ0lGSUVEIn19",
   "signatures": [
      {
         "keyid": "",
         "sig": "MEUCIDGBc8J4tpGaSyTcdHecnZOa725Tja4tozwtXNr6jSz7AiEA8xOts32aCmDs3TaUTyV8Tiv461gblt+ysfCT2OH9DFU="
      }
   ]
}

```

migmartri (changes requested):
See my latest comment.

migmartri (approved):

jiparis:
Thanks, @migmartri. Fixed!

PR #729, opened 2024-04-29
fix(cli): sanitize remote url

migmartri:
In some cases, like the case of Gitlab, the CI will inject a remote with some temporary credentials to pull from the private repositories. This meant that the attestations had those temporary credentials in it as part of the `git.head` object in the attestation.

This patch updates the attestation cli to make sure the URI is sanitized before being ingested.


javirln (approved):
LGTM!

PR #731, opened 2024-04-30
feat(ci): Declarative Chainloop contracts

javirln:
This PR creates a GitHub workflow that runs in in two scenarios:
- Every time there is a push in `main` branch
- Every day at midnight

Monitors changes in the `.github/workflows/contracts` and applies a `chainloop wf contract update ...` if a contract has been updated.


javirln:
Let's decide if we want this to run on pull request and pushes on `main` to remove associated code.

.github/workflows/sync_contracts.yml
```diff
@@ -0,0 +1,46 @@
...
+      CHAINLOOP_API_TOKEN: ${{ secrets.CHAINLOOP_ROBOT_ACCOUNT_SYNC_CONTRACTS }}
+    steps:
+      - name: Install Chainloop
+        run: |
+          curl -sfL https://raw.githubusercontent.com/chainloop-dev/chainloop/01ad13af08950b7bfbc83569bea207aeb4e1a285/docs/static/install.sh | bash -s -- --version v${{ env.CHAINLOOP_VERSION }}
+
+      - uses: actions/checkout@v4.1.4
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
+      - name: Get changed files
```
migmartri, line 30: I do not think this is what we want. We want to go through the whole directory and apply update always. note that revisions will not bump if the content hasn't changed

javirln: Regardless of whether the have changed or not?

migmartri (approved):

PR #732, opened 2024-04-30
feat(controlplane): verify that dependent attestations exist

### LINKED ISSUE #719: add ATTESTATION kind as material type

migmartri:
Follow up from attestation kind 

closes #719

### DIFF
--- a/app/controlplane/internal/biz/workflowrun.go
+++ b/app/controlplane/internal/biz/workflowrun.go
@@ -24,8 +24,10 @@ import (
 
 	"github.com/chainloop-dev/chainloop/app/controlplane/internal/pagination"
 	"github.com/chainloop-dev/chainloop/internal/attestation"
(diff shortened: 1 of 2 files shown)


jiparis (approved):
Looks great to me.

Issue #735, opened 2024-04-30
validate that the dependent attestation exist before we store the referrer


migmartri:
refs #719

PR #745, opened 2024-05-03
chore: validate dependent attestation in referrer endpoint

migmartri:
This patch updates the ingestion process for the referrer graph to make sure that attached attestations node exists in advance.

Closes https://github.com/chainloop-dev/chainloop/issues/735


jiparis (approved):
Thanks! looks good to me.

Issue #635, opened 2024-03-26
Add support for CSAF_* material types


danlishka:
We [currently support](https://docs.chainloop.dev/reference/operator/contract#material-schema) CSAF_VEX. This task aims to add support for [other profiles](https://docs.oasis-open.org/csaf/csaf/v2.0/cs03/csaf-v2.0-cs03.html#4-profiles).

- [Examples](https://github.com/oasis-tcs/csaf/tree/master/csaf_2.0/examples/csaf)
- [FAQ](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/guidance/faq.md)
- [GitHub Repository](https://github.com/oasis-tcs/csaf/tree/master)

migmartri:
I think that the @bitnami team did some work with supporting CSAF in go. 

Do you know if this is true and if their work is available? cc/ @juan131

juan131:
Yes @migmartri 

We did some contributions to https://github.com/csaf-poc/csaf_distribution
We also did some contributions to Trivy & Grype, see:

- https://github.com/aquasecurity/trivy/pull/5535
- https://github.com/aquasecurity/trivy/pull/5923
- https://github.com/anchore/grype/pull/1826

PR #748, opened 2024-05-06
refactor(materials): Add custom types for different CSAF profiles

javirln:
This PR introduces a refactor that extracts the different CSAF profiles with _partial_ validations from the specification found here: https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#4-profiles

The implementation does not attempt to add a validation library but rather some basic validations to the common CSAF document.

Refers #635

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -77,7 +77,7 @@ require (
 	github.com/invopop/jsonschema v0.7.0
 	github.com/jackc/pgx/v5 v5.5.4
 	github.com/muesli/reflow v0.3.0
-	github.com/openvex/go-vex v0.2.5
+	github.com/openvex/go-vex v0.2.6-0.20240506020847-46399d425c68
 	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
 	google.golang.org/genproto/googleapis/api v0.0.0-20240401170217-c3f982113cda
 	google.golang.org/genproto/googleapis/bytestream v0.0.0-20240318140521-94a12d6c2237
@@ -135,7 +135,7 @@ require (
 	github.com/mitchellh/go-testing-interface v1.14.1 // indirect
 	github.com/nozzle/throttler v0.0.0-20180817012639-2ea982251481 // indirect
 	github.com/oklog/run v1.1.0 // indirect
-	github.com/package-url/packageurl-go v0.1.1 // indirect
+	github.com/package-url/packageurl-go v0.1.2 // indirect
 	github.com/pjbgf/sha1cd v0.3.0 // indirect
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
 	github.com/pkg/xattr v0.4.9 // indirect
--- a/internal/attestation/crafter/materials/csaf.go
+++ b/internal/attestation/crafter/materials/csaf.go
@@ -1,0 +1,152 @@
+//
+// Copyright 2024 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package materials
+
+import (
+	"fmt"
+
+	"github.com/openvex/go-vex/pkg/csaf"
+)
+
+// CSAFValidator is an interface for validating CSAF documents
+type CSAFValidator interface {
+	ValidateDocument() error
+}
+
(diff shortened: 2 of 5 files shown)


javirln:
Another alternative solution to this is to rely solely on the implementation of github.com/openvex/go-vex/pkg/csaf for the custom types and validations of the new materials.

javirln:
It's not that we don't have the latest version that we do but rather that the latest release on their end with the needed changed, failed to be released.

javirln:
Updated the library to a specific commit and used library types + custom ones.

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,152 @@
...
+type Tracking struct {
+	csaf.Tracking
+	RevisionHistory []RevisionHistoryItem `json:"revision_history"`
+	Status          string                `json:"status"`
+	Version         string                `json:"version"`
+}
+
+type CSAFDocument struct {
+	csaf.CSAF
+	Document struct {
```
jiparis, line 46: Why redefining this struct? and the fields below? aren't they already in the embedded CSAF struct with the correct fields and annotations?

internal/attestation/crafter/materials/csafvex.go
```diff
@@ -44,13 +45,22 @@ func NewCSAFVEXCrafter(materialSchema *schemaapi.CraftingSchema_Material, backen
...
 func (i *CSAFVEXCrafter) Craft(ctx context.Context, filepath string) (*api.Attestation_Material, error) {
 	i.logger.Debug().Str("path", filepath).Msg("decoding CSAF VEX file")
-	doc, err := csaf.Open(filepath)
-	// parse doesn't fail if the provided file is a valid JSON, but not a valid CSAF VEX file
-	if err != nil || doc.Document.Title == "" {
-		if err != nil {
-			i.logger.Debug().Err(err).Msg("error decoding file")
-		}
 
+	fh, err := os.Open(filepath)
```
jiparis, line 49: I'm probably lacking some context about why why `csaf.Open` is not a good option anymore. If I understand correctly, the purpose is to differentiate the different CSAF types, that's why here you are decoding to a `Vex` instead of the generic `CSAF` type defined in the library.
I'm not sure if this is counterproductive, since here we are overriding or duplicating some logic already present in the library.

internal/attestation/crafter/materials/csaf.go
javirln: Yes, the default csaf.CSAF document does not have some fields that as per the specification must exist. If you see the following line, we are doing some kind of inheritance + override of fields so we can add the missing ones.

internal/attestation/crafter/materials/csafvex.go
javirln: The issue is that `csaf.Open` returns a `csaf.CSAF` document but since we are overriding the struct with our custom type, it's not valid anymore. In any case csaf.Open only decodes the file into a csaf.CSAF document and does not perform any validation.

But, the main reason is that we working now with our own types that inherit from the library.

javirln: This is open to debate in case we want a minimum of validation since right now VEX files only checks the `Title` is set.

migmartri (approved):
I think it makes sense to me. There are a couple of things that would improve this patch.

1 - add some more commentary on the `csaf` file indicating that this is not a material type but a auxiliary library. In fact, I think it would be ok if you put it inside its own package.
2 - It would be great to have some tests with test-data documents that run those validations.

go.mod
```diff
@@ -77,7 +77,7 @@ require (
 	github.com/invopop/jsonschema v0.7.0
 	github.com/jackc/pgx/v5 v5.5.4
 	github.com/muesli/reflow v0.3.0
-	github.com/openvex/go-vex v0.2.5
+	github.com/openvex/go-vex v0.2.6-0.20240506020847-46399d425c68
```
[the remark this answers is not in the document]
javirln: I updated the library manually. The reason behind it is that in recent commits there are a bunch of types that are implemented which are useful since we don't have to implement them ourselves.

AFAIK, no cross-reference more than the base CSAF document which them inherit the rest of the profiles from the specification.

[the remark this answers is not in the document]
migmartri: but openvex is not CSAF vex is it? Or what am I missing?

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,152 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package materials
+
+import (
+	"fmt"
+
+	"github.com/openvex/go-vex/pkg/csaf"
```
migmartri, line 21: OpenVEX and CSAF Vex are different specifications, aren't they?

go.mod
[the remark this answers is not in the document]
migmartri: Yeah, I see, I replied below, this openvex library seems to have some types for csaf.

jiparis (approved):
Looks good. thanks!

javirln:
Closing the issue as we will implement its support based on a json schema. No need for a refactor.

PR #749, opened 2024-05-06
feat(materials): Add support for CSAF 2.0 and 2.1 schemas and remaining CSAF_* materials

javirln:
This pull request builds upon the changes introduced in [PR #751](https://github.com/chainloop-dev/chainloop/pull/751) to implement CSAF version 2.0 and 2.1 support within Chainloop. This update incorporates three new CSAF Chainloop materials, expanding upon the existing CSAF [VEX profile](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#45-profile-5-vex):

- [Security Advisory](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#44-profile-4-security-advisory)
- [Informational Advisory](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#43-profile-3-informational-advisory)
- [Security Incident Response](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#42-profile-2-security-incident-response)

This implementation adheres to the CSAF 2.0 and 2.1 JSON schema verification standards, accessible [here](https://github.com/oasis-tcs/csaf/tree/master/csaf_2.0). Furthermore, it refactors the `CSAFVEXCrafter` to a more generalized `CSAFCrafter`, aligning with the JSON specification 2.0 and 2.1.

With this update, Chainloop contracts can now process the newly introduced materials, ensuring validation against the schema. Notably, there is no requirement to specify the `csaf_version` within the file, as Chainloop dynamically determines the appropriate schema for validation.

```yaml
schemaVersion: v1
runner:
  type: GITHUB_ACTION
materials:
  - type: CSAF_INFORMATIONAL_ADVISORY
    name: informational-advisory
  - type: CSAF_SECURITY_ADVISORY
    name: security-advisory
  - type: CSAF_SECURITY_INCIDENT_RESPONSE
    name: security-incident-response
```

References: #635

This update enhances the versatility and compatibility of Chainloop with the latest CSAF standards.


internal/attestation/crafter/materials/csaf_security_indicent_response.go
```diff
@@ -0,0 +1,68 @@
...
+	return &CSAFSecurityIncidentResponseCrafter{
+		backend:       backend,
+		crafterCommon: &crafterCommon{logger: l, input: materialSchema},
+	}, nil
+}
+
+func (i *CSAFSecurityIncidentResponseCrafter) Craft(ctx context.Context, filepath string) (*api.Attestation_Material, error) {
+	i.logger.Debug().Str("path", filepath).Msg("decoding CSAF Security Incident Response file")
+
+	fh, err := os.Open(filepath)
```
migmartri, line 49: There seems to be quite a bit of common code here. The one that opens a file from a path and marshals it into a document. 

Should we move it to the shared library `vex`?

internal/schemavalidators/schemavalidators.go
```diff
@@ -93,3 +126,39 @@ func ValidateCycloneDX(data interface{}, version CycloneDXVersion) error {
...
+		}
+
+		errs = multierror.Append(errs, err)
+	} else {
+		return nil
+	}
+
+	err = compiledCSAFSchemas[CSAFVersion2_0].Validate(data)
+	if err != nil {
+		if err := errorIsJSONFormat(err); err != nil {
```
jiparis, line 147: this might not be needed, as it's already checked above.

jiparis (approved):
Looks great, just a couple of minor comments.

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,73 @@
...
+	i.logger.Debug().Str("path", filepath).Msg("decoding CSAF file")
+	f, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("can't open the file: %w - %w", err, ErrInvalidMaterialType)
+	}
+
+	var v interface{}
+	if err := json.Unmarshal(f, &v); err != nil {
+		i.logger.Debug().Err(err).Msg("error decoding file")
+		return nil, fmt.Errorf("invalid casf file: %w", ErrInvalidMaterialType)
```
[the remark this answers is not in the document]
javirln: Yes, my bad, fixing it!

internal/schemavalidators/schemavalidators.go
javirln: The point is that if we don't add it with the multi error, the same error regarding the format will appear twice on this way, if that's the problem, it will just exist on the first try.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -91,6 +91,9 @@ message CraftingSchema {
 
       // Chainloop attestation coming from a different workflow.
       ATTESTATION = 12;
+      CSAF_INFORMATIONAL_ADVISORY = 13;
```
migmartri, line 94: Great! nitpick, Just for visiblity, could you bring closer `CSAF_VEX = 8;` so they are all grouped :)

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/chainloop-dev/chainloop/internal/schemavalidators"
+	"github.com/rs/zerolog"
+)
+
+type CSAFCrafter struct {
+	backend *casclient.CASBackend
+	*crafterCommon
+}
+
+func NewCSAFCrafter(materialSchema *schemaapi.CraftingSchema_Material, backend *casclient.CASBackend, l *zerolog.Logger) (*CSAFCrafter, error) {
```
migmartri, line 36: Is there any way we could differentiate these material types?

What I mean is that one of the purposes of a crafter is to make sure we offer guardrails to developers so they provide exactly what they were asked for.

So in this case, if SecOps asks for a `Security incident response` I should not accept a `vex` file for example.

Is there any way we can differentiate those profiles? I could think of 1) doing the generic validation and 2) check if they are the right kind of profile. Is what possible?

migmartri (commented):
Awesome! 

I'd like to see if we can differentiate the different profiles. We could reuse a single crafter but we might need to do additional checks.

migmartri:
On another note, let's not forget to document these new material types in the readme and the docs before release. Probably better in another patch

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
javirln: Sure thing! :)

internal/attestation/crafter/materials/csaf.go
javirln: I completely agree I believe it's a bit counterproductive right now. We could do several things:
- Generic validation: Current approach, we just get the material and if non of the desired ones, raise a generic error.
- Individual crafter specification that just call this one, each of them with the individual material check.
- Fail on the material check, try to infer the profile from the document. This is a best effort since the `document.category` is not a required field but rather a suggestion to be set upon documentation, example:
> The value of /document/category SHALL be csaf_security_incident_response.
> The value of /document/category SHALL be csaf_informational_advisory.
> The value of /document/category SHALL be csaf_security_advisory.
> The value of /document/category SHALL be csaf_vex.

javirln: I've refactored it so each time uses a different `constructor` that validates the proper material type and fails with sense.

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,105 @@
...
+func NewCSAFInformationalAdvisoryCrafter(materialSchema *schemaapi.CraftingSchema_Material, backend *casclient.CASBackend, l *zerolog.Logger) (*CSAFCrafter, error) {
+	if materialSchema.Type != schemaapi.CraftingSchema_Material_CSAF_INFORMATIONAL_ADVISORY {
+		return nil, fmt.Errorf("material type is not CSAF_INFORMATIONAL_ADVISORY format")
+	}
+
+	return baseCSAFCrafter(materialSchema, backend, l)
+}
+
+// NewCSAFSecurityIncidentResponseCrafter creates a new CSAF Security Incident Response crafter
+func NewCSAFSecurityIncidentResponseCrafter(materialSchema *schemaapi.CraftingSchema_Material, backend *casclient.CASBackend, l *zerolog.Logger) (*CSAFCrafter, error) {
```
migmartri, line 57: mm, I am not sure I understand. Will this fail if I provide a VEX file but I say that's a INCENT_REPORT?

internal/attestation/crafter/materials/testdata/csaf_security_advisory.json
```diff
@@ -0,0 +1,319 @@
+{
+  "document": {
+    "aggregate_severity": {
+      "namespace": "https://access.redhat.com/security/updates/classification/",
+      "text": "Critical"
+    },
+    "category": "csaf_security_advisory",
```
migmartri, line 7: shouldn't you check for the `category`?

javirln: The problem with the category is that it would be just a best effort since it's not a compulsory parameter.

migmartri: but as you said in another comment, should we check that it's either the value we want or empty?

javirln: We could do that but it won't guarantee us that the document is valid and that would need to be done on the `Craft` method rather than on the initializer.

javirln: Oki, changes applied I believe it shall be ok now.

migmartri (approved):
Awesome thanks!

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var v interface{}
+	if err := json.Unmarshal(f, &v); err != nil {
+		i.logger.Debug().Err(err).Msg("error decoding file")
+		return nil, fmt.Errorf("invalid CSAF file: %w", ErrInvalidMaterialType)
+	}
+
+	// Validate the CSAF file against the specified category.
+	document, docExists := v.(map[string]interface{})["document"]
```
migmartri, line 105: it would have been more clear if we use a custom struct with only the document and category objects.

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,125 @@
...
+	if docExists {
```
migmartri, line 106: if you use a custom struct I do not think you would have needed to do all these if checks and castings.

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,125 @@
...
+		documentMap, docMapOk := document.(map[string]interface{})
+		category, categoryExists := documentMap["category"].(string)
+
+		if docMapOk && categoryExists && category != "" && category != i.category {
```
migmartri, line 110: I'd have done this validation after you do the base validation, that way you can guarantee that it's a base doc, and then use a custom struct to do a custom validation.

Does it make sense?

migmartri (approved):
Left some comments for a future refactoring. For now I'd merge it.

Thanks!

internal/attestation/crafter/materials/csaf.go
javirln: hmm but the custom struct won't have the remaining fields to be passed down to the validator no? Maybe I'm missing something :/

internal/attestation/crafter/materials/csaf.go
```diff
@@ -0,0 +1,125 @@
+//
+// Copyright 2023 The Chainloop Authors.
```
migmartri, line 2: 2024 :)

internal/attestation/crafter/materials/csaf.go
migmartri: it will not, that's why I mentioned doing this validation after the other one. You might require another marshall process though.

Anyways, not a big deal, feel free to merge it :)

Issue #752, opened 2024-05-07
Allow API tokens to create robot accounts


jiparis:
To ease the automatic onboarding of workflows, organizations might configure an API token globally, and then creating ephemeral robot accounts in each workflow run. This is the closest approach to a refresh-token/api-token schema without needing to setup a full OAuth2 provider. 

The benefit of this approach would be:
* attestations are still only done through robot accounts
* organizations don't need to create a single robot account for each workflow (which could be problematic in orgs with dozens of jobs)

This is a rough summary of the needed changes:
[] add RobotAccount policies to API tokens
[] add expiration time to RAs
[] add a cleaup process for expired RAs (making them ephemeral, or one-time use)

migmartri:
Is there anything left here?

javirln:
> Is there anything left here?

All done here, closing the issue.

migmartri:
There were (are) two things missing
- document the attestaiton process with API tokens in the docs
- use it in one of our workflows

Are these completed?

javirln:
> There were (are) two things missing
> 
> * document the attestaiton process with API tokens in the docs
> * use it in one of our workflows
> 
> Are these completed?

They are in progress.

migmartri:
ok, are you tracking them in another task or smth? Should we reopen this issue? Whatever you prefer :)

PR #754, opened 2024-05-08
feat(policies): Add attestation as new resource type and workflow create permission

javirln:
This PR adds a new resource type `attestation` with three new permissions for: `create, read and list`. Additionally it create a new permissions for `workflow` resource, `create`.

The addition to the default policy list for API Tokens and to the server URL mapping will come in a different PR.

Refs #752


jiparis (approved):
LGTM!

PR #763, opened 2024-05-09
feat(attestations): support API tokens for attestations

javirln:
This PR creates a global middleware that accepts multiple JWT providers. The goal of such middleware is to be able to try to validate an incoming JWT in a shared way so multiple tokens types have the same entry points.

Additionally a new middleware was created to setup a `RobotAccount` from an `API Token`. Some attestation proto request were updated to optionally send the workflow name.

Refers #752


app/controlplane/internal/data/workflow.go
```diff
@@ -218,6 +218,26 @@ func (r *WorkflowRepo) FindByID(ctx context.Context, id uuid.UUID) (*biz.Workflo
...
+func (r *WorkflowRepo) FindByName(ctx context.Context, name string) (*biz.Workflow, error) {
+	workflow, err := r.data.db.Workflow.Query().
+		Where(workflow.DeletedAtIsNil(), workflow.Name(name)).
+		WithContract().WithOrganization().
+		Only(ctx)
+	if err != nil && !ent.IsNotFound(err) {
+		return nil, err
+	} else if workflow == nil {
+		return nil, nil
+	}
```
jiparis, line 230: you can use the other pattern we are already applying in other places, where the error translation is done in the repository layer:
```suggestion
	if err != nil {
		if ent.IsNotFound(err) {
			return nil, biz.NewErrNotFound("workflow")
		}
		return nil, err
	}
```

app/controlplane/internal/server/grpc.go
```diff
@@ -192,16 +194,52 @@ func craftMiddleware(opts *Opts) []middleware.Middleware {
...
-			},
-				jwtMiddleware.WithSigningMethod(robotaccount.SigningMethod),
-				jwtMiddleware.WithClaims(func() jwt.Claims { return &robotaccount.CustomClaims{} }),
+			// 1 - Extract information from the JWT by using the claims
+			usercontext.WithJWTTokenMiddleware(
+				// Robot account provider
+				usercontext.WithTokenProvider(
+					usercontext.WithProviderKey(usercontext.RobotAccountProviderKey),
+					usercontext.WithClaims(func() jwt.Claims { return &robotaccount.CustomClaims{} }),
+					usercontext.WithVerifyAudienceFunc(func(token *jwt.Token) bool {
```
jiparis, line 203: Look into `ClaimsValidator` interface. It's provided by the library to perform this kind of custom validations.

app/controlplane/internal/server/grpc.go
```diff
@@ -192,16 +194,52 @@ func craftMiddleware(opts *Opts) []middleware.Middleware {
...
-			jwtMiddleware.Server(func(_ *jwt.Token) (interface{}, error) {
-				// TODO: add support to multiple signing methods and keys
-				return []byte(opts.AuthConfig.GeneratedJwsHmacSecret), nil
-			},
-				jwtMiddleware.WithSigningMethod(robotaccount.SigningMethod),
-				jwtMiddleware.WithClaims(func() jwt.Claims { return &robotaccount.CustomClaims{} }),
+			// 1 - Extract information from the JWT by using the claims
+			usercontext.WithJWTTokenMiddleware(
+				// Robot account provider
+				usercontext.WithTokenProvider(
```
jiparis, line 200: Probably you can move this (and the one velow) to a function for better clarity.

app/controlplane/internal/usercontext/jwt_middleware.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			return nil, ErrWrongContext
+		}
+	}
+}
+
+// runProviderValidator runs the token parser for the given provider
+func runProviderValidator(provider providerOption, jwtToken string) (*jwt.Token, error) {
+	if provider.keyFunc == nil {
```
migmartri, line 176: this is based on some other code right? I want to make sure we are not reinventing the wheel :)

app/controlplane/internal/server/grpc.go
```diff
@@ -192,16 +194,52 @@ func craftMiddleware(opts *Opts) []middleware.Middleware {
...
+					usercontext.WithProviderKey(usercontext.RobotAccountProviderKey),
```
migmartri, line 201: if providerkey is required we might want to add it as a parameter, not as an option.

app/controlplane/internal/usercontext/jwt_middleware.go
```diff
@@ -0,0 +1,231 @@
...
+					err       error
+				)
+
+				tokenProviderLen := len(o.tokenProviders)
+				for idx, provider := range o.tokenProviders {
+					tokenInfo, err = runProviderValidator(provider, jwtToken)
+
+					// Check if it's the last provider and still failed
+					if err != nil {
+						if idx != tokenProviderLen-1 {
```
migmartri, line 142: I find this comparison confusing. what about `if I < tokenProviderLen`?

app/controlplane/internal/usercontext/jwt_middleware.go
```diff
@@ -0,0 +1,231 @@
...
+							continue
+						}
+
+						return nil, err
```
migmartri, line 146: I tend to wrap errors to add tracing, I'd suggest you do the same `fmt.Errorf("... %w", err)`

app/controlplane/internal/usercontext/jwt_middleware.go
```diff
@@ -0,0 +1,231 @@
...
+	ErrTokenExpired           = errorsAPI.Unauthorized(reason, "JWT token has expired")
+	ErrTokenParseFail         = errorsAPI.Unauthorized(reason, "Fail to parse JWT token ")
+	ErrUnSupportSigningMethod = errorsAPI.Unauthorized(reason, "Wrong signing method")
+	ErrWrongContext           = errorsAPI.Unauthorized(reason, "Wrong context for middleware")
+	ErrNeedTokenProvider      = errorsAPI.Unauthorized(reason, "Token provider is missing")
+)
+
+type JWTAuthContext struct {
+	Claims         jwt.Claims
+	IsAPIToken     bool
```
migmartri, line 55: Why do we need two booleans? What about just the providerKey? which then can be compared to the const you have above?

app/controlplane/internal/usercontext/apitoken_middleware.go
```diff
@@ -101,6 +102,98 @@ func WithCurrentAPITokenAndOrgMiddleware(apiTokenUC *biz.APITokenUseCase, orgUC
...
+	}
+
+	// Note: Expiration time does not need to be checked because that's done at the JWT
+	// verification layer, which happens before this middleware is called
+	if token.RevokedAt != nil {
+		return nil, errors.New("API token revoked")
+	}
+
+	// Find the associated organization
+	org, err := orgUC.FindByID(ctx, token.OrganizationID.String())
```
migmartri, line 167: do you need to load the org? Can't you just use the `token.OrganizationID.String()` note that at this point the token is authenticated already

app/controlplane/internal/usercontext/apitoken_middleware.go
```diff
@@ -101,6 +102,98 @@ func WithCurrentAPITokenAndOrgMiddleware(apiTokenUC *biz.APITokenUseCase, orgUC
...
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving the organization: %w", err)
+	} else if org == nil {
+		return nil, errors.New("organization not found")
+	}
+
+	var workflowName string
```
migmartri, line 174: this might not be worth having here, I am wondering if we should just load the workflow in the service, it might be more transparent and easier to understand.

migmartri (commented):
Hey! 

Good stuff, a couple of comments.

- I think we could put the new middleware in its own package
- I am thinking that it might not be worth using a middleware to find the workflow by name. It might be better if in the case of non-robot-accounts, the returned current-robot-account returns empty workflow. And then, in the service we load it, what do you think?
- It would be great to document the new middleware, to know what's based on some other code and so on.
- I am curious why you think we need to return workflow_name in some API endpoints

Thanks!

app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -74,6 +76,7 @@ message AttestationServiceStoreRequest {
   // encoded DSEE envelope
   bytes attestation = 1 [(buf.validate.field).bytes.min_len = 1];
   string workflow_run_id = 2 [(buf.validate.field).string = {min_len: 1}];
+  string workflow_name = 3;
```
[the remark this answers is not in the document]
javirln: Hmm It's added on the request, if you expand GitHub preview, it's added on the `AttestationServiceStoreRequest`.

app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -91,6 +94,7 @@ message AttestationServiceCancelRequest {
     not_in: [0]
   }];
   string reason = 3;
+  string workflow_name = 4;
```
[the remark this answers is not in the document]
javirln: Same here

app/controlplane/internal/data/workflow.go
javirln: Ok, I'll also revisit where I taken it from because might be wrong as well.

app/controlplane/internal/server/grpc.go
javirln: ClaimsValidator still needs to be casted to the proper struct if it has a custom struct. We are validating after with the interface with `VerifyAudience`.

app/controlplane/internal/usercontext/jwt_middleware.go
javirln: It makes sense, I'll leverage the use of the provider keys and export them to do the validation.

app/controlplane/internal/usercontext/jwt_middleware.go
```diff
@@ -0,0 +1,231 @@
...
+	op := &providerOption{}
+	for _, opt := range opts {
+		opt(op)
+	}
+	return func(o *options) {
+		o.tokenProviders = append(o.tokenProviders, *op)
+	}
+}
+
+func WithJWTTokenMiddleware(opts ...JWTOption) middleware.Middleware {
```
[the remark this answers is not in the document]
javirln: This code is the structure of kratos to create custom middleware, this part specifically was not taken from anywhere. The inner logic of the jwt middleware yes.

app/controlplane/internal/usercontext/jwt_middleware.go
javirln: This is base on the jwtmiddleware.Serve inner logic. A few problems and why we can not reuse their code directly:
- Such middleware whenever it success, sends the request to the following middleware in the stack, making it impossible to inject custom data on the context. This change, returns a token info rather than continuing with the next middleware.
- The jwt parsing happens in a block, not in any extracted function we can include along with its validation.

Reference: https://github.com/go-kratos/kratos/blob/9e66ac2f5bcb9ab18d9b8d378c5b3233c7bb0a73/middleware/auth/jwt/jwt.go#L88

Our approach:
- Reuse parsing and validation of JWT as common as possible.
- Return a token that is valid and ready to be used, rather than continue with the next middleware.
- Use that token and the configured provider to set custom flags and information.

app/controlplane/internal/usercontext/apitoken_middleware.go
javirln: If we move that information to the services, they will be the one in charge of from a workflowName, take the workflowID. This middleware here won't make much sense then because it will be only setting the `OrgID` since the token ID was recently removed.

jiparis: Moving it to the service means that service might receive "incomplete" requests. In the other side, doing database queries from the middlewares doesn't sound the right place. It would be nice if we figure out an intermediate solution, something like a common service initialization phase.

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -0,0 +1,268 @@
...
+	ErrMissingKeyFunc         = errorsAPI.Unauthorized(reason, "keyFunc is missing")
+	ErrTokenInvalid           = errorsAPI.Unauthorized(reason, "Token is invalid")
+	ErrTokenExpired           = errorsAPI.Unauthorized(reason, "JWT token has expired")
+	ErrTokenParseFail         = errorsAPI.Unauthorized(reason, "Fail to parse JWT token ")
+	ErrUnSupportSigningMethod = errorsAPI.Unauthorized(reason, "Wrong signing method")
+	ErrWrongContext           = errorsAPI.Unauthorized(reason, "Wrong context for middleware")
+)
+
+// NewAttestationRobotAccountProvider return the configuration to validate and verify token issued for Robot Accounts
+func NewAttestationRobotAccountProvider(signingSecret string, expectedAudience []string) JWTOption {
```
jiparis: `signingSecret` and `expectedAudience` are implementation details for each provider and maybe shouldn't be exposed as parameters (as there's nothing to parameterize). The caller doesn't know which audience or secret is each provider using. Do you think it makes sense?

Same comment for the other provider.

javirln: The expectedAudience we can indeed move it to the initializer function the signingSecret however no because it's being injected when the service is being run by the AuthConfig.

app/controlplane/internal/usercontext/apitoken_middleware.go
jiparis: If we decided to go through the middleware path, instead of trying to cast the `req interface{}` to each known request type, we could use reflection to check if the field exists (I don't see any other option).
```
message := req.(ProtoMessage)
message.ProtoReflect().Get(fieldDescriptor)
```

app/controlplane/internal/biz/workflow.go
```diff
@@ -211,6 +212,14 @@ func (uc *WorkflowUseCase) FindByID(ctx context.Context, workflowID string) (*Wo
 	return uc.wfRepo.FindByID(ctx, workflowUUID)
 }
 
+func (uc *WorkflowUseCase) FindByName(ctx context.Context, workflowName string) (*Workflow, error) {
+	if workflowName == "" {
+		return nil, NewErrValidationStr("empty workflow name")
+	}
+
+	return uc.wfRepo.FindByName(ctx, workflowName)
```
migmartri, line 220: We need to make sure the workflow we find is in our organization. Note how we have `FindByIDInOrg`

app/controlplane/internal/usercontext/apitoken_middleware.go
migmartri: TLDR

Personally, but take it with a grain of salt since I still need to review this patch thoroughly, I'd rather go to the approach where the services figure out how to get the workflow. And the middleware just a) `validate the API token JWT`, b) `checks that the token is not revoked` and c) `stores the org id in the robot-account (soon to be renamed) object` 

TLDR/

I see, to me, a rule of thumb is designing the solution from the POV of the API. I'm not too fond of the current approach (requiring name so the middleware can complete the stuff) because it forces us to make suboptimal API choices https://github.com/chainloop-dev/chainloop/pull/766

Knowing what I know now, I'd rather let the services figure out what's missing if needed, for example I can see that to get the workflowID

- att init (and contract get): let's take it from the provided workflow_name if needed 
- att push or get creds: get it from the workflowRunID

As you can see, there are different parameters in different API endpoints from where we can find the workflowID, so I think it's reasonable to let this logic be handled by the service layer, it's not generic enough.

> If we move that information to the services, they will be the one in charge of from a workflowName, take the workflowID. This middleware here won't make much sense then because it will be only setting the OrgID since the token ID was recently removed.

I think the middleware even in that case makes sene. It will do two things, validate that the token is not revoked and set the value of the OrgID. OrgID is very important to make sure we scope the calls to the biz layer.

app/controlplane/internal/service/attestation.go
```diff
@@ -361,6 +361,31 @@ func (s *AttestationService) GetUploadCreds(ctx context.Context, req *cpAPI.Atte
...
+	}
+
+	// If WorkflowID is not set and no workflowName is provided, return BadRequest.
+	if workflowName == "" && robotAccount.WorkflowID == "" {
+		return nil, errors.BadRequest("bad request", "when using an API Token, workflow name is required as parameter")
+	}
+
+	// If WorkflowID is not set, retrieve it using the workflow name.
+	if robotAccount.WorkflowID == "" {
+		workflow, err := s.workflowUseCase.FindByName(ctx, workflowName)
```
migmartri, line 379: For the case where we want to find the workflow from the name, we need to take into account the organization, since organization is the only element that we can trust (it came from the token) in the request.

otherwise, people could just circumvent these API calls, by providing other orgs names right?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -0,0 +1,269 @@
...
+	return func(o *providerOption) {
+		o.verifyAudienceFunc = f
+	}
+}
+
+type options struct {
+	tokenProviders []providerOption
+}
+
+func WithTokenProvider(providerKey string, opts ...TokenProviderOption) JWTOption {
```
migmartri, line 151: WithTokenProvider doens't need to be exposed anymore, let's make it downcase no?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	return func(o *options) {
+		o.tokenProviders = append(o.tokenProviders, *op)
+	}
+}
+
+// WithAttestationTokenMiddleware creates a custom JWT middleware that configured with different token providers
+// tries to run all validations from an incoming token. If you of the providers matches the expected audience
+// it gets parsed and sent down to the next middleware
+func WithAttestationTokenMiddleware(opts ...JWTOption) middleware.Middleware {
```
migmartri, line 166: We might want to indicate in the name that this is a multi-selector kind of middleware.

Also, the package is already called `attmiddleware` so we do not need to have the full name in the function.

What about `JWTMulti` or similar so we express two things, this is about handling the `jwt` (that's not shown anywhere anymore), and that's a multi selector.

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -0,0 +1,269 @@
...
+				}
+			}
+
+			return nil, ErrWrongContext
+		}
+	}
+}
+
+// runProviderValidator runs the token parser for the given provider
+func runProviderValidator(provider providerOption, jwtToken string) (*jwt.Token, error) {
```
migmartri, line 213: please add more commentary on where this code comes from. Thanks!

app/controlplane/internal/service/attestation.go
javirln: I've replaced it by the new `FindByNameInOrg` which does exactly that. Takes the orgID from the token.

app/controlplane/internal/data/workflow.go
```diff
@@ -203,10 +203,32 @@ func (r *WorkflowRepo) FindByID(ctx context.Context, id uuid.UUID) (*biz.Workflo
...
+	// Not efficient, we need to do a query limit = 1 grouped by workflowID
+	lastRun, err := getLastRun(ctx, workflow)
+	if err != nil {
+		return nil, err
+	}
+
+	return entWFToBizWF(workflow, lastRun)
+}
+
+func (r *WorkflowRepo) FindByName(ctx context.Context, name string) (*biz.Workflow, error) {
```
migmartri, line 222: so we do not need this method anymore do we?

app/controlplane/internal/biz/workflow.go
```diff
@@ -47,6 +47,7 @@ type WorkflowRepo interface {
 	GetOrgScopedByName(ctx context.Context, orgID uuid.UUID, workflowName string) (*Workflow, error)
 	IncRunsCounter(ctx context.Context, workflowID uuid.UUID) error
 	FindByID(ctx context.Context, workflowID uuid.UUID) (*Workflow, error)
+	FindByName(ctx context.Context, workflowName string) (*Workflow, error)
```
migmartri, line 50: we do not need this right?

app/controlplane/internal/data/workflow.go
javirln: No we don't, just removed it.

app/controlplane/internal/service/attestation.go
```diff
@@ -361,6 +361,31 @@ func (s *AttestationService) GetUploadCreds(ctx context.Context, req *cpAPI.Atte
 	return &cpAPI.AttestationServiceGetUploadCredsResponse{Result: resp}, nil
 }
 
+// completeRobotAccount auxiliary method to complete a RobotAccount instance. This is needed since when coming from
+// an API Token request, the only information we have set is the workflow name not the ID
+func (s *AttestationService) completeRobotAccount(ctx context.Context, workflowName string) (*usercontext.RobotAccount, error) {
```
migmartri, line 366: I find this method a little bit unnecessary now that it's used in one place only. It feels like too much to mutate an object to just get the workflowID. It could be simpler to add some of that code to the actual caller, and focus on that, get the workflowID not to mutate the robot account object. Does it make sense?

app/controlplane/internal/usercontext/apitoken_middleware.go
```diff
@@ -101,6 +102,72 @@ func WithCurrentAPITokenAndOrgMiddleware(apiTokenUC *biz.APITokenUseCase, orgUC
...
+				return handler(ctx, req)
+			}
+
+			genericClaims, ok := authInfo.Claims.(jwt.MapClaims)
+			if !ok {
+				return nil, errors.New("error mapping the claims")
+			}
+
+			// We've received an API-token
+			if genericClaims.VerifyAudience(apitoken.Audience, true) {
```
migmartri, line 127: shouldn't this fail otherwise?

migmartri (changes requested):
ptal at my comment on the audience check in the new APItoken middleware. I want to confirm it's correct

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -0,0 +1,276 @@
...
+func NewAPITokenProvider(signingSecret string) JWTOption {
+	return withTokenProvider(
+		APITokenProviderKey,
+		WithVerifyAudienceFunc(func(token *jwt.Token) bool {
+			claims, ok := token.Claims.(jwt.MapClaims)
+			if !ok {
+				return false
+			}
+			for _, aud := range []string{apitoken.Audience} {
+				if claims.VerifyAudience(aud, true) {
```
migmartri, line 90: this code could be simplified. 

`return claims.VerifyAudience(apitoken.Audience, true)`

app/controlplane/internal/service/attestation.go
javirln: Yes, it makes total sense!

app/controlplane/internal/usercontext/apitoken_middleware.go
javirln: Actually this check here can be deleted. The audience is already check previously on the `JWTMulti` middleware and we should be able to reach this point if and only if, the audience is valid.

I'll return an error.

migmartri (commented):
Some minor comments, lgtm overall!

app/controlplane/internal/usercontext/apitoken_middleware.go
migmartri: It's ok to keep it in two places. I lean towards defensive programming for these core pieces.

app/controlplane/internal/service/attestation.go
```diff
@@ -90,7 +90,22 @@ func NewAttestationService(opts *NewAttestationServiceOpts) *AttestationService
 func (s *AttestationService) GetContract(ctx context.Context, req *cpAPI.AttestationServiceGetContractRequest) (*cpAPI.AttestationServiceGetContractResponse, error) {
 	robotAccount := usercontext.CurrentRobotAccount(ctx)
 	if robotAccount == nil {
-		return nil, errors.NotFound("not found", "robot account not found")
+		return nil, errors.NotFound("not found", "neither robot account nor API token found")
+	}
+
+	// If WorkflowID is not set and no workflowName is provided, return BadRequest.
+	if req.GetWorkflowName() == "" && robotAccount.WorkflowID == "" {
```
migmartri, line 97: much simpler now, thanks :)

migmartri (approved):
Awesome! 🚀

PR #765, opened 2024-05-09
chore(controlplane): remove robor-account relatioship from workflowrun

migmartri:
Since we are moving to different authz methods, we no longer want to store the ra ID in the workflowrun.

This patch removes its references, some context:

- First I updated the ent/schemas of workflowrun and robotaccount and removed the relationship
- This means that after doing `make generate` a migration (and some generated code) that drops the column is created.
- From that point onward, it was about updating that `data`, and `biz`


jiparis (approved):
Awesome, thanks!

PR #766, opened 2024-05-10
refactor(cli): Use attestation state to pass workflow-name

javirln:
This PR leverages the changes introduced in https://github.com/chainloop-dev/chainloop/pull/763 to alway send the workflow-name along with the workflowID since the cli request might be happening with a API token.

This change ensures the CLI keeps working with both Robot Accounts and API tokens in a transparent manner to the end user. Only an initial workflow-name may be pass to attestation init if using and API token. Example:

```bash
$ chainloop attestation init --token $CHAINLOOP_ROBOT_ACCOUNT --wofklow-name wf-name

```


javirln:
https://github.com/chainloop-dev/chainloop/pull/763 needs to be merged first.

app/cli/cmd/attestation_init.go
```diff
@@ -50,7 +51,7 @@ func newAttestationInitCmd() *cobra.Command {
 			}
 
 			// Initialize it
-			attestationID, err := a.Run(cmd.Context(), contractRevision)
+			attestationID, err := a.Run(cmd.Context(), contractRevision, workflowName)
```
migmartri, line 54: could we detect if you are providing a robot-account and a name and fail by telling you that you can not use name (or that name is ignored) in that case?

app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -91,6 +100,9 @@ message AttestationServiceCancelRequest {
     not_in: [0]
   }];
   string reason = 3;
+  // This parameter is not needed by Robot Account since they have the workflowID embedded.
+  // API Tokens will send the parameter explicitly
+  string workflow_name = 4;
```
migmartri, line 105: I am struggling to understand why we need workflow_name in this situation. If you have the `run` and the organization in the token, aren't you supposed to also be able to know the workflow and make sure you have access to it?

app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -74,6 +80,9 @@ message AttestationServiceStoreRequest {
   // encoded DSEE envelope
   bytes attestation = 1 [(buf.validate.field).bytes.min_len = 1];
   string workflow_run_id = 2 [(buf.validate.field).string = {min_len: 1}];
+  // This parameter is not needed by Robot Account since they have the workflowID embedded.
+  // API Tokens will send the parameter explicitly
+  string workflow_name = 3;
```
migmartri, line 85: same here, I do not know why we need the name

app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -137,6 +149,9 @@ message WorkflowRunServiceViewResponse {
 
 message AttestationServiceGetUploadCredsRequest {
   string workflow_run_id = 1;
+  // This parameter is not needed by Robot Account since they have the workflowID embedded.
+  // API Tokens will send the parameter explicitly
+  string workflow_name = 2;
```
migmartri, line 154: same here, it looks like an unnecessary parameter.

migmartri (changes requested):
I might not undertand something but looking at this API, it feel like we are adding some additional requirements like `providing name in most of the API calls`, probably because of the way it's implemented in the backend.

I'd like to stop and re-evaluate this, it looks weird to me that the API should require `name` anywhere expect `init` and maybe `contract retrieval`. For the rest, you already have a runID and an organization, you should not need to pass the name afaic

Thanks!

javirln:
The whole reasoning of adding the `workflow_name` to the rest of APIs besides from the `GetContract` is to consolidate the Attestation operation by using the name rather than the ID, as we wanted to start using names in the resources.

In any case, we can just forget about that change and leave it strictly to `GetContract` where there it's compulsory.

app/cli/cmd/attestation_init.go
javirln: I've added warning message indicating it will be ignored.

migmartri (approved):
Minor comments, thanks

app/cli/cmd/attestation_init.go
```diff
@@ -38,6 +39,13 @@ func newAttestationInitCmd() *cobra.Command {
 		Annotations: map[string]string{
 			useWorkflowRobotAccount: "true",
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if os.Getenv(robotAccountEnvVarName) != "" && workflowName != "" {
```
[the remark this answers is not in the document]
javirln: One of my concerns is that the API token will be set on the env variable $CHAINLOOP_ROBOT_ACCOUNT because that's how it's named on the `attestation` command so this print will be always shown on the CLI when the `workflow_name` is set.

Is that what we want, to always show the message on API Tokens?

PR #771, opened 2024-05-13
feat(release): Add missing amd64 binaries

javirln:
This PR adds missing amd64 cli binaries to the release contract that were missing.

Refs #731

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -106,7 +106,7 @@ jobs:
           # Binaries x86_64
           # TODO: add the rest of binaries
           # NOTE that we are not making the attestation fail if the material is not found. We will fail on "att push"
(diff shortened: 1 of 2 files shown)


migmartri (approved):
Should we add the source code as artifact next?

javirln:
> Should we add the source code as artifact next?

Yes, I would do it in a different PR.

.github/workflows/release.yaml
```diff
@@ -106,7 +106,7 @@ jobs:
           # Binaries x86_64
           # TODO: add the rest of binaries
           # NOTE that we are not making the attestation fail if the material is not found. We will fail on "att push"
-          echo -n '${{ steps.release.outputs.artifacts }}' | jq -r '.[] | select(.type=="Binary" and .goos=="linux") | { "name": "\(.extra.ID)-\(.goos)-\(.goarch)", "path":"\(.path)"} | @base64' | while read i; do
+          echo -n '${{ steps.release.outputs.artifacts }}' | jq -r '.[] | select(.type=="Binary") | { "name": "\(.extra.ID)-\(.goos)-\(.goarch)", "path":"\(.path)"} | @base64' | while read i; do
```
danlishka, line 109: what if we start parsing the contract .github/workflows/contracts/chainloop-vault-release.yml instead ?

PR #772, opened 2024-05-14
feat(ci): Add source code tarball

javirln:
This PR adds Chainloop source code as a new attestation artifact.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,6 +27,7 @@ jobs:
       CHAINLOOP_ROBOT_ACCOUNT: ${{ secrets.CHAINLOOP_ROBOT_ACCOUNT }}
       CONTAINER_IMAGE_CP: ghcr.io/chainloop-dev/chainloop/control-plane:${{ github.ref_name }}
       CONTAINER_IMAGE_CAS: ghcr.io/chainloop-dev/chainloop/artifact-cas:${{ github.ref_name }}
+      GH_TOKEN: ${{ github.token }}
     steps:
       - name: Install Cosign
         uses: sigstore/cosign-installer@ef6a6b364bbad08abd36a5f8af60b595d12702f8 # main
@@ -41,7 +42,7 @@ jobs:
         run: |
           sudo wget -q https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -O /usr/local/bin/jq
           sudo chmod u+x /usr/local/bin/jq
-
+#
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
@@ -111,6 +112,16 @@ jobs:
               BINARY_PATH=$(echo "${i}" | base64 --decode | jq -r ${1} .path)
               chainloop attestation add --name ${BINARY_NAME} --value ${BINARY_PATH} || true
             done
+
+      - name: Add Attestation Artifacts (source code)
+        run: |
+          # When the trigger of the action is a release, github.ref contains refs/tags/<tag_name>
+          # Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
+
+          tag=$(echo -n ${{github.ref}} | cut -d / -f3)
+          version=$(echo -n $tag | sed 's/v//g')
+          gh release download $tag -A tar.gz -D /tmp
+          chainloop attestation add --name source-code --value "/tmp/chainloop-$version.tar.gz"
 
       - name: Finish and Record Attestation
         if: ${{ success() }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -112,6 +112,11 @@ jobs:
               chainloop attestation add --name ${BINARY_NAME} --value ${BINARY_PATH} || true
             done
 
+      - name: Add Attestation Artifacts (source code)
```
migmartri, line 115: related to a comment from @danlishka this morning. Is there any way to find this source code tgz from the github release itself instead of crafting our own?

image

javirln: Hmm I'll try to read about it and know where is that taken from. Maybe that's generated after the release process or in a different step that we don't have access.

javirln:
Updated with the retrieval of the source code from the release GitHub page.

.github/workflows/release.yaml
```diff
@@ -112,6 +112,15 @@ jobs:
...
+        run: |
+          # When the trigger of the action is a release, github.ref contains refs/tags/<tag_name>
+          # Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
+
+          tag=$(echo -n ${{github.ref}} | cut -d / -f3)
+          curl -sL -o /tmp/source-code.tar.gz "https://github.com/chainloop-dev/chainloop/archive/refs/tags/$tag.tar.gz"
+          chainloop attestation add --name source-code --value /tmp/source-code.tar.gz
```
migmartri, line 122: Interesting, it would be great to keep the file format too.

We could use the `gh` CLI, which I think it's installed in the runner, and as long as you have `permissions: read` in the job you should be able to use it.

For example, 

You can download the release source code by doing 
```
gh release download v0.87.1 -A tar.gz 
=> chainloop-0.87.1.tar.gz
```

In the future, we could basically download all the assets and attest them in chainloop.

```
$ gh release download v0.87.1 -D test 
$ ls test
cas.cyclonedx.json                            chainloop-cli-0.87.1-darwin-arm64.tar.gz      chainloop-cli-0.87.1-linux-amd64.tar.gz.sig  checksums.txt                cosign.pub
chainloop-cli-0.87.1-darwin-amd64.tar.gz      chainloop-cli-0.87.1-darwin-arm64.tar.gz.sig  chainloop-cli-0.87.1-linux-arm64.tar.gz      checksums.txt.sig
chainloop-cli-0.87.1-darwin-amd64.tar.gz.sig  chainloop-cli-0.87.1-linux-amd64.tar.gz       chainloop-cli-0.87.1-linux-arm64.tar.gz.sig  controlplane.cyclonedx.json
```

migmartri:
Please, while working on a feature, make sure you mark it as draft. I want to make sure I do not merge smth that I've approved before by mistake. 

Thanks!

.github/workflows/release.yaml
javirln: That's cool, I've added the needed changes and it's using now the GitHub CLI.

migmartri (approved):

.github/workflows/release.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
       CHAINLOOP_ROBOT_ACCOUNT: ${{ secrets.CHAINLOOP_ROBOT_ACCOUNT }}
       CONTAINER_IMAGE_CP: ghcr.io/chainloop-dev/chainloop/control-plane:${{ github.ref_name }}
       CONTAINER_IMAGE_CAS: ghcr.io/chainloop-dev/chainloop/artifact-cas:${{ github.ref_name }}
+      GH_TOKEN: ${{ github.token }}
```
migmartri, line 30: Are you sure you need to expose the env variable? I believe it also works with GITHUB_TOKEN https://cli.github.com/manual/gh_help_environment

migmartri (changes requested):

.github/workflows/release.yaml
javirln: I've added the env variable that was recommended by the CLI when running it and it stated that was GH_TOKEN :/

migmartri: what do you mean that it was recommended? A warning message or smth?

I can see this in their docs

> GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com API requests. Setting this avoids being prompted to authenticate and takes precedence over previously stored credentials.

javirln: >what do you mean that it was recommended? A warning message or smth?

Precisely it was this error message: https://github.com/chainloop-dev/chainloop/actions/runs/9079530767/job/24948940556

>gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}

migmartri: Interesting, thanks for the clarification.

FTR for the future, the error is 

> gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:

PR #773, opened 2024-05-14
chore(docs): Add CSAF contract example

javirln:
Updates de documentation examples with how CSAF profiles can be added to a Chainloop contract.

### DIFF
--- a/docs/examples/contracts/csaf/contract.yaml
+++ b/docs/examples/contracts/csaf/contract.yaml
@@ -1,6 +1,4 @@
 schemaVersion: v1
-runner:
-  type: GITHUB_ACTION
 materials:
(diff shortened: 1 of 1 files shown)


docs/examples/contracts/csaf/contract.yaml
```diff
@@ -0,0 +1,16 @@
+schemaVersion: v1
+runner:
+  type: GITHUB_ACTION
```
migmartri, line 3: you can remove the runner, or is there any reason you want it to be here? Note that the runner can be auto-detected.

migmartri (approved):

docs/examples/contracts/csaf/contract.yaml
javirln: No specific reason but since it's meant to be generic, let's remove to now give the impression it's tied up to that specific runner.

Issue #780, opened 2024-05-15
CLI telemetry


migmartri:
We'd like to add some basic, anonymous, optional telemetry in the CLI that could help us understand what commands are being used.

- ID must always be protected and anonymous, we could use https://github.com/denisbrodbeck/machineid
- We must support https://consoledonottrack.com/
- We must not record sensitive information. Probably not arguments.

PR #781, opened 2024-05-15
feat(posthog): Add Telemetry client Posthog

javirln:
This PR creates a new implementation of a Tracker client based on Posthog.

Users can deactivate telemetry by setting the env variable `DO_NOT_TRACK=1`, it follows the convention at: https://consoledonottrack.com/

Information being tracked:
- System information (OS, ARCH)
- Command ran
- CI Runner information if detected
- CLI version
- Hashed Control Plane URL
- Anonymous device ID
- User and Organization if logged in

Refs: #780

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -28,6 +28,8 @@ builds:
     ldflags:
       - "{{ .Env.COMMON_LDFLAGS }}"
       - -X github.com/chainloop-dev/chainloop/app/cli/cmd.Version={{ .Version }}
+      - -X github.com/chainloop-dev/chainloop/app/cli/cmd.posthogAPIKey={{ .Env.POSTHOG_API_KEY }}
+      - -X github.com/chainloop-dev/chainloop/app/cli/cmd.posthogEndpoint={{ .Env.POSTHOG_ENDPOINT }}
     targets:
       - darwin_amd64
       - darwin_arm64
--- a/app/cli/cmd/root.go
+++ b/app/cli/cmd/root.go
@@ -16,14 +16,18 @@
 package cmd
 
 import (
-	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
+	"sync"
 
 	"github.com/adrg/xdg"
 	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
+	"github.com/chainloop-dev/chainloop/app/cli/internal/telemetry"
+	"github.com/chainloop-dev/chainloop/app/cli/internal/telemetry/posthog"
 	"github.com/chainloop-dev/chainloop/internal/grpcconn"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/rs/zerolog"
@@ -53,12 +57,16 @@ const (
 	userAudience         = "user-auth.chainloop"
 	//nolint:gosec
 	apiTokenAudience = "api-token-auth.chainloop"
+	// Follow the convention stated on https://consoledonottrack.com/
+	doNotTrackEnv = "DO_NOT_TRACK"
 )
 
-type AuthenticationToken struct{}
-type ParsedToken struct {
-	ID   string
-	Type string
+var telemetryWg sync.WaitGroup
+
+type parsedToken struct {
+	id        string
+	orgID     string
+	tokenType string
 }
 
 func NewRootCmd(l zerolog.Logger) *cobra.Command {
@@ -92,14 +100,27 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 
 			actionOpts = newActionOpts(logger, conn)
 
-			// For telemetry reasons we parse the token to know the type of token is being used when executing the CLI
-			// Once we have the token type we can send it to the telemetry service by injecting it on the context
-			token, err := parseToken(apiToken)
-			if err != nil {
-				logger.Debug().Err(err).Msg("parsing token for telemetry")
-			}
+			if !isTelemetryDisabled() {
+				logger.Debug().Msg("Telemetry enabled, to disable it use DO_NOT_TRACK=1")
 
-			cmd.SetContext(context.WithValue(cmd.Context(), AuthenticationToken{}, token))
+				telemetryWg.Add(1)
+				go func() {
+					defer telemetryWg.Done()
+
+					// For telemetry reasons we parse the token to know the type of token is being used when executing the CLI
+					// Once we have the token type we can send it to the telemetry service by injecting it on the context
+					token, err := parseToken(apiToken)
+					if err != nil {
+						logger.Debug().Err(err).Msg("parsing token for telemetry")
+						return
+					}
+
+					err = recordCommand(cmd, token)
+					if err != nil {
+						logger.Debug().Err(err).Msg("sending command to telemetry")
+					}
+				}()
+			}
 
 			return nil
 		},
@@ -143,6 +164,16 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 
 func init() {
 	cobra.OnInitialize(initConfigFile)
+	// Using the cobra.OnFinalize because the hooks don't work on error
+	cobra.OnFinalize(func() {
+		// In some cases the command is faster than the telemetry, in that case we wait
+		telemetryWg.Wait()
+	})
+}
+
+// isTelemetryDisabled checks if the telemetry is disabled by the user or if we are running a development version
+func isTelemetryDisabled() bool {
+	return os.Getenv(doNotTrackEnv) == "1" || os.Getenv(doNotTrackEnv) == "true" || Version == devVersion
 }
 
 func initLogger(logger zerolog.Logger) (zerolog.Logger, error) {
@@ -239,46 +270,125 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
 // 3. API token
 // Each one of them have an associated audience claim that we use to identify the type of token. If the token is not
 // present, nor we cannot match it with one of the expected audience, return nil.
-func parseToken(token string) (*ParsedToken, error) {
+func parseToken(token string) (*parsedToken, error) {
 	if token == "" {
-		return &ParsedToken{}, nil
+		return nil, nil
 	}
 
 	// Create a parser without claims validation
 	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
 
 	// Parse the token without verification
-	parsedToken, _, err := parser.ParseUnverified(token, jwt.MapClaims{})
+	t, _, err := parser.ParseUnverified(token, jwt.MapClaims{})
 	if err != nil {
 		return nil, err
 	}
 
-	// Extract claims
-	claims := parsedToken.Claims.(jwt.MapClaims)
-
-	// Get the audience claim
-	if val, ok := claims["aud"]; ok && val != nil {
-		// Chainloop tokens have only one audience in an array
-		aud, ok := val.([]interface{})
-		if !ok {
-			return nil, nil
-		}
-		if len(aud) == 0 {
-			return nil, nil
-		}
-
-		switch aud[0].(string) {
-		case apiTokenAudience:
-			return &ParsedToken{Type: "api-token"}, nil
-		case userAudience:
-			userID := claims["user_id"].(string)
-			return &ParsedToken{Type: "user", ID: userID}, nil
-		case robotAccountAudience:
-			return &ParsedToken{Type: "robot-account"}, nil
-		default:
-			return nil, nil
-		}
+	// Extract generic claims otherwise, we would have to parse
+	// the token again to get the claims for each type
+	claims, ok := t.Claims.(jwt.MapClaims)
+	if !ok {
+		return nil, nil
 	}
 
-	return nil, nil
+	// Get the audience claim
+	val, ok := claims["aud"]
+	if !ok || val == nil {
+		return nil, nil
+	}
+
+	// Ensure audience is an array of interfaces
+	// Chainloop only has one audience per token
+	aud, ok := val.([]interface{})
+	if !ok || len(aud) == 0 {
+		return nil, nil
+	}
+
+	// Initialize parsedToken
+	pToken := &parsedToken{}
+
+	// Determine the type of token based on the audience.
+	switch aud[0].(string) {
+	case apiTokenAudience:
+		pToken.tokenType = "api-token"
+		if tokenID, ok := claims["jti"].(string); ok {
+			pToken.id = tokenID
+		}
+		if orgID, ok := claims["org_id"].(string); ok {
+			pToken.orgID = orgID
+		}
+	case userAudience:
+		pToken.tokenType = "user"
+		if userID, ok := claims["user_id"].(string); ok {
+			pToken.id = userID
+		}
+	case robotAccountAudience:
+		pToken.tokenType = "robot-account"
+		if tokenID, ok := claims["jti"].(string); ok {
+			pToken.id = tokenID
+		}
+		if orgID, ok := claims["org_id"].(string); ok {
+			pToken.orgID = orgID
+		}
+	default:
+		return nil, nil
+	}
+
+	return pToken, nil
+}
+
+var (
+	// Posthog API key and endpoint to be injected in the build process
+	posthogAPIKey   = ""
+	posthogEndpoint = ""
+)
+
+// recordCommand sends the command to the telemetry service
+func recordCommand(executedCmd *cobra.Command, authInfo *parsedToken) error {
+	telemetryClient, err := posthog.NewClient(posthogAPIKey, posthogEndpoint)
+	if err != nil {
+		logger.Debug().Err(err).Msgf("creating telemetry client: %v", err)
+	}
+
+	cmdTracker := telemetry.NewCommandTracker(telemetryClient)
+	tags := telemetry.Tags{
+		"cli_version":      Version,
+		"cp_url_hash":      hashControlPlaneURL(),
+		"chainloop_source": "cli",
+	}
+
+	// It tries to extract the token from the context and add it to the tags. If it fails, it will ignore it.
+	if authInfo != nil {
+		tags["token_type"] = authInfo.tokenType
+		tags["user_id"] = authInfo.id
+		tags["org_id"] = authInfo.orgID
+	}
+
+	if err = cmdTracker.Track(executedCmd.Context(), extractCmdLineFromCommand(executedCmd), tags); err != nil {
+		return fmt.Errorf("sending event: %w", err)
+	}
+
+	return nil
+}
+
+// extractCmdLineFromCommand returns the full command hierarchy as a string from a cobra.Command
+func extractCmdLineFromCommand(cmd *cobra.Command) string {
+	var cmdHierarchy []string
+	currentCmd := cmd
+	// While the current command is not the root command, keep iteration.
+	// This is done to get the full hierarchy of the command and remove the root command from the hierarchy.
+	for currentCmd.Use != "chainloop" {
+		cmdHierarchy = append([]string{currentCmd.Use}, cmdHierarchy...)
+		currentCmd = currentCmd.Parent()
+	}
+
+	cmdLine := strings.Join(cmdHierarchy, " ")
+	return cmdLine
+}
+
+// hashControlPlaneURL returns a hash of the control plane URL
+func hashControlPlaneURL() string {
+	url := viper.GetString("control-plane.API")
+
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(url)))
 }
--- a/app/cli/cmd/root_test.go
+++ b/app/cli/cmd/root_test.go
@@ -25,38 +25,68 @@ func TestParseToken(t *testing.T) {
 	tests := []struct {
 		name  string
 		token string
-		want  ParsedToken
+		want  *parsedToken
 	}{
 		{
 			name:  "robot account",
-			token: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJvcmdfaWQiOiI5M2QwMjI3NS04NTNjLTRhZDYtOWQ2MC04ZjU2MmIxMjNmZDIiLCJ3b3JrZmxvd19pZCI6IjM1ZTZkOGIwLWE0OGYtNDFmYS05YmU3LWQ1OTM5YjJkZGUyNiIsImlzcyI6ImNwLmNoYWlubG9vcCIsImF1ZCI6WyJhdHRlc3RhdGlvbnMuY2hhaW5sb29wIl0sImp0aSI6IjQ4YWVhMWNiLTk5MGUtNDM2OS1hOTFhLTczNTIzNzk1NjhiNSJ9.aYPAPK-AauJpxRGEapV2ejwxhyhmFej6S79Ni-xJ4Q0",
-			want: ParsedToken{
-				Type: "robot-account",
+			token: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJvcmdfaWQiOiJkZGRiODIwMS1lYWI2LTRlNjEtOTIwMS1mMTJiNDdjMDE4OTIiLCJ3b3JrZmxvd19pZCI6ImY0NmIzMDc5LTMwOGYtNGIwNC1hYjYwLTY3NDNmOGUzMGQzMyIsImlzcyI6ImNwLmNoYWlubG9vcCIsImF1ZCI6WyJhdHRlc3RhdGlvbnMuY2hhaW5sb29wIl0sImp0aSI6IjkzODI5NDE1LTA1ODYtNDFkYS05NTJkLTkyYTRjNDk1ZWEyMCJ9.3Af2C62PiODbEknhv47j0LHLuAgWLqvTrfmIzFjwPCM",
+			want: &parsedToken{
+				id:        "93829415-0586-41da-952d-92a4c495ea20",
+				tokenType: "robot-account",
+				orgID:     "dddb8201-eab6-4e61-9201-f12b47c01892",
 			},
 		},
 		{
 			name:  "user account",
 			token: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiYmMwYjIxOTktY2E4NS00MmFiLWE4NTctMDQyZTljMTA5ZDQzIiwiaXNzIjoiY3AuY2hhaW5sb29wIiwiYXVkIjpbInVzZXItYXV0aC5jaGFpbmxvb3AiXSwiZXhwIjoxNzE1OTM1MjUwfQ.ounYshGtagtYQsVIzNeE0ztVYRXrmjFSpdmaTF4QvyY",
-			want: ParsedToken{
-				ID:   "bc0b2199-ca85-42ab-a857-042e9c109d43",
-				Type: "user",
+			want: &parsedToken{
+				id:        "bc0b2199-ca85-42ab-a857-042e9c109d43",
+				tokenType: "user",
 			},
 		},
 		{
 			name:  "api token",
-			token: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJjcC5jaGFpbmxvb3AiLCJhdWQiOlsiYXBpLXRva2VuLWF1dGguY2hhaW5sb29wIl0sImp0aSI6IjE0NTQ5MzUwLTFjNGItNDdlYi05NDZkLWY2MjJhZWYyMDk0MyJ9.BPzuNxSwx10h22fJ3ocAOEIjsq9OOlk9p8fSoCwqSmM",
-			want: ParsedToken{
-				Type: "api-token",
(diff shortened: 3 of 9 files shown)


app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 			return nil
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			// Do not track telemetry if the env variable is set
+			var telemetryClient telemetry.Client
+			if os.Getenv(doNotTrackEnv) == "1" || os.Getenv(doNotTrackEnv) == "true" {
+				logger.Debug().Msg("Telemetry is disabled")
+			} else {
+				var err error
+				telemetryClient, err = posthog.NewClient(posthogAPIKey, posthogEndpoint)
```
migmartri, line 102: so you do not want to crash on purpose correct?

app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 			return nil
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			// Do not track telemetry if the env variable is set
+			var telemetryClient telemetry.Client
+			if os.Getenv(doNotTrackEnv) == "1" || os.Getenv(doNotTrackEnv) == "true" {
```
migmartri, line 98: it would be nice if the detection could be extracted as a function `isTelemetryDisabled() bool`

app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
...
+			var cmdHierarchy []string
+			currentCmd := cmd
+			// While the current command is not the root command, keep iteration.
+			// This is done to get the full hierarchy of the command and remove the root command from the hierarchy.
+			for currentCmd.Use != "chainloop" {
+				cmdHierarchy = append([]string{currentCmd.Use}, cmdHierarchy...)
+				currentCmd = currentCmd.Parent()
+			}
+
+			cmdLine := strings.Join(cmdHierarchy, " ")
```
migmartri, line 119: and the command line generation also to be extracted to a function too so it could potentially be tested.

app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
...
+
+			err := cmdTracker.Track(cmd.Context(), cmdLine, telemetry.Tags{
+				"cli_version": Version,
```
migmartri, line 122: it would be great if we could also capture the conttrolplaneURL but hash it anonymously. We have the controlplane URI in viper, I guess we should be able to find it.

app/cli/internal/telemetry/posthog/posthog.go
```diff
@@ -53,14 +53,20 @@ type Tracker struct {
 
 // TrackEvent sends an event to the PostHog server.
 func (p *Tracker) TrackEvent(_ context.Context, eventName string, id string, tags telemetry.Tags) error {
+	defer p.client.Close()
```
migmartri, line 56: why is this needed? To send the event once the function finishes?

app/cli/cmd/root.go
```diff
@@ -47,6 +50,13 @@ const (
 	appName                 = "chainloop"
 	//nolint:gosec
 	tokenEnvVarName = "CHAINLOOP_TOKEN"
+	doNotTrackEnv   = "DO_NOT_TRACK"
```
migmartri, line 53: add a comment with a link to https://consoledonottrack.com/

app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 			return nil
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			// Do not track telemetry if the env variable is set
+			var telemetryClient telemetry.Client
+			if os.Getenv(doNotTrackEnv) == "1" || os.Getenv(doNotTrackEnv) == "true" {
+				logger.Debug().Msg("Telemetry is disabled")
+			} else {
+				var err error
```
migmartri, line 101: I'd add a debug statement saying "Telemetry enabled, to disable it use DO_NOT_TRACK=1"

app/cli/cmd/root.go
```diff
@@ -83,6 +93,38 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 			return nil
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+			// Do not track telemetry if the env variable is set
+			var telemetryClient telemetry.Client
+			if os.Getenv(doNotTrackEnv) == "1" || os.Getenv(doNotTrackEnv) == "true" {
+				logger.Debug().Msg("Telemetry is disabled")
```
migmartri, line 99: we might not need this debug statement but instead put one when it's enabled so they know how to disable it. It will also remove the if/else

migmartri (commented):
Cool!

app/cli/cmd/root.go
migmartri: and also the command status. Isn't it a way to detect if the command returned an error? We capture errors somewhere else globally to do some manual handling, i.e show you that you need to re-login

app/cli/internal/telemetry/posthog/posthog.go
javirln: Yes and to ensure the client is properly closed.

app/cli/cmd/root.go
javirln: I have added both of them computing the hash of the control plane url

app/cli/cmd/root.go
javirln: Yes, I don't want to crash if the telemetry fails, I think that could disturb users and automations

javirln (commented):
Refactored the code to be able to check the exit code as it was not trivial since cobra hooks are not executed when a command fails.

In any case I think it's pretty neat, we have everything we wanted.

app/cli/main.go
```diff
@@ -16,25 +16,75 @@
...
+	// Run the command and get the executed command
+	executedCmd, err := rootCmd.ExecuteC()
+	if err != nil {
+		var msg string
+		msg, exitCode = errorInfo(err, logger)
 		logger.Error().Msg(msg)
-		os.Exit(exitCode)
+	}
+
+	runTelemetry(logger, executedCmd, exitCode)
```
jiparis, line 61: I see that only succeeded commands will be tracked.
Do you thins this could be moved to the `PreRunE` (or PostRun) method in the rootCmd command instead?

javirln: No no, it will track both successful and unsuccessful commands that's why it's moved into the main package. The problem with the `PostRun(E)` is that they don't run if there has been an error. Moving the logic to the main caller, `rootCmd.Execute` give us the chance to hook it up to know when there has been an error and track it as well.

Additionally to know what's the actual command that has been run we moved it to use `rootCmd.ExecuteC` which returns the executed command therefore in this way we have the whole picture :)

migmartri: but to be clear, there is a bug in that code that doesn't do that right? That was a catch from @jiparis.

jiparis: Sorry, my bad, I didn't notice the removal of `os.Exit(exitCode)`. Ignore my comment.

jiparis (approved):

javirln:
Secrets on the repository updated already to the proper project.

.goreleaser.yml
```diff
@@ -28,6 +28,8 @@ builds:
     ldflags:
       - "{{ .Env.COMMON_LDFLAGS }}"
       - -X github.com/chainloop-dev/chainloop/app/cli/cmd.Version={{ .Version }}
+      - -X main.posthogAPIKey={{ .Env.POSTHOG_API_KEY }}
```
migmartri, line 31: Let's make sure we have a page documenting this change that goes up in the same Chainloop release :)

app/cli/cmd/root.go
```diff
@@ -269,12 +269,14 @@ func parseToken(token string) (*ParsedToken, error) {
 
 		switch aud[0].(string) {
 		case apiTokenAudience:
-			return &ParsedToken{Type: "api-token"}, nil
+			tokenID := claims["jti"].(string)
```
migmartri, line 272: you might want to protect this casting a little bit
```
tokenID, ok := claims ...
if !ok {
return nil, nil
}
```

app/cli/cmd/root_test.go
```diff
@@ -31,6 +31,7 @@ func TestParseToken(t *testing.T) {
 			name:  "robot account",
 			token: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJvcmdfaWQiOiI5M2QwMjI3NS04NTNjLTRhZDYtOWQ2MC04ZjU2MmIxMjNmZDIiLCJ3b3JrZmxvd19pZCI6IjM1ZTZkOGIwLWE0OGYtNDFmYS05YmU3LWQ1OTM5YjJkZGUyNiIsImlzcyI6ImNwLmNoYWlubG9vcCIsImF1ZCI6WyJhdHRlc3RhdGlvbnMuY2hhaW5sb29wIl0sImp0aSI6IjQ4YWVhMWNiLTk5MGUtNDM2OS1hOTFhLTczNTIzNzk1NjhiNSJ9.aYPAPK-AauJpxRGEapV2ejwxhyhmFej6S79Ni-xJ4Q0",
 			want: ParsedToken{
+				ID:   "48aea1cb-990e-4369-a91a-7352379568b5",
```
migmartri, line 34: thanks for adding tests.

app/cli/cmd/root.go
```diff
@@ -269,12 +269,14 @@ func parseToken(token string) (*ParsedToken, error) {
...
+			return &ParsedToken{Type: "api-token", ID: tokenID}, nil
 		case userAudience:
 			userID := claims["user_id"].(string)
 			return &ParsedToken{Type: "user", ID: userID}, nil
 		case robotAccountAudience:
-			return &ParsedToken{Type: "robot-account"}, nil
+			tokenID := claims["jti"].(string)
+			return &ParsedToken{Type: "robot-account", ID: tokenID}, nil
```
migmartri, line 279: btw, organization has been added and will be available in both robot-account and api-token once #795 is merged :)

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+
 func main() {
 	// Couldn't find an easier way to disable the timestamp
 	logger := zerolog.New(zerolog.ConsoleWriter{Out: os.Stderr, FormatTimestamp: func(interface{}) string { return "" }})
 	rootCmd := cmd.NewRootCmd(logger)
 
-	if err := rootCmd.Execute(); err != nil {
-		msg, exitCode := errorInfo(err, logger)
+	// Set the default exit code to 0, in case of an error it will be updated
+	exitCode := 0
```
migmartri, line 51: nitpick. Leveraging default values would have helped in this case too. i.e `var exitCode int`

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+	if err != nil {
+		var msg string
+		msg, exitCode = errorInfo(err, logger)
 		logger.Error().Msg(msg)
-		os.Exit(exitCode)
+	}
+
+	runTelemetry(logger, executedCmd, exitCode)
+
+	os.Exit(exitCode)
```
migmartri, line 63: do we want to `os.Exit` even if we have 0 as returned code? I know it's the last statement of the program but I am not sure if there is an additional consequence of doing this instead of letting the program finish as before

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+}
+
+// runTelemetry sends the telemetry data to the configured endpoint
+func runTelemetry(logger zerolog.Logger, executedCmd *cobra.Command, exitCode int) {
```
migmartri, line 67: I'd call this method, `recordCommand` to be more specific of the scope of this telemetry

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+	var telemetryClient telemetry.Client
+
+	if !isTelemetryDisabled() {
```
migmartri, line 70: tbh, at this point is probably better that we wrap this `runTelemetry` method within `!isTelemetryDisabled`, that way we do not run any useless compute like generating all those tags and so on if we know it's not going to happen.

In other words, I'd do something like

if !isTelemetryDisabled() { runTelemetry(...) }

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+		var err error
+		telemetryClient, err = posthog.NewClient(posthogAPIKey, posthogEndpoint)
+		if err != nil {
+			logger.Debug().Err(err).Msgf("creating telemetry client: %v", err)
```
migmartri, line 74: why don't we return here?

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+	}
+
+	if parsedToken := executedCmd.Context().Value(cmd.AuthenticationToken{}); parsedToken != nil {
+		tags["token_type"] = parsedToken.(*cmd.ParsedToken).Type
+		tags["user_id"] = parsedToken.(*cmd.ParsedToken).ID
+	}
+
+	err := cmdTracker.Track(executedCmd.Context(), extractCmdLineFromCommand(executedCmd), tags)
+	if err != nil {
+		logger.Debug().Err(err).Msgf("tracking command: %v", err)
```
migmartri, line 94: same here, we could return technically

app/cli/internal/telemetry/telemetry.go
```diff
@@ -49,28 +50,56 @@ func NewCommandTracker(client Client) *CommandTracker {
...
+		result[k] = v
+	}
+	for k, v := range userTags {
+		result[k] = v
+	}
+	return result
+}
+
+// Determines the user ID using available information or generates a random one.
+func determineUserID(tags map[string]string) string {
```
migmartri, line 84: it would be great to add some tests for this method to help explain what it does.

app/cli/internal/telemetry/telemetry.go
```diff
@@ -49,28 +50,56 @@ func NewCommandTracker(client Client) *CommandTracker {
...
+	// Get the machine ID.
+	machineID, _ := machineid.ProtectedID("chainloop")
+	tags["machine_id"] = machineID
+
+	// Check if user ID is provided in tags.
```
migmartri, line 89: add a comment saying that this is not the case on the unauthenticated scenario

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+
+	cmdTracker := telemetry.NewCommandTracker(telemetryClient)
+	tags := telemetry.Tags{
+		"cli_version": cmd.Version,
+		"exit_code":   fmt.Sprint(exitCode),
+		"cp_url_hash": hashControlPlaneURL(),
+		"source":      "cli",
+	}
+
+	if parsedToken := executedCmd.Context().Value(cmd.AuthenticationToken{}); parsedToken != nil {
```
migmartri, line 87: add a comment saying that this context information might not be here if the user is not authenticated, and that's ok, expected and handled by the underlying library.

migmartri (approved):
Added some minor comments :)

app/cli/main.go
```diff
@@ -16,25 +16,82 @@
...
+		}
+		logger.Debug().Msg("Telemetry enabled, to disable it use DO_NOT_TRACK=1")
+	}
+
+	cmdTracker := telemetry.NewCommandTracker(telemetryClient)
+	tags := telemetry.Tags{
+		"cli_version": cmd.Version,
+		"exit_code":   fmt.Sprint(exitCode),
+		"cp_url_hash": hashControlPlaneURL(),
+		"source":      "cli",
```
migmartri, line 84: I'd rename this to `chainloop_source` so we can differentiate it from other common analytic property. What do you think?

app/cli/main.go
javirln: I know but the other option I think it's weird:
```golang
if !isTelemetryDisabled() {
  recordCommand(logger, executedCmd, exitCode)
}

if exitCode != 0 {
  os.Exit(exitCode)
}
```

app/cli/main.go
javirln: Sure

app/cli/main.go
javirln: We actually could, I'll change it.

app/cli/main.go
javirln: Sure.

app/cli/cmd/root.go
javirln: Awesome

.goreleaser.yml
javirln: Yes, I still have pending add a documentation page.

app/cli/main.go
migmartri: yeah, I agree, my question was more out of curiosity, to know if there was a difference between letting the program finish or calling exit(0), it might not be.

javirln: Ahh sorry! To be honest I'm not sure, I guess it won't have any different :/

app/cli/main.go
```diff
@@ -16,25 +16,86 @@
...
+	}
+
+	os.Exit(exitCode)
+}
+
+// recordCommand sends the command to the telemetry service
+func recordCommand(logger zerolog.Logger, executedCmd *cobra.Command, exitCode int) {
+	telemetryClient, err := posthog.NewClient(posthogAPIKey, posthogEndpoint)
+	if err != nil {
+		logger.Debug().Err(err).Msgf("creating telemetry client: %v", err)
```
migmartri, line 72: this debug entry seems to be shown even if debug level is not enabled. 

```
./bin/chainloop wf ls
ERR authentication required, please run "chainloop auth login"
DBG creating telemetry client: invalid configuration, API Key and endpoint URL are required error="invalid configuration, API Key and endpoint URL are required"
```

not sure why, or if this is the expected behavior.

app/cli/main.go
```diff
@@ -16,25 +16,86 @@
 package main
 
 import (
+	"crypto/sha256"
```
migmartri, line 19: could you bump the year on the header?

app/cli/main.go
migmartri: ok, the problem is that this logger is not correctly initialized the one in root is

```
func initLogger(logger zerolog.Logger) (zerolog.Logger, error) {
	lvl := zerolog.InfoLevel
	if flagDebug {
		lvl = zerolog.DebugLevel
	}

	return logger.Level(lvl), nil
}
```

This means that the logger I believe is in debug mode out of the box, not great I think. Not sure how to solve it though. At least we should make sure that these errors that are not meant to scare the user are not shown.

migmartri: I can see that a way to do this would be to newRootCommand to return also the initialized logger, to main can use it accordingly

migmartri: not trivial, we could leave it for now but create a task to propagate the level to main, wdyt?

javirln: Yeah...I was giving a try this morning precisely with what you mention of returning the logger but it seems it does not take into account the previously set level...
I'll create a task for it.

app/cli/main.go
javirln: Sure

app/cli/main.go
migmartri: well,not really, do you also see this message on every command?

```
./bin/chainloop wf ls                                                  
DBG Telemetry enabled, to disable it use DO_NOT_TRACK=1
```

migmartri (changes requested):
Do not show debug messages by default, especially the one about telemetry enabled

app/cli/main.go
javirln: Yes in all commands, it does not matter the value of the debug flag it will always be shown.

javirln:
> Do not show debug messages by default, especially the one about telemetry enabled

I've moved the message to the `root` command. I was not able to set the proper flag from outside not making the logger to be consistent with the logging level. I even tried the global setting from zerlog without luck.

migmartri (commented):
Thanks,

I tried to figure out the `flag` issue and boy it's quite confusing, ..., I kinda managed to fix it but I do not like my solution. In am ok going with your solution for now but we should add a task to fix it. 

Re telemetry performance, I'll take a quick look and see if I find anything, once done I think I'll be ready to approve this and move forward. Thanks!

app/cli/internal/telemetry/telemetry.go
```diff
@@ -49,28 +50,57 @@ func NewCommandTracker(client Client) *CommandTracker {
...
+		result[k] = v
+	}
+	for k, v := range userTags {
+		result[k] = v
+	}
+	return result
+}
+
+// DetermineUserID Determines the user ID using available information or generates a random one.
+func DetermineUserID(tags Tags) string {
```
migmartri, line 84: this function probably doesn't need to be exposed. I am assuming it is because the test package uses `_test`. In that case I'd ask the question if we need the `test` file to use a different package or just treat it as unit tests in the `telemetry` package.

What do you think?

app/cli/main.go
```diff
@@ -16,25 +16,86 @@
...
+		"chainloop_source":  "cli",
+		"execution_time_ms": fmt.Sprint(executionTime.Milliseconds()),
+	}
+
+	// It tries to extract the token from the context and add it to the tags. If it fails, it will ignore it.
+	contextToken := executedCmd.Context().Value(cmd.AuthenticationToken{})
+	if contextToken != nil {
+		parsedToken, ok := contextToken.(*cmd.ParsedToken)
+		if ok {
+			tags["token_type"] = parsedToken.Type
```
migmartri, line 91: I get a panic when using locally a CHAINLOOP_API_TOKEN during an attestation

```
./bin/chainloop att init --replace
ERR rpc error: code = InvalidArgument desc = when using an API Token, workflow name is required as parameter
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x201ea00]

goroutine 1 [running]:
main.recordCommand(0xc000f47508, 0x1, 0x18fffa5e)
        /home/migmartri/work/chainloop/chainloop/app/cli/main.go:91 +0x3a0
main.main()
        /home/migmartri/work/chainloop/chainloop/app/cli/main.go:64 +0x2c8
```

migmartri: Ok, the problem was that my token didn't have `org_id` since the current live backend doesn't have it yet. 

but I think this is surfacing another problem, that the `parseToken` method can return 3 values, `nil`, empty token `&ParsedToken{}` or a full-fledged token. 

We should probably only return `nil` or a token and handle it accordingly.

app/cli/cmd/root.go
```diff
@@ -269,12 +283,31 @@ func parseToken(token string) (*ParsedToken, error) {
 
 		switch aud[0].(string) {
 		case apiTokenAudience:
-			return &ParsedToken{Type: "api-token"}, nil
+			tokenID, ok := claims["jti"].(string)
```
migmartri, line 286: related to another of my comments, I think we should improve this method, it relies too much in casting, causing for example to return `nil` if the token doesn't have a property.

I think this could be improved by 

a - leveraging `claims.VerifyAudience(aud, true) bool`
b - once found, we could map to their actual claims so they are no longer a map of interface and can be accessed properly. 

I think this way the function will be less brittle, what do you think?

app/cli/cmd/root_test.go
```diff
@@ -44,8 +45,9 @@ func TestParseToken(t *testing.T) {
 		},
 		{
 			name:  "api token",
```
migmartri, line 47: you should probably try a token that's missing some properties. That's a case that could happen in the future and will happen today (old API tokens without org)

javirln:
Refactored all code with:
- Moved all logic from `main` to `root`.
- Use a separated go routine to send the telemetry on the `persistentPreRunE`. Use a `sync.WaitGroup` to wait for it in case it's slow.
- Refactored pase of the jwt token in a generic way, otherwise we would have to cast the token to each type of claim coupling backend and CLI due to types being on the backend.
- Added more tests with empty fields and random tokens.

app/cli/main.go
```diff
@@ -31,9 +31,12 @@ func main() {
 	logger := zerolog.New(zerolog.ConsoleWriter{Out: os.Stderr, FormatTimestamp: func(interface{}) string { return "" }})
 	rootCmd := cmd.NewRootCmd(logger)
 
+	// Run the command
 	if err := rootCmd.Execute(); err != nil {
+		var msg string
```
migmartri, line 36: we can probably revert all the changes in this file. For example, having the msg var  doesn't seem to make sense anymore.

app/cli/cmd/root.go
```diff
@@ -53,12 +57,16 @@ const (
...
 )
 
-type AuthenticationToken struct{}
-type ParsedToken struct {
-	ID   string
-	Type string
+var telemetryWg sync.WaitGroup
+
+type parsedToken struct {
+	ID    string
```
migmartri, line 67: just noticed, these properties do not need to be uppercase either

app/cli/cmd/root.go
```diff
@@ -92,15 +100,26 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
...
+				}
+
+				telemetryWg.Add(1)
+				go func() {
+					err := recordCommand(cmd, token)
+					if err != nil {
+						logger.Debug().Err(err).Msg("sending command to telemetry")
+					}
+
+					telemetryWg.Done()
```
migmartri, line 119: this probably should be called using `defer` `defer telemetryWg.Done()`

app/cli/cmd/root.go
```diff
@@ -92,15 +100,26 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
...
-			// For telemetry reasons we parse the token to know the type of token is being used when executing the CLI
-			// Once we have the token type we can send it to the telemetry service by injecting it on the context
-			token, err := parseToken(apiToken)
-			if err != nil {
-				logger.Debug().Err(err).Msg("parsing token for telemetry")
+			if !isTelemetryDisabled() {
+				logger.Debug().Msg("Telemetry enabled, to disable it use DO_NOT_TRACK=1")
+				// For telemetry reasons we parse the token to know the type of token is being used when executing the CLI
+				// Once we have the token type we can send it to the telemetry service by injecting it on the context
+				token, err := parseToken(apiToken)
```
migmartri, line 107: just realized that this can probably be inside the `subroutine` too

app/cli/internal/telemetry/telemetry.go
migmartri, line 84: `DetermineUserID` can be downcase if you move the tests to the same package

migmartri (approved):

app/cli/cmd/root.go
```diff
@@ -239,46 +268,125 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
 // 3. API token
 // Each one of them have an associated audience claim that we use to identify the type of token. If the token is not
 // present, nor we cannot match it with one of the expected audience, return nil.
-func parseToken(token string) (*ParsedToken, error) {
+func parseToken(token string) (*parsedToken, error) {
```
jiparis, line 271: Not for this PR, but most probably we can have a common interface. Something like 
```
type Token interface {
  func Parse(token string) (*parsedToken, error)
}
```
And then moving the "extracting" logic to their own implementations.

app/cli/cmd/root.go
```diff
@@ -239,46 +268,125 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
...
+			pToken.OrgID = orgID
 		}
+	default:
+		return nil, nil
+	}
+
+	return pToken, nil
+}
+
+var (
```
jiparis, line 338: Not important: Do you think we can move the below code to its own file `telemetry.go` or similar just for readability?

app/cli/internal/telemetry/telemetry.go
javirln: I've added it into a separated packaged because if added to the same package it creates an import cycle between the mocks and the implementation:
```
FAIL	github.com/chainloop-dev/chainloop/app/cli/internal/telemetry [setup failed]
package github.com/chainloop-dev/chainloop/app/cli/internal/telemetry
	imports github.com/chainloop-dev/chainloop/app/cli/internal/telemetry/mocks
	imports github.com/chainloop-dev/chainloop/app/cli/internal/telemetry: import cycle not allowed in test
```

migmartri: You are right, this is the problem of using the mocks generated by mockery :(

A workaround is to create another file, `telemerty_unit_test.go` or similar and have these unit tests. 

In general when I end up writing code in a specific way just so it works in the tests, I realize that I might be doing smth wrong and re-evaluate. But up to you. 

Thansk!

app/cli/cmd/root.go
javirln: I would do it if we really support more telemetry and figure out the problem with the logger since right now it's coupled to it.

app/cli/internal/telemetry/telemetry.go
javirln: >A workaround is to create another file, telemerty_unit_test.go or similar and have these unit tests.

But the problem would still be the same, the method will still be public right?

I mean for me, I wouldn't even create a test for this specific function since they are covered by the tests on the `Track` method that includes it.

PR #782, opened 2024-05-16
feat(authentication): Parse auth token on root command

javirln:
Parse the stored authentication token to know what kind of user is using the CLI.

This is a best effort approach.

Refs: #780

### DIFF
--- a/app/cli/cmd/root.go
+++ b/app/cli/cmd/root.go
@@ -51,7 +51,8 @@ const (
 	tokenEnvVarName      = "CHAINLOOP_TOKEN"
 	robotAccountAudience = "attestations.chainloop"
 	userAudience         = "user-auth.chainloop"
-	apiTokenAudience     = "api-token-auth.chainloop"
+	//nolint:gosec
+	apiTokenAudience = "api-token-auth.chainloop"
 )
 
 type AuthenticationToken struct{}
--- a/app/cli/cmd/root.go
+++ b/app/cli/cmd/root.go
@@ -92,12 +92,13 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 
 			actionOpts = newActionOpts(logger, conn)
 
+			// For telemetry reasons we parse the token to know the type of token is being used when executing the CLI
+			// Once we have the token type we can send it to the telemetry service by injecting it on the context
 			token, err := parseToken(apiToken)
 			if err != nil {
 				logger.Debug().Err(err).Msg("parsing token for telemetry")
 			}
 
-			// Inject the authentication token into the command context
 			cmd.SetContext(context.WithValue(cmd.Context(), AuthenticationToken{}, token))
 
 			return nil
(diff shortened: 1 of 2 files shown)


app/cli/cmd/root.go
```diff
@@ -211,3 +227,39 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
 	// loaded from config file, previously stored via "auth login"
 	return viper.GetString(confOptions.authToken.viperKey), nil
 }
+
+// ParseToken the token and return the type of token
+func ParseToken(token string) (*ParsedToken, error) {
```
migmartri, line 232: why is it an exposed function?

app/cli/cmd/root.go
```diff
@@ -211,3 +227,39 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
...
+	parsedToken, _, err := parser.ParseUnverified(token, jwt.MapClaims{})
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract claims
+	claims := parsedToken.Claims.(jwt.MapClaims)
+
+	// Check for specific claims and return accordingly
+	if val, ok := claims["workflow_id"]; ok && val != nil {
```
migmartri, line 250: do we have any other way to detect it? Could the audience or the issuer give you the hint? I thought the issuer actually would tell you what it is.

app/cli/cmd/root.go
javirln: To be unit tested.

app/cli/cmd/root.go
javirln: The issuer is the same for all tokens.

We can potentially try the audience which values are:
- api-token-auth.chainloop
- user-auth.chainloop
- attestations.chainloop

But then the mapping of Claims make it harder but let me have a try.

migmartri (approved):

app/cli/cmd/root.go
```diff
@@ -211,3 +231,48 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
...
+		switch aud[0].(string) {
+		case apiTokenAudience:
+			return &ParsedToken{Type: "api-token"}, nil
+		case userAudience:
+			userID := claims["user_id"].(string)
+			return &ParsedToken{Type: "user", ID: userID}, nil
+		case robotAccountAudience:
+			return &ParsedToken{Type: "robot-account"}, nil
+		default:
+			return &ParsedToken{}, nil
```
migmartri, line 273: return nil, nil?

migmartri (approved):

app/cli/cmd/root.go
```diff
@@ -211,3 +232,53 @@ func loadControlplaneAuthToken(cmd *cobra.Command) (string, error) {
...
+		if !ok {
+			return nil, nil
+		}
+		if len(aud) == 0 {
+			return nil, nil
+		}
+
+		switch aud[0].(string) {
+		case apiTokenAudience:
+			return &ParsedToken{Type: "api-token"}, nil
```
migmartri, line 272: my bad, shouldn't we add the identifier of the token? That way we can know how much for example a single token is being used in an org?

migmartri: and I mean the identifier of both the API_TOKEN and the RobotAccount

javirln: Do you see it as the `ID` on the `ParsedToken` or a new field `TokenID`?

migmartri: mm, I was thinking that maybe using the current ID in the ParsedToken? Why do you think we should do it as TokenID?

javirln: No no, to have different fields to filter by on the UI but we can merge them into `ID` since we have the type of token to know if it's a user.

migmartri: ahh, I see, no idea tbh. You probably know more.

Issue #783, opened 2024-05-16
attestation push and reset does not work on API Tokens


javirln:
It seems the attestation push and attestation reset fails when using the API tokens due to not being able to find a workflow ID.

API Tokens does not have a workflow ID associated.

PR #784, opened 2024-05-16
fix(attestation): Parse correctly workflowrun ids

javirln:
This patch fixes the attestation process when using API tokens since they were failing due to workflow run id not being available.

Refs: #783

### DIFF
--- a/app/controlplane/internal/service/attestation.go
+++ b/app/controlplane/internal/service/attestation.go
@@ -213,18 +213,17 @@ func (s *AttestationService) Store(ctx context.Context, req *cpAPI.AttestationSe
 		return nil, errors.NotFound("not found", "robot account not found")
 	}
 
-	if robotAccount.WorkflowID == "" {
-		wr, err := s.wrUseCase.GetByIDInOrgOrPublic(ctx, robotAccount.OrgID, req.WorkflowRunId)
-		if err != nil {
-			return nil, handleUseCaseErr(err, s.log)
-		}
-		robotAccount.WorkflowID = wr.Workflow.ID.String()
-	} else {
-		// Check that provided workflowRun belongs to workflow encoded in the robot account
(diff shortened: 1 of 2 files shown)


app/controlplane/internal/service/attestation.go
```diff
@@ -213,9 +213,17 @@ func (s *AttestationService) Store(ctx context.Context, req *cpAPI.AttestationSe
 		return nil, errors.NotFound("not found", "robot account not found")
 	}
 
-	// Check that provided workflowRun belongs to workflow encoded in the robot account
-	if exists, err := s.wrUseCase.ExistsInWorkflow(ctx, robotAccount.WorkflowID, req.WorkflowRunId); err != nil || !exists {
-		return nil, errors.NotFound("not found", "workflowRun not found")
+	if robotAccount.WorkflowID == "" {
```
migmartri, line 216: I thought that we said that we were ok with just checking if it belongs to an org, in both cases.

app/controlplane/internal/service/attestation.go
```diff
@@ -213,9 +213,17 @@ func (s *AttestationService) Store(ctx context.Context, req *cpAPI.AttestationSe
...
+		wr, err := s.wrUseCase.GetByIDInOrgOrPublic(ctx, robotAccount.OrgID, req.WorkflowRunId)
```
migmartri, line 217: I do not think this method is what we want. As the name suggests, this will return a run  that belongs to you or it's public, we do not want you to be able to attest public workflows.

The point of the check is to make sure that the run is from one of your orgs.

migmartri (changes requested):

javirln:
Reworked the method to use a proper way to find Workflow Runs that matches the given organization (from API token or Robot Account).

javirln:
Closing in favor of https://github.com/chainloop-dev/chainloop/pull/790

PR #793, opened 2024-05-16
feat(attestations): handle authentication combinations

migmartri:
This code improves the user feedback by returning an error when the user is using a robot-account-based attestation and a workflow-name at the same time.

I do it by refactoring a little bit the code to explicitly check the kind of authentication is received and handle both cases. Note that I required to do a change in the CLI too so send the `name` only if only the user provided it.

API token auth missing name

```
go run main.go --insecure att init --replace                         
ERR rpc error: code = InvalidArgument desc = when using an API Token, workflow name is required as parameter
```

using ROBOT account with name

```
go run main.go --insecure att init --replace --workflow-name fo
ERR rpc error: code = InvalidArgument desc = workflow name is not compatible with robot-account based attestations
```


javirln (approved):

jiparis (approved):
Looks good!

PR #795, opened 2024-05-16
chore: add organizationID to API token

migmartri:
The generated API JWT token didn't contain the organization ID in it, which made it non-consistent with the robot-account token.

This patch a) adds the organization to the token and b) makes sure that the middleware does the proper parsing 

refs #780


javirln (approved):
LGTM

PR #797, opened 2024-05-17
feat(cli): improve attestation push result in table format

### LINKED ISSUE #750: Improve `att push` output
now, `att push` supports both `-o json` and regular view. This task is about improving the default output to show something similar we show in our labs trust report

https://github.com/chainloop-dev/chainloop/actions/runs/8938087539#summary-24551614195

I believe that that output is the status of the attestation right before pushing it. We should also show the digest of the pushed att and maybe the duration.

jiparis:
This PR improves the way `att push` result is shown. It uses now the status action logic (in `full` mode) before running the actual push to gather all the information from the current attestation.
```
✗ go run app/cli/main.go --insecure --token $TOKEN att push --key ./cosign.key
Enter password for private key:
INF push completed
│ Initialized At    │ 17 May 24 11:11 UTC                  │
│ Attestation ID    │ efb770a4-203f-4bba-9310-50272e39de26 │
│ Name              │ mywf                                 │
│ Project           │ myproj                               │
│ Contract Revision │ 2                                    │
┌─────────────────────────────────────────────────────────────────────────────────────┐
│ Materials                                                                           │
├───────────┬─────────────────────────────────────────────────────────────────────────┤
│ Name      │ foo                                                                     │
│ Type      │ ARTIFACT                                                                │
│ Set       │ Yes                                                                     │
│ Required  │ Yes                                                                     │
│ Is output │ Yes                                                                     │
│ Value     │ contract.json                                                           │
│ Digest    │ sha256:7d2681cd27240354f571b662108ca73c78fb2a6130aa6dc16e04af8905e4c361 │
└───────────┴─────────────────────────────────────────────────────────────────────────┘
INF Attestation Digest: sha256:eea3b1f58ed98b8b548321f850a751b0a8f066fc2cc3734781225a8d6fbc6608
```

Closes #750

### DIFF
--- a/app/cli/cmd/attestation_init.go
+++ b/app/cli/cmd/attestation_init.go
@@ -77,7 +77,7 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
-			return encodeOutput(res, func(result *action.AttestationStatusResult) error { return attestationStatusTableOutput(res, false) })
+			return encodeOutput(res, simpleStatusTable)
 		},
 	}
 
--- a/app/cli/cmd/attestation_init.go
+++ b/app/cli/cmd/attestation_init.go
@@ -77,7 +77,7 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
-			return encodeOutput(res, attestationStatusTableOutput)
+			return encodeOutput(res, func(result *action.AttestationStatusResult) error { return attestationStatusTableOutput(res, false) })
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


app/cli/cmd/attestation_init.go
```diff
@@ -77,7 +77,7 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
-			return encodeOutput(res, attestationStatusTableOutput)
```
migmartri, line 80: another way of doing this a little bit more declarative could have been to have two aliases.

```go
func emptyStatusTable(result *action.AttestationStatusResult) {
attestationStatusTableOutput)(result, false)
}


func fullStatusTable(result *action.AttestationStatusResult) {
attestationStatusTableOutput)(result, true)
}
```

Then you could keep the return similar `return encodeOutput(res, emptyStatusTable)`

migmartri (approved):

migmartri (approved):
Thank you!

PR #775, opened 2024-05-14
feat(ci): Use reusable workflow creation on Helm package

javirln:
This PR leverages the resolvable workflow on Chainloop labs ([link](https://github.com/chainloop-dev/labs/blob/main/.github/workflows/chainloop_onboard.yml)) that onboards a new Workflow created from the context of the GitHub Action, if it does not exist.

Additionally, it uses a global API Token to perform all steps in the attestation process.

Refs https://github.com/chainloop-dev/platform/issues/630

PR #802, opened 2024-05-20
Revert "feat(ci): Use reusable workflow creation on Helm package (#775)"

migmartri:
Reverting until we deploy the fixed backend

This reverts commit b889590bc9c00d670fc0648e8ac91d0a241311de.


javirln (approved):
I believe the signing is failing.

PR #814, opened 2024-05-20
feat(posthog): Inject Posthog credentials

javirln:
This patch adds public Posthog API key that points to Chainloop instance so it can be included in external builds such as Homebrew.

[Refs: #676](https://github.com/chainloop-dev/platform/issues/676)


jiparis (approved):
Looks good, thanks!

migmartri (approved):

Issue #785, opened 2024-05-16
Consume GitHub releases


migmartri:
We want to be able to attest a gitHub release in an easy way. 

Ideally, we could have a reusable workflow/action that reacts to a `release` event and performs an attestation of all the published artifacts.

As explained here https://github.com/chainloop-dev/chainloop/pull/772#discussion_r1599939348 we can download all the artifacts using the GH CLI, and then we could push them to Chainloop by using a `CHAINLOOP_TOKEN`

We could even automatically create a workflow if needed using also our new reusable workflow for that purpose cc/ @jiparis 

```[tasklist]
### Tasks
- [ ] https://github.com/chainloop-dev/chainloop/issues/786
- [ ] https://github.com/chainloop-dev/chainloop/issues/816
- [ ] https://github.com/chainloop-dev/chainloop/issues/788
- [ ] https://github.com/chainloop-dev/chainloop/issues/787
- [ ] https://github.com/chainloop-dev/chainloop/issues/844
- [x] add guard to check it's a release
- [ ] https://github.com/chainloop-dev/chainloop/issues/849
- [ ] https://github.com/chainloop-dev/chainloop/issues/852
- [ ] Be able to configure the output of contract-less materials
- [x] Modify existing release contracts?
- [ ] https://github.com/chainloop-dev/chainloop/issues/866
- [ ] https://github.com/chainloop-dev/chainloop/issues/867
- [ ] https://github.com/chainloop-dev/chainloop/issues/868
```

javirln:
One of the issues found when deploying the reusable workflow into our pipelines is the fact that a workflow cannot be triggered by other workflow if the token used is the default GitHub token. That provoked that the `build and package` workflow didn't the `release` workflow since, what triggers a release within `build and package` is `goreleaser` using a GitHub Token.

More information regarding the issue can be found here: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

This does not mean the workflow does not work, it only means that in order to be triggered by another workflow it needs to use a personal access token, manual releases of course trigger the workflow.

migmartri:
Should we support workflow dispatch? That one seems to be supported.

javirln:
Additionally to the `release` trigger? 
How do you think it would be the flow for our use case? Once `build and package` is finished, us manually going to the workflow and trigger it? In that case we need to check that the github.ref_name is still the tag

migmartri:
[...]

PR #815, opened 2024-05-21
feat(attestation): Allow contract-less materials

javirln:
This patch allows users to add materials to a contract without the need of it to be specified on it by passing the material's `kind` inline. 

Old API is still working as it used before, a new `--kind` is added to tell the CLI the material type. There are constraints:
- Both `--name` and `--kind` cannot be empty at the same time.
- Both `--name` and `--kind` can be set at the same time but `--kind` will be ignored and a warning message will be printed to the user if `--debug` is activated

Collisions between material's names and the ones outside the contract are prevented by the generation of a random name.

Example:

Given the following contract:
```bash
$ chainloop --insecure attestation init --replace --workflow-name wf-test
│ Initialized At    │ 21 May 24 07:23 UTC                  │
│ Attestation ID    │ 6e9d2e76-fdc0-4493-a344-4cf44a2b7bf2 │
│ Name              │ wf-test                              │
│ Team              │ founding                             │
│ Project           │ core                                 │
│ Contract Revision │ 3                                    │
┌────────────────────────┐
│ Materials              │
├───────────┬────────────┤
│ Name      │ one-file   │
│ Type      │ ARTIFACT   │
│ Set       │ No         │
│ Required  │ Yes        │
│ Is output │ Yes        │
├───────────┼────────────┤
│ Name      │ other-file │
│ Type      │ EVIDENCE   │
│ Set       │ No         │
│ Required  │ Yes        │
│ Is output │ Yes        │
└───────────┴────────────┘
```
There are two expected materials of kind `ARTIFACT` and `EVIDENCE`. With the changes we can perform the following:
```bash
$ chainloop --insecure attestation add --value controlplane.cyclonedx.json  --kind SBOM_CYCLONEDX_JSON
```
We are explicitly saying which is the kind (`SBOM_CYCLONEDX_JSON`) we want to apply to material that is not found on the contract. After fulfilling the required materials, the ones stated on the contract, we can perform the attestation as usual:
```bash
$ chainloop --insecure attestation push --key cosign.key
Enter password for private key:
INF push completed
│ Initialized At    │ 21 May 24 06:37 UTC                  │
│ Attestation ID    │ 57b8dc22-d84d-40c3-a04c-8948231b3134 │
│ Name              │ wf-test                              │
│ Team              │ founding                             │
│ Project           │ core                                 │
│ Contract Revision │ 3                                    │
┌─────────────────────────────────────────────────────────────────────────────────────┐
│ Materials                                                                           │
├───────────┬─────────────────────────────────────────────────────────────────────────┤
│ Name      │ one-file                                                                │
│ Type      │ ARTIFACT                                                                │
│ Set       │ Yes                                                                     │
│ Required  │ Yes                                                                     │
│ Is output │ Yes                                                                     │
│ Value     │ go.mod                                                                  │
│ Digest    │ sha256:29773f085c46a33efcb6cdb185f6ec30ce1c4ca708b860708cd055b70488ef4d │
├───────────┼─────────────────────────────────────────────────────────────────────────┤
│ Name      │ other-file                                                              │
│ Type      │ EVIDENCE                                                                │
│ Set       │ Yes                                                                     │
│ Required  │ Yes                                                                     │
│ Is output │ Yes                                                                     │
│ Value     │ LICENSE.md                                                              │
│ Digest    │ sha256:c71d239df91726fc519c6eb72d318ec65820627232b2f796219e87dcf35d0ab4 │
├───────────┼─────────────────────────────────────────────────────────────────────────┤
│ Name      │ material-1716273574284654000                                            │
│ Type      │ SBOM_CYCLONEDX_JSON                                                     │
│ Set       │ Yes                                                                     │
│ Required  │ No                                                                      │
│ Is output │ Yes                                                                     │
│ Value     │ controlplane.cyclonedx.json                                             │
│ Digest    │ sha256:a6bc29d7a2d8d9f6df12a86ee4c86c58189d77bb6ded9487330c39f46ee00d9a │
└───────────┴─────────────────────────────────────────────────────────────────────────┘
Attestation Digest: sha256:61832d846b870d01647a384c3df49e3c75fd87f45821c9295d97ab91d5bae198
```

Refs #785


app/cli/cmd/attestation_add.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 func newAttestationAddCmd() *cobra.Command {
-	var name, value string
+	var name, value, kind string
 	var artifactCASConn *grpc.ClientConn
 	var annotationsFlag []string
```
migmartri, line 33: we should probably add some examples of how to use this command now. See https://github.com/chainloop-dev/chainloop/blob/90d5a932ca32a11ce0262a21e4440fa84d3f147f/app/cli/cmd/attestation_push.go#L32 for some examples

app/cli/cmd/attestation_add.go
```diff
@@ -47,6 +47,16 @@ func newAttestationAddCmd() *cobra.Command {
 		Annotations: map[string]string{
 			useWorkflowRobotAccount: "true",
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			switch {
+			case name == "" && kind == "":
+				return fmt.Errorf("both --name and --kind cannot be empty")
+			case name != "" && kind != "":
+				logger.Warn().Msg("both --name and --kind are provided, --kind will be ignored")
```
migmartri, line 55: I personally lean towards rapid and clear user feedback. In this case I think it would be more useful to fail.

app/cli/cmd/attestation_add.go
```diff
@@ -47,6 +47,16 @@ func newAttestationAddCmd() *cobra.Command {
 		Annotations: map[string]string{
 			useWorkflowRobotAccount: "true",
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			switch {
+			case name == "" && kind == "":
+				return fmt.Errorf("both --name and --kind cannot be empty")
```
migmartri, line 53: this doesn't sound right, but it might be. fwiw, I'd have wrote it as 

either --name or --kind needs to be set

app/cli/cmd/attestation_add.go
```diff
@@ -90,13 +100,12 @@ func newAttestationAddCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&name, "name", "", "name of the material to be recorded")
```
migmartri, line 102: maybe be more explicit here in the description of the flag.

`name of the material as shown in the contract`

app/cli/cmd/attestation_add.go
```diff
@@ -90,13 +100,12 @@ func newAttestationAddCmd() *cobra.Command {
...
 	cmd.Flags().StringVar(&value, "value", "", "value to be recorded")
-	err = cmd.MarkFlagRequired("value")
+	err := cmd.MarkFlagRequired("value")
 	cobra.CheckErr(err)
 	cmd.Flags().StringSliceVar(&annotationsFlag, "annotation", nil, "additional annotation in the format of key=value")
 	flagAttestationID(cmd)
+	cmd.Flags().StringVar(&kind, "kind", "", "kind of the material to be recorded")
 
 	// Optional OCI registry credentials
 	cmd.Flags().StringVar(&registryServer, "registry-server", "", fmt.Sprintf("OCI repository server, ($%s)", registryServerEnvVarName))
```
migmartri, line 111: not for this Pr, but we might want to add supporting `output` flag too

migmartri (commented):
Not finished the review yet, just leaving some comments for now so you might be able to make progress 

Thansk!

internal/attestation/crafter/crafter.go
```diff
@@ -463,8 +463,32 @@ func (c *Crafter) ResolveEnvVars(ctx context.Context, attestationID string) erro
 	return nil
 }
 
-// Inject material to attestation state
-func (c *Crafter) AddMaterial(ctx context.Context, attestationID, key, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
+// AddMaterialContractFree adds a material to the crafting state without checking the contract schema.
+// This is useful for adding materials that are not defined in the schema.
+// The name of the material is automatically calculated to conform the API contract.
+func (c *Crafter) AddMaterialContractFree(ctx context.Context, attestationID, kind, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
```
migmartri, line 469: I like this name and thanks for the documentation on the method :)

internal/attestation/crafter/crafter.go
```diff
@@ -487,6 +511,12 @@ func (c *Crafter) AddMaterial(ctx context.Context, attestationID, key, value str
 	}
 
 	// 3 - Craft resulting material
+	return c.addMaterial(ctx, m, attestationID, key, value, casBackend, runtimeAnnotations)
+}
+
+// addMaterials adds the incoming material m to the crafting state
+func (c *Crafter) addMaterial(ctx context.Context, m *schemaapi.CraftingSchema_Material, attestationID, key, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
```
migmartri, line 518: You might don't need to pass the key since `m` has a `name` correct?

internal/attestation/crafter/crafter.go
```diff
@@ -463,8 +463,32 @@ func (c *Crafter) ResolveEnvVars(ctx context.Context, attestationID string) erro
...
+	if err := c.requireStateLoaded(); err != nil {
+		return err
+	}
+
+	// 1 - Try to parse incoming type to a known kind
+	m := schemaapi.CraftingSchema_Material{}
+	if val, found := schemaapi.CraftingSchema_Material_MaterialType_value[kind]; found {
+		m.Type = schemaapi.CraftingSchema_Material_MaterialType(val)
+	} else {
+		return fmt.Errorf("%q not found as available kind", kind)
```
migmartri, line 479: can we return in the error available kinds are

```
"%q kind not found. Available options are %q", ...
```

internal/attestation/crafter/crafter.go
```diff
@@ -463,8 +463,32 @@ func (c *Crafter) ResolveEnvVars(ctx context.Context, attestationID string) erro
...
+// The name of the material is automatically calculated to conform the API contract.
+func (c *Crafter) AddMaterialContractFree(ctx context.Context, attestationID, kind, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
+	if err := c.requireStateLoaded(); err != nil {
+		return err
+	}
+
+	// 1 - Try to parse incoming type to a known kind
+	m := schemaapi.CraftingSchema_Material{}
+	if val, found := schemaapi.CraftingSchema_Material_MaterialType_value[kind]; found {
+		m.Type = schemaapi.CraftingSchema_Material_MaterialType(val)
```
migmartri, line 477: awesome, glad that you are using this :)

internal/attestation/crafter/crafter.go
```diff
@@ -463,8 +463,32 @@ func (c *Crafter) ResolveEnvVars(ctx context.Context, attestationID string) erro
...
 }
 
-// Inject material to attestation state
-func (c *Crafter) AddMaterial(ctx context.Context, attestationID, key, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
+// AddMaterialContractFree adds a material to the crafting state without checking the contract schema.
+// This is useful for adding materials that are not defined in the schema.
+// The name of the material is automatically calculated to conform the API contract.
+func (c *Crafter) AddMaterialContractFree(ctx context.Context, attestationID, kind, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) error {
+	if err := c.requireStateLoaded(); err != nil {
+		return err
```
migmartri, line 471: I know this is not from this PR, but you should probably return fmt.Errorf here for consistency.

app/cli/internal/action/attestation_status.go
```diff
@@ -99,26 +99,14 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
-			if err := setMaterialValue(cm, materialResult.Material); err != nil {
-				return nil, err
-			}
-			materialResult.Set = true
-			materialResult.Tag = cm.GetContainerImage().GetTag()
-		}
+	// Temporal map to keep track of visited materials
+	visitedMaterials := make(map[string]struct{})
+	if err := action.populateContractMaterials(res, visitedMaterials); err != nil {
+		return nil, err
```
migmartri, line 105: fmt.Errorf please

app/cli/internal/action/attestation_status.go
```diff
@@ -150,6 +138,61 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
+		// No need to check for name collisions, as it is not defined in the contract schema and it's
+		// autogenerated by the crafter
+		materialResult := &AttestationStatusResultMaterial{
+			Material: &Material{
+				Name:        name,
+				Type:        m.GetMaterialType().String(),
+				Annotations: stateAnnotationToAction(m.Annotations),
+			},
+			// No need to check if the material is optional or not, as it is not defined in the contract schema
+			IsOutput: true, Required: false,
```
migmartri, line 182: add a todo saying that `isOutput` this will be configurable someday

Out of curiosity, why did you choose to make this output `true`?

app/cli/internal/action/attestation_status.go
```diff
@@ -150,6 +138,61 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
+		}
+
+		if err := setMaterialValue(m, materialResult.Material); err != nil {
+			return err
```
migmartri, line 186: fmt.Errorf thanks!

app/cli/internal/action/attestation_status.go
```diff
@@ -150,6 +138,61 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
+				Annotations: pbAnnotationsToAction(m.Annotations),
+			},
+			IsOutput: m.Output, Required: !m.Optional,
+		}
+
+		if cm, found := action.c.CraftingState.Attestation.Materials[m.Name]; found {
+			if err := setMaterialValue(cm, materialResult.Material); err != nil {
+				return err
+			}
+			materialResult.Set = true
```
migmartri, line 156: it might make sense to add these two lines (set and tag) inside `setMaterialValue`

app/cli/internal/action/attestation_status.go
```diff
@@ -99,26 +99,14 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
-
-		// If it has been added already we load the value
-		if cm, found := c.CraftingState.Attestation.Materials[m.Name]; found {
-			if err := setMaterialValue(cm, materialResult.Material); err != nil {
-				return nil, err
-			}
-			materialResult.Set = true
-			materialResult.Tag = cm.GetContainerImage().GetTag()
-		}
+	// Temporal map to keep track of visited materials
```
migmartri, line 102: I'd add a comment explaining why we need to do lookups on basically two collections.

We want to make sure we show all the expected materials, that's why we load them from the contract

but also want to show additional items that might not be in the contract once they are set.

app/cli/internal/action/attestation_status.go
```diff
@@ -150,6 +138,61 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
+		}
+
+		res.Materials = append(res.Materials, *materialResult)
+		visitedMaterials[m.Name] = struct{}{}
+	}
+	return nil
+}
+
+// populateAdditionalMaterials populates the materials that are not defined in the contract schema
+func (action *AttestationStatus) populateAdditionalMaterials(res *AttestationStatusResult, visitedMaterials map[string]struct{}) error {
```
migmartri, line 167: These functions could start using some testing.

app/cli/cmd/attestation_add.go
```diff
@@ -90,13 +100,12 @@ func newAttestationAddCmd() *cobra.Command {
...
 	cmd.Flags().StringVar(&name, "name", "", "name of the material to be recorded")
-	err := cmd.MarkFlagRequired("name")
-	cobra.CheckErr(err)
 	cmd.Flags().StringVar(&value, "value", "", "value to be recorded")
-	err = cmd.MarkFlagRequired("value")
+	err := cmd.MarkFlagRequired("value")
 	cobra.CheckErr(err)
 	cmd.Flags().StringSliceVar(&annotationsFlag, "annotation", nil, "additional annotation in the format of key=value")
 	flagAttestationID(cmd)
+	cmd.Flags().StringVar(&kind, "kind", "", "kind of the material to be recorded")
```
migmartri, line 108: could we show here what options are available?

I believe you did something similar already somewhere else.

migmartri (commented):
Looks good to me, added some comments but none of them are feature related.

Good job!

Note: I will approve once we improve the user feedback on the command

internal/attestation/crafter/crafter.go
javirln: Hmm yes, it's included in `m` we can use it.

internal/attestation/crafter/crafter.go
javirln: Ran into a few problems that I'm glad I finally overcome them :D

internal/attestation/crafter/crafter.go
javirln: Sure thing, let's add it.

app/cli/internal/action/attestation_status.go
javirln: > Out of curiosity, why did you choose to make this output true?
I thought we might wanted them to be shown in the final graph however, thinking now I think that's the default behavior right?

Shall we set it to false?

app/cli/internal/action/attestation_status.go
javirln: I'll try to add them

app/cli/internal/action/attestation_status.go
javirln: Sure

app/cli/cmd/attestation_add.go
javirln: Yep, I'll add them

jiparis: +1 I think we can print the list of supported material types. It should be easy from the proto enum.

jiparis (commented):
It's looking great, @javirln.

app/cli/cmd/attestation_add.go
```diff
@@ -47,6 +48,23 @@ func newAttestationAddCmd() *cobra.Command {
 		Annotations: map[string]string{
 			useWorkflowRobotAccount: "true",
 		},
+		Example: `  chainloop attestation add --attestation-id <attestation-id> --name <material-name> --value <material-value> --kind <material-kind> --annotation key=value,key2=value2
```
migmartri, line 51: I'd not add `--attestation-id <attestation-id>` since this is a feature not enabled by default, you will need to `init` with `remote-state`

migmartri: In fact, that doesn't seem an example, since kind and name are incompatible. If we want to show all the options, it might make more sense to use it somewhere else, is there an `usage` option?

app/cli/internal/action/attestation_status.go
migmartri: TBH I do not know, it's probably sensible to stick to the `false` default.

In any case I believe this should be configurable in the future, probably passing a flag. Could you create a follow up task for that?

app/cli/internal/action/attestation_status.go
```diff
@@ -150,6 +136,78 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string)
...
+			},
+			// No need to check if the material is optional or not, as it is not defined in the contract schema
+			// TODO: Make IsOutput configurable
+			IsOutput: true, Required: false,
+		}
+
+		if err := setMaterialValue(m, materialResult); err != nil {
+			return fmt.Errorf("setting material value: %w", err)
+		}
+		materialResult.Set = true
```
migmartri, line 203: did you miss removing these lines from here?

app/cli/internal/action/attestation_status.go
```diff
@@ -163,16 +221,35 @@ func pbAnnotationsToAction(in []*pbc.Annotation) []*Annotation {
...
+			Name:  k,
+			Value: v,
+		})
+	}
+
+	return res
+}
+
+func setMaterialValue(w *v1.Attestation_Material, o *AttestationStatusResultMaterial) error {
+	setCommonFields := func(value, hash, tag string) {
```
migmartri, line 239: Wouldn't be more Go idiomatic to just mutate the struct like this at the top of the function?

```
o.Set = true
o.Tag = v.GetContainerImage().GetTag()
```

The current implementation is strange because it seems it's meant to give you control on what to pass depending on the value, but then you do not use that control (you are always passing the `getContainerImage().GetTag()`

internal/attestation/crafter/crafter.go
```diff
@@ -43,6 +43,28 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// attestationCraftingMaterial is a map of material types to their corresponding schemaapi.CraftingSchema_Material_MaterialType
+var attestationCraftingMaterial = func() map[string]schemaapi.CraftingSchema_Material_MaterialType {
+	res := make(map[string]schemaapi.CraftingSchema_Material_MaterialType)
+	for k, v := range schemaapi.CraftingSchema_Material_MaterialType_value {
+		if k != "MATERIAL_TYPE_UNSPECIFIED" {
+			res[strings.Replace(k, "MATERIAL_TYPE_", "", 1)] = schemaapi.CraftingSchema_Material_MaterialType(v)
```
migmartri, line 51: Do you need to do a replace here?

```
CraftingSchema_Material_MaterialType_value = map[string]int32{
		"MATERIAL_TYPE_UNSPECIFIED":       0,
		"STRING":                          1,
		"CONTAINER_IMAGE":                 2,
		"ARTIFACT":                        3,
		"SBOM_CYCLONEDX_JSON":             4,
		"SBOM_SPDX_JSON":                  5,
		"JUNIT_XML":                       6,
		"OPENVEX":                         7,
		"HELM_CHART":                      10,
		"SARIF":                           9,
		"EVIDENCE":                        11,
		"ATTESTATION":                     12,
		"CSAF_VEX":                        8,
		"CSAF_INFORMATIONAL_ADVISORY":     13,
		"CSAF_SECURITY_ADVISORY":          14,
		"CSAF_SECURITY_INCIDENT_RESPONSE": 15,
	}
```

internal/attestation/crafter/crafter.go
```diff
@@ -43,6 +43,28 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// attestationCraftingMaterial is a map of material types to their corresponding schemaapi.CraftingSchema_Material_MaterialType
+var attestationCraftingMaterial = func() map[string]schemaapi.CraftingSchema_Material_MaterialType {
```
migmartri, line 47: why do you need to return `schemaapi.CraftingSchema_Material_MaterialType`? Is it using anywhere?

internal/attestation/crafter/crafter.go
```diff
@@ -43,6 +43,28 @@ import (
...
+	for k, v := range schemaapi.CraftingSchema_Material_MaterialType_value {
+		if k != "MATERIAL_TYPE_UNSPECIFIED" {
+			res[strings.Replace(k, "MATERIAL_TYPE_", "", 1)] = schemaapi.CraftingSchema_Material_MaterialType(v)
+		}
+	}
+	return res
+}
+
+// ListAvailableMaterialKind returns a list of available material kinds
+func ListAvailableMaterialKind() []string {
```
migmartri, line 58: btw, this method could be added to the crafting schema package. 

So you could do `schemaapi.ListAvailableMaterialKinds()`

migmartri (commented):
Awesome, added some minor comments.

migmartri (approved):
Awesome!

app/cli/internal/action/attestation_status.go
javirln: Sure, task created and field changed.

app/cli/internal/action/attestation_status.go
javirln: Yep my bad, removed.

app/cli/internal/action/attestation_status.go
javirln: Sure thing, I'll add to the bottom to avoid nil pointers.

internal/attestation/crafter/crafter.go
javirln: Moved to `schemaapi` package.

internal/attestation/crafter/crafter.go
javirln: Moved to `schemaapi` package.

app/cli/cmd/attestation_add.go
javirln: There is an `Example` fields where this is loaded. I'll remove the first option all together since you're right it can be confusing.

PR #820, opened 2024-05-22
feat(attestation): Allow the auto discovery of material's kind

javirln:
This patch ease the addition of new materials to an attestation by enabling by default the auto discovery of materials' kind. If the material is not on the contract nor a specific kind is flagged, the CLI will try to match the material with any of the registered kinds.

**Please note this is a best effort.**

Example of usage. Given the following contract:
```bash
$ chainloop --insecure attestation init --replace --workflow-name wf-test
WRN API contacted in insecure mode
INF Attestation initialized! now you can check its status or add materials to it
┌───────────────────┬──────────────────────────────────────┐
│ Initialized At    │ 22 May 24 13:38 UTC                  │
├───────────────────┼──────────────────────────────────────┤
│ Attestation ID    │ 583553ef-d051-4c41-aec4-a4cdd725bf89 │
│ Name              │ wf-test                              │
│ Team              │ founding                             │
│ Project           │ core                                 │
│ Contract Revision │ 3                                    │
└───────────────────┴──────────────────────────────────────┘
┌────────────────────────┐
│ Materials              │
├───────────┬────────────┤
│ Name      │ one-file   │
│ Type      │ ARTIFACT   │
│ Set       │ No         │
│ Required  │ Yes        │
│ Is output │ Yes        │
├───────────┼────────────┤
│ Name      │ other-file │
│ Type      │ EVIDENCE   │
│ Set       │ No         │
│ Required  │ Yes        │
│ Is output │ Yes        │
└───────────┴────────────┘
```
Let's add the compulsory materials:
```bash
$ chainloop --insecure attestation add --value go.mod --name one-file
WRN API contacted in insecure mode
INF material added to attestation

$ chainloop --insecure attestation add --value LICENSE.md --name other-file
WRN API contacted in insecure mode
INF material added to attestation
```
And finally let's try to discover one material without specifying its type:
```bash
$ chainloop --insecure attestation add --value controlplane.cyclonedx.json
WRN API contacted in insecure mode
INF material kind detected kind=SBOM_CYCLONEDX_JSON
INF material added to attestation
```
As a result we can see how it's added to the result:
```bash
$ chainloop --insecure attestation push --key cosign.key
WRN API contacted in insecure mode
Enter password for private key:
INF push completed
┌───────────────────┬──────────────────────────────────────┐
│ Initialized At    │ 22 May 24 13:38 UTC                  │
├───────────────────┼──────────────────────────────────────┤
[...]

PR #822, opened 2024-05-22
feat(action): Create new action to consume GitHub releases

javirln:
This patch uses changes introduced in https://github.com/chainloop-dev/chainloop/pull/820 to auto discover materials based on the GitHub releases' assets.

Changes:
- Uses the workflow chainloop-vault-release
- Rename the file from `github_release` to `release`
- Attests everything included on the GitHub Release

Refs #785

### DIFF
--- a/.github/workflows/github_release.yaml
+++ b/.github/workflows/github_release.yaml
@@ -54,15 +54,6 @@ jobs:
           gh release download $tag -A tar.gz -D /tmp
           chainloop attestation add --value "/tmp/chainloop-$version.tar.gz"
 
-          # Include control-plane image
-          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/control-plane:$tag"
-
-          # Include cas image
-          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/artifact-cas:$tag"
-
-          # Include cli image
-          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/cli:$tag"
-
       - name: Finish and Record Attestation
         if: ${{ success() }}
         run: |
--- a/.github/workflows/github_release.yaml
+++ b/.github/workflows/github_release.yaml
@@ -1,0 +1,83 @@
+name: GitHub Release Attestation
+
+on:
+  release:
+
+permissions:
+  contents: read
+
+jobs:
+  # This reusable workflow inspects if the given workflow_name exists on Chainloop. If the Workflow does not exist
+  # it will create one with an empty contract ready for operators to be filled. Otherwise, if found, it will just
+  # be ignored and the process will continue. For this to work it's using a pre-created API Token
+  onboard_workflow:
+    name: Onboard Chainloop Workflow
+    uses: chainloop-dev/labs/.github/workflows/chainloop_onboard.yml@4173e015dbd5dc2a8802555c268da63d57bbe576
+    with:
+      project: "chainloop"
+      workflow_name: "chainloop-github-release"
+    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/github_release.yaml
```diff
@@ -0,0 +1,80 @@
...
+          chainloop attestation add --value "/tmp/chainloop-$version.tar.gz"
+
+          # Include control-plane image
+          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/control-plane:$tag"
+
+          # Include cas image
+          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/artifact-cas:$tag"
+
+          # Include cli image
+          chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/cli:$tag"
```
javirln, line 61: Not sure if we can get the url from somewhere else. It shouldn't change but the tag.

.github/workflows/github_release.yaml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+  # This reusable workflow inspects if the given workflow_name exists on Chainloop. If the Workflow does not exist
+  # it will create one with an empty contract ready for operators to be filled. Otherwise, if found, it will just
+  # be ignored and the process will continue. For this to work it's using a pre-created API Token
+  onboard_workflow:
+    name: Onboard Chainloop Workflow
+    uses: chainloop-dev/labs/.github/workflows/chainloop_onboard.yml@4173e015dbd5dc2a8802555c268da63d57bbe576
+    with:
+      project: "chainloop"
+      workflow_name: "chainloop-github-release"
```
migmartri, line 18: what would happen if you remove the name?

.github/workflows/github_release.yaml
```diff
@@ -0,0 +1,71 @@
...
+      api_token: ${{ secrets.CHAINLOOP_API_TOKEN }}
+
+  release:
+    name: Record release from GitHub
+    runs-on: ubuntu-latest
+    needs: onboard_workflow
+    permissions:
+      packages: write
+    env:
+      CHAINLOOP_VERSION: 0.88.2
```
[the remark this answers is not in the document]
migmartri: you can remove it

migmartri (approved):

.github/workflows/github_release.yaml
javirln: It will be created automagically.

migmartri:
We need to decide how this is going to co-exist/replace with our current release job/attestation. Do you have any thoughts on that transition?

javirln:
> We need to decide how this is going to co-exist/replace with our current release job/attestation. Do you have any thoughts on that transition?

I had a look at it the other day and we are indeed adding some values twice. We could leave the old `release` to the following:
- Create all the binaries with `goreleaser`
- Attest plugins since they are not part of the GitHub releases
- Generate all SBOMS from source code
- Create PR for Chart bump

`goreleaser` will take care of the majority of the heavy lifting by creating all binaries and pushing images. From the GitHub workflow we could:
- Attest all binaries except plugins
- Attest all SBOMS generated on the previous step part for the GitHub release
- Attest all container images

javirln:
Other option on the table:
- Remove duplicates on `release` workflow
- Use this reusable workflow as parte of the release one
- Add in different attestation the missing pieces such as the plugins binaries and container images (they would be removed from the GitHub release workflow)

That would also require changes on the `release` contract as well.

javirln:
I would remove the container images from this one, open a new PR to include it on the `Release`. Please keep in mind that this will create two different attestations that might not be related more from the commit.

migmartri (approved):

PR #827, opened 2024-05-24
feat(cli): support kms signing with cosign providers

jiparis:
This PR loads some cosign providers to support KMS signing in the CLI out of the box. It enables AWS, GCP, Azure and Vault signing.
For example, this is how AWS KMS signing would look like during `attestation push` command (assuming client has setup AWS credentials properly):

```
chainloop --token $TOKEN att push --key awskms:///arn:aws:kms:us-east-1:ACCOUNT:key/KEY_ID
```


migmartri (approved):
Awesome!

Issue #840, opened 2024-05-29
Current org is not being set with API tokens


jiparis:
While playing with keyless signing, I was getting consistent "missing org" errors in the different API calls. 

After some debugging, I noticed that `requireCurrentOrg` method is not finding it, although the claim is present in the context.

It looks like the context value is not being set in the robot account / api token middlewares (I confirmed that `WithCurrentAPITokenAndOrgMiddleware` is only called in the user token middleware)

jiparis:
Some findings:
* `WithCurrentUserAndOrgMiddleware` is in charge of setting the organization from the membership and store it in the `currentOrgCtxKey{}` key in the context. This key is used in the `requireCurrentOrg` method. This middleware is only run in user token contexts
* In the other side, `WithAttestatiWithAttestationContextFromRobotAccount` does a similar thing for robot accounts and API tokens, but instead, takes the OrgID from a claim and stores it in a `RobotAccount` object in the `currentRobotAccountCtxKey{}` context key. 
Only non-attestation APIs can use `requireCurrentOrg`. Attestation APIs should use `CurrentRobotAccount` method instead. 
The issue comes from the fact that I was wronly using `requireCurrentOrg` in a new attestation API. So we can close this task, but we should be aware of this when we refactor API permissions.

PR #841, opened 2024-05-29
chore(chainloop): embed error in when org ID cannot be retrieved

jiparis:
I was silently getting a "context deadline exceeded" error but it was being reported as a "error extracting organization from APIToken". This will improve it for easier debugging:
```
✗ go run app/cli/main.go --insecure --token $TOKEN att init --replace --workflow-name mywf
ERR rpc error: code = DeadlineExceeded desc = error extracting organization from APIToken: error retrieving the API token: finding token: getting APIToken: context deadline exceeded
exit status 1
```
Refs #840


migmartri (approved):
I am curious why we are getting context deadline though.

PR #842, opened 2024-05-29
fix(release-workflow): Add back container images for attestation

javirln:
This patch adds back the the attestation for all three container images.

It additionally sets a condition to not add the Chainloop Attestation URL if the release notes has it already.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,6 +53,15 @@ jobs:
           version=$(echo -n $tag | sed 's/v//g')
           gh release download $tag -A tar.gz -D /tmp
           chainloop attestation add --value "/tmp/chainloop-$version.tar.gz"
+          
+           # Include control-plane image
+           chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/control-plane:$tag"
+          
+           # Include cas image
+           chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/artifact-cas:$tag"
+
+           # Include cli image
+           chainloop attestation add --value "ghcr.io/chainloop-dev/chainloop/cli:$tag"
 
       - name: Finish and Record Attestation
         id: attestation-push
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -53,6 +53,15 @@ jobs:
           version=$(echo -n $tag | sed 's/v//g')
```
migmartri, line 53: out of curiosity, why do we need to do a sed here?

migmartri (approved):

.github/workflows/release.yaml
javirln: Because the downloaded source code file does not include the tag `v0.89.0` but the version `0.89.0`:
```
$ gh release download v0.89.0 -A tar.gz /tmp
$ ls -la /tmp/
total 38288
drwxrwxrwt  8 root     wheel       256 May 29 14:22 .
drwxr-xr-x  6 root     wheel       192 May 22 12:57 ..
-rw-r--r--@ 1 javirln  wheel  18898136 May 29 14:22 chainloop-0.89.0.tar.gz
```

.github/workflows/release.yaml
```diff
@@ -81,6 +90,11 @@ jobs:
           chainloop_release_url="## Chainloop Attestation"$'\n'"[View the attestation of this release](https://app.chainloop.dev/attestation/${{ steps.attestation-push.outputs.attestation_sha }})"
           current_notes=$(gh release view ${{ github.ref }} --json body -q '.body')
           
-          modified_notes="$chainloop_release_url"$'\n\n'"$current_notes"
+          # Only add the attestation link if it doesn't already exist
+          if ! echo "$current_notes" | grep -q "## Chainloop Attestation"; then
+            modified_notes="$chainloop_release_url"$'\n\n'"$current_notes"
+          else
+            modified_notes="$current_notes"
```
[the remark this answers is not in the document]
javirln: Ok, it makes sense

.github/workflows/release.yaml
migmartri: So, you can't use $tag for the container image tag do you?

javirln: The container image tag does contain the tag: https://github.com/chainloop-dev/chainloop/pkgs/container/chainloop%2Fcontrol-plane/220543906?tag=v0.89.0

migmartri: My bad Javi, I was looking at the other `sed`. Sorry about the noise and thanks for the clarification.

Issue #829, opened 2024-05-26
Document Ory setup


migmartri:
We could take a look at https://github.com/ory and document its usage

PR #843, opened 2024-05-29
feat(docs): Document Ory Hydra as OAuth2 provider

javirln:
This patch adds documentation regarding the configuration needed to use Ory Hydra as OAuth2 provider.

Closes https://github.com/chainloop-dev/chainloop/issues/829

### DIFF
--- a/docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
+++ b/docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
@@ -1,0 +1,62 @@
+---
+title: Use Ory Hydra as OAuth2 provider
+---
+
+A requirement to run your own Chainloop instance, is to have a way of authenticate users. One of the most common ways to do this is by using OAuth2.
+Ory Hydra is a popular OAuth2 provider that can be used to authenticate users in your Chainloop instance once connected to an external OIDC provider.
+
+## Pre-requisites
+
+To configure your Chainloop instance with Ory Hydra you'll need the following information:
+
+- Ory Hydra instance running
+- Access to your OICD provider configuration
+
+## Configure Ory Hydra
+
+First, you'll need to have an Ory Hydra instance running. You can follow the [official documentation](https://www.ory.sh/hydra/docs/) to set up your own instance.
+Then simply create a new OAuth2 client in your Ory Hydra instance. You can do this by running the following command:
+
+```bash
+$  hydra create oauth2-client --name "ACME Solutions"  --grant-type authorization_code,refresh_token --response-type code  --scope openid,email,profile  --redirect-uri http://CHAINLOOP_INSTANCE_URL/auth/callback --endpoint https://ORY_HYDRA_URL
+```
+
+Chainloop client will only request `openid`, `email` and `profile` scopes.
+
+Relevant information that can be noted from the command signature is:
+- `name`: The name of the OAuth2 client
+- `grant-type`: The grant type of the client it needs to be set `authorization_code` and `refresh_token`
+- `response-type`: The response type of the client: `code`
+- `scope`: The scopes that the client will request: `openid`, `email`, `profile`
+- `redirect-uri`: The redirect URI of the client: Whenever is the Chainloop instance URL plus `/auth/callback`
+- `endpoint`: The endpoint of the Ory Hydra instance
+
+Once the command is run, it will give back something similar to the following:
+
+```bash
+CLIENT ID       b028840e-8c54-4d01-91b9-eb2c4aa6fc0e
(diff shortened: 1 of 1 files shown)


docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Use Ory Hydra as OAuth2 provider
```
migmartri, line 2: Should we say OIDC (OpenID Connect) instead of Oauth2?

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
```diff
@@ -0,0 +1,62 @@
...
+---
+
+A requirement to run your own Chainloop instance, is to have a way of authenticate users. One of the most common ways to do this is by using OAuth2.
```
migmartri, line 5: I'd say that's the only way we support.

I mean rephrasing it to smth like

```
A requirement to run your own Chainloop instance is to configure an OIDC provider to authenticate users who interact with the control plane.
```

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
```diff
@@ -0,0 +1,62 @@
...
+Ory Hydra is a popular OAuth2 provider that can be used to authenticate users in your Chainloop instance once connected to an external OIDC provider.
```
migmartri, line 6: > once connected to an external OIDC provider.

not sure I understand this part. Does this mean that hydra needs to be connected to an external OIDC? Or that Hydra is an OIDC IDp?

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+```bash
+$  hydra create oauth2-client --name "ACME Solutions"  --grant-type authorization_code,refresh_token --response-type code  --scope openid,email,profile  --redirect-uri http://CHAINLOOP_INSTANCE_URL/auth/callback --endpoint https://ORY_HYDRA_URL
+```
+
+Chainloop client will only request `openid`, `email` and `profile` scopes.
+
+Relevant information that can be noted from the command signature is:
+- `name`: The name of the OAuth2 client
+- `grant-type`: The grant type of the client it needs to be set `authorization_code` and `refresh_token`
```
migmartri, line 28: do we need refresh token? I do not think so.

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
```diff
@@ -0,0 +1,62 @@
...
+- `response-type`: The response type of the client: `code`
+- `scope`: The scopes that the client will request: `openid`, `email`, `profile`
+- `redirect-uri`: The redirect URI of the client: Whenever is the Chainloop instance URL plus `/auth/callback`
+- `endpoint`: The endpoint of the Ory Hydra instance
+
+Once the command is run, it will give back something similar to the following:
+
+```bash
+CLIENT ID       b028840e-8c54-4d01-91b9-eb2c4aa6fc0e
+CLIENT SECRET   V.Erxy5_f5zYvBHXi4fZWrB6tw
```
migmartri, line 38: I'd redact that to not to trigger secret scanners

migmartri (approved):

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
javirln: I've removed the sentence but, Hydra needs to be connected to an external IDP

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
javirln: What do we have on dex?

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
javirln: Sure

docs/docs/guides/deployment/hydra-oauth2/hydra-oauth2.mdx
migmartri: Not sure about dex but Auth0, which we use in prod seems to by default indeed add `refresh_token`. I guess it's fine although I do not think we keep any kind of refresh token in the client, we could potentially disable that grant and should be fine. 

For now, until we test that, I'd leave it as is

Thanks

PR #856, opened 2024-05-30
fix(workflow): Add personal access token

javirln:
This patch passes the personal access token as `GITHUB_TOKEN` env variable.


javirln:
Closing the issue since we are not going to handling this for the time being.

PR #858, opened 2024-05-31
chore(auth): set organization in auth token flow

jiparis:
In #840, we noticed that orgID wasn't being set in the API TOKEN authn flow. This PR adds it to the context, in the same way it's done for the user authn flow.

Refs #840


app/controlplane/internal/server/grpc.go
```diff
@@ -50,15 +50,15 @@ import (
...
 type Opts struct {
 	// UseCases
-	UserUseCase          *biz.UserUseCase
-	RobotAccountUseCase  *biz.RobotAccountUseCase
-	CASBackendUseCase    *biz.CASBackendUseCase
-	CASClientUseCase     *biz.CASClientUseCase
-	IntegrationUseCase   *biz.IntegrationUseCase
-	ReferrerUseCase      *biz.ReferrerUseCase
-	APITokenUseCase      *biz.APITokenUseCase
-	OrganizationUserCase *biz.OrganizationUseCase
```
migmartri, line 60: good catch, thanks

app/controlplane/internal/usercontext/robotaccount_middleware.go
```diff
@@ -104,6 +104,13 @@ func WithAttestationContextFromRobotAccount(robotAccountUseCase *biz.RobotAccoun
 				return nil, errors.New("error retrieving the organization from the auth token")
 			}
 
+			org, err := orgUseCase.FindByID(ctx, orgID)
+			if err != nil {
+				return nil, err
```
migmartri, line 109: for the error, I´d either wrap it, or hide it altogether. About hiding it, I've been doing it in these kind authentication-related code, but I am not sure it's needed here.

In any case, pick one, and replace the naked error :)

migmartri (approved):
My only comment is about replacing the naked error

PR #859, opened 2024-05-31
feat(keyless): signing use case

jiparis:
This is the first of a set of PRs to implement keyless signing in chainloop vault and CLI.

This PR focuses on the implementation of the signing use case. The use case receives a certificate request and signs it with a configured certificate authority. It uses some of the Sigstore logic for this purpose.

### DIFF
--- a/app/controlplane/internal/biz/signing.go
+++ b/app/controlplane/internal/biz/signing.go
@@ -42,7 +42,7 @@ func NewChainloopSigningUseCase(ca ca.CertificateAuthority) *SigningUseCase {
 	return &SigningUseCase{CA: ca}
 }
 
-func (s *SigningUseCase) CreateSigningCert(ctx context.Context, orgId string, csrRaw []byte) ([]string, error) {
+func (s *SigningUseCase) CreateSigningCert(ctx context.Context, orgID string, csrRaw []byte) ([]string, error) {
 	var publicKey crypto.PublicKey
 
 	if len(csrRaw) == 0 {
@@ -67,7 +67,7 @@ func (s *SigningUseCase) CreateSigningCert(ctx context.Context, orgId string, cs
 	}
 
 	// Create certificate from CA provider (no Signed Certificate Timestamps for now)
-	csc, err := s.CA.CreateCertificate(ctx, NewChainloopPrincipal(orgId), publicKey)
+	csc, err := s.CA.CreateCertificate(ctx, NewChainloopPrincipal(orgID), publicKey)
 	if err != nil {
 		return nil, err
 	}
@@ -88,24 +88,24 @@ func (s *SigningUseCase) CreateSigningCert(ctx context.Context, orgId string, cs
 }
 
 type ChainloopPrincipal struct {
-	orgId string
+	orgID string
 }
 
 var _ identity.Principal = (*ChainloopPrincipal)(nil)
 
-func NewChainloopPrincipal(orgId string) *ChainloopPrincipal {
-	return &ChainloopPrincipal{orgId: orgId}
+func NewChainloopPrincipal(orgID string) *ChainloopPrincipal {
+	return &ChainloopPrincipal{orgID: orgID}
 }
 
 func (p *ChainloopPrincipal) Name(_ context.Context) string {
-	return p.orgId
+	return p.orgID
 }
 
 func (p *ChainloopPrincipal) Embed(_ context.Context, cert *x509.Certificate) error {
 	// no op.
 	// TODO: Chainloop might have their own private enterprise number with the Internet Assigned Numbers Authority
 	// 		 to embed its own identity information in the resulting certificate
-	cert.Subject = pkix.Name{Organization: []string{p.orgId}}
+	cert.Subject = pkix.Name{Organization: []string{p.orgID}}
 
 	return nil
 }
--- a/app/controlplane/internal/biz/signing.go
+++ b/app/controlplane/internal/biz/signing.go
@@ -1,0 +1,111 @@
+//
+// Copyright 2024 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


app/controlplane/internal/biz/signing.go
```diff
@@ -0,0 +1,111 @@
...
+	CA ca.CertificateAuthority
+}
+
+var _ SigningCertCreator = (*SigningUseCase)(nil)
+
+func NewChainloopSigningUseCase(ca ca.CertificateAuthority) *SigningUseCase {
+	return &SigningUseCase{CA: ca}
+}
+
+func (s *SigningUseCase) CreateSigningCert(ctx context.Context, orgID string, csrRaw []byte) ([]string, error) {
```
migmartri, line 45: It would be nice to add a little bit of documentation in this method. Mostly what it does and returns.

app/controlplane/internal/biz/signing.go
```diff
@@ -0,0 +1,111 @@
...
+	"errors"
+
+	"github.com/sigstore/fulcio/pkg/ca"
+	"github.com/sigstore/fulcio/pkg/identity"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+type SigningCertCreator interface {
+	// CreateSigningCert creates a signing certificate from and returns the certificate chain
+	CreateSigningCert(context.Context, string, []byte) ([]string, error)
```
migmartri, line 32: what's this interface for? Note that in general we are not creating interfaces in the use-cases. The interfaces that we have there are on the `repo` layer.

migmartri (commented):
Awesome!

app/controlplane/internal/biz/signing.go
jiparis: I was assuming that we wanted to have interfaces at this level as well (to create test mocks, etc.). But maybe it's something we can do for the whole layer in a future change. I'll keep it simple for now.

app/controlplane/internal/biz/signing.go
```diff
@@ -0,0 +1,111 @@
...
+	var publicKey crypto.PublicKey
+
+	if len(csrRaw) == 0 {
+		return nil, errors.New("csr cannot be empty")
+	}
+
+	// Parse CSR
+	csr, err := cryptoutils.ParseCSR(csrRaw)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
jiparis: Sure

app/controlplane/internal/biz/signing.go
```diff
@@ -0,0 +1,111 @@
...
+	return append([]string{finalPEM}, finalChainPEM...), nil
+}
+
+type ChainloopPrincipal struct {
+	orgID string
+}
+
+var _ identity.Principal = (*ChainloopPrincipal)(nil)
+
+func NewChainloopPrincipal(orgID string) *ChainloopPrincipal {
```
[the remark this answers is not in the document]
jiparis: No, and the struct can be private as well. I'll fix it.

app/controlplane/internal/biz/signing_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type signingUseCaseTestSuite struct {
+	suite.Suite
+	uc  *biz.SigningUseCase
+	csr []byte
+}
+
+func (s *signingUseCaseTestSuite) TestSigningUseCase_CreateSigningCert() {
```
migmartri, line 41: nitpick: I think this is smth I asked @javirln once, but is using `_` i.e `TestSigningUseCase_CreateSigningCert` go idiomatic?

app/controlplane/internal/biz/signing_test.go
```diff
@@ -0,0 +1,98 @@
...
+		certChain, err := s.uc.CreateSigningCert(context.TODO(), "myorgid", s.csr)
+		s.NoError(err)
+
+		// assert 2 certificates: signing certificate + chain (only one)
+		s.Len(certChain, 2)
+
+		// check cert contents
+		cert, err := cryptoutils.UnmarshalCertificatesFromPEM([]byte(certChain[0]))
+		s.NoError(err)
+		s.Assert().Len(cert, 1)
```
migmartri, line 57: s.Len

app/controlplane/internal/biz/signing_test.go
```diff
@@ -0,0 +1,98 @@
...
+		s.Assert().Equal("myorgid", cert[0].Subject.Organization[0])
```
migmartri, line 58: s.Equal

app/controlplane/internal/biz/signing_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func TestSuite(t *testing.T) {
+	suite.Run(t, new(signingUseCaseTestSuite))
+}
+
+func (s *signingUseCaseTestSuite) SetupTest() {
+	csr, err := createCSR()
+	if err != nil {
+		s.Require().NoError(err)
```
migmartri, line 69: I do not think you need to put this inside the if err!=nil

app/controlplane/internal/biz/signing_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	s.csr = csr
+
+	ca, err := ephemeralca.NewEphemeralCA()
+	if err != nil {
+		s.Require().NoError(err)
```
migmartri, line 75: same here

migmartri (approved):

app/controlplane/internal/biz/signing_test.go
jiparis: It's the suggestion from the IDE, I believe it's some "idiomatic", but I don't think there is a written rule.

app/controlplane/internal/biz/signing_test.go
jiparis: yes, it was a leftover. good catch

PR #860, opened 2024-05-31
feat(keyless): signing service

jiparis:
This is the second PR for keyless signing. In this case, it's the service that exposes the feature implemented in the use case from https://github.com/chainloop-dev/chainloop/pull/859 (ignore changes from the use case)

It takes a CSR from the CLI and returns a full certificate chain, signed by the signing authority.

Note that the service is not yet configured nor exposed. It will be done in the next PR.


app/controlplane/api/controlplane/v1/signing.proto
```diff
@@ -0,0 +1,40 @@
...
+
+
+syntax = "proto3";
+
+package controlplane.v1;
+
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1;v1";
+
+service SigningService {
+  rpc SigningCert (SigningCertRequest) returns (SigningCertResponse);
```
migmartri, line 24: nitpick.

- Document what this method does above this line. 
- Add rename it to something more like `GenerateSigningCert` to indicate that that a new cert will be done.

app/controlplane/api/controlplane/v1/signing.proto
```diff
@@ -0,0 +1,40 @@
...
+}
+
+message SigningCertRequest {
+  bytes certificate_signing_request = 1;
```
migmartri, line 28: btw, you can add here validations too, 

for example

```
  bytes certificate_signing_request = 1 [(buf.validate.field).bytes.min_len = 1];
```

https://buf.build/bufbuild/protovalidate/docs/main:buf.validate#buf.validate.BytesRules

app/controlplane/internal/biz/signing.go
```diff
@@ -0,0 +1,111 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"errors"
+
+	"github.com/sigstore/fulcio/pkg/ca"
+	"github.com/sigstore/fulcio/pkg/identity"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+type SigningCertCreator interface {
```
migmartri, line 30: I understand that this code can be ignored

app/controlplane/internal/service/signing.go
```diff
@@ -0,0 +1,58 @@
...
+var _ v1.SigningServiceServer = (*SigningService)(nil)
+
+func NewSigningService(signing biz.SigningCertCreator) *SigningService {
+	return &SigningService{
+		signing: signing,
+	}
+}
+
+func (s *SigningService) SigningCert(ctx context.Context, req *v1.SigningCertRequest) (*v1.SigningCertResponse, error) {
+	if len(req.GetCertificateSigningRequest()) == 0 {
```
migmartri, line 43: see my comment about validating at the API layer.

migmartri (approved):

PR #861, opened 2024-05-31
feat(keyless): expose signing service though GRPC

jiparis:
This is the continuation of https://github.com/chainloop-dev/chainloop/pull/860. In this PR all pieces are glued and configured through the command line.

Only File CA signing backend is supported currently. A development CA key and cert are provided for testing purposes.

refs #865


app/controlplane/cmd/main.go
```diff
@@ -218,3 +224,16 @@ func initSentry(c *conf.Bootstrap, logger log.Logger) (cleanupFunc func(), err e
 func newProtoValidator() (*protovalidate.Validator, error) {
 	return protovalidate.New()
 }
+
+func newSigningCA(_ context.Context, c *conf.Bootstrap, logger log.Logger) (ca.CertificateAuthority, error) {
```
migmartri, line 228: instead of conf.Bootstrap it might make sense to have `CertificateAuthority` instead, that way you simplify potential testing.

app/controlplane/cmd/main.go
```diff
@@ -218,3 +224,16 @@ func initSentry(c *conf.Bootstrap, logger log.Logger) (cleanupFunc func(), err e
...
+	// File
+	if c.GetCertificateAuthority().GetFileCa() != nil {
+		fileCa := c.GetCertificateAuthority().GetFileCa()
+		_ = logger.Log(log.LevelInfo, "msg", "Keyless: File CA configured")
+		return fileca.NewFileCA(fileCa.GetCertPath(), fileCa.GetKeyPath(), fileCa.GetKeyPass(), false)
+	}
+
+	// No CA configured, keyless will be deactivated.
+	_ = logger.Log(log.LevelInfo, "msg", "Keyless NOT configured")
```
migmartri, line 237: `Keyless Signing` (add signing)

app/controlplane/internal/conf/conf.proto
```diff
@@ -118,3 +121,21 @@ message Auth {
     string redirect_url_scheme = 4;
   }
 }
+
+message CA {
+  oneof ca {
+    FileCA file_ca = 1;
+    KmsCA kms_ca = 2;
```
migmartri, line 128: kms is not supported yet right? Should we remove it?

migmartri (approved):

jiparis:
Added a guard to prevent using the API when CA not configured

PR #863, opened 2024-05-31
feat(keyless): document keyless deployment

jiparis:
Added some instructions on how to deploy the controlplane in keyless mode.

### DIFF
--- a/docs/docs/guides/deployment/keyless/keyless.mdx
+++ b/docs/docs/guides/deployment/keyless/keyless.mdx
@@ -1,0 +1,49 @@
+---
+title: Configure your deployment for keyless signing
+---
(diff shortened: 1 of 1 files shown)


jiparis:
Tested locally: no formatting issues.

migmartri (commented):
I am not sure we 

1 - need to document this, since in development keyless works out of the box and we might be able to do the same in the helm chart
2 - if we document it, we probably should do it in the helm chart readme instead.

Let's revisit this patch once we know how we are going to deploy this.

jiparis:
Closing, as we documented the chart instead. https://github.com/chainloop-dev/chainloop/pull/878

PR #879, opened 2024-06-04
feat(docs): add cookie consent script

tknura:
This patch adds the cookie consent script which should display the cookie consent banner at the bottom of the page. `autoBlock=on` should prevent other scripts to load cookies until being accepted
 
Zrzut ekranu 2024-06-4 o 17 56 33


migmartri (approved):
thanks

Issue #891, opened 2024-06-05
phase out robot accounts


migmartri:
[robot-accounts](https://docs.chainloop.dev/getting-started/workflow-definition#robot-account-creation) are being replaced by API tokens 

This means changes in the CLI workflow especially around attestation.

```[tasklist]
### Tasks
- [ ] https://github.com/chainloop-dev/chainloop/issues/889
- [x] Deprecate ra sub-command
- [ ] https://github.com/chainloop-dev/chainloop/issues/904
- [x] Add warning on authentication collision
- [ ] https://github.com/chainloop-dev/chainloop/issues/905
- [ ] make sure that lab attestation uses this key and the report works (digest link)
- [x] Remove robot-account information from doc overviews
- [ ] https://github.com/chainloop-dev/chainloop/issues/892
- [ ] https://github.com/chainloop-dev/chainloop/issues/893
```

PR #897, opened 2024-06-05
fix: load chainloop_token gracefully if auth login has happened before

migmartri:
To prevent unexpected credentials overrides in the case the user has previously logged in explicitly but has an env variable, the system will not give priority to the user authentication instead and will show a warning message.

In addition to that, the credentials load logic has been improved to being able to detect when credentials are explicitly set vs loaded from env variables, and notify the user accordingly. 

Token provided

```
chainloop wf ls --token $T
INF API token provided to the command line
...
```

with env var instead CHAINLOOP_TOKEN=foobar

```
chainloop wf ls           
...
```

refs #891

### DIFF
--- a/app/cli/cmd/root.go
+++ b/app/cli/cmd/root.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023 The Chainloop Authors.
+// Copyright 2024 The Chainloop Authors.
 //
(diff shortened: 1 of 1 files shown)


javirln (approved):
Awesome!

PR #906, opened 2024-06-05
fix(docs): fix code-blocks, theme and admonitions

### LINKED ISSUE #904: docs: fix code-block and admonitions
They seem to have stopped working after the upgrade to 3.x

migmartri:
When I did the upgrade to 3.x I overlooked a couple of requirements from their [migration guide](https://docusaurus.io/docs/migration/v3)

- I missed upgrading all the dependencies `@mdx-js/react` and `prism-react-renderer`.
- I missed that the theme now loads via esmodules  

After the fix, we have now properly configured code blocks (with copy button cc/ @jiparis) and attention banners.

image

> NOTE: the rest of the updates in the dependencies is just because I upgraded to the latest version of docusaurus 

closes #904

### DIFF
--- a/docs/docs/intro/intro.mdx
+++ b/docs/docs/intro/intro.mdx
@@ -8,12 +8,6 @@ title: Introduction
 import MailingListForm from "../partials/_mailing-list-form.mdx";
 import Image from "@theme/IdealImage";
 
-:::info UPDATE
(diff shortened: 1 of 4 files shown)


danlishka (approved):
LGTM

PR #907, opened 2024-06-06
feat: raw token output in CLI

migmartri:
This patch adds an additional output flag value `token` to the `api-token create` command that outputs just the raw API token.

This is useful for automation and one-liner quickstart

```
export CHAINLOOP_TOKEN=$(chainlooop org api-token create --name quickstart -o token)
```

refs #891

### DIFF
--- a/app/cli/cmd/organization_apitoken_create.go
+++ b/app/cli/cmd/organization_apitoken_create.go
@@ -45,6 +45,11 @@ func newAPITokenCreateCmd() *cobra.Command {
 				return fmt.Errorf("creating API token: %w", err)
 			}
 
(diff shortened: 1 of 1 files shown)


javirln (approved):
Approving it but one question. The fact that we are using `fmt.Print` is intentional to send the output to stdout right?

migmartri:
> Approving it but one question. The fact that we are using `fmt.Print` is intentional to send the output to stdout right?

That's correct, to be used as in the example

javirln:
Ok so for the future refactors, we need to think on leveraging the `cmd.Stdout` and `cmd.Stderr` in order to redirect the output to introduce tests.

PR #920, opened 2024-06-07
chore(grpc): Cleanup autogenerate protos

javirln:
This patch runs `make generate` on the backend and adds missing generate proto definitions to the source code.

### DIFF
--- a/app/controlplane/api/gen/frontend/google/protobuf/descriptor.ts
+++ b/app/controlplane/api/gen/frontend/google/protobuf/descriptor.ts
@@ -4,6 +4,98 @@ import _m0 from "protobufjs/minimal";
 
 export const protobufPackage = "google.protobuf";
 
+/** The full set of known editions. */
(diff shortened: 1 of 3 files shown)


migmartri (commented):
I am surprised about this change, do you know when they started appearing? What change caused this new files to be generated?

Also, I believe we have some tests to make sure all the files are generated otherwise fail, do you know if this can get caught?

Thanks!

javirln:
> I am surprised about this change, do you know when they started appearing? What change caused this new files to be generated?
> 
> Also, I believe we have some tests to make sure all the files are generated otherwise fail, do you know if this can get caught?
> 
> Thanks!

This is indeed caught by the tests and it seems on the CI that's failing right now. I'm not sure if it's a version problem between the CI and local since it seems the CI is not able to get the files properly.

javirln:
Closing the PR since it must be a local problem and not related to the repo itself.

Issue #683, opened 2024-04-22
chore(workflows): do not allow name update


migmartri:
Because of moving to a more declarative approach based on the name of the resources, we don't want to allow users to change the name of workflows.

The goal of the task are:
- Remove the option from the CLI to update the name by passing the flag `--name`
- Modify GRPC services to don't accept the `Name` as parameter

javirln:
I will do it in a breaking changing way, meaning not deprecation notice will be added to the CLI nor the field will be marked as deprecated on the service.

PR #921, opened 2024-06-07
feat(workflows): Remove the option of updating the workflow name

javirln:
This PR eliminates the ability for users to update the name of a workflow, as we are transitioning to a declarative approach where names are immutable.

Refs #683


javirln:
This PR contains unrelated changes that we'll be remove once this PR is merged: https://github.com/chainloop-dev/chainloop/pull/920

migmartri (commented):
Thanks Javi, you can go one step down by updating also the database schema and making `name` immutable.

https://github.com/chainloop-dev/chainloop/blob/570c4ef13afe8baa4596e8e2e722af84a795d72c/app/controlplane/internal/data/ent/schema/workflowcontract.go#L38

Thanks!

javirln:
I've mad the column immutable and did a migration to update possible `null` values + setting it required on the DB.

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
```diff
@@ -0,0 +1,6 @@
+-- Add default value to "name" column in "workflows" table if it is null
+UPDATE "workflows" SET "name" = CONCAT(split_part("id"::text, '-', 1)) WHERE "name" IS NULL;
```
javirln, line 2: Only setting the default vale as part of the ID. Originally I used a concatenation of project + id but the project might be null as well although neither on the CLI and UI is optional, I wanted to play safe here.

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
```diff
@@ -0,0 +1,6 @@
...
+-- Create index "workflows_name_organization_id" to table: "workflows"
+CREATE UNIQUE INDEX "workflows_name_organization_id" ON "workflows" ("name", "organization_id") WHERE (deleted_at IS NULL);
```
migmartri, line 4: I am confused. Hasn't this already been done?

From db schema.

```
"workflow_name_organization_id" UNIQUE, btree (name, organization_id) WHERE deleted_at IS NULL
```

migmartri (changes requested):

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
migmartri: How did you create the migration? In theory no migration should have been created, I believe the constraints are already in place. Or am I missing something?

also, did that migration work for you? What happened when you ran the controlplane locally? Did it work?

javirln: What I did was:
- Create changes on `app/controlplane/internal/data/ent/schema/workflow.go`
- Ran `make generate`
- Create an empty migration by `make migration_new`
- Wrote the changes you see on the file
- Run `make migration_sync`

I've ran the application and the migration worked without problem. But double checking the code it seems the index was already created but I couldn't see it on my schema, which is weird.

In any case, I'll remove those changes since they are not needed.

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
javirln: Yes, they were there but I don't know why my schema was not registering those changes and I couldn't see the index that's why I created.

Doing a hard reset fixed the issue. Sorry for the noise.

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
migmartri: thanks, let's figure out why your environment was different. 

Can you 

- go to `main`
- docker compose down (also removing the docker volumes)
- docker compose up.
- Load the psql console `docker compose exec postgresql psql -U postgres controlplane` 
- `d workflows`?

app/cli/internal/action/workflow_update.go
```diff
@@ -30,18 +30,19 @@ func NewWorkflowUpdate(cfg *ActionsOpts) *WorkflowUpdate {
 }
 
 type NewWorkflowUpdateOpts struct {
```
migmartri, line 32: this is not from your PR, but at first it seems that I had a brain fart when naming this struct probably because I copied it from the creation flow. `NewWorkflowUpdateOpts` should probably be called `WorkflowUpdateOpts`

migmartri (approved):
LGTM!

app/controlplane/internal/data/ent/migrate/migrations/20240610060712.sql
javirln: Yes, I can see the index now, it might have been the outdated volume on the DB. Thanks Migue!

app/cli/internal/action/workflow_update.go
javirln: Let me do the change!

PR #927, opened 2024-06-08
fix(attestation): do not load cas-backend on dry-mode

### LINKED ISSUE #926: dry-run attestation fails when adding materials
When using the attestation process in `dry-run` and using the new `API tokens` approach, the upload fails to find the default cas-backend.

With that setup, there is not `workflowRunID` and the attestation goes through this deprecated path, which in turns seems to fail to find the default cas-backend

https://github.com/chainloop-dev/chainloop/blob/588375f5f568a71501331fb7dea89fa37095dc55/app/controlplane/internal/service/attestation.go#L325-L332

migmartri:
When running an attestation in dry-mode, we should not upload files either and hence we do not need to ask for the cas-backend information to the server.

This PR makes sure that the controlplane `GetUploadCreds` is only called if regular attestation mode is engaged.

fixes #926

### DIFF
--- a/app/controlplane/internal/service/attestation.go
+++ b/app/controlplane/internal/service/attestation.go
@@ -318,35 +318,20 @@ func (s *AttestationService) GetUploadCreds(ctx context.Context, req *cpAPI.Atte
 
 	// Find the CAS backend associated with this workflowRun, that's the one that will be used to upload the materials
 	// NOTE: currently we only support one backend per workflowRun but this will change in the future
-
-	// DEPRECATED: if no workflow run is provided, we use the default repository
-	// Maintained for compatibility reasons with older versions of the CLI
-	var backend *biz.CASBackend
-	if req.WorkflowRunId == "" {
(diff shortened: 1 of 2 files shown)


app/controlplane/internal/service/attestation.go
```diff
@@ -318,36 +318,21 @@ func (s *AttestationService) GetUploadCreds(ctx context.Context, req *cpAPI.Atte
...
+		return nil, handleUseCaseErr(err, s.log)
+	} else if wRun == nil {
+		return nil, errors.NotFound("not found", "workflow run not found")
+	}
 
-	// DEPRECATED: if no workflow run is provided, we use the default repository
-	// Maintained for compatibility reasons with older versions of the CLI
-	var backend *biz.CASBackend
-	if req.WorkflowRunId == "" {
-		s.log.Warn("DEPRECATED: using main repository to get upload creds")
```
migmartri, line 326: I also removed this deprecated path since it's not needed.

Issue #796, opened 2024-05-17
minimum CLI Version check in the backend


migmartri:
The CLI and backend are rapidly evolving and currently there is no check to make sure CLI/Controlplane are compatible.

This is problematic since for example @jp-gouin ran into the following issue 

```
- attestation.workflow.organization: value length must be at least 1 characters [string.min_len]
```

which was related to using a version of the CLI (0.83.0) that was not sending the required organization information in the attestation, requirement that was added in  0.87.

jiparis:
Implementation:
* Add a `minumunVersion` to the `/infoz` enpoint response
* Block CLI command only when there is a requirement in the backend (the case described in this task)

Another useful feature would be to add the used CLI version and digest to the local crafting state, so that these issues can be easily debugged. We would use this metadata in the render phase when doing `att push`.

PR #932, opened 2024-06-10
fix(attestation): push on dry-run

migmartri:
There was a problem while trying to `att push` an attestation that was initialized in `dry-run` mode. The validation error was

```
ERR invalid crafting state: validation error:
 - attestation.workflow.organization: value length must be at least 1 character [string.min_len]
```

This patch skips that schema validation when running in `dry-run` mode since in this mode it's know to not to be fully complete.

refs #796

### DIFF
--- a/app/cli/cmd/attestation_init.go
+++ b/app/cli/cmd/attestation_init.go
@@ -81,6 +81,10 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
+			if res.DryRun {
+				logger.Info().Msg("The attestation is being crafted in dry-run mode. It will not get stored once rendered")
+			}
+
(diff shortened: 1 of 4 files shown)


app/cli/cmd/attestation_init.go
```diff
@@ -81,6 +81,10 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
+			if res.DryRun {
+				logger.Info().Msg("The attestation is being crafted in dry-run mode. It will not get stored once rendered")
```
migmartri, line 85: this is just removing the dry-run output from stdout and move it to stderr as we are doing lately in the rest of the code.

javirln (approved):
Thanks Migue!

PR #935, opened 2024-06-11
feat(chart): Replace Vault dependency to use Bitnami chart

javirln:
This patch replaces the Hashicorp Vault dependency with Bitnami's version of Vault to align with their [CONTRIBUTING guidelines](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

The primary difference between the old and new Chart is the support for development mode in Vault, particularly for local development instances. Bitnami's chart is sufficiently flexible, allowing us to enable development mode by adjusting a few options, such as the default `args` of the server and `extraEnvVars`.

Additionally, this patch includes the addition of license headers and updates the Chart.yaml version and executed Bitnami's [README generator tool](https://github.com/bitnami/readme-generator-for-helm).


jiparis (approved):

deployment/chainloop/values.yaml
```diff
@@ -809,12 +812,29 @@ postgresql:
...
+## @param vault.server.extraEnvVars[0].name Root token for the vault server
+## @param vault.server.extraEnvVars[0].value The value of the root token. Default: notasecret
+## @param vault.server.extraEnvVars[1].name Address to listen on development mode
+## @param vault.server.extraEnvVars[1].value The address to listen on. Default: [::]:8200
 vault:
   server:
-    dev:
-      enabled: true
-      devRootToken: "notapassword"
+    args: [
```
migmartri, line 825: Could you add a comment saying that this is only used when enabled `development` mode in the chart? 

Thanks!

migmartri (approved):
Awesome @javirln thanks!

deployment/chainloop/values.yaml
javirln: Sure, I'll add it to the top of the `vault` key!

Issue #936, opened 2024-06-11
specify  user in our docker images


migmartri:
https://docs.docker.com/scout/policy/#default-non-root-user

PR #937, opened 2024-06-11
build images with username

migmartri:
builds all our container images indicating a non-root user `1001`

Next, we can release these images and give them a try in our helm chart, which will probably require some changes to pick the right usercontext

refs #936


javirln (approved):
LGTM

Issue #914, opened 2024-06-06
Implement keyless verification


jiparis:
Currently, keyless signing is in production in experimental mode, as generated attestations are not yet verifiable (because generated certificate is not stored).

This task is for implementing the full verification scenario, following best practices.

jiparis:
We will adopt sigstore approach for storing and verifying local bundles.

jiparis:
See https://github.com/chainloop-dev/chainloop/issues/990 for storing the verification material.

PR #943, opened 2024-06-12
feat(sign): option to generate sigstore bundle after signing

jiparis:
This PR supports generating Sigstore bundles in `att push` command. The goal is to have all the verification material downloaded locally (except the root certificate) for later verification, mostly in key-less scenarios.

It would populate the verificationMaterial with either the ephemeral certificate (key-less) in DER format, or the public key in PEM format used for signing, so that they can be used right away.

Key-less example:
```
> go run app/cli/main.go --insecure --token $TOKEN att init --replace --name mywf
...
> go run app/cli/main.go --insecure --token $TOKEN att add --name foo --value evidence.txt
> go run app/cli/main.go --insecure --debug --token $TOKEN att push --bundle mybundle.json
...
DBG generating in-toto statement
DBG generating a keyless signer
DBG generating new certificate request
INF generating Sigstore bundle mybundle.json
...
```
```
> cat mybundle.json | jq -r .dsseEnvelope.payload | base64 --decode
{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"chainloop.workflow.mywf","digest":{"sha256":"5d6cfc4f3154275e04867bb2ba6991d6cc5cb0e26b1c67caff8a969400f557b2"}}
...
```
Inspecting the certificate
```
> cat mybundle.json | jq -r ".verificationMaterial.x509CertificateChain.certificates.[].rawBytes" | base64 --decode > eph.crt
> openssl x509 -noout -text -in eph.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            72:5c:b7:4b:a0:ea:b6:2a:e5:71:cc:5a:63:4e:8d:b1:a5:6a:20:22
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=ES, ST=Some-State, O=Chainloop, OU=chainloop, CN=chainloop
...
```

Note that verification is not yet implemented.

Refs #914


app/cli/internal/action/attestation_push.go
```diff
@@ -133,7 +134,7 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
 
 	wrappedSigner := signer.GetSigner(action.keyPath, action.Logger, pb.NewSigningServiceClient(action.CPConnection))
 	renderer, err := renderer.NewAttestationRenderer(action.c.CraftingState, action.cliVersion, action.cliDigest, wrappedSigner,
-		renderer.WithLogger(action.Logger))
+		renderer.WithLogger(action.Logger), renderer.WithBundle(action.bundlePath))
```
migmartri, line 137: this naming is a little bit cofusing to me, it feel like you are providing a bundle when what you are saying is that it should output somewhere right?

In that case I'd call it `WithBundleOutput`, or `WithBundleOutputPath` wdyt?

internal/attestation/renderer/renderer.go
```diff
@@ -87,15 +106,85 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
 		return nil, err
 	}
 
-	signedAtt, err := ab.signer.SignMessage(bytes.NewReader(rawStatement))
+	signedAtt, err := ab.dsseSigner.SignMessage(bytes.NewReader(rawStatement))
 	if err != nil {
 		return nil, fmt.Errorf("signing message: %w", err)
 	}
 
-	var dseeEnvelope dsse.Envelope
```
migmartri, line 95: oops, thanks for fixing the typo :)

internal/attestation/renderer/renderer.go
```diff
@@ -87,15 +106,85 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
...
+		err = os.WriteFile(ab.bundlePath, json, 0600)
+		if err != nil {
+			return nil, fmt.Errorf("writing bundle: %w", err)
+		}
+	}
+
+	return &dsseEnvelope, nil
+}
+
+func (ab *AttestationRenderer) envelopeToBundle(dsseEnvelope dsse.Envelope) (*protobundle.Bundle, error) {
```
migmartri, line 139: It would be great to add some unit tests on this. I believe somewhere else I did some `golden` tests that rely on `snapshots` of the output. In other words, you can put envelopes as inputs and bundles as outputs, and the test just check those.

internal/attestation/renderer/renderer.go
```diff
@@ -87,15 +106,85 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
...
+	// DSSE Envelope is already base64 encoded, we need to decode to prevent it from being encoded twice
+	payload, err := base64.StdEncoding.DecodeString(dsseEnvelope.Payload)
```
migmartri, line 141: this feels unnecessary. Is there any way we can transform, or use (pass to the function) &dsse2.Envelope instead of the old dsse.Envelope?

internal/attestation/renderer/renderer.go
```diff
@@ -87,15 +106,85 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
...
+	case *cosign.Signer:
+		pk, err := signature.PublicKeyPem(v)
+		if err != nil {
+			return nil, fmt.Errorf("getting public key: %w", err)
+		}
+
+		bundle.VerificationMaterial.Content = &protobundle.VerificationMaterial_PublicKey{PublicKey: &v12.PublicKeyIdentifier{
+			Hint: string(pk),
+		}}
+	case *chainloopsigner.Signer:
```
migmartri, line 171: could this be encapsulated in the signer interface itself?

basically something like `(s ChainloopSigner) VerificationMaterial`?

migmartri (approved):
Cool, left some notes :)

internal/attestation/renderer/renderer.go
jiparis: Yes I was thinking that. Let me give it a look

migmartri: Please mark the Pr as draft if you plan to do any changes. Thanks!

internal/attestation/renderer/renderer.go
jiparis: They are just different implementations of the same thing: `dsse2.Envelope` is the protobuf definition from Sigstore. Our implementation is using the one from Secure Systems Lab instead. We'd need to change all our code to adopt the new one. We can do it in another PR.

internal/attestation/renderer/renderer.go
jiparis: I'll do it as part of a separate PR, as it requires defining a new interface and moving some packages.

Issue #942, opened 2024-06-12
support auto-onboarding users to organization


migmartri:
It would be helpful if the chainloop instance could be configured so new users are automatically onboarded to a given organization(s).

For example this could be part of the configuration file and could look like

```bash
autoOnboarding:
  - org: my-org
    role: viewer
  - org: my-other-org
    role: admin
```

Once configured, when the user logs in, the system will

- Check if the org exist, if it does not, it will create it
- Create a membership for the current user to that organization with that role.

What do you think?

javirln:
> Create a membership for the current user to that organization with that role.

There's a slight problem with this approach. If a user already belongs to the organization with one role, attempting to create a new membership with a different role will trigger a database constraint. This constraint ensures that only one role can exist for a given user and organization, which is logical.

If we want to update the role, we face another issue: the user attempting to update the role cannot be the same user who currently holds the membership.

My suggestion is as follows: If no membership exists, we apply the desired role. If this succeeds, great. If the user already has a membership for that organization, it may require an admin to update the role, and we do nothing further.

migmartri:
> If we want to update the role, we face another issue: the user attempting to update the role cannot be the same user who currently holds the membership.

can you elaborate? Why can't our system change the role?

migmartri:
In any case I think it's ok to not to override existing roles and take this as an `onboarding` for new users. Not a way to have a source of truth in code. Does it make sense?

javirln:
> > If we want to update the role, we face another issue: the user attempting to update the role cannot be the same user who currently holds the membership.
> 
> can you elaborate? Why can't our system change the role?

Yes, basically is the following constraint: https://github.com/chainloop-dev/chainloop/blob/84ad7ed5807dd52b8b9d171855b739390333d372/app/controlplane/pkg/biz/membership.go#L159

> In any case I think it's ok to not to override existing roles and take this as an onboarding for new users. Not a way to have a source of truth in code. Does it make sense?

Yep, it makes total sense! :D

migmartri:
> Yes, basically is the following constraint:

I see, but btw, remember, we can add as many use-cases as we need. In this case. That specific one is suited for users updating others, but we could have created another one that doesn't have that constraint. In any case, I think that skipping an update makes more sense in this case.

Issue #945, opened 2024-06-12
improve allowlist middleware


migmartri:
Currently, there is an allowlist mechanism that can be set inn the backend, but the message shown to the user can not be configured. It could be useful to have a way to customize the information shown in the CLI to the users so they have clear call to actions, such as following a link to request access.
```[tasklist]
### Tasks
- [ ] support custom message
- [ ] support specific routes
```

migmartri:
```
allowList:
  rules: ["foo@bar.com", "@baz.com"]
  customErrorMsg: "this is limited please visit http://foobar.com/request-access"
  selectRoutes: ["controlplane.v1.chainloop.dev/Create...."]
```

PR #951, opened 2024-06-13
feat(auto-onboarding): Allow auto-onboarding on organizations

javirln:
This patch introduces a new configuration option that enables a Chainloop instance to automatically onboard new users into predefined organizations with specific roles.

For example:
```yaml
# config.devel.yaml

# Previous existing configuration

onboarding:
  - name: "read-only-demo"
    role: "viewer"
  - name: "read-write-demo"
    role: "owner"
```

With this configuration, new and existing users will be automatically onboarded into the `read-only-demo` and `chainloop` organizations with the roles of `viewer` and `owner`, respectively. When users log in to Chainloop for the first time, the server will:

- Find or create the specified organizations
- Create memberships for users in these organizations if they do not already exist. Existing memberships will not be modified.

Refs #942


app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -58,6 +58,9 @@ message Bootstrap {
     // https://github.com/chainloop-dev/chainloop/blob/126f47b6c0803eac844b8e3e1a21d582f00e4dc6/app/artifact-cas/internal/service/download.go#L34
     string download_url = 3;
   }
+
+  // Configuration for onboarding users in organizations with specific roles
+  OnboardingSpec onboarding = 10;
```
migmartri, line 63: any reason you want to have a two level object here? 

```
onboarding:
  `auto_onboard_organizations`: ...
```

wouldn't something like

`repeated OnboardingSpec auto_onboarding = 10` be enough?

That will render a simpler yaml I think

```yaml
auto_onboarding:
  - name: foo
    role: bar
```

migmartri (commented):
More feedback to come

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
javirln: No strong reason, just made it two object in case we need to add more functionalities in the future, we have the `OnboardinSpec` already there.

javirln: In any case, I can flatten it, not a big deal.

Makefile
```diff
@@ -13,7 +13,7 @@ api:
 .PHONY: config
 # generate config proto
 config:
-	cd ./internal/credentials/api && buf generate
+	cd ./pkg/credentials/api && buf generate
```
migmartri, line 16: did I miss this? Sorry and thanks!

app/controlplane/configs/config.devel.yaml
```diff
@@ -58,3 +58,9 @@ auth:
 #   enabled: true
 #   allowed_orgs:
 #     - deadbeef
+
+onboarding:
```
migmartri, line 62: I'd not add this feature by default in development, otherwise we will not be able to test the `no orgs` available scenario. What do you think?

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -135,3 +138,11 @@ message CA {
     string key_pass = 3;
   }
 }
+
+// OnboardingSpec is a configuration to automatically onboard users in organizations with specific roles
+message OnboardingSpec {
+  // name is the name of the organization
```
migmartri, line 144: `name of the organization` or is there a reason to do `x is the x` ?

app/controlplane/internal/service/auth.go
```diff
@@ -289,6 +301,61 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) (
 	return http.StatusTemporaryRedirect, nil
 }
 
+// autoOnboardOnOrganizations onboards the user to the organizations specified in the config
+// If the organization does not exist, it will be created with an inline CAS backend
+// If the user is not already a member of the organization, a membership will be created with the role specified in the config
+func autoOnboardOnOrganizations(ctx context.Context, svc *AuthService, u *biz.User) error {
+	for _, spec := range svc.onboardingConfig {
+		// Ensure the organization exists or create it if it doesn't
+		org, err := svc.orgUseCase.FindOrCreate(ctx, spec.GetName())
```
jiparis, line 310: You might need to introduce some locking here. This code might be run by different threads or pods, so it must be reentrant.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -135,3 +138,11 @@ message CA {
...
+  string name = 1;
+  // role is the role to assign to the user. For the moment, only "viewer" is supported
+  string role = 2;
```
migmartri, line 147: what if we use the proto definition we have for a role?

https://github.com/chainloop-dev/chainloop/blob/3f2f0ead388c46c5bdba98a8823832dbb557d5a4/app/controlplane/api/controlplane/v1/response_messages.proto#L145-L149

Then, we can extract this code to be part of the proto package

https://github.com/chainloop-dev/chainloop/blob/3f2f0ead388c46c5bdba98a8823832dbb557d5a4/app/controlplane/internal/service/user.go#L98

something like `(r pb.MembershipRole) ToAuthz ...`

and then we will not need the mapping we do

app/controlplane/internal/service/auth.go
```diff
@@ -242,6 +249,11 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) (
 		svc.log.Infow("msg", "new user associated to an org", "org_id", currentOrg.ID, "user_id", u.ID)
 	}
 
+	// Auto onboard the user to the organizations specified in the config
+	if err := autoOnboardOnOrganizations(ctx, svc, u); err != nil {
```
migmartri, line 253: nitpick. `AutoOnboardOrganizations` sound better to me.

migmartri: this whole method could probably be part of a biz user-case instead no? The less logic we have in svc the better, and this module is already pretty overloaded. 

so yes, I'd move this method to biz under organization biz use-cases

the signature could be smth like 
```
func (s *..usecase) autoOnboardOnOrganizations(ctx context.Context, orgNames []string, userID string) error
```

note how I am removing depending on objects and instead make sure we pass just the attributes, i.e userID instead of the whole user, this will help with testing later on.

app/controlplane/internal/service/auth.go
```diff
@@ -289,6 +301,61 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) (
...
+		// Ensures the user is a member of the organization with the specified role
+		if err := ensureMembership(ctx, svc, org.ID, u.ID, role); err != nil {
+			return sl.LogAndMaskErr(err, svc.log)
+		}
+	}
+	return nil
+}
+
+// determineRole maps the config role to the appropriate authz.Role
+func determineRole(configRole string) authz.Role {
```
migmartri, line 327: as mentioned before, we might not need this if we use the proto definitions that already exist + the mapping we already have.

app/controlplane/internal/service/auth.go
```diff
@@ -289,6 +301,61 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) (
...
+		return authz.RoleViewer
+	case AutoOnboardingOwnerRole:
+		return authz.RoleOwner
+	default:
+		return authz.RoleViewer
+	}
+}
+
+// ensureMembership ensures the user is a member of the organization with the specified role
+func ensureMembership(ctx context.Context, svc *AuthService, orgID, userID string, role authz.Role) error {
```
migmartri, line 341: same, this should be in biz.

app/controlplane/pkg/biz/organization.go
```diff
@@ -197,6 +198,32 @@ func (uc *OrganizationUseCase) FindByID(ctx context.Context, id string) (*Organi
 	return org, nil
 }
 
+// FindByName finds an organization by name.
+func (uc *OrganizationUseCase) FindByName(ctx context.Context, name string) (*Organization, error) {
```
migmartri, line 202: once you move the other code to biz, you might not need to expose this use-case nor the one below. You'll still need the repo though.

migmartri (commented):
Some comments on code structuring and how to use what we already have. 

Thanks!

migmartri (changes requested):
requesting changes per our conversation offline, thanks

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
javirln: Happy to move the enum as par of the new configuration. The problem comes when extracting that the role mapping, that it creates an importing cycle difficult to solve.
```
2024/06/14 14:33:22 running ent codegen: entc/load: package command-line-arguments
	imports github.com/chainloop-dev/chainloop/app/controlplane/pkg/data/ent/schema
	imports github.com/chainloop-dev/chainloop/app/controlplane/internal/authz
	imports github.com/chainloop-dev/chainloop/app/controlplane/internal/conf/controlplane/config/v1
	imports github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1
	imports github.com/chainloop-dev/chainloop/app/controlplane/internal/authz: import cycle not allowed
```

migmartri: I see, can the mapping be moved to biz? instead of svc? That might help with calling it from the new use-case.

javirln: Yep, I believe that would do the trick, I'll try it. Thanks.

javirln:
All changes addressed:
- Moved logic to biz
- Moved all integration tests to `organization_integration_tests`
- Example of configuration the sample config
- Created an alternative method of creating users otherwise integration tests won't make sense. Another option was to use a repo on the integration tests but, not super happy about it.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -135,3 +139,11 @@ message CA {
     string key_pass = 3;
   }
 }
+
+// OnboardingSpec is a configuration to automatically onboard users in organizations with specific roles
+message OnboardingSpec {
+  // Name of the organization
+  string name = 1;
```
migmartri, line 146: for next time since we are not being consistent. We could add the validations too `[(buf.validate.field).string.min_len = 1];`

app/controlplane/internal/service/user.go
```diff
@@ -95,19 +95,6 @@ func (s *UserService) DeleteMembership(ctx context.Context, req *pb.DeleteMember
 	return &pb.DeleteMembershipResponse{}, nil
 }
 
-func pbRoleToBiz(r pb.MembershipRole) authz.Role {
```
migmartri, line 98: thanks :)

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +264,75 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
 	// Delete the organization
 	return uc.orgRepo.Delete(ctx, orgUUID)
 }
+
+// AutoOnboardOrganizations creates the organizations specified in the onboarding config and assigns the user to them
+// with the specified role if they are not already a member.
+func (uc *OrganizationUseCase) AutoOnboardOrganizations(ctx context.Context, userID string) error {
```
migmartri, line 270: I am wondering if it makes sense to pass the onboarding config as a parameter here instead. That way there is no dependency injection dependency in the module and can be tested more easily. But either way, it's fine :)

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +264,75 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
...
+	}
+
+	return nil
+}
+
+// ensureOrganizationExists ensures that an organization specified by the onboarding configuration exists.
+// If the organization does not exist, it creates it.
+func (uc *OrganizationUseCase) ensureOrganizationExists(ctx context.Context, spec *conf.OnboardingSpec) (*Organization, error) {
+	// Ensure the organization exists or create it if it doesn't
+	org, err := uc.FindByName(ctx, spec.GetName())
```
migmartri, line 303: why are you using the use-case here instead of the existing underlying repo? I'd say that this private method should use the private repo instead. It will also make error handling easier.

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +264,75 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
 	// Delete the organization
 	return uc.orgRepo.Delete(ctx, orgUUID)
 }
+
+// AutoOnboardOrganizations creates the organizations specified in the onboarding config and assigns the user to them
+// with the specified role if they are not already a member.
+func (uc *OrganizationUseCase) AutoOnboardOrganizations(ctx context.Context, userID string) error {
+	// Parse user UUID
```
migmartri, line 271: can uc.onboardingConfig be nil?

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +264,75 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
...
+		}
+	}
+
+	return org, nil
+}
+
+// ensureUserMembership ensures that a user is a member of the specified organization with the appropriate role.
+// If the membership does not exist, it creates it.
+func (uc *OrganizationUseCase) ensureUserMembership(ctx context.Context, orgUUID, userUUID uuid.UUID, spec *conf.OnboardingSpec) error {
+	m, err := uc.membershipRepo.FindByOrgAndUser(ctx, orgUUID, userUUID)
```
migmartri, line 321: as you see, in this case you use a repo not the use-case. Which i think it's correct.

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +264,75 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
...
+	if err != nil && !errors.As(err, &ErrNotFound{}) {
```
migmartri, line 322: are you sure this error handling is correct? I think that method returns `m=nil` if not found, not ErrNotFound?

app/controlplane/pkg/biz/organization_integration_test.go
```diff
@@ -268,3 +272,88 @@ func (s *OrgIntegrationTestSuite) SetupTest() {
...
+	}))
+
+	s.setupUsersAndOrganization(ctx, assert)
+	s.setupMembership(ctx, assert)
+}
+
+func (s *AuthOnboardingTestSuite) setupUsersAndOrganization(ctx context.Context, assert *assert.Assertions) {
+	var err error
+	s.org, err = s.Organization.Create(ctx, "onboarded-org")
+	assert.NoError(err)
```
migmartri, line 302: for this kind of setup, you might want to use `require.NoError` instead of `assert` that way we stop the execution of the tests.

app/controlplane/pkg/biz/organization_integration_test.go
```diff
@@ -268,3 +272,88 @@ func (s *OrgIntegrationTestSuite) SetupTest() {
...
+			Name: "testing-org",
+			Role: v1.MembershipRole_MEMBERSHIP_ROLE_ORG_VIEWER,
+		},
+	}))
+
+	s.setupUsersAndOrganization(ctx, assert)
+	s.setupMembership(ctx, assert)
+}
+
+func (s *AuthOnboardingTestSuite) setupUsersAndOrganization(ctx context.Context, assert *assert.Assertions) {
```
migmartri, line 299: You do not need to pass `assert *assert.Assertions` nor use `assert := assert.New(t)` the suite already has it i.e `s.NoError`

app/controlplane/pkg/biz/organization_integration_test.go
```diff
@@ -268,3 +272,88 @@ func (s *OrgIntegrationTestSuite) SetupTest() {
...
+	assert.NoError(err)
+}
+
+func (s *AuthOnboardingTestSuite) setupMembership(ctx context.Context, assert *assert.Assertions) {
+	var err error
+	s.m, err = s.Membership.Create(ctx, s.org.ID, s.usr1.ID, biz.WithMembershipRole(authz.RoleViewer))
+	assert.NoError(err)
+}
+
+func (s *AuthOnboardingTestSuite) TestAutoOnboardOrganizations() {
```
migmartri, line 317: I'd also test

- The user find or create method calling this new method
- calling this method twice
- calling this without configuration

app/controlplane/pkg/biz/user.go
```diff
@@ -90,6 +96,15 @@ func (uc *UserUseCase) DeleteUser(ctx context.Context, userID string) error {
 	return uc.userRepo.Delete(ctx, userUUID)
 }
 
+func (uc *UserUseCase) CreateByEmail(ctx context.Context, email string) (*User, error) {
```
migmartri, line 99: I have mixed feelings of creating methods just for testing (same about findbyname), it feels upside down to me. 

In fact I think in this case this is wrong and can be confusing, it's not trivial to know that `createbyemail` will not add auto-onboarding. 

Personally, I've have passed an extra optional argument `WithDisableAutoOnboarding` or make it part of the the userUsecase struct`

migmartri (commented):
Thanks Javi, looks good in general!

app/controlplane/pkg/biz/user.go
javirln: I would wanted it to be more explicit on the auto but adding it to the FindOrCreate does not harm much as long as we always remember that the autonboarding exists there.

I've added an optional parameter to the method to disable the auto onboarding. Please note that I didn't create a regular Builder with all Its options because I do believe it's too verbose for just one configuration.

javirln (commented):
I've removed the methods `FinByName` and `CreateByEmail`, the dependencies on the private methods and reworked on the tests

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +248,79 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
...
+			return fmt.Errorf("failed to ensure user membership: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// ensureOrganizationExists ensures that an organization specified by the onboarding configuration exists.
+// If the organization does not exist, it creates it.
+func (uc *OrganizationUseCase) ensureOrganizationExists(ctx context.Context, spec *conf.OnboardingSpec) (*Organization, error) {
```
migmartri, line 285: Not a blocker but in general I prefer to make the arguments as scooped down as possible. For example instead of the whole spec,  here I think you just need the name string, and in the other one you need the role.

migmartri (approved):
LGTM!

app/controlplane/pkg/biz/organization.go
jiparis: 👍  I agree. It would be simpler to just pass the name of the org.

app/controlplane/pkg/biz/user.go
```diff
@@ -90,15 +96,30 @@ func (uc *UserUseCase) DeleteUser(ctx context.Context, userID string) error {
 	return uc.userRepo.Delete(ctx, userUUID)
 }
 
-func (uc *UserUseCase) FindOrCreateByEmail(ctx context.Context, email string) (*User, error) {
+// FindOrCreateByEmail finds or creates a user by email. By default, it will auto-onboard the user
+// to the organizations defined in the configuration. If disableAutoOnboarding is set to true, it will
+// skip the auto-onboarding process.
+func (uc *UserUseCase) FindOrCreateByEmail(ctx context.Context, email string, disableAutoOnboarding ...bool) (*User, error) {
```
jiparis, line 102: Why `...bool`? How many `bools` are we expecting here?

app/controlplane/pkg/biz/organization.go
```diff
@@ -244,3 +248,79 @@ func (uc *OrganizationUseCase) Delete(ctx context.Context, id string) error {
...
+	// Ensure the organization exists or create it if it doesn't
+	org, err := uc.orgRepo.FindByName(ctx, spec.GetName())
+	if err != nil {
+		return nil, fmt.Errorf("failed to find organization: %w", err)
+	} else if org == nil {
```
jiparis, line 290: I might be wrong, but I remember we talked offline about about not creating orgs, and assume they existed previously.

jiparis (approved):
Thanks! Just a couple of minor comments.

app/controlplane/pkg/biz/organization.go
javirln: I agree with you but in this case, we need the whole spec, the expected roles as well, so effectively what's being passed in the name of the orgs and the roles it should be applied.

app/controlplane/pkg/biz/user.go
javirln: It's an optional parameter otherwise we would have to add that parameter to all occurrences.

app/controlplane/pkg/biz/organization.go
javirln: As far as I remember if it was because of the possible collisions it was solved on the DB. In any case we can always come back and remove it.

Maybe when we have https://github.com/chainloop-dev/chainloop/issues/945 done.

migmartri: True, I'd do it in another smaller follow up PR

Issue #922, opened 2024-06-07
Pod fails to start on OpenShift: Error: runc create failed: unable to start container process: eaccess /atlas: permission denied


hanygirgis:
I'm trying to deploy Chainloop on an OpenShift environment, using a command similar to the following: 

helm install chainloop oci://ghcr.io/chainloop-dev/charts/chainloop \
    --set global.openshift=true \
    --set development=true \
    --set controlplane.auth.oidc.url=*url* \
    --set controlplane.auth.oidc.clientID=*client* \
    --set controlplane.auth.oidc.clientSecret=*secret* \
    --set controlplane.image.repository=*local*/chainloop/control-plane \
    --set controlplane.image.tag=v0.91.0 \
    --set controlplane.migration.image.repository=*local*/chainloop/control-plane-migrations \
    --set controlplane.migration.image.tag=v0.91.0 \
    --set cas.image.repository=*local*/chainloop/artifact-cas \
    --set cas.image.tag=v0.91.0
    
However, the **chainloop-controlplane** pod fails during initialization (Init:CreateContainerError) with error:
    
_Error: container create failed: time="2024-06-07T12:16:22Z" level=error msg="runc create failed: unable to start container process: eaccess /atlas: permission denied"_

Any idea how to solve this issue?

migmartri:
Hi @hanygirgis.

Do you mind sharing some more information about your openshift environment? We are in the process of adding support of customizable `podsecuritycontext` that might be the culprit of your issue.

Thanks!

hanygirgis:
Hi @migmartri,

I'm using this version:
```
OpenShift version: 4.13.27
Kubernetes version: v1.26.11
```
A suggestion is to maybe used the same `global.openshift` boolean flag that is used by Vault to configure the `podsecuritycontext`, to allow openshift compatibility with just one setting (if possible).

Thanks!

migmartri:
That's a great pointer! Summoning @javirln, who's currently working on some Helm Chart changes to discuss next steps.

In any case, @hanygirgis we are open to contributions if you want to give it a try, we'll be happy to assist, either here at Github or also in our [Slack](https://join.slack.com/t/chainloop-community/shared_invite/zt-2k34dvx3r-u85uGP_KiLC6ic5Wy4aRnQ) :)

Thanks

javirln:
> Thanks for the invitation : ) Sure, I'll be glad to contribute.

Hello @hanygirgis :)

[...]

PR #953, opened 2024-06-13
chore: upgrade ORM to support postgresql 15

migmartri:
the latest version of ent supports postgresql 15 and 16.

Before this change, the CP was erroring at boot

```
2024-06-13T21:43:22.318Z [INFO]  plugin: plugin process exited: path=/plugins/chainloop-plugin-smtp pid=30                                                                                                                                                                                
2024-06-13T21:43:22.318Z [DEBUG] plugin: plugin exited                                                                                                                                                                                                                                    
INFO msg=watcher's ctx cancel : context canceled                                                                                                                                                                                                                                          
panic: failed to create adapter: postgres: unexpected number of rows: 1                    
```

refs #922


javirln (approved):
Thanks Migue! The changes are compatible not only with the Chart version but also with RDS right?

migmartri:
> Thanks Migue! The changes are compatible not only with the Chart version but also with RDS right?

the [compatibility notes](https://github.com/ent/ent/blob/master/doc/md/dialects.md#postgresql) don't tell me otherwise.

I'll give it a test though, thanks!

PR #971, opened 2024-06-17
feat(cli): output format to export PAE-encoded payload

### LINKED ISSUE #968: feat(cli): Verify Chainloop attestations with cosign verify-blob
Problem:
Chainloop uses DSSE Envelope format for attestations. The [DSSE protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol) states that the payload must be encoded using the "Pre-Authenticated Encoding" (PAE) before signing. The PAE just adds a header to the payload, and then the result is signed:
```
PAE(type, body) = "DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(body) + SP + body
+               = concatenation
SP              = ASCII space [0x20]
"DSSEv1"        = ASCII [0x44, 0x53, 0x53, 0x45, 0x76, 0x31]
LEN(s)          = ASCII decimal encoding of the byte length of s, with no leading zeros
```

The final result is a DSSE envelope with a `payloadType`, a `payload`, and a set of `signature`s:
```
{
   "payload": "base64payload",
   "signatures": [
      {
         "keyid": "",
         "sig": "MEUCIQD2lrUoBR/WhFDFl9WarHUOAn1xL9jnARLNRnvrlx/JagIgflAUm/PEENJEUBIXUG7dyH651puqS+7KRwHJj2viPKQ="
      }
   ]
}
```
According to [the protocol](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol), verifiers should form a PAE-encoded version of the payload, and then verify it against the signatures. Chainloop already does this, but external verifiers should make their own implementations as well.

For example, if we wanted to verify an attestation with Cosign, we'd need to do something like this:
```
# Get the attestation
> chainloop wf run describe --digest sha256:8e0bb3ce5b2aa98fb1aaab2cb9e8ba8015e9802d2a6d0ddb75aa09a96750bc85 --output attestation > att.json
# Extract the signature:
> cat att.json | jq -r '.signatures[].sig' > sig.txt
# Extract the payload and add the PAE header (works for MacOS shell, Linux command options might vary):
> cat att.json | jq -r .payload | base64 --decode > payload.txt
> sed -i .bak "1s/^/DSSEv1 28 application\/vnd.in-toto+json $(stat -f%z payload.txt) /" payload.txt
# Finally, verify with cosign
> cosign -d verify-blob --signature sig.txt --insecure-ignore-tlog --key cosign.pub payload.txt
WARNING: Skipping tlog verification is an insecure practice that lacks of transparency and auditability verification for the blob.
Verified OK
```

The goal of this task is to make this verification easier to improve the interoperability of both tools. 
The proposal is to add a new output format for `wf run describe` that outputs the PAE-encoded payload. For example:
```
# Get signature
> chainloop wf run describe --digest sha256:8e0bb3ce5b2aa98fb1aaab2cb9e8ba8015e9802d2a6d0ddb75aa09a96750bc85 --output attestation | jq -r '.signatures[].sig' > sig.txt
# Get PAE-endoded payload
> chainloop wf run describe --digest sha256:8e0bb3ce5b2aa98fb1aaab2cb9e8ba8015e9802d2a6d0ddb75aa09a96750bc85 --output payload-pae > payload.txt
# Verify
> cosign -d verify-blob --signature sig.txt --insecure-ignore-tlog --key cosign.pub payload.txt
```

jiparis:
This PR implements the `payload-pae` output format to obtain the PAE-encoded payload for easier verification:
```
go run app/cli/main.go --insecure wf run describe --digest sha256:8e0bb3ce5b2aa98fb1aaab2cb9e8ba8015e9802d2a6d0ddb75aa09a96750bc85  --output payload-pae > payload.pae.txt
```
payload.pae.txt would look like:
```
DSSEv1 28 application/vnd.in-toto+json 1414 {"_type":"https://in-toto.io/Statement/v1","subject" ...
```
Then, it can be easily matched against the signature with Cosign:
```
cosign -d verify-blob --signature sig.txt --insecure-ignore-tlog --key cosign.pub payload.pae.txt
```
Closes #968

### DIFF
--- a/app/cli/cmd/workflow_workflow_run_describe.go
+++ b/app/cli/cmd/workflow_workflow_run_describe.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
@@ -26,11 +27,13 @@ import (
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/jedib0t/go-pretty/v6/text"
 	"github.com/muesli/reflow/wrap"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/spf13/cobra"
 )
 
 const formatStatement = "statement"
 const formatAttestation = "attestation"
+const formatPayloadPAE = "payload-pae"
 
 func newWorkflowWorkflowRunDescribeCmd() *cobra.Command {
 	var runID, attestationDigest, publicKey string
@@ -58,7 +61,7 @@ func newWorkflowWorkflowRunDescribeCmd() *cobra.Command {
 				return err
 			}
 
-			return encodeAttestationOutput(res)
+			return encodeAttestationOutput(res, os.Stdout)
 		},
 	}
 
@@ -73,7 +76,7 @@ func newWorkflowWorkflowRunDescribeCmd() *cobra.Command {
 	}
 
 	// Override default output flag
-	cmd.InheritedFlags().StringVarP(&flagOutputFormat, "output", "o", "table", "output format, valid options are table, json, attestation or statement")
+	cmd.InheritedFlags().StringVarP(&flagOutputFormat, "output", "o", "table", "output format, valid options are table, json, attestation, statement or payload-pae")
 
 	return cmd
 }
@@ -193,7 +196,7 @@ func predicateV1Table(att *action.WorkflowRunAttestationItem) {
 	}
 }
 
-func encodeAttestationOutput(run *action.WorkflowRunItemFull) error {
+func encodeAttestationOutput(run *action.WorkflowRunItemFull, writer io.Writer) error {
 	// Try to encode as a table or json
 	err := encodeOutput(run, workflowRunDescribeTableOutput)
 	// It was correctly encoded, we are done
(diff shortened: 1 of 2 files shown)


app/cli/cmd/workflow_workflow_run_describe.go
```diff
@@ -19,18 +19,21 @@ import (
...
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/jedib0t/go-pretty/v6/text"
 	"github.com/muesli/reflow/wrap"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/spf13/cobra"
 )
 
 const formatStatement = "statement"
 const formatAttestation = "attestation"
+const formatPayloadPAE = "payload-pae"
```
migmartri, line 36: can you add a comment here explaining what's this for?

app/cli/cmd/workflow_workflow_run_describe.go
```diff
@@ -217,7 +220,18 @@ func encodeAttestationOutput(run *action.WorkflowRunItemFull) error {
...
 	case formatAttestation:
 		return encodeJSON(run.Attestation.Envelope)
+	case formatPayloadPAE:
+		return describePayload(run, writer)
 	default:
 		return ErrOutputFormatNotImplemented
 	}
 }
+
+func describePayload(run *action.WorkflowRunItemFull, writer io.Writer) error {
```
migmartri, line 230: this name can be a little but confusing to me, why is it called `describePayload`?

migmartri (approved):

app/cli/cmd/workflow_workflow_run_describe.go
jiparis: I agree, I'll change it.

PR #984, opened 2024-06-19
fix: return empty list of memberships instead of error

migmartri:
```
grpcurl --plaintext -H "authorization: Bearer $TOKEN"  localhost:9000  controlplane.v1.UserService.ListMemberships
{
  
}
```

```
go run main.go --insecure org ls -o json                                                                          
[]
```


jiparis (approved):

javirln (commented):
If we add that endpoint, when a user is logged in without an organization, it will raise a JWT missing error

javirln (approved):

PR #986, opened 2024-06-19
fix: do not create org during auto-onboarding

### LINKED ISSUE #985: auto-onboarding follow-ups
- putting an invalid role shouldn't crash at runtime

```
2024-06-19T09:34:19.565+0200    INFO    {"kind": "server", "component": "grpc", "operation": "/controlplane.v1.StatusService/Infoz", "args": "", "code": 0, "reason": "", "stack": "", "latency": 0.000208941}
2024-06-19T09:34:24.776+0200    ERROR   {"component": "service", "msg": "failed to auto-onboard user: failed to ensure user membership: failed to create membership: ent: validator failed for field \"Membership.role\": membership: invalid enum value for role field: \"\""}
github.com/go-kratos/kratos/contrib/log/zap/v2.(*Logger).Log
```

- we should not create an org if it does not exist

migmartri:
Closes #985


javirln (approved):

app/controlplane/pkg/biz/organization_integration_test.go
```diff
@@ -275,9 +275,9 @@ func (s *OrgIntegrationTestSuite) SetupTest() {
 
 type AuthOnboardingTestSuite struct {
 	testhelpers.UseCasesEachTestSuite
-	usr, usr1 *biz.User
-	org       *biz.Organization
-	m         *biz.Membership
+	userWithoutOrg, userInOrg *biz.User
```
migmartri, line 278: made the variables more explicit to understand a little bit better who is who

app/controlplane/pkg/biz/organization_integration_test.go
```diff
@@ -286,109 +286,100 @@ func (s *AuthOnboardingTestSuite) SetupTest() {
 
 	s.TestingUseCases = testhelpers.NewTestingUseCases(t, testhelpers.WithOnboardingConfiguration([]*conf.OnboardingSpec{
 		{
-			Name: "testing-org",
+			Name: "non-existing-org",
```
migmartri, line 289: added the option of an org that doesn't exist.

PR #989, opened 2024-06-19
feat(chart): Allow onboarding and allow list on values.yaml

### LINKED ISSUE #987: helm chart support of allow-list and auto-onboarding

javirln:
This patch adds support for configuring the onboarding and the new allow list on the Chart from the values.yaml.

Closes #987


deployment/chainloop/templates/controlplane/config.configmap.yaml
```diff
@@ -10,13 +10,13 @@ metadata:
...
 data:
-  {{- if .Values.controlplane.allowList }}
+  {{- if .Values.controlplane.auth.allowList }}
   allow_list.yaml: |
     auth:
       allow_list:
-        {{- range .Values.controlplane.allowList }}
-        - {{ . | quote }}
-        {{- end }}
+        rules: {{ .Values.controlplane.auth.allowList.rules | toYaml | nindent 10 }}
```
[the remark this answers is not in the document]
javirln: Hmm no strong reason, I just wanted to make sure that I was going to render what I wanted. I can change it.

deployment/chainloop/values.yaml
```diff
@@ -124,6 +124,11 @@ controlplane:
     enabled: false
     allowedOrgs: []
 
+  ## @param controlplane.onboarding Configure the auto onboarding feature for selected organizations
+  ## @extra controlplane.onboarding.name Name of the organization to onboard
+  ## @extra controlplane.onboarding.role Role of the organization to onboard
+  onboarding:
```
migmartri, line 130: this should contain an example

deployment/chainloop/templates/controlplane/config.configmap.yaml
```diff
@@ -41,4 +41,6 @@ data:
       download_url: {{ include "chainloop.cas.external_url" . }}/download
     plugins_dir: {{ .Values.controlplane.pluginsDir }}
     referrer_shared_index:
-      {{- toYaml .Values.controlplane.referrerSharedIndex | nindent 6 }}
\ No newline at end of file
+      {{- toYaml .Values.controlplane.referrerSharedIndex | nindent 6 }}
+    onboarding:
```
migmartri, line 45: we could hide this option if not set.

deployment/chainloop/templates/controlplane/config.configmap.yaml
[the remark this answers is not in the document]
migmartri: either way let's be consistent since it seems that the onboarding one is implemented differently. Let's pick one :)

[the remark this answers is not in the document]
migmartri: I think smth like this should do the trick

```
  allow_list.yaml: |
    auth:
      allow_list:
        {{- toYaml .Values.controlplane.auth.allowList | nindent 8 }}
```

deployment/chainloop/values.yaml
javirln: Added.

deployment/chainloop/values.yaml
```diff
@@ -182,6 +187,14 @@ controlplane:
       clientID: ""
       clientSecret: ""
 
+    ## @param controlplane.auth.allowList.rules List of domains or emails to allow
+    ## @param controlplane.auth.allowList.selectedRoutes List of selected routes to allow
+    ## @param controlplane.auth.allowList.customMessage Custom message to display when a user is not allowed
+    allowList:
```
migmartri, line 194: if we do not want to render this by default we should comment this out and set instead of @param, @extra

migmartri: same about onboarding

javirln: I didn't know that was an option...I'll do it.

deployment/chainloop/values.yaml
```diff
@@ -124,6 +124,12 @@ controlplane:
     enabled: false
     allowedOrgs: []
 
+  ## @param controlplane.onboarding[0].name Name of the organization to onboard
+  ## @param controlplane.onboarding[0].role Role of the organization to onboard
+  onboarding:
+    - name: "read-only-demo"
```
migmartri, line 130: oh, this shouldn't be here

migmartri (changes requested):

deployment/chainloop/values.yaml
javirln: That was the example no?

migmartri: this is not an example, these are the default values. Rendering it without any override will add this onboarding 

```diff
       cas_robot_account_private_key_path: "/secrets/cas.private.key"
@@ -479,6 +479,10 @@
     referrer_shared_index:
       allowedOrgs: []
       enabled: false
+    
+    onboarding:
+      - name: read-only-demo
+        role: MEMBERSHIP_ROLE_ORG_VIEWER

```

migmartri: And yes, I'll comment out onboarding and allowList in values.yaml.

Then the render should show no difference compared to a previous version of the chart

deployment/chainloop/values.yaml
```diff
@@ -182,6 +188,14 @@ controlplane:
       clientID: ""
       clientSecret: ""
 
+    ## @extra controlplane.auth.allowList.rules List of domains or emails to allow
+    ## @extra controlplane.auth.allowList.selectedRoutes List of selected routes to allow
```
migmartri, line 192: indicate that if not set means that it applies to all routes

migmartri (approved):
Please bump the chart version too. Probaby to `.2` since I have .1 in the works.

deployment/chainloop/values.yaml
javirln: Done.

PR #991, opened 2024-06-19
fix(chart): Fix conflict and bump chart and images

javirln:
This patch fixes de conflict and bumps the image versions.


migmartri (approved):
thanks

PR #1000, opened 2024-06-19
chore: store cas-backend bytes in DB

migmartri:
Store the pre-defined max bytes in the database column on creation. 

Updates existing entries with current pre-defined values


jiparis (approved):
Great!. Please, create a follow up task to have this configured from settings.

PR #1001, opened 2024-06-19
feat: upload size server-size validation

### LINKED ISSUE #999: validate max CAS-backend blob size server side
Currently, this check is done client-side before adding material types. 

We should implement two more checks

- One in `chainloop artifact upload` too
- One server side on the CAS proxy. Something we could do is to add the max size to the jwt token used to talk to the CAS proxy and make the check there.

migmartri:
We now make sure that the uploads (both from attestation or from `artifacts upload`) are restricted by the cas-backend size.

This is done by injecting the maxSizebytes in the JWT used to communicate with the cas-backend proxy. Then the proxy uses this value during the upload and continuously check for overflows during the reception of the chunk. 

If it goes over the limit, it rejects the upload.

```
2024-06-19T23:15:11.925+0200    DEBUG   {"component": "service", "msg": "upload chunk received", "digest": "49bc20df15e412a64472421e13fe86ff1c5165e18b2afccf160d4dc19fe68a14", "currentSize": 104857600, "maxSize": 104857600, "chunkSize": 1048576}
2024-06-19T23:15:11.926+0200    ERROR   {"component": "service", "msg": "max size of upload exceeded: current=105906176, max=104857600"}
``` 

closes #999

### DIFF
--- a/app/artifact-cas/internal/server/grpc_test.go
+++ b/app/artifact-cas/internal/server/grpc_test.go
@@ -96,7 +96,7 @@ func TestJWTAuthFunc(t *testing.T) {
 
 			b, err := robotaccount.NewBuilder(opts...)
 			require.NoError(t, err)
-			token, err := b.GenerateJWT("backend-type", "secret-id", tc.audience, robotaccount.Downloader)
+			token, err := b.GenerateJWT("backend-type", "secret-id", tc.audience, robotaccount.Downloader, 0)
 			require.NoError(t, err)
 
 			// add bearer token to context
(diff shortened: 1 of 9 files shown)


jiparis:
> This is done by injecting the maxSizebytes in the JWT
So all the CAS configuration is injected this way when generating the token, right? this is great to avoid unnecessary DB accesses during uploads (and probably the proxy doesn't have access to it anyways).

migmartri:
> > This is done by injecting the maxSizebytes in the JWT
> 
> So all the CAS configuration is injected this way when generating the token, right? this is great to avoid unnecessary DB accesses during uploads (and probably the proxy doesn't have access to it anyways).

That's correct, the CAS backend just receives the allowed max size and trusts it since it's part of the JWT they can unpack and verify with their public key.

PR #1002, opened 2024-06-20
fix(auth): Allow users to delete account even without org

### LINKED ISSUE #998: To be able to delete your account even if you don't have organizations
```
chainloop auth delete-account
ERR you are not part of any organization, please run "chainloop organization create --name ORG_NAME" to create one
```

javirln:
This patch allows users to delete their accounts even if they don't belong to any organization.

Closes #998

### DIFF
--- a/app/controlplane/internal/server/grpc.go
+++ b/app/controlplane/internal/server/grpc.go
@@ -256,7 +256,7 @@ func requireRobotAccountMatcher() selector.MatchFunc {
 
 // Matches all operations that require to have a current organization
 func requireAllButOrganizationOperationsMatcher() selector.MatchFunc {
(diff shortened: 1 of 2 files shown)


migmartri (approved):
this patch seems to have two things, about deleting the account and about seeing the context. 

Are these changes related?

javirln:
> this patch seems to have two things, about deleting the account and about seeing the context.
> 
> Are these changes related?

Yes, the CLI when trying to delete first, tries to get the current context of the user https://github.com/chainloop-dev/chainloop/blob/d458117ae53c556aca16e7a3ca8e0dc89287724e/app/cli/cmd/auth_delete_account.go#L34 to later on, delete the account:
https://github.com/chainloop-dev/chainloop/blob/d458117ae53c556aca16e7a3ca8e0dc89287724e/app/cli/cmd/auth_delete_account.go#L47

jiparis (approved):

PR #1005, opened 2024-06-20
fix: allow leaving account if invitation sent

migmartri:
In the scenario where a user sent an invitation to another, and that person stayed in the organization. The user then couldn't delete the account.

This fixes it by making sure that the invitations are removed.

```
*ent.ConstraintError
ent: constraint failed: ERROR: update or delete on table "users" violates foreign key constraint "org_invitations_users_sender" on table "org_invitations" (SQLSTATE 23503)
```


jiparis (approved):
In general, I'd prefer logical deletes and anonymizing instead (for auditing, reporting, etc). We can review it if you want.

migmartri:
> In general, I'd prefer logical deletes and anonymizing instead (for auditing, reporting, etc). We can review it if you want.

I agree, let's revisit this later. So far, our constants are focused on consistency.

PR #1024, opened 2024-06-24
fix(cli): wait for the command to finish

migmartri:
This patch fixes a bug that made `xdg-open` not to return error and hence not triggering the headless login process

### DIFF
--- a/app/cli/cmd/auth_login.go
+++ b/app/cli/cmd/auth_login.go
@@ -167,20 +167,28 @@ func (a *app) handleCallback(w http.ResponseWriter, r *http.Request) {
 }
 
 func openbrowser(url string) error {
-	var err error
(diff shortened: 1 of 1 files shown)


app/cli/cmd/auth_login.go
```diff
@@ -167,20 +167,28 @@ func (a *app) handleCallback(w http.ResponseWriter, r *http.Request) {
...
-		err = fmt.Errorf("unsupported platform")
+		return fmt.Errorf("unsupported platform")
 	}
 
-	return err
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	if err := cmd.Wait(); err != nil {
```
migmartri, line 187: this was the missing piece :)

javirln (approved):

Issue #1027, opened 2024-06-24
finalize migration to name from IDs


migmartri:
Currently, most of our resources include a `name` column that's unique and follows some dns-type convetions which makes them suitable to replace our current IDs in most operations.

you can see this for example in `workflow describe` which accepts both `name` and `id`

The goal of this task is to 

- review the current state of things with regards of the use of names
- make sure `names` are immutable after creation
- add support for `name` as identifier for operations flags
- remove IDs from tables in the CLI if there is a `name` counterpart. 
```[tasklist]
### Tasks
- [x] Overview https://docs.google.com/document/d/15AWmc4yoHXWwWX_XGFPBbGLBrAdCBDYLIq9D7YD80to/edit
- [ ] https://github.com/chainloop-dev/chainloop/issues/680
- [ ] https://github.com/chainloop-dev/chainloop/issues/681
- [ ] https://github.com/chainloop-dev/chainloop/issues/682
- [ ] https://github.com/chainloop-dev/chainloop/issues/1034
- [ ] https://github.com/chainloop-dev/chainloop/issues/1035
- [ ] https://github.com/chainloop-dev/chainloop/issues/1038
- [ ] https://github.com/chainloop-dev/chainloop/issues/1040
- [ ] https://github.com/chainloop-dev/chainloop/issues/1041
- [ ] https://github.com/chainloop-dev/chainloop/issues/1042
- [ ] https://github.com/chainloop-dev/chainloop/issues/1043
- [x] review clients / cli before release
- [ ] Review documentation
```

PR #1049, opened 2024-06-27
chore(contracts): create workflow with contract name

migmartri:
Supports creating workflows by providing an existing contract name instead of the ID. Refs #1027


javirln (approved):
I think there is some lint errors.

PR #1051, opened 2024-06-28
chore(integrations); move from id to name

migmartri:
- Attach by workflow and integration name
- Filter attachments by workflow name

refs #1027


app/controlplane/pkg/data/orgmetrics.go
```diff
@@ -43,10 +43,10 @@ func NewOrgMetricsRepo(data *Data, l log.Logger) biz.OrgMetricsRepo {
 }
 
 func (repo *OrgMetricsRepo) RunsTotal(ctx context.Context, orgID uuid.UUID, tw time.Duration) (int32, error) {
-	total, err := orgScopedQuery(repo.data.DB, orgID).
+	total, err := orgScopedQuery(repo.data.DB.Debug(), orgID).
 		QueryWorkflows().
 		QueryWorkflowruns().
-		Where(workflowrun.CreatedAtGTE(time.Now().Add(-tw))).
+		Where(workflowrun.CreatedAtIn(time.Now(), time.Now().Add(-tw))).
```
javirln, line 49: This is leftover right?

javirln, line 49: These are leftovers right?

javirln (changes requested):
I think there are leftovers needed to be removed 👀

app/controlplane/pkg/data/orgmetrics.go
migmartri: Oh man, sorry about that. Great catch! ;)

javirln (approved):

PR #1052, opened 2024-06-28
feat(orgmetrics): Accept start and end date for metrics

javirln:
This patch modifies the OrgMetrics use case to receive a start and end date which are more flexible that a time.Duration. Do all the changes up until the repository.

Ent does not have a between operator for dates that's the reason of using gte and lte. In operator does not produce the proper SQL query:

`In` operator:
```sql
SELECT COUNT(DISTINCT "workflow_runs"."id")
FROM "workflow_runs"
         JOIN (SELECT "workflows"."id"
               FROM "workflows"
                        JOIN (SELECT "organizations"."id" FROM "organizations" WHERE "organizations"."id" = $1) AS "t1"
                             ON "workflows"."organization_id" = "t1"."id") AS "t1"
              ON "workflow_runs"."workflow_workflowruns" = "t1"."id"
WHERE "workflow_runs"."created_at" IN ($2, $3)
```

`gte` and `lte` operators:
```sql
SELECT COUNT(DISTINCT "workflow_runs"."id")
FROM "workflow_runs"
         JOIN (SELECT "workflows"."id"
               FROM "workflows"
                        JOIN (SELECT "organizations"."id" FROM "organizations" WHERE "organizations"."id" = $1) AS "t1"
                             ON "workflows"."organization_id" = "t1"."id") AS "t1"
              ON "workflow_runs"."workflow_workflowruns" = "t1"."id"
WHERE "workflow_runs"."created_at" >= $2
  AND "workflow_runs"."created_at" <= $3
```

The difference being, `in` tries to matches within the specific items on the list where the `gte` and `lte` operators try to match within the range between those dates.

### DIFF
--- a/app/controlplane/pkg/biz/orgmetrics.go
+++ b/app/controlplane/pkg/biz/orgmetrics.go
@@ -17,6 +17,7 @@ package biz
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/go-kratos/kratos/v2/log"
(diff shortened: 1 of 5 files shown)


app/controlplane/pkg/biz/orgmetrics.go
```diff
@@ -30,12 +30,12 @@ type OrgMetricsUseCase struct {
 
 type OrgMetricsRepo interface {
 	// Total number of runs within the provided time window (from now)
-	RunsTotal(ctx context.Context, orgID uuid.UUID, timeWindow time.Duration) (int32, error)
+	RunsTotal(ctx context.Context, orgID uuid.UUID, timeWindow TimeWindow) (int32, error)
```
migmartri, line 33: in general I tend to pass pointers to struct instead of a copy of the struct itself. Should we change that in both the params on the repo methods and the use-case ones?

app/controlplane/pkg/biz/orgmetrics.go
```diff
@@ -48,11 +48,17 @@ type ByStatusCount struct {
...
+
 func NewOrgMetricsUseCase(r OrgMetricsRepo, l log.Logger) (*OrgMetricsUseCase, error) {
 	return &OrgMetricsUseCase{logger: log.NewHelper(l), repo: r}, nil
 }
 
-func (uc *OrgMetricsUseCase) RunsTotal(ctx context.Context, orgID string, timeWindow time.Duration) (int32, error) {
+func (uc *OrgMetricsUseCase) RunsTotal(ctx context.Context, orgID string, timeWindow TimeWindow) (int32, error) {
 	orgUUID, err := uuid.Parse(orgID)
 	if err != nil {
 		return 0, err
```
migmartri, line 64: we should probably add a timewindow validation, that `From` and `To` is set.

app/controlplane/pkg/biz/orgmetrics.go
```diff
@@ -48,11 +48,17 @@ type ByStatusCount struct {
 	Count  int32
 }
 
+// TimeWindow represents in time.Time format not in time.Duration
+type TimeWindow struct {
+	StartDate time.Time
```
migmartri, line 53: I'd not call it `Date` since it's a time.

Maybe smth like `From`, `To`?

migmartri (approved):

app/controlplane/pkg/biz/orgmetrics.go
javirln: Will do that yes.

app/controlplane/pkg/biz/orgmetrics.go
javirln: Adding them.

migmartri:
@javirln thanks for explaining the nuances on the `ent` query, I appreciate it :)

javirln:
@migmartri Added validations and custom error handling

migmartri (approved):
Nice!

PR #1054, opened 2024-06-28
feat(docs): Document EJBCA support for signing

jiparis:
This PR adds instructions to using EJBCA as ephemeral certificate issuer.
I've also moved the `--bundle` option to its own section.


migmartri (approved):
LGTM!

It would be great if you could also update the k8s guide by bringing the latest change by running

```
yarn run docusaurus download-remote-deployment-readme
```

Thanks!

jiparis:
> yarn run docusaurus download-remote-deployment-readme

The README won't pass the linter. I'll send a separate PR for it.

Issue #122, opened 2023-05-22
Support policies for attestations


migmartri:
Currently, no verification is done when an attestation is received in the control plane. Creation and validation checks are done in the client side, but not on the server side. 

We should allow operators to attach `rego` or potentially `cue` policies to their contracts and these should get evaluated during the reception of the attestation.  

We created a task #35 which will get superseded by this functionality since 35 is in fact a policy check that some materials exist and that the runner type is correct.

re: implementation

- We can start with making this check an enforcement, meaning that if the policy check fails the `att push` command will too since the check for the first version could be sync.
- We could implement this check as a fan-out integration. Create it by default in your org and automatically attach it to each of your workflows.

About the policy format.

We should probably aim towards Open Policy Agent (OPA) and leverage (conftest) logic. We should take a look at policy-controller [way of doing this](https://docs.sigstore.dev/policy-controller/overview/#configuring-policy-that-validates-attestations)

On the UX side of things we could allow attaching a policy to our contract today.

example from policy-controller that could map our current contract too.

using `cue` 
```yaml
policy:
        type: cue
        data: |
          predicateType: "https://cosign.sigstore.dev/attestation/v1"
          predicate: "foobar e2e test"
```

or rego

```yaml
policy:
        type: rego
        data: |
          package sigstore
          default isCompliant = false
          isCompliant {
            input.predicateType == "https://cosign.sigstore.dev/attestation/v1"
            input.predicate == "foobar e2e test"
          }
```

Note: If we go ahead with the implementation based on an integration. We would need to extend the integrations model to read from its result and apply it to the workflow runs. Currently they are just fire and forget.

```[tasklist]
### Tasks
- [x] Propose policy API (protos)
```

PR #1068, opened 2024-07-03
feat(policies): first draft for a policies API

jiparis:
This is a first draft to add policies support to contracts and attestations.

Given a policy in CRD-like format:
```
kind: Policy
metadata:
  name: my-policy
spec:
  kind: ATTESTATION
  path: policy.rego
```
It's intended to be applied to an `ATTESTATION` material (hence the `kind` filter). The actual policy code it's placed in the Rego file `policy.rego`. For instance, in this example, it would look for an `approval=true` annotation in the attestation.

Then, this policy can be applied to any contract:
```
schemaVersion: v1
materials:
  - type: ATTESTATION
    name: qa-attestation
policies:
  - ref: my-policy
    selector: 
      name: qa-attestation
    with:
      name: author
      value: qa-team@chainloop.dev
```

Meaning that it would check that the `qa-attestation` material  (filtered by the `selector` field) passes `my-policy` policy, setting additional argument `author` so that the rego policy checks that attestation field too.

Refs #122


app/controlplane/api/workflowcontract/v1/policy.proto
```diff
@@ -0,0 +1,62 @@
...
+syntax = "proto3";
+
+package workflowcontract.v1;
+
+import "buf/validate/validate.proto";
+import "workflowcontract/v1/crafting_schema.proto";
+
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1";
+
+message Policy {
```
migmartri, line 25: Thanks for adding all the validations! :)

app/controlplane/api/workflowcontract/v1/policy.proto
```diff
@@ -0,0 +1,62 @@
...
+  string api_version = 1 [(buf.validate.field).string.const = "workflowcontract.chainloop.dev/v1"];
+  string kind = 2 [(buf.validate.field).string.const = "Policy"];
+
+  Metadata metadata = 3 [(buf.validate.field).required = true];
+  PolicySpec spec = 4 [(buf.validate.field).required = true];
+}
+
+message Metadata {
+  // the name of the policy
+  string name = 3 [(buf.validate.field) = {
```
migmartri, line 35: sweet

app/controlplane/api/workflowcontract/v1/policy.proto
```diff
@@ -0,0 +1,62 @@
...
+    cel: {
+      message: "must contain only lowercase letters, numbers, and hyphens.",
+      expression: "this.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",
+      id: "name.dns-1123",
+    },
+    required: true,
```
migmartri, line 41: I believe required is the default behavior no?

app/controlplane/api/workflowcontract/v1/policy.proto
```diff
@@ -0,0 +1,62 @@
...
+  }];
+}
+
+message PolicySpec {
+  // path to a policy script. It might consist of a URI reference
+  string path = 1 [(buf.validate.field).required = true];
+
+  // stage at which this policy will be run.
+  // Only "push" is supported currently and this field will be ignored
+  PolicyStage stage = 2;
```
migmartri, line 51: You can add a validation that checks that the value is not unspecified

app/controlplane/api/workflowcontract/v1/policy.proto
```diff
@@ -0,0 +1,62 @@
...
+      expression: "this.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",
+      id: "name.dns-1123",
+    },
+    required: true,
+  }];
+}
+
+message PolicySpec {
+  // path to a policy script. It might consist of a URI reference
+  string path = 1 [(buf.validate.field).required = true];
```
migmartri, line 47: we might want to do here `oneof source` that for now it can only be a path, tomorrow it could be embed string.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -103,3 +105,32 @@ message Annotation {
   // This value can be set in the contract or provided during the attestation
   string value = 2;
 }
+
+// A policy to be applied to this contract
+message PolicyAttachment {
+  // policy reference. It must consist of an identifier or URI reference
+  string ref = 1 [(buf.validate.field).string.min_len = 1];
```
migmartri, line 112: what do you see as first ref, a local path ref? if so we might want to validate that's file://

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -103,3 +105,32 @@ message Annotation {
...
+  bool disabled = 3;
+
+  repeated PolicyArgument with = 4;
+
+  message MaterialSelector {
+    // material name
+    string name = 1;
+
+    // material type
+    CraftingSchema.Material.MaterialType kind = 2;
```
migmartri, line 128: does this even make sense? I thought that policies by itself already define the material kind?

migmartri (approved):
LGTM! Left some comments!

Thanks

app/controlplane/api/workflowcontract/v1/policy.proto
jiparis: You are right, the regular expression already takes care.

app/controlplane/api/workflowcontract/v1/policy.proto
jiparis: I wanted it to be optional. That's why I added the comment about it being ignored.

migmartri: You can still make it optional I think, either optional or a valid value

jiparis: Let me check

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
jiparis: Yes, maybe we don't need it yet. I was thinking of the case of "apply this policy to all materials of this type". The policy itself could be something generic, like "check that the format is json". But we might be making it too complex. I'll remove this.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
jiparis: I was thinking just the policy `name`, since they need to exist previously.

jiparis: Actually, this could evolve to a `oneof` with `name`, `path`, and `embedded` options.

jiparis: Adding the oneof already to be prepared.

app/controlplane/api/workflowcontract/v1/policy.proto
jiparis: It won't work. If it's not provided, it will be assigned the default value which is `0` (unspecified). So both are equivalent.

PR #1069, opened 2024-07-03
chore(docs): improve signing documentation

### LINKED ISSUE #1067: improve signing comparison page
[This page](https://docs.chainloop.dev/reference/signing) is great, but it would be better if we could indicate there which on of those options require client-side configuration and which one do not (keyless). Maybe shown in another column?  

cc/ @jiparis

jiparis:
This PR adds some clarification on how the different signing methods must be used.

Closes #1067


docs/docs/reference/signing.md
```diff
@@ -15,27 +15,37 @@ Some of these signing methods are inherited from the amazing Sigstore community
...
+| KMS               | `--key awskms://<KeyID>`                                                                               | `--key awskms://<KeyID>`                                                   | 
+| PKCS#11           | `--key pkcs11://<KeyId>`                                                                               | `--key pkcs11://<KeyId>`                                                   |
+| Kubernetes secret | `--key k8s://<namespace>/<secretName>` (where `cosign.key` and `cosign.password` secrets are expected) | `--key k8s://<namespace>/<secretName>` (where `cosign.pub` is expected)    |
+| Gitlab secret     | `--key gitlab://<project>` (it will look for `COSIGN_PRIVATE_KEY`, `COSIGN_PASSWORD` variables)        | `--key gitlab://<project>` (it will look for `COSIGN_PUBLIC_KEY` variable) |
+
+### Keyless signing
+These methods don't require any special setup in the client. You must make sure you get the CA certificate chain out-of-band, as it will be required for the verification command.
```
migmartri, line 30: > You must make sure you get the CA certificate chain out-of-band, as it will be required for the verification command.

I am not sure I understand what this means. It might be easier to explain that this requires configuration in controlplane during deployment, and point to the Hel chart section maybe?

migmartri (approved):

docs/docs/reference/signing.md
jiparis: It's related to the verification in the CLI. You always need the root certificate (no matter the CA you are using), which must be provided by your organization.

PR #1080, opened 2024-07-08
chore: upgrade Go to 1.22.4

jiparis:
Upgrade Go version to fix CVE-2024-24791, as well as Cors to 1.11.0 to fix GHSA-mh55-gqvf-xfwm

PR #1082, opened 2024-07-08
chore(go): update Go version in build pipelines

jiparis:
Follow up of #1080 bumping golang in the pipelines and Docker files


javirln (approved):
LGTM

PR #1086, opened 2024-07-10
feat(policies): evaluate policies and store the result in attestation

jiparis:
Note that this feature is not yet integrated in any chainloop command. This PR includes:

```[tasklist]
- [x] complete evaluation of policies against a crafting state (before rendering and signing)
- [x] include evaluation results in the attestation
- [x] support inline materials
- [x] support embedded policies
- [x] support CAS materials
```

The missing part (it will be done in a different PR) is to connect policy verification during `chainloop attestation push` command.

Please check the test suite for the different use cases and examples.


pkg/policies/policies.go
```diff
@@ -16,13 +16,179 @@
...
+}
+
+// Verify verifies that the statement is compliant with the policies present in the schema
+func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation, error) {
+	violations := make([]*engine.PolicyViolation, 0)
+	policies := pv.state.GetInputSchema().GetPolicies()
+	for _, policyAtt := range policies {
+		if policyAtt.Disabled {
+			// policy is disabled
+			// TODO: WARN.
```
javirln, line 52: Shall we add the warn?

pkg/policies/policies.go
```diff
@@ -16,13 +16,179 @@
...
+	}
+	jsonContent, err := crafter.LoadJSONBytes(rawData, filepath.Ext(reference))
+	if err != nil {
+		return nil, fmt.Errorf("loading policy spec: %w", err)
+	}
+	var policy v1.Policy
+	if err := protojson.Unmarshal(jsonContent, &policy); err != nil {
+		return nil, fmt.Errorf("unmarshalling policy spec: %w", err)
+	}
+	return &policy, nil
```
javirln, line 110: Shall we validate it before returning?
```golang
validator, err := protovalidate.New()
	if err != nil {
		return err
	}

	err = validator.Validate(policy)
	if err != nil {
		return err
	}
```

pkg/policies/policies.go
```diff
@@ -16,13 +16,179 @@
...
+	"path/filepath"
+
 	v1 "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
+	"github.com/chainloop-dev/chainloop/internal/attestation/crafter"
+	v12 "github.com/chainloop-dev/chainloop/internal/attestation/crafter/api/attestation/v1"
+	"github.com/chainloop-dev/chainloop/internal/casclient"
 	"github.com/chainloop-dev/chainloop/pkg/policies/engine"
 	"github.com/chainloop-dev/chainloop/pkg/policies/engine/rego"
+	"github.com/sigstore/cosign/v2/pkg/blob"
+	"google.golang.org/protobuf/encoding/protojson"
```
javirln, line 32: This is super nitpick sorry...could you organize the imports like:
```
standard library

chainloop and current source code

third party dependencies
```

internal/attestation/crafter/api/attestation/v1/crafting_state.proto
```diff
@@ -88,6 +88,26 @@ message Attestation {
...
+// A policy executed against an attestation
+message Policy {
+  // The policy name from the policy spec
+  string name = 1;
+  // The attachment as in the contract, with arguments and any other metadata
+  workflowcontract.v1.PolicyAttachment attachment = 2 [(buf.validate.field).required = true];
+  // The policy script body (rego)
+  string body = 3 [(buf.validate.field).required = true];
+  // The policy violations, if any
+  repeated Violation violations = 4;
```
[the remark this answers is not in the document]
jiparis: Yes, we will send all violations embedded in the attestation. If there is none, it will be an empty list.

pkg/policies/policies.go
jiparis: It's exactly that way. Otherwise the linter would complain (I also have goimports configured locally)

pkg/policies/policies.go
jiparis: In the next PR, I'm injecting the logger. I'll do it there. Thanks for the heads up.

pkg/policies/policies.go
jiparis: well, it seems goimports is not configured properly, checking ...

jiparis: Done!

PR #1087, opened 2024-07-11
feat(policies): evaluate policies on attestation push command

### LINKED ISSUE #122: Support policies for attestations

jiparis:
This PR is a follow up that connects all the pieces to evaluate any policy present in the schema before pushing the attestation. The PR includes:
```[tasklist]
### features
- [x] Evaluating policies against materials during `att add`
- [x] Evaluating policies against statement (not signed) during `att push`
- [x] validate schema in crafter (during `contract create` and `contract update`)
- [x] Storing evaluation results in the predicate, included in the signed attestation, and sent to the controlplane.
``` 

Policies can be specified and linked to any schema. For example, this policy can be used to evaluate any material of type SBOM_SPDX_JSON, and check that was created with Syft.
```
# File policy-made-with-syft.yaml
kind: Policy
metadata:
  name: made-with-syft
spec:
  type: SBOM_SPDX_JSON
  embedded: |
    package main

    import future.keywords.in

    deny[msg] {
      not made_with_syft

      msg := "Not made with syft"
    }

    made_with_syft {
      some creator in input.creationInfo.creators
      contains(creator, "syft")
    }
```
Policy code can also be detached and referenced from the spec. I'll update the documentation in a separate PR.
Then, it can be linked to any contract:
```
# my-schema.yaml
schemaVersion: v1
materials:
  - name: sbom
    type: SBOM_SPDX_JSON
policies:
  materials:
    - ref: "policy-made-with-syft.yaml" 
  attestation:
    - ref: "policy-author.yaml"
```
Note the `policies.materials` and `policies.attestation` section, denoting policies to be applied to all materials or the whole attestation (before signing) accordingly. In the case of materials, since the policy has a `type` in the specification, it will be applied to all materials that matches that type. There are other options and filters to flexibilize the policy selection criteria, allowing to apply policies to specific materials. For example, it can be filtered by name:
```
...
policies:
  materials:
    - ref: "policy-made-with-syft.yaml" 
      selector:
         name: "sbom" # only apply the policy to the "sbom" material
```
If there is no `type` in the policy spec, it will be applied regardless the material has a compatible type or not. This is useful for custom validations and artifacts.

Closes #122


javirln (approved):
Awesome!!

internal/attestation/renderer/chainloop/chainloop.go
```diff
@@ -73,18 +75,6 @@ type ProvenancePredicateCommon struct {
 	RunnerURL  string            `json:"runnerURL,omitempty"`
 	// Custom annotations
 	Annotations map[string]string `json:"annotations,omitempty"`
-	// Applied policies
-	Policies []PolicyPredicate `json:"policies,omitempty"`
-}
-
-// PolicyPredicate represents a policy that has been run against an attestation
-type PolicyPredicate struct {
```
jiparis, line 81: Moved outside the common section.

pkg/policies/policies.go
```diff
@@ -204,3 +170,35 @@ func policyViolationsToAttestationViolations(violations []*engine.PolicyViolatio
 	}
 	return
 }
+
+// We need to create a connection for every single artifact, because it depends on the digest
+func (pv *PolicyVerifier) getCASClient(role pb.CASCredentialsServiceGetRequest_Role, digest string) (*casclient.Client, error) {
```
migmartri, line 175: I believe we have some code that does this already, so we could just pass the connection ready to the action?

migmartri: https://github.com/migmartri/chainloop/blob/40f96e6edc5c4f3bba347b30b44af88614dd8e30/app/cli/cmd/artifact.go#L38-L62

app/cli/cmd/attestation_push.go
```diff
@@ -64,7 +65,10 @@ func newAttestationPushCmd() *cobra.Command {
 			a, err := action.NewAttestationPush(&action.AttestationPushOpts{
 				ActionsOpts: actionOpts, KeyPath: pkPath, BundlePath: bundle,
 				CLIVersion: info.Version, CLIDigest: info.Digest,
-				SignServerCAPath: signServerCAPath,
+				SignServerCAPath:   signServerCAPath,
+				CASURI:             viper.GetString(confOptions.CASAPI.viperKey),
```
migmartri, line 69: Why don't we just pass the CAS connection? We have already code in the CMD package to generate the connection https://github.com/migmartri/chainloop/blob/40f96e6edc5c4f3bba347b30b44af88614dd8e30/app/cli/cmd/artifact.go#L38-L62

you can see an example in artifact_download command

app/cli/internal/action/attestation_add.go
```diff
@@ -20,12 +20,13 @@ import (
 	"errors"
 	"fmt"
 
+	"google.golang.org/grpc"
```
migmartri, line 23: Are you using any specific sorting plugin?

pkg/policies/policies.go
```diff
@@ -177,7 +138,12 @@ func (pv *PolicyVerifier) getMaterialPayload(ctx context.Context, m *v12.Attesta
 	// Use the CAS to look for the material
 	var b bytes.Buffer
 	w := bufio.NewWriter(&b)
-	err := pv.cas.Download(ctx, w, m.GetArtifact().GetDigest())
+
+	client, err := pv.getCASClient(pb.CASCredentialsServiceGetRequest_ROLE_DOWNLOADER, m.GetArtifact().GetDigest())
+	if err != nil {
+		return nil, err
+	}
+	err = client.Download(ctx, w, m.GetArtifact().GetDigest())
```
migmartri, line 146: if we happen to do this validation in `attestation add` we would not need to download the file.

Downloading at push when the file is still there doesn't seem optimal and potentially problematic with big files.

pkg/policies/policies.go
```diff
@@ -85,68 +97,17 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
-	}, nil
-}
-
 // load the subject of the policy.
-func (pv *PolicyVerifier) loadSubject(ctx context.Context, attachment *v1.PolicyAttachment, spec *v1.Policy, state *v12.CraftingState) ([]byte, error) {
+func (pv *PolicyVerifier) loadSubject(ctx context.Context, attachment *v1.PolicyAttachment, spec *v1.Policy) ([]byte, error) {
+	state := pv.state
+
 	// Load the affected material or attestation, and checks if the expected name and type match
 	name := attachment.GetSelector().GetName()
```
migmartri, line 112: To me the name should be a filter between all the types that a policy affects to, not the differentiator between a material policy or an attestation policy. WDYT?

pkg/policies/policies.go
```diff
@@ -177,7 +138,12 @@ func (pv *PolicyVerifier) getMaterialPayload(ctx context.Context, m *v12.Attesta
 	// Use the CAS to look for the material
 	var b bytes.Buffer
 	w := bufio.NewWriter(&b)
-	err := pv.cas.Download(ctx, w, m.GetArtifact().GetDigest())
+
+	client, err := pv.getCASClient(pb.CASCredentialsServiceGetRequest_ROLE_DOWNLOADER, m.GetArtifact().GetDigest())
```
migmartri, line 142: are you handling when the materials are stored inline?

migmartri (commented):
Thanks @jiparis for the PR, looking promising. I've not finished the review yet but I want to leave some early thoughts.

- I am not convinced that running the material policies on push is a good idea, especially taking into account that we download the materials again. Why can't we run the policies on `att add` and store the violation in the crafting state?
- we should revisit the use of `name` in the attachment, the way it's implemented today will not allow it working with contractless materials. To me, the `kind` is in the policy and name would be an optional selector to scope down the result.

Let me know you thoughts, happy to chat tomorrow :)

internal/attestation/crafter/crafter_test.go
```diff
@@ -209,6 +220,24 @@ func (s *crafterSuite) TestLoadSchema() {
...
+			wantErr:      true,
+		},
+		{
+			name:         "missing script",
+			contractPath: "testdata/contracts/with_policy_missing_rego.yaml",
+			wantErr:      true,
+		},
+		{
+			name:         "rego policy",
+			contractPath: "testdata/contracts/with_rego.yaml",
```
migmartri, line 239: should this test somehow check the result? `want`?

internal/attestation/renderer/chainloop/chainloop.go
```diff
@@ -177,18 +165,18 @@ func ExtractPredicate(envelope *dsse.Envelope) (NormalizablePredicate, error) {
 	// 2 - Extract the Chainloop predicate from the in-toto statement
 	switch statement.PredicateType {
 	case PredicateTypeV02:
-		var predicate *ProvenancePredicateV02
+		var predicate ProvenancePredicateV02
```
migmartri, line 168: out of curiosity, why replacing the pointers?

internal/attestation/renderer/chainloop/v02.go
```diff
@@ -165,9 +169,43 @@ func (r *RendererV02) predicate() (*structpb.Struct, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error converting policies to values: %w", err)
+	}
+	// Store it in a ListValue struct.
+	predicate.Fields["policies"] = &structpb.Value{Kind: &structpb.Value_ListValue{ListValue: &structpb.ListValue{Values: policyValues}}}
+
 	return predicate, nil
 }
 
+func policiesToValues(policies []*v1.Policy) ([]*structpb.Value, error) {
```
migmartri, line 183: could be worse, I think you did pretty well here :)

migmartri (commented):
more feedback

internal/attestation/renderer/chainloop/v02.go
```diff
@@ -165,9 +169,43 @@ func (r *RendererV02) predicate() (*structpb.Struct, error) {
 		return nil, fmt.Errorf("error unmarshaling predicate: %w", err)
 	}
 
+	// Policies have oneofs, which are not compatible with the regular json.Marshal. We need to do it separately
+	policyValues, err := policiesToValues(r.att.Policies)
+	if err != nil {
+		return nil, fmt.Errorf("error converting policies to values: %w", err)
+	}
+	// Store it in a ListValue struct.
+	predicate.Fields["policies"] = &structpb.Value{Kind: &structpb.Value_ListValue{ListValue: &structpb.ListValue{Values: policyValues}}}
```
migmartri, line 178: do you need to store the result as structpb? I thought we had a custom predicate that we were mapping ourselves?

pkg/policies/policies.go
```diff
@@ -85,68 +97,17 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
 		pv.state.Attestation.Policies = append(pv.state.Attestation.Policies, &v12.Policy{
 			Name:       spec.Metadata.Name,
 			Attachment: policyAtt,
-			Body:       string(script.Source),
 			Violations: policyViolationsToAttestationViolations(res),
```
migmartri, line 100: we should probably also store when a policy was successfully ran not only when it's a violation, what do you think?

app/cli/internal/action/attestation_add.go
jiparis: Just `goimports --local`

app/cli/cmd/attestation_push.go
jiparis: We can't, because CAS connection is specific for a given digest, and here we wanted to download "any" material, so we need to create specific connections for each.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -144,3 +147,49 @@ message PolicyAttachment {
     string value = 2 [(buf.validate.field).required = true];;
   }
 }
+
+// Represents a policy to be applied to a material or attestation
+message Policy {
```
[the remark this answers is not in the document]
jiparis: To avoid import cycles. In particular `policy.proto` needed material types, and crafting_schema needed policies. We could extract the common part to a third proto, but that would require several unrelated changes throughout all the code.

internal/attestation/renderer/chainloop/v02.go
jiparis: Just because we have some `oneof`s in the policies, which is incompatible with structpb conversion.

pkg/policies/policies.go
jiparis: In that case `Violations` will be empty, but the policy will still be there.

pkg/policies/policies.go
jiparis: I like it. We'll give priority to the type, and then filter by name if it's provided. I'll send another revision of the PR.

pkg/policies/policies.go
jiparis: Yes, in the policies logic, it checks if the material is inline, and in that case, it retrieves the content from the `Content` field.

pkg/policies/policies.go
jiparis: Yes, I agree.

pkg/policies/policies.go
jiparis: It needs the digest, which depends on each material. We might want to revisit that restriction, but currently we must create a new client for each.

jiparis:
Please put reviews on hold, I'll send another revision with requested changes.

internal/attestation/renderer/chainloop/chainloop.go
jiparis: I think at some point during my changes it needed to be preallocated. Not sure if that's still the case.

internal/attestation/crafter/materials/materials.go
```diff
@@ -207,3 +210,30 @@ func Craft(ctx context.Context, materialSchema *schemaapi.CraftingSchema_Materia
 
 	return m, nil
 }
+
+// LoadJSONBytes Extracts raw data in JSON format from different sources, i.e cue or yaml files
+func LoadJSONBytes(rawData []byte, extension string) ([]byte, error) {
```
jiparis, line 215: I've moved this to avoid import loops.

app/cli/cmd/attestation_push.go
```diff
@@ -19,10 +19,11 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
```
migmartri, line 26: These changes make the PR very noisy, you mentioned that you use `goimports` but I can not reproduce the change locally with latest version of goimports.

Are you sure you do not have some additional configuration in the editor?

```
# From main
goimports -w app/cli/cmd/attestation_push.go 
git status
[empty]
```

I think we need to figure this out before we merge, otherwise we'll be doing back and forth continuously

jiparis: Ok, let me undo these accidental changes, since they are not related.

internal/attestation/crafter/crafter_test.go
jiparis: Done!

internal/attestation/crafter/testdata/contracts/with_policy_embedded.yaml
```diff
@@ -0,0 +1,4 @@
+schemaVersion: "v1"
+policies:
+  attestation:
+    - ref: testdata/contracts/policy_embedded.yaml
```
migmartri, line 4: is this embedded?

internal/attestation/crafter/testdata/contracts/with_rego.yaml
```diff
@@ -0,0 +1,4 @@
+schemaVersion: "v1"
+policies:
+  attestation:
```
migmartri, line 3: in the future we should also have some examples with `material` attachments, all of them in this directory seem to be attestation.

internal/attestation/crafter/crafter.go
```diff
@@ -212,9 +184,32 @@ func LoadSchema(pathOrURI string) (*schemaapi.CraftingSchema, error) {
 		return nil, err
 	}
 
+	// Load, validate policies, and embed them in the schema
+	if err := validateAndPreloadPolicies(schema.GetPolicies().GetMaterials()); err != nil {
```
migmartri, line 188: Isn't this embedding the policy which we said we didn't want to do?

internal/attestation/crafter/crafter.go
```diff
@@ -577,13 +572,20 @@ func (c *Crafter) addMaterial(ctx context.Context, m *schemaapi.CraftingSchema_M
 		return fmt.Errorf("validation error: %w", err)
 	}
 
+	// Validate policies
+	pv := policies.NewPolicyVerifier(c.CraftingState.InputSchema, c.logger)
+	policyResults, err := pv.VerifyMaterial(ctx, mt, value)
```
migmartri, line 577: This is very clear now, thanks :)

pkg/policies/policies.go
```diff
@@ -16,61 +16,111 @@
...
 type PolicyVerifier struct {
-	state *v12.CraftingState
-	cas   casclient.Downloader
+	schema *v1.CraftingSchema
+	logger *zerolog.Logger
 }
 
-func NewPolicyVerifier(state *v12.CraftingState, client casclient.Downloader) *PolicyVerifier {
+func NewPolicyVerifier(schema *v1.CraftingSchema, logger *zerolog.Logger) *PolicyVerifier {
 	// only Rego engine is currently supported
```
migmartri, line 47: does this comment make sense anymore? I can't see any reference to engine

pkg/policies/policies.go
```diff
@@ -16,61 +16,111 @@
...
-			// policy is disabled
-			// TODO: WARN.
-			continue
+// VerifyMaterial applies all required policies to a material
+func (pv *PolicyVerifier) VerifyMaterial(ctx context.Context, material *v12.Attestation_Material, artifactPath string) ([]*v12.Policy, error) {
+	result := make([]*v12.Policy, 0)
+	policies, err := pv.requiredPoliciesForMaterial(material)
+	if err != nil {
+		return nil, fmt.Errorf("error getting required policies for material: %w", err)
+	}
```
migmartri, line 57: nitpick, add empty line

pkg/policies/policies.go
```diff
@@ -16,61 +16,111 @@
...
-	violations := make([]*engine.PolicyViolation, 0)
-	policies := pv.state.GetInputSchema().GetPolicies()
-	for _, policyAtt := range policies {
-		if policyAtt.Disabled {
-			// policy is disabled
-			// TODO: WARN.
-			continue
+// VerifyMaterial applies all required policies to a material
+func (pv *PolicyVerifier) VerifyMaterial(ctx context.Context, material *v12.Attestation_Material, artifactPath string) ([]*v12.Policy, error) {
+	result := make([]*v12.Policy, 0)
```
migmartri, line 53: I have a problem with the naming here, it seems that the term `Policy` is being overused for both the actual policy and for evaluations of the policy on a material for example.

Wouldn't make sense to change the name of this one to `PolicyEvaluation`?

pkg/policies/policies.go
```diff
@@ -79,29 +129,131 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
+		_, err = base64.StdEncoding.Decode(rawMaterial, []byte(envelope.Payload))
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode attestation material: %w", err)
+		}
+	}
+
+	return rawMaterial, nil
+}
+
+func (pv *PolicyVerifier) requiredPoliciesForMaterial(material *v12.Attestation_Material) ([]*v1.PolicyAttachment, error) {
```
migmartri, line 188: Could you add a comment on what this method does? I mean, the different logic paths it goes through.

In another patch, it would be great to test this method so it's used as documentation too.

pkg/policies/policies.go
```diff
@@ -79,29 +129,131 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
+}
+
+func getMaterialContent(material *v12.Attestation_Material, artifactPath string) ([]byte, error) {
+	var rawMaterial []byte
+	var err error
+
+	// nolint: gocritic
+	if material.InlineCas {
+		rawMaterial = material.GetArtifact().GetContent()
+	} else if artifactPath == "" {
```
migmartri, line 163: nitpick. I find easier to read separated statements

```
if artifactPath == "" {
return nil, errors...
}

```
rawMaterial, err = os.ReadFile(artifactPath)
		if err != nil {
			return nil, fmt.Errorf("failed to read material content: %w", err)
		}
```

File: pkg/policies/policies.go [hunk 25]
```diff
@@ -79,29 +129,131 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
+		return nil, errors.New("artifact path required")
+	} else {
+		// read content from local filesystem
+		rawMaterial, err = os.ReadFile(artifactPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read material content: %w", err)
+		}
+	}
+	// special case for ATTESTATION materials, the statement needs to be extracted from the dsse wrapper.
+	if material.MaterialType == v1.CraftingSchema_Material_ATTESTATION {
```
migmartri, line 173: Good catch

pkg/policies/policies.go
```diff
@@ -79,29 +129,131 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
+		var envelope dsse.Envelope
+		if err := json.Unmarshal(rawMaterial, &envelope); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal attestation material: %w", err)
+		}
+
+		_, err = base64.StdEncoding.Decode(rawMaterial, []byte(envelope.Payload))
```
migmartri, line 179: can you double check that we do not have a method that does this already? I remember creating a common method for marshalling/unmarshalling from JSON to Statement

pkg/policies/policies.go
```diff
@@ -16,61 +16,111 @@
...
+		subject, err := getMaterialContent(material, artifactPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load material content: %w", err)
+		}
+
+		pv.logger.Info().Msgf("evaluating policy '%s' against material '%s'", spec.Metadata.Name, material.GetArtifact().GetId())
+
+		// verify the policy
+		ng := getPolicyEngine(spec)
+		res, err := ng.Verify(ctx, script, subject)
```
migmartri, line 81: nitpick, instead of `res` call it violations?

pkg/policies/policies.go
```diff
@@ -79,29 +129,131 @@ func (pv *PolicyVerifier) Verify(ctx context.Context) ([]*engine.PolicyViolation
...
+		pvs = append(pvs, &v12.Policy_Violation{
+			Subject: violation.Subject,
+			Message: violation.Violation,
+		})
+	}
+	return
+}
+
+// LoadPolicySpec loads and validates a policy spec from a contract
+func LoadPolicySpec(attachment *v1.PolicyAttachment) (*v1.Policy, error) {
```
migmartri, line 242: I like it! 

Some testing later on would be ideal

internal/attestation/renderer/chainloop/v02.go
```diff
@@ -38,6 +37,8 @@ const PredicateTypeV02 = "chainloop.dev/attestation/v0.2"
 type ProvenancePredicateV02 struct {
 	*ProvenancePredicateCommon
 	Materials []*intoto.ResourceDescriptor `json:"materials,omitempty"`
+	// Map materials and policies
+	Policies map[string][]*v1.Policy `json:"policies,omitempty"`
```
migmartri, line 41: as I mentioned, `Policy` might not be a good name here. PolicyEval?

internal/attestation/renderer/renderer.go
```diff
@@ -134,6 +152,38 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
 	return &dsseEnvelope, nil
 }
 
+func addPolicyResults(statement *intoto.Statement, policyResults []*v1.Policy) error {
```
migmartri, line 155: I'd add a comment explaining why this method needs to leverage structpb and so on

migmartri (approved):
Very nice!

Approving modulo making sure that we are not embedding the policies on contract create/update

internal/attestation/crafter/crafter.go
jiparis: I removed the embedding logic, but I should change the method name 😞 . Let me change it.

internal/attestation/crafter/testdata/contracts/policy_embedded.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: workflowcontract.chainloop.dev/v1
+kind: Policy
```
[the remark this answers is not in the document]
jiparis: makes sense, I'll do it.

pkg/policies/policies.go
jiparis: Makes all sense. I'll create a specific struct for this.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -144,3 +155,50 @@ message PolicyAttachment {
...
+
+    // embedded source code (only Rego supported currently)
+    string embedded = 2;
+
+    option (buf.validate.oneof).required = true;
+  };
+
+  // stage at which this policy will be run.
+  // Only "push" is supported currently and this field will be ignored
+  PolicyStage stage = 3;
```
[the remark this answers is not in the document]
jiparis: Right, it's not yet implemented.

internal/attestation/crafter/testdata/contracts/with_policy_embedded.yaml
jiparis: The code in the policy, not the policy in the contract.

pkg/policies/policies.go
jiparis: Yes, it's tested thoroughly in policies_test.go. I'll add the comment.

internal/attestation/renderer/renderer.go
```diff
@@ -134,6 +152,48 @@ func (ab *AttestationRenderer) Render() (*dsse.Envelope, error) {
...
+	err = json.Unmarshal(jsonPredicate, &p)
+	if err != nil {
+		return fmt.Errorf("unmarshalling predicate: %w", err)
+	}
+
+	// insert policy evaluations
+	if p.PolicyEvaluations == nil {
+		p.PolicyEvaluations = make(map[string][]*v1.PolicyEvaluation)
+	}
+	p.PolicyEvaluations["ATTESTATION"] = policyResults
```
jotadrilo, line 177: You should use the string representation of the MaterialType enum to avoid future inconsistencies

```suggestion
	p.PolicyEvaluations[schemaapi.CraftingSchema_Material_ATTESTATION.String()] = policyResults
```

jiparis: Ey, thanks for the suggestion @jotadrilo . But in this case it's not a material, but policies applied to the whole attestation. I know the naming is confusing, we might choose a different key for this particular case. I'll give it a try.

pkg/policies/policies.go
```diff
@@ -177,7 +177,7 @@ func getMaterialContent(material *v12.Attestation_Material, artifactPath string)
 			return nil, fmt.Errorf("failed to unmarshal attestation material: %w", err)
 		}
 
-		_, err = base64.StdEncoding.Decode(rawMaterial, []byte(envelope.Payload))
+		rawMaterial, err = envelope.DecodeB64Payload()
```
migmartri, line 180: told you :)

internal/attestation/renderer/renderer.go
jotadrilo: But you are mapping materials and policies, right? In other groups, you are grouping policies by their material name.

If this specific grouping has a different semantic, I'd better use a different field in the in-toto predicate (v2), or add a reserved MaterialType name for this kind of grouping.

pkg/policies/policies.go
jiparis: Yes! And it also implements URLEncoding which is is supported by DSSE in addition to StdEncoding.

internal/attestation/renderer/renderer.go
jiparis: Yes, we were evaluating those options. We'll revisit this in next iterations, as it might be part of a batch with other improvements. Thanks!

PR #1094, opened 2024-07-15
feat(orgmetrics): Allow to optionally pass time window to queries

javirln:
This patch allows to optionally pass a `TimeWindow` struct to the `OrgMetricsRepository` in order to limit the results. This is useful since it's possible we want to get all data from DB rather than scoping them within a time window.

### DIFF
--- a/app/controlplane/pkg/biz/orgmetrics.go
+++ b/app/controlplane/pkg/biz/orgmetrics.go
@@ -17,7 +17,6 @@ package biz
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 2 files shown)


migmartri:
Do we still need this change after our conversation yesterday? 

I am a little bit concerned about opening the time windows in these kind of operations that are costly

jiparis:
Is this really needed? do we want to potentially list all existing workflow runs for an organization? I think some sort of pagination should be forced in all endpoints.

PR #1095, opened 2024-07-16
feat(api-token): Add org_name to token payload

javirln:
This patch adds the `org_name` to the API Tokens payload for an easy identification along with the `org_id`. 
Please note the new field is not used anywhere yet.

```shell
$ export CHAINLOOP_TOKEN=$OLD_TOKEN
$ go run app/cli/main.go --insecure wf run ls                                                               
there are no workflow runs yet
```

```shell
$ export CHAINLOOP_TOKEN=$NEW_TOKEN
$ go run app/cli/main.go --insecure wf run ls                                                               
there are no workflow runs yet
```

Refs: PFM-713


app/controlplane/pkg/biz/apitoken.go
```diff
@@ -135,8 +139,14 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description
 		return nil, fmt.Errorf("storing token: %w", err)
 	}
 
+	// Retrieve the organization
+	org, err := uc.orgUseCase.FindByID(ctx, orgID)
```
migmartri, line 143: You should move this above line 134 so you do not create a token if the org can't be found.

migmartri (approved):
Thanks,

I'd rather that next time you split these kind of PRs into parts.

- Add name to token
- Send it in telemetry

Imagine that there is smth wrong with the telemetry changes and we'll not be able to rollback in this case since we might have some tokens with wrong state (unknown name property)

javirln:
Pushed new version removing all telemetry changes and tackling feedback.

migmartri (approved):

Issue #1098, opened 2024-07-17
Prometheus collector


javirln:
The idea of the issue is for Chainloop to have an authenticated endpoint where users can retrieve Prometheus metrics about their organizations.

One base collector with a defined set of metrics that, for the first version, runs every time there is a request to the endpoint and gather the metrics. It needs to be authenticated with a valid Chainloop API Token.

Endpoint: `/prom/{org_name}/metrics`.

This collector will potentially be the default collector for all Prometheus registries defined by the Chainloop's business logic to expose metrics to customers, that is something like the following:
```golang
func (rm *ChainloopRegistryManager) AddRegistry(reg *ChainloopPrometheusRegistry) {
	collector := metrics.NewBaseChainloopCollector(reg.ID, rm.orgmetrics)

	reg.MustRegister(collector)
	rm.Registries = append(rm.Registries, reg)
}
```
In the above example there is the concept of RegistryManager which is in charged of:

- Keep track of all Prometheus registries being created for users within the lifecycle of the control plane.
- Register new Prometheus integrations by periodically checking the Integrations table assigning them the custom Chainloop collector.

migmartri:
@javirln could you please create a subtask about making sure permissions for pulling data from prometheus is mapped in casbin and checked in the endpoint?

Thanks!

PR #1099, opened 2024-07-17
feat(http): Create jwt middlewares for http servers

javirln:
This patch adds two jwt middleware to validate JWT tokens in http servers, one for tokens that are part of a query param and other from authorization headers.

Refs #1098

### DIFF
--- a/pkg/middlewares/http/jwt.go
+++ b/pkg/middlewares/http/jwt.go
@@ -1,0 +1,125 @@
+//
+// Copyright 2024 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/middlewares/http/jwt.go
```diff
@@ -0,0 +1,125 @@
...
+// AuthFromQueryParam is a middleware that extracts the token from the query parameter and verifies it
+func AuthFromQueryParam(keyFunc jwt.Keyfunc, claimsFunc ClaimsFunc, signingMethod jwt.SigningMethod, next nhttp.Handler) nhttp.Handler {
+	return nhttp.HandlerFunc(func(w http.ResponseWriter, r *nhttp.Request) {
+		token := r.URL.Query().Get("t")
+		if token == "" {
+			nhttp.Error(w, "missing token", nhttp.StatusUnauthorized)
+			return
+		}
+
+		claims, err := verifyAndMarshalJWT(token, keyFunc, claimsFunc, signingMethod)
```
migmartri, line 47: it feels like all the code from here down could be extracted to a common function?

The only difference between the two middleware is from where it gets the token from.

migmartri (approved):

PR #1106, opened 2024-07-18
feat(policies): document policies

jiparis:
This PR adds a new documentation page, as well as some policy examples to play with.

Refs #122

### DIFF
--- a/docs/docs/reference/policies.md
+++ b/docs/docs/reference/policies.md
@@ -1,0 +1,103 @@
+---
+title: Policies
+---
+
+Starting with Chainloop [0.93.8](https://github.com/chainloop-dev/chainloop/releases/tag/v0.93.8), operators can attach policies to contracts. 
+These policies will be evaluated against all attestation metadata and materials in any workflow with that contract.
+
+### Policy specification
+A policy can be defined in a YAML document, like this:
+```yaml
+# sbom-licenses.yaml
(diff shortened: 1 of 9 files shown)


docs/docs/reference/policies.md
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+schemaVersion: v1
+materials:
+  - name: sbom
+    type: SBOM_CYCLONEDX_JSON
+  - name: another-sbom
+    type: SBOM_CYCLONEDX_JSON
+policies:
+  materials: # policies applied to materials
+    - ref: sbom-licenses.yaml # (1)
```
migmartri, line 56: It's not clear if this is a local file that needs to be somewhere or a name? Did we end up adding file://... to the ref?

migmartri (approved):
I like it.

The part that seems handwavy and potentially missing is the different kind of references a policy attachment can have.

i.e how to point to a local policy and or a remote policy.

It might make sense to add an example of a remote policy that points to one of the policies you added as well in this patch :)

docs/docs/reference/policies.md
jiparis: It's using the `blob.LoadFileOrURL` method from cosign, which defaults to the local filesystem if no `http://`, `https://` or `env://` is provided.

jiparis: I'll make sure it's clear in the docs.

PR #1113, opened 2024-07-18
feat(chart): Change template to allow prometheus integration config

javirln:
This patch allows to include a set of organizations as part of the configuration to activate the Prometheus integration:

Refs https://github.com/chainloop-dev/chainloop/issues/1098

### DIFF
--- a/deployment/chainloop/templates/controlplane/config.configmap.yaml
+++ b/deployment/chainloop/templates/controlplane/config.configmap.yaml
@@ -44,7 +44,7 @@ data:
     onboarding:
       {{- toYaml .Values.controlplane.onboarding | nindent 6 }}
     {{- end }}
-    {{ if .Values.controlplane.prometheus_integration }}
+    {{ if .Values.controlplane.prometheus_org_metrics }}
     prometheus_integration:
-      {{- toYaml .Values.controlplane.prometheus_integration | nindent 6 }}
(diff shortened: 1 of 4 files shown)


deployment/chainloop/values.yaml
```diff
@@ -144,6 +144,10 @@ controlplane:
   #    - name: "read-only-demo"
   #      role: "MEMBERSHIP_ROLE_ORG_VIEWER"
 
+  ## @extra controlplane.prometheus_integration.org_name Name of the organization to integrate with Prometheus
```
migmartri, line 147: this documentation doesnt seem right. org_name is a property of an object in the array.

I'd just document the parent one 

> @extra controlplane.prometheus_integration Expose prometheus metrics for your organization

deployment/chainloop/templates/controlplane/config.configmap.yaml
```diff
@@ -43,4 +43,8 @@ data:
     {{ if .Values.controlplane.onboarding }}
     onboarding:
       {{- toYaml .Values.controlplane.onboarding | nindent 6 }}
-    {{- end }}
\ No newline at end of file
+    {{- end }}
+    {{ if .Values.controlplane.prometheus_integration }}
+    prometheus_integration:
```
migmartri, line 48: prometheus_integration sound very generic, this could mean anything.

What about `prometheus_org_metrics`

migmartri (approved):

PR #1132, opened 2024-07-24
chore(dagger): update module to dagger v0.12

jpadams:
- Update engine version `to v0.12.2`
- internal `import "main/internal/dagger"`
- Update types: `X` -> `dagger.X`
- Update `Debug()` signature since `Terminal()` now returns `dagger.Container`: https://github.com/dagger/dagger-go-sdk/commit/b15a9be1ce2844276f36f6d24f95070b38cdebd3


migmartri (approved):
Thanks a lot @jpadams!

We'll upgrade the chainloop CLI inside and re-release the module in another PR.

migmartri:
Hi @jpadams, it seems that the linter its failing because it loads a previous version of the engine that doesn't support go 1.22.

Could you please update this line https://github.com/chainloop-dev/chainloop/blob/103742fb98c26cc4ff9a9fb1cc2a900304c12e1a/.github/workflows/lint.yml#L63 to include `DAGGER_VERSION=0.12.2` instead of `0.11.0`

Thanks!

migmartri (changes requested):
See my previous comment on updating the dagger version using our linter in the CI.

migmartri (approved):
Thank you!

Issue #973, opened 2024-06-18
bug: remote state attestations seems to suffer from race conditions


migmartri:
I've managed to run into race conditions while running an attestation that leverages remote state (`--remote-state`) with parallel. The result was that some items had not been added to the final attestation.

migmartri:
I'd suggest to look into https://en.wikipedia.org/wiki/Optimistic_concurrency_control where

- get state API endpoint will also return the `sha` of the content
- once the client wants to update will send that original `sha`
- the server, in a transaction before updating it will make sure it's the same one, otherwise will reject the update
- the client will need to handle this case and retry accordingly

PR #1133, opened 2024-07-24
feat: add support for optimistic concurrency check

migmartri:
This is the first step (API, biz, data layers) to enable optimistic concurrency in a remote state. It contains:

- The state API now returns the digest of the current state.
- This digest needs to be sent back during an update; if it mismatches with the content stored, it will return a known conflict error.
- This new error will be handled in another PR by the CLI, which will retry accordingly.

This digest check happens in a DB transaction now, including the use of `SELECT FOR UPDATE`, which required enabling locking in `ent`, which is the reason of the `data/ent` changes in this PR. 

Example of read response.

image

Write also supports an optional `baseDigest` that will be used for checking

ref #973


app/controlplane/api/controlplane/v1/attestation_state.proto
```diff
@@ -17,18 +17,19 @@ syntax = "proto3";
 
 package controlplane.v1;
 
-option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1;v1";
-
-import "buf/validate/validate.proto";
 import "attestation/v1/crafting_state.proto";
+import "buf/validate/validate.proto";
```
migmartri, line 21: buf format seems to make these changes now, I am not sure why

app/controlplane/pkg/data/attestationstate.go
```diff
@@ -48,26 +51,69 @@ func (r *AttestationStateRepo) Initialized(ctx context.Context, runID uuid.UUID)
 	return exists, nil
 }
 
-func (r *AttestationStateRepo) Save(ctx context.Context, runID uuid.UUID, state []byte) error {
-	err := r.data.DB.WorkflowRun.UpdateOneID(runID).SetAttestationState(state).Exec(ctx)
+// baseDigest, when provided will be used to check that it matches the digest of the state currently in the DB
+// if the digests do not match, the state has been modified and the caller should retry
+func (r *AttestationStateRepo) Save(ctx context.Context, runID uuid.UUID, state []byte, baseDigest string) (err error) {
```
migmartri, line 56: this method has the meat of the implementation, it makes sure that 

in a transaction

- Reads the state stored in DB and blocks the row for update
- Calculates the sha of its content and compares it with the baseDigest optionally provided
- if it doesn't match, it returns a ConflictError and cancels the transaction, unblocking the row.
- If it does, we update the column and unblock the row

app/controlplane/api/controlplane/v1/attestation_state.proto
```diff
@@ -68,3 +74,8 @@ message AttestationStateServiceResetRequest {
 
 message AttestationStateServiceResetResponse {}
 
+enum AttestationStateError {
+  ATTESTATION_STATE_ERROR_UNSPECIFIED = 0;
+  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409
+  ATTESTATION_STATE_ERROR_CONFLICT = 1 [(errors.code) = 409];
```
jiparis, line 80: Nice! we can set error codes in the spec!

app/controlplane/pkg/data/attestationstate.go
```diff
@@ -48,26 +51,69 @@ func (r *AttestationStateRepo) Initialized(ctx context.Context, runID uuid.UUID)
...
+
+	defer func() {
+		// Unblock the row if there was an error
+		if err != nil {
+			_ = tx.Rollback()
+		}
+	}()
+
+	// compared the provided digest with the digest of the state in the DB
+	// TODO: make digest check mandatory on updates
```
jiparis, line 70: I guess we can do this when all clients have been updated, right?

jiparis (approved):

app/controlplane/pkg/data/attestationstate.go
migmartri: Correct, for now it's not required.

PR #1135, opened 2024-07-25
chore(controlplane): move config proto to public package

### LINKED ISSUE #1134: Move all .proto files to public API package
The goal is to move all remaining proto specs to API packages, if they are intended to be consumed externally.

jiparis:
Move this .proto and generated code to the `api` package, since other public packages depend on this one. This way, third party chainloop integrations can fully use the API.

Closes #1134


app/controlplane/api/controlplane/v1/conf.proto
```diff
@@ -22,7 +22,7 @@ import "controlplane/v1/response_messages.proto";
 import "credentials/v1/config.proto";
 import "google/protobuf/duration.proto";
 
-option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/internal/conf/controlplane/config/v1;conf";
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1;v1";
 
 message Bootstrap {
```
migmartri, line 27: I am not convinced about moving the whole file, the reason been that `Bootstrap` inside is specifically about this app.

What we can do is to move the required sub-structs (like we did for example with credentials).

jiparis: Ok, I'll check what are those required substructs.

jiparis: I've finally moved only OnboardingSpec message, which is set as a dependency for some use cases.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -168,19 +168,6 @@ message CA {
   }
 }
 
-// OnboardingSpec is a configuration to automatically onboard users in organizations with specific roles
-message OnboardingSpec {
-  // Name of the organization
-  string name = 1 [(buf.validate.field).string.min_len = 1];
-  // Role to assign to the user
-  controlplane.v1.MembershipRole role = 2 [
```
migmartri, line 176: I think in reality what you want to extract is membership role?

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -61,7 +61,7 @@ message Bootstrap {
   }
 
   // Configuration for onboarding users in organizations with specific roles
-  repeated OnboardingSpec onboarding = 10;
+  repeated onboarding_spec.v1.OnboardingSpec onboarding = 10;
```
migmartri, line 64: I do not think that underscore in package names are go idiomatic.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -168,19 +168,6 @@ message CA {
   }
 }
 
-// OnboardingSpec is a configuration to automatically onboard users in organizations with specific roles
-message OnboardingSpec {
```
migmartri, line 172: onboarding spec is still very specific from controlplane app, what is the minimum common denomitator that needs to be used somewhere else?

app/controlplane/pkg/conf/controlplane/config/v1/config.proto
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+syntax = "proto3";
+package controlplane.config.v1;
```
migmartri, line 18: you probably need to place the buf.yaml to `conf` not o `conf/controlplane`

jiparis: Yes, that did the trick.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -61,7 +61,7 @@ message Bootstrap {
   }
 
   // Configuration for onboarding users in organizations with specific roles
-  repeated OnboardingSpec onboarding = 10;
+  repeated config.v1.OnboardingSpec onboarding = 10;
```
migmartri, line 64: out of curiosity, do you need to import or will be considered the same package?

migmartri (approved):
Thanks

I am curious if you still need the import or at compilation proto considers all the files from the same package regardless of the location.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
jiparis: Yes, it's needed. I've tried removing the import and it doesn't work with and without `config.v1.` prefix. It seems they are not considered the same package after all.

jiparis: However, it works leaving the import and removing the prefix. That makes more sense to me. They are considered the same package, but it needs some help finding the new message definitions.

Issue #1151, opened 2024-07-29
standardize Helm Chart with Bitnami Practices 


migmartri:
From [https://github.com/bitnami/charts/pull/27100#issuecomment-2256043379](https://github.com/bitnami/charts/pull/27100#issuecomment-2256043379)

NOTE: It's essential to evaluate and discuss each breaking change we might add to the chart.
```[tasklist]
### Tasks
- [x] Openshift settings
- [x] hardcoded ports in configmap/secrets
- [x] auto-scaling
- [x] network policies
- [ ] tls.existingSecret? https://github.com/bitnami/charts/pull/27100#discussion_r1705381329
- [x] fix hpa i.e  averageUtilization: {{ .Values.controlplane.autoscaling.hpa.targetMemory }}
- [x] hardcode container names "We usually hardcode a container name, in this case we could use 'controlplane'"
```

migmartri:
So the chart has now been merged in Bitnami. Awesome! Great work @javirln!

I believe the last batch of feedback included some missing features in our chart that were applied in their branch, but not in ours. I think some of those changes should be merged also in our chart. @javirln do you have any suggestion on which ones should we add back?

PR #1155, opened 2024-07-29
feat(deployment): run development mode with dex

migmartri:
removes the requirement of providing oidc settings to deployment in development mode.

Now that chart will take either the provided oidc settings or dex settings if we are running in development

Development mode, no settings

```
helm template chainloop . --set development=true | grep oidc -A 3 
      oidc:    
        domain: "http://chainloop-dex:5556/dex"
        clientID: "chainloop-dev"
        clientSecret: "ZXhhbXBsZS1hcHAtc2VjcmV0"

``` 

development mode with overrides

```
helm template chainloop . \
  --set controlplane.auth.oidc.url=https://chainloop-local.us.auth0.com/ \
  --set development=true | grep -A 3 oidc 
      oidc:    
        domain: "https://chainloop-local.us.auth0.com/"
        clientID: "chainloop-dev"
        clientSecret: "ZXhhbXBsZS1hcHAtc2VjcmV0"
```

and standard mode will require the oauth options as usual

Note, there are some removals in the config settings because there was an option not been used.

ref #1151


app/controlplane/configs/config.devel.yaml
```diff
@@ -56,7 +56,6 @@ data:
 auth:
   oidc:
     domain: "http://0.0.0.0:5556/dex"
-    redirect_url_scheme: "http"
```
migmartri, line 59: this option was not being used

javirln (approved):

PR #1167, opened 2024-07-30
chore(prometheus): update histogram variable reference

migmartri:
This patch ensures that the histogram metric is a local reference not a shared one.

### DIFF
--- a/app/controlplane/pkg/metrics/prometheus/manager.go
+++ b/app/controlplane/pkg/metrics/prometheus/manager.go
@@ -20,26 +20,26 @@ import (
 )
 
 type ChainloopRegistryManager struct {
-	Registries map[string]*registry.PrometheusRegistry
+	registries map[string]*registry.PrometheusRegistry
 }
 
 func NewChainloopRegistryManager() *ChainloopRegistryManager {
 	return &ChainloopRegistryManager{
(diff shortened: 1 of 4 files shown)


app/controlplane/pkg/metrics/prometheus/manager.go
```diff
@@ -20,26 +20,26 @@ import (
 )
 
 type ChainloopRegistryManager struct {
-	Registries map[string]*registry.PrometheusRegistry
```
migmartri, line 23: while testing, I realized that this property didn't need to be exposed.

app/controlplane/pkg/metrics/prometheus/registry/registry.go
```diff
@@ -49,6 +42,14 @@ func NewPrometheusRegistry(name string, gatherer collector.ChainloopMetricsGathe
 
 	reg.MustRegister(bcc)
 
+	// It's important to register this variable in the factory since it's a different one per registry
+	workflowRunDurationSeconds := prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
migmartri, line 46: this is the main fix of the PR

javirln (approved):

jiparis (approved):

PR #1168, opened 2024-07-31
fix(grpc): Bump golang grcp dependency

javirln:
This patch bumps the version of golang grpc `v1.62.1` to `v1.64.1` to fix: https://github.com/chainloop-dev/chainloop/security/dependabot/55


javirln:
Checking the changes on the release page of google.com/grpc I cannot see any breaking change.

PR #1176, opened 2024-08-01
fix(dagger): Bump google.com/grpc dependency

javirln:
This patch fixes the dependency of google.com/grpc to a patched version.

`dagger develop` and manual change on `go.mod` has been performed.

This fixes: https://github.com/chainloop-dev/chainloop/security/dependabot/55

### DIFF
--- a/dagger.json
+++ b/dagger.json
@@ -2,5 +2,5 @@
   "name": "chainloop",
   "sdk": "go",
   "source": "extras/dagger",
-  "engineVersion": "v0.12.2"
+  "engineVersion": "v0.12.3"
 }
--- a/extras/dagger/go.mod
+++ b/extras/dagger/go.mod
@@ -42,5 +42,5 @@ require (
 	go.opentelemetry.io/otel/trace v1.27.0
 	golang.org/x/exp v0.0.0-20231110203233-9a3e6036ecaa
 	golang.org/x/sync v0.7.0
-	google.golang.org/grpc v1.64.0
+	google.golang.org/grpc v1.64.1
 )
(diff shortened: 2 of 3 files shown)


dagger.json
```diff
@@ -2,5 +2,5 @@
   "name": "chainloop",
   "sdk": "go",
   "source": "extras/dagger",
-  "engineVersion": "v0.12.2"
+  "engineVersion": "v0.11.6"
```
migmartri, line 5: this is a downgrade

migmartri (changes requested):

dagger.json
migmartri: your dagger client is probably outdated, and dagger develop made that change.

javirln: Let me try upgrade my dagger installation and see if that's the issue.

migmartri:
> manual change on go.mod has been performed.

What do you mean by manual? `go get` ?

javirln:
> > manual change on go.mod has been performed.
> 
> What do you mean by manual? `go get` ?

Changing the value from 1.64.0 to 1.64.1

dagger.json
javirln: Yep, that was the issue, I had the installation outdated.

migmartri:
> > > manual change on go.mod has been performed.
> > 
> > 
> > What do you mean by manual? `go get` ?
> 
> Changing the value from 1.64.0 to 1.64.1

why manually and not `go get`. I am genuinely curious since `go get` my also calculate transitive dependency updates or am I wrong?

javirln:
> > > > manual change on go.mod has been performed.
> > > 
> > > 
> > > What do you mean by manual? `go get` ?
> > 
> > 
> > Changing the value from 1.64.0 to 1.64.1
> 
> why manually and not `go get`. I am genuinely curious since `go get` my also calculate transitive dependency updates or am I wrong?

I mean not strong reason, let me do it using go get to see if it creates new transition dependencies.

PR #1181, opened 2024-08-01
fix(chart): Fixes OIDC default configuration

javirln:
This patch fixes the helper function that injects the Dex configuration.

#1151 https://github.com/bitnami/charts/pull/27100


migmartri (approved):
Good catch, sorry about that 🤦🏼

Issue #1178, opened 2024-08-01
Add material name validation when attesting


javirln:
Now end users can provide an optional name to the materials being attested. We need to ensure the material names comply with our name constrains and that they are validated.

This didn't happen before since the material name were defined on the contracts that indeed had validations and the ones being attested, are two different types.

PR #1184, opened 2024-08-02
feat(attestation): Add name constraints to attestation materials name

javirln:
This patch adds the needed validations to the material's names so it coherent with the rest of names in resources.

Before:
```bash
$ chainloop --insecure attestation add --value ./internal/attestation/crafter/materials/testdata/valid-chart.tgz --name aWDAWDawd
INF material kind detected kind=HELM_CHART
```
Now:
```bash
$ chainloop --insecure attestation add --value ./internal/attestation/crafter/materials/testdata/valid-chart.tgz --name aWDAWDawd
ERR adding material: failed to auto-discover material kind: validating material: validation error:
 - name: must contain only lowercase letters, numbers, and hyphens. [name.dns-1123]
```

https://github.com/chainloop-dev/chainloop/issues/1178


migmartri (approved):
thanks!

PR #1192, opened 2024-08-05
feat(chart): Document Dex users on README.md

javirln:
This patch updates the documentation so the pre configured static users on development mode are written on the README.md.

#1151 https://github.com/bitnami/charts/pull/27100


migmartri (approved):
thanks

PR #1204, opened 2024-08-12
feat(chart): Modify TLS config param

javirln:
This patch modifies the name of the TLS config param.

Diff:
```diff
diff --git a/old.yml b/new.yml
index 7bb317e..bb56aef 100644
--- a/old.yml	
+++ b/new.yml	
@@ -441,7 +441,7 @@ metadata:
 type: Opaque
 data:
   # We store it also as a different key so it can be reused during upgrades by the common.secrets.passwords.manage helper
-  generated_jws_hmac_secret: "dkNTVk5SaTVrWg=="
+  generated_jws_hmac_secret: "MWh5U2p1eVVnVw=="
   db_migrate_source: "cG9zdGdyZXM6Ly9jaGFpbmxvb3A6Y2hhaW5sb29wcHdkQGNoYWlubG9vcC1wb3N0Z3Jlc3FsOjU0MzIvY2hhaW5sb29wLWNwP3NzbG1vZGU9ZGlzYWJsZQ=="
 stringData:
   config.secret.yaml: |
@@ -465,7 +465,7 @@ stringData:
       # HMAC key used to sign the JWTs generated by the controlplane
       # The helper returns the base64 quoted value of the secret
       # We need to remove the quotes and then decoding it so it's compatible with the stringData stanza
-      generated_jws_hmac_secret: "vCSVNRi5kZ"
+      generated_jws_hmac_secret: "1hySjuyUgW"
 
       # Private key used to sign the JWTs meant to be consumed by the CAS
       cas_robot_account_private_key_path: "/secrets/cas.private.key"
@@ -1582,7 +1582,7 @@ spec:
     metadata:
       annotations:
         checksum/config: 7ebb40b21af918bef7decf41a5d6bbfcbdcab8a7b275e06922e7174f4ecfd98d
-        checksum/secret-config: 27987b6cef61e3d5de9c475a14e38d0f827fa1e895fccc0adbe02b88a2222226
+        checksum/secret-config: 1d5c3ebbfabb0b512373e7d4bab0ce81800aeb3ecf1866fd9fa4deb56957880c
         checksum/cas-private-key: 0bb2f0ef78066c0debed29e25af937f54deaa1c53b7b278afe0719308d271ef6
         kubectl.kubernetes.io/default-container: controlplane
       labels:

```

Refs #1151

### DIFF
--- a/deployment/chainloop/Chart.yaml
+++ b/deployment/chainloop/Chart.yaml
@@ -7,7 +7,7 @@ description: Chainloop is an open source software supply chain control plane, a 
 
 type: application
 # Bump the patch (not minor, not major) version on each change in the Chart Source code
-version: 1.87.0
+version: 1.87.1
 # Do not update appVersion, this is handled automatically by the release process
 appVersion: v0.95.5
 
--- a/deployment/chainloop/README.md
+++ b/deployment/chainloop/README.md
@@ -529,7 +529,7 @@ chainloop config save \
 | `controlplane.containerPorts.http`             | controlplane HTTP container port                                                                | `8000`            |
 | `controlplane.containerPorts.grpc`             | controlplane gRPC container port                                                                | `9000`            |
 | `controlplane.containerPorts.metrics`          | controlplane prometheus metrics container port                                                  | `5000`            |
-| `controlplane.tlsConfig.secret.name`           | name of a secret containing TLS certificate to be used by the controlplane grpc server.         | `""`              |
+| `controlplane.tls.existingSecret`              | Existing secret name containing TLS certificate to be used by the controlplane grpc server      | `random`          |
 | `controlplane.pluginsDir`                      | Directory where to look for plugins                                                             | `/plugins`        |
 | `controlplane.referrerSharedIndex`             | Configure the shared, public index API endpoint that can be used to discover metadata referrers |                   |
 | `controlplane.referrerSharedIndex.enabled`     | Enable index API endpoint                                                                       | `false`           |
@@ -713,15 +713,15 @@ chainloop config save \
 
 ### Artifact Content Addressable (CAS) API
 
-| Name                         | Description                                                                             | Value             |
-| ---------------------------- | --------------------------------------------------------------------------------------- | ----------------- |
-| `cas.replicaCount`           | Number of replicas                                                                      | `2`               |
-| `cas.image.registry`         | Image registry                                                                          | `REGISTRY_NAME`   |
-| `cas.image.repository`       | Image repository                                                                        | `REPOSITORY_NAME` |
-| `cas.containerPorts.http`    | controlplane HTTP container port                                                        | `8000`            |
-| `cas.containerPorts.grpc`    | controlplane gRPC container port                                                        | `9000`            |
-| `cas.containerPorts.metrics` | controlplane prometheus metrics container port                                          | `5000`            |
-| `cas.tlsConfig.secret.name`  | name of a secret containing TLS certificate to be used by the controlplane grpc server. | `""`              |
+| Name                         | Description                                                                                | Value             |
+| ---------------------------- | ------------------------------------------------------------------------------------------ | ----------------- |
+| `cas.replicaCount`           | Number of replicas                                                                         | `2`               |
+| `cas.image.registry`         | Image registry                                                                             | `REGISTRY_NAME`   |
+| `cas.image.repository`       | Image repository                                                                           | `REPOSITORY_NAME` |
+| `cas.containerPorts.http`    | controlplane HTTP container port                                                           | `8000`            |
+| `cas.containerPorts.grpc`    | controlplane gRPC container port                                                           | `9000`            |
+| `cas.containerPorts.metrics` | controlplane prometheus metrics container port                                             | `5000`            |
+| `cas.tls.existingSecret`     | Existing secret name containing TLS certificate to be used by the controlplane grpc server | `""`              |
 
 ### CAS Networking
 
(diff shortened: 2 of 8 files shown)


deployment/chainloop/templates/cas/configmap.yaml
```diff
@@ -21,7 +21,7 @@ data:
         # grpc downloads/uploads don't require this because they don't have timeouts
         timeout: 300s
       grpc:
-        {{- if .Values.cas.tlsConfig.secret.name  }}
+        {{- if .Values.cas.tls.existingSecret  }}
```
migmartri, line 24: I am worried this is a breaking change. We have users that are using this option today to enable e2e tls encryption. 

Could we support both options? 

Basically document the new one but if the user has the other one in the values.yaml file also render these?

deployment/chainloop/README.md
```diff
@@ -521,25 +521,25 @@ chainloop config save \
...
-| `controlplane.replicaCount`                    | Number of replicas                                                                              | `2`               |
-| `controlplane.image.registry`                  | Image registry                                                                                  | `REGISTRY_NAME`   |
-| `controlplane.image.repository`                | Image repository                                                                                | `REPOSITORY_NAME` |
-| `controlplane.containerPorts.http`             | controlplane HTTP container port                                                                | `8000`            |
-| `controlplane.containerPorts.grpc`             | controlplane gRPC container port                                                                | `9000`            |
-| `controlplane.containerPorts.metrics`          | controlplane prometheus metrics container port                                                  | `5000`            |
-| `controlplane.tlsConfig.secret.name`           | name of a secret containing TLS certificate to be used by the controlplane grpc server.         | `""`              |
```
migmartri, line 532: the new description is not good, it doesn't explain what's the cert is for.

migmartri (changes requested):

deployment/chainloop/templates/cas/configmap.yaml
javirln: If would then personally won't add this option since it's just naming convention by Bitnami to be aligned with their chart which result in the same functionality with different naming.

migmartri: Right, but what if a user wants to start using a Bitnami Chart, e.g., because of VAC? I'd give them an easy ramp.  

It doesn't seem hard to me to create two helpers, one called `cas.tls-enabled` that makes an `or of those two conditions` and another one called `cas-tls-secret-name` or smth like that?

deployment/chainloop/values.yaml
```diff
@@ -150,6 +150,15 @@ controlplane:
...
+  ## TLS configuration
+  ##
+  tls:
+    ## @param controlplane.tls.existingSecret Existing secret with TLS certificates. The secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
+    ## NOTE: When it's set it will disable secret creation.
+    ##
+    existingSecret: ""
+
   ## @param controlplane.tlsConfig.secret.name name of a secret containing TLS certificate to be used by the controlplane grpc server.
   tlsConfig:
```
migmartri, line 163: I remove this from the values.yaml file, it's a deprecated option, we do not need to document it. Note that once you remove it you'll need to update your helper to check for the nil struct.

deployment/chainloop/values.yaml
```diff
@@ -919,6 +928,15 @@ cas:
     grpc: 9000
     metrics: 5000
 
+  ##
+  ## TLS configuration
+  ##
```
migmartri, line 933: please update the description to clearly explain what this option is for, similar to what we had before

> name of a secret containing TLS certificate to be used by the controlplane grpc server. The secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.

deployment/chainloop/Chart.yaml
```diff
@@ -7,7 +7,7 @@ description: Chainloop is an open source software supply chain control plane, a
 
 type: application
 # Bump the patch (not minor, not major) version on each change in the Chart Source code
-version: 1.86.14
+version: 1.86.17
```
migmartri, line 10: make sure you bump this compared to the one in main.

deployment/chainloop/Chart.yaml
```diff
@@ -7,7 +7,7 @@ description: Chainloop is an open source software supply chain control plane, a
...
 # Do not update appVersion, this is handled automatically by the release process
 appVersion: v0.95.3
```
migmartri, line 12: same here, this got bumped.

deployment/chainloop/values.yaml
```diff
@@ -932,17 +926,11 @@ cas:
   ## TLS configuration
   ##
   tls:
-    ## @param cas.tls.existingSecret Existing secret with TLS certificates. The secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
-    ## NOTE: When it's set it will disable secret creation.
+    ## @param cas.tls.existingSecret Existing secret name containing TLS certificate to be used by the controlplane grpc server
```
migmartri, line 929: cas grpc Server

deployment/chainloop/values.yaml
```diff
@@ -154,16 +154,10 @@ controlplane:
...
+    ## NOTE: When it's set it will disable secret creation. The secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
     ##
-    existingSecret: ""
-
-  ## @param controlplane.tlsConfig.secret.name name of a secret containing TLS certificate to be used by the controlplane grpc server.
-  tlsConfig:
-    secret:
-      # the secret must contains 2 keys: tls.crt and tls.key respectively containing the certificate and private key.
-      name: ""
+    existingSecret: "random"
```
migmartri, line 160: remove "random" since the other one doesn't have it and it's not meant to be random is it?

deployment/chainloop/Chart.yaml
javirln: Sure, local one went rogue, sorry for that.

migmartri (approved):

Issue #1197, opened 2024-08-12
Add indexes to workflows and workflow runs


javirln:
There are some queries that can be improved in terms of performance by adding some indexes, for example the retrieval of workflow runs based on a workflow, currently, the SQL plan looks like to retrieve last 3 workflows runs of a workflow:
```sql
SELECT "workflow_runs"."id",
               "workflow_runs"."created_at",
               "workflow_runs"."finished_at",
               "workflow_runs"."state",
               "workflow_runs"."reason",
               "workflow_runs"."run_url",
               "workflow_runs"."runner_type",
               "workflow_runs"."attestation",
               "workflow_runs"."attestation_digest",
               "workflow_runs"."attestation_state",
               "workflow_runs"."contract_revision_used",
               "workflow_runs"."contract_revision_latest"
        FROM "workflow_runs"
        WHERE EXISTS (SELECT "workflows"."id"
                      FROM "workflows"
                      WHERE "workflow_runs"."workflow_workflowruns" = "workflows"."id"
                        AND "workflows"."id" = 'f09a517b-6975-489b-956e-14446b30b0d5')
        ORDER BY "workflow_runs"."created_at" DESC
        LIMIT 3
```
```
Limit  (cost=1970.54..1970.55 rows=3 width=229)
  ->  Sort  (cost=1970.54..1970.55 rows=3 width=229)
        Sort Key: workflow_runs.created_at DESC
        ->  Nested Loop  (cost=0.00..1970.52 rows=3 width=229)
              ->  Seq Scan on workflows  (cost=0.00..6.28 rows=1 width=16)
                    Filter: (id = 'f09a517b-6975-489b-956e-14446b30b0d5'::uuid)
              ->  Seq Scan on workflow_runs  (cost=0.00..1964.21 rows=3 width=245)
                    Filter: (workflow_workflowruns = 'f09a517b-6975-489b-956e-14446b30b0d5'::uuid)
```

Just by adding an index on the `workflow_workflowruns`, improves:
```
Limit  (cost=18.43..18.44 rows=3 width=229)
  ->  Sort  (cost=18.43..18.44 rows=3 width=229)
        Sort Key: workflow_runs.created_at DESC
        ->  Nested Loop  (cost=0.29..18.40 rows=3 width=229)
              ->  Seq Scan on workflows  (cost=0.00..6.28 rows=1 width=16)
                    Filter: (id = 'f09a517b-6975-489b-956e-14446b30b0d5'::uuid)
[...]

PR #1205, opened 2024-08-12
feat(workflow): Add missing index on workflows

javirln:
This patch adds a missing index on the foreign key `workflow_workflowruns` from the `workflow_runs` table.

Refs #1197


migmartri (commented):
I am a little bit confused about the order these files have appeared here. 

Could you elaborate a little of how you are adding these db changes? Are you running any generate command?

javirln:
> I am a little bit confused about the order these files have appeared here.
> 
> Could you elaborate a little of how you are adding these db changes? Are you running any generate command?

1. Changed `app/controlplane/pkg/data/ent/schema/workflowrun.go`
2. Run `make -C app/controlplane migration_sync`
3. Run `make generate`

migmartri (approved):

migmartri:
> > I am a little bit confused about the order these files have appeared here.
> > Could you elaborate a little of how you are adding these db changes? Are you running any generate command?
> 
> 1. Changed `app/controlplane/pkg/data/ent/schema/workflowrun.go`
> 2. Run `make -C app/controlplane migration_sync`
> 3. Run `make generate`

ok, makes sense

I usually run `make generate` before and then `make run` which happens to run `sync`. I am wondering if we should add `migration_sync` as a dependent target of `make generate` what do you think?

javirln:
> > > I am a little bit confused about the order these files have appeared here.
> > > Could you elaborate a little of how you are adding these db changes? Are you running any generate command?
> > 
> > 
> > 
> > 1. Changed `app/controlplane/pkg/data/ent/schema/workflowrun.go`
> > 2. Run `make -C app/controlplane migration_sync`
> > 3. Run `make generate`
> 
> ok, makes sense
> 
> I usually run `make generate` before and then `make run` which happens to run `sync`. I am wondering if we should add `migration_sync` as a dependent target of `make generate` what do you think?

We can add it in a separated PR yes! :)

PR #1213, opened 2024-08-12
feat(policies): support policies with arguments

### LINKED ISSUE #1212: Parameterizable policies
PolicyAttachements already have a `with` attribute to allow adding arguments to the policy instance. 

A first version of this feature could consist of just adding sets of key/values to the input object in Rego policies, without any sort of additional schemas nor validation.

A future improvement could also provide an input schema for additional arguments.

jiparis:
This PR adds support for setting policy arguments as simple key/value pairs. 

Given this policy if the attestation developer is the expected one:
```
kind: Policy
metadata:
  name: workflow
spec:
  type: ATTESTATION
  embedded: |
    package main

    import rego.v1

    deny contains msg if {
    	not valid_developer

    	msg := sprintf("Invalid developer found: %s", [input.args.email])
    }

    valid_developer if {
    	some subject in input.subject
    	subject.annotations["author.email"] == input.args.email
    }
```
We can pass the `input.args.email` argument with the following syntax in the contract:
```
schemaVersion: v1
policies:
  attestation:
    - ref: policy_with_args.yaml
      with:
         email: devel@chainloop.dev
```

It supports value arrays through multiline strings or CSV, for example:
```
policies:
  attestation:
    - ref: policy_with_args.yaml
      with:
         emails: devel1@chainloop.dev,devel2@chainloop.dev,devel3@chainloop.dev
         versions: |
            v14.3
            v14.4
            v15.3
```
Closes #1212


danlishka (approved):
This is perfect. It means we will be able to specify a list of developer emails, licenses, or CVE levels?

pkg/policies/engine/rego/rego.go
```diff
@@ -31,10 +31,12 @@ import (
 type Rego struct {
 }
 
+const InputArgs = "args"
```
migmartri, line 34: do you need to expose this const?

migmartri (approved):
Thanks @jiparis!

I have a question. So we are making a map of `any`, meaning that if I pass an array of emails for example in the attachment, that will be an array in the rego environment too?

It might be useful to add this in the test too

jiparis:
> Thanks @jiparis!
> 
> I have a question. So we are making a map of `any`, meaning that if I pass an array of emails for example in the attachment, that will be an array in the rego environment too?
> 
> It might be useful to add this in the test too

The engine supports it. But it has been exposed as a `map[string]string` in the Protobuf spec. We might need to move it to a protobuf `Any`, but not sure about its compatibility with, for example, arrays (should it be specified with a `repeated Any`?

pkg/policies/engine/rego/rego.go
jiparis: No, changing it.

jiparis:
> This is perfect. It means we will be able to specify a list of developer emails, licenses, or CVE levels?

Yes, exactly. I just added support for multivalue arguments.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -149,18 +150,14 @@ message PolicyAttachment {
   // set to true to disable this rule
   bool disabled = 4;
 
-  repeated PolicyArgument with = 5;
+  // optional arguments for policies
+  map<string, string> with = 5;
+//  bytes with = 5;
```
migmartri, line 155: spurious comment

I'd also add a comment here showing some examples of how to use it, containing single, multiline or comma separated? what do you think?

migmartri, line 155: spurious comment

I'd also add a comment here showing some examples of how to use it, containing single, multiline or comma separated? what do you think?

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -18,6 +18,7 @@ syntax = "proto3";
 package workflowcontract.v1;
 
 import "buf/validate/validate.proto";
+import "google/protobuf/any.proto";
```
migmartri, line 21: not needed

pkg/policies/engine/rego/rego.go
```diff
@@ -49,6 +51,16 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte)
 		return nil, fmt.Errorf("failed to parse input: %w", err)
 	}
 
+	// add arguments
+	if args != nil {
```
migmartri, line 55: it's not clear to me if you are doing any processing for single, comma-separated or multiline now?

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
jiparis: Yes, sounds good.

pkg/policies/engine/rego/rego.go
jiparis: This code is just embedding whatever comes as args (`map[string]any`) into the `input` object, so that it can be used as `input.args.foo` from the Rego code.

migmartri: but is it understanding that a multi-line or comma separated values are arrays?

jiparis: Yes, that parsing is done in the changes from policies.go. I've added a bunch of tests for all cases.

PR #1216, opened 2024-08-13
feat(cli): increase retry time

migmartri:
I noticed in the wild that 10 seconds was not enough in some cases to recover from conflicts. This patch increases the upper limit so I'll watch it

### DIFF
--- a/app/cli/cmd/attestation_add.go
+++ b/app/cli/cmd/attestation_add.go
@@ -110,7 +110,7 @@ func newAttestationAddCmd() *cobra.Command {
 
 					return nil
 				},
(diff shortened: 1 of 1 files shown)


app/cli/cmd/attestation_add.go
```diff
@@ -110,7 +110,7 @@ func newAttestationAddCmd() *cobra.Command {
 
 					return nil
 				},
-				backoff.NewExponentialBackOff(backoff.WithMaxElapsedTime(10*time.Second)),
+				backoff.NewExponentialBackOff(backoff.WithMaxElapsedTime(3*time.Minute)),
```
migmartri, line 113: the number is arbirtrary tbh, the default value is 60 minutes with I think it's too much

javirln (approved):

PR #1188, opened 2024-08-03
feat(policies): support external policies

jiparis:
This PR adds the support for external policies referenced by name, and optionally a prefix with the provider name:
```
schemaVersion: v1
materials:
policies:
  attestation:
    - name: sbom-present
    # or - name: myprovider://sbom-present
```

the policy will be retrieved at `attestation init` time from the provider. It's expected it to expose an endpoint (configured in chainloop `policy_providers` configuration section, with the policy name as a path parameter. For example, if the url is set to `https://myhost/policies`, Chainloop will ask for `https://myhost/policies/sbom-present`. The resulting payload will provide the policy in JSON format, just like this:
```
{
    "policy": {
        "apiVersion": "workflowcontract.chainloop.dev/v1",
        "kind": "Policy",
        "metadata": {
            "name": "sbom-present",
            "description": "Checks a SBOM is present in the attestation materials",
            "annotations": {
                "category": "sbom"
            }
        },
        "spec": {
            "embedded": "package main\n  \n# Verifies there is a SBOM material, even if not enforced by contract\n\nimport future.keywords.contains\nimport future.keywords.in\n\ndeny[msg] {\n  not has_sbom\n  msg := \"missing SBOM material\"\n}\n\n# Collect all material types\nkinds contains kind {\n  some material in input.predicate.materials\n  kind := material.annotations[\"chainloop.material.type\"]\n}\n  \nhas_sbom {\n  values := [\"SBOM_SPDX_JSON\",\"SBOM_CYCLONEDX_JSON\"]      \n  kinds[_] == values[_]\n}\n",
            "type": "ATTESTATION"
        }
    }
}
```

Example CLI execution (see the policy violation at the end):
```
✗ chainloop wf contract update --name embedded --contract test/my-contract.yaml
WRN API contacted in insecure mode
WRN Both user credentials and $CHAINLOOP_TOKEN set. Ignoring $CHAINLOOP_TOKEN.
INF Contract updated!
┌────────────────────────────────────────────┐
│ Contract                                   │
├──────────────────────┬─────────────────────┤
│ Name                 │ embedded            │
├──────────────────────┼─────────────────────┤
│ Description          │                     │
├──────────────────────┼─────────────────────┤
│ Associated Workflows │                     │
├──────────────────────┼─────────────────────┤
│ Revision number      │ 28                  │
├──────────────────────┼─────────────────────┤
│ Revision Created At  │ 03 Aug 24 17:27 UTC │
└──────────────────────┴─────────────────────┘
[...]

Issue #1180, opened 2024-08-01
Support remote policies


jiparis:
The goal of this task is to add support for organization owned policies, stored in an external repository.

Currently, any policy author can specify remote policies by providing an external public URL. But this might not be enough for private or builtin policies, that organizations don't want to expose publicly, but still want to keep in a central repository. For this purpose, we propose allowing the configuration of "Policy providers", consisting of an external service that will provide the policy material on request.

Policies will be referenced by name, optionally prepending a provider name:

```
apiVersion: workflowcontract.chainloop.dev/v1
kind: Policy
metadata:
  name: sarif-errors
spec:
  ref: policy-name # or chainloop://policy-name
```

On `contract create`, Chainloop would query the provider to verify that the reference is valid.
On `contract describe`, Chainloop will query the provider to retrieve the policy material and embed it in the resulting payload. This way, the CLI doesn't need to know anything about providers, since all the policies will be already in the contract.

The policy provider service would accept Chainloop API_TOKEN for Bearer authentication.

migmartri:
@jiparis we need to update the helm chart to support configuring providers correct?

jiparis:
correct. Let me follow up a quick PR

jiparis:
Please check [https://github.com/chainloop-dev/chainloop/pull/1224](https://github.com/chainloop-dev/chainloop/pull/1224)

PR #1218, opened 2024-08-13
policies provider framework

jiparis:
This PR adds a framework to resolve policies from external providers. Providers are registered through configuration, and they will resolve using a HTTP call to the external provider.

Note that this PR supersedes https://github.com/chainloop-dev/chainloop/pull/1188, as it will be splitted in different PRs for easier review.

Refs #1180


app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -65,6 +65,16 @@ message Bootstrap {
 
   // Configuration to enable Prometheus integration for the specified organizations
   repeated PrometheusIntegrationSpec prometheus_integration = 11;
+
+  // providers for external policy repositories
+  repeated PolicyProvider policy_providers = 12;
+}
+
+message PolicyProvider {
+  string name = 1 [(buf.validate.field).string.min_len = 1];
```
migmartri, line 74: we should probably add some validations to these names since they will be used as prefixes. We could use the same validation we use for other names.

app/controlplane/internal/policies/policyprovider.go
```diff
@@ -0,0 +1,79 @@
...
+	name, host string
+	isDefault  bool
+}
+
+type ProviderResponse struct {
+	Policy map[string]any `json:"policy"`
+}
+
+// Resolve calls the remote provider for retrieving a policy
+func (p *PolicyProvider) Resolve(policyName string, token string) (*schemaapi.Policy, error) {
```
migmartri, line 39: let's add some simple validations that make sure that `name` and `token` are provided.

app/controlplane/internal/policies/registry.go
```diff
@@ -0,0 +1,66 @@
...
+
+	conf "github.com/chainloop-dev/chainloop/app/controlplane/internal/conf/controlplane/config/v1"
+)
+
+// Registry manages policy providers
+type Registry struct {
+	providers map[string]*PolicyProvider
+}
+
+func NewRegistry(conf ...*conf.PolicyProvider) *Registry {
```
migmartri, line 29: should this factory also return an error and make sure that only one is set as default and error otherwise?

app/controlplane/internal/policies/registry.go
```diff
@@ -0,0 +1,66 @@
...
+	if name == "" {
+		return r.DefaultProvider()
+	}
+
+	return r.providers[name]
+}
+
+// GetProviderFromReference finds a provider given a policy reference in the form of `provider://policy`.
+// it returns the default provider if none found
+func (r *Registry) GetProviderFromReference(ref string) *PolicyProvider {
```
migmartri, line 60: mm, I'd rather that this fails if the ref is invalid, i.e doesn't have two parts. Defaulting to default seems unintuitive, also, defaultProvider is not required to exist is it?

migmartri (approved):

app/controlplane/internal/policies/registry.go
jiparis: right. I'll make this method more restricted.

jiparis: I've removed the method, since the representation of the reference doesn't belong to this context, and it's not used anyways.

migmartri (approved):

PR #1220, opened 2024-08-13
feat(policies): load and evaluate remote policies on attestation operations

jiparis:
This is the closing PR for #1180 

This change adds the capability to load policies from remote providers, by using the `chainloop://[provider-name/]policy-name` syntax in the `ref` field in contracts.

Note that for the feature to fully work, at least one provider must be configured in the control plane deployment. If no provider is set in the `ref` field, the default provider will be tried.
Previously, only policies coming from local filesystem and public HTTPS were supported.

Note that I've removed policy validations during contract craft operations, since they might not be present during contract creation.

Example execution:
```
> chainloop wf contract update --name embedded --contract test/my-contract.yaml
INF Contract updated!
┌────────────────────────────────────────────┐
│ Contract                                   │
├──────────────────────┬─────────────────────┤
│ Name                 │ embedded            │
├──────────────────────┼─────────────────────┤
│ Description          │                     │
├──────────────────────┼─────────────────────┤
│ Associated Workflows │                     │
├──────────────────────┼─────────────────────┤
│ Revision number      │ 30                  │
├──────────────────────┼─────────────────────┤
│ Revision Created At  │ 13 Aug 24 23:11 UTC │
└──────────────────────┴─────────────────────┘
┌────────────────────────────────────────────────────┐
│ {                                                  │
│   "schemaVersion": "v1",                           │
│   "policies": {                                    │
│     "materials": [                                 │
│       {                                            │
│         "ref": "chainloop://cyclonedx-banned-agpl" │
│       }                                            │
│     ],                                             │
│     "attestation": [                               │
│       {                                            │
│         "ref": "chainloop://sbom-present"          │
│       }                                            │
│     ]                                              │
│   }                                                │
│ }                                                  │
└────────────────────────────────────────────────────┘
```

```
✗ chainloop att init --name policytest --replace
│ Initialized At    │ 13 Aug 24 23:11 UTC                  │
│ Attestation ID    │ 4967926c-b527-4a91-98ae-e868ff06f31b │
│ Name              │ policytest                           │
│ Project           │ myproject                            │
│ Contract Revision │ 30                                   │
```

```
✗ chainloop att push
INF evaluating policy 'sbom-present' on attestation
WRN found policy violations (sbom-present) for statement
WRN  - missing SBOM material
INF push completed
│ Initialized At    │ 13 Aug 24 23:27 UTC                  │
│ Attestation ID    │ 31604952-2a73-4309-bc0d-30690c841e28 │
│ Name              │ policytest                           │
│ Project           │ myproject                            │
│ Contract Revision │ 30                                   │
```

### DIFF
--- a/pkg/policies/loader.go
+++ b/pkg/policies/loader.go
@@ -1,0 +1,101 @@
+//
+// Copyright 2024 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policies
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	pb "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1"
+	v1 "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
+	"github.com/chainloop-dev/chainloop/pkg/attestation/crafter/materials"
+	"github.com/sigstore/cosign/v2/pkg/blob"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// Loader defines the interface for policy loaders from contract attachments
+type Loader interface {
+	Load(context.Context, *v1.PolicyAttachment) (*v1.Policy, error)
+}
+
+// EmbeddedLoader returns embedded policies
+type EmbeddedLoader struct{}
+
+func (e *EmbeddedLoader) Load(_ context.Context, attachment *v1.PolicyAttachment) (*v1.Policy, error) {
+	return attachment.GetEmbedded(), nil
+}
+
+// URLLoader loader loads policies from filesystem and HTTPS references
+type URLLoader struct{}
+
+func (l *URLLoader) Load(_ context.Context, attachment *v1.PolicyAttachment) (*v1.Policy, error) {
+	reference := attachment.GetRef()
+
+	// look for the referenced policy spec (note: loading by `name` is not supported yet)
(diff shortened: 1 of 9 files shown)


pkg/policies/loader.go
```diff
@@ -0,0 +1,116 @@
...
+	Client pb.AttestationServiceClient
+}
+
+var remotePolicyCache = make(map[string]*v1.Policy)
+var cacheMutex sync.Mutex
+
+func (c *ChainloopLoader) Load(ctx context.Context, attachment *v1.PolicyAttachment) (*v1.Policy, error) {
+	ref := attachment.GetRef()
+
+	cacheMutex.Lock()
```
migmartri, line 82: what's this for? to pull the policy only once? maybe the mutext should be in the ChainloopLoader struct?

pkg/policies/loader.go
```diff
@@ -0,0 +1,116 @@
...
+	defer cacheMutex.Unlock()
+
+	if remotePolicyCache[ref] != nil {
+		return remotePolicyCache[ref], nil
+	}
+
+	parts := strings.SplitN(ref, "://", 2)
```
migmartri, line 89: wrapping this ref parsing into a function with some tests would help quite a bit

migmartri (approved):
LGTM!

migmartri (commented):
my only comment is that we might not be ready to make the switch to default provider since if we merge this, it would mean that all the policies that we have today up there will behave like remote providers when they were supposed to be files?

What if we do this in two steps

1 - we require passing a protocol and no protocol still means `file`
2 - we rewrite our examples, policies and so on to include file://
3 - we replace what no protocol means 

wdyt?

pkg/policies/loader.go
jiparis: Ok, I'll do it in a follow up PR

jiparis:
> my only comment is that we might not be ready to make the switch to default provider since if we merge this, it would mean that all the policies that we have today up there will behave like remote providers when they were supposed to be files?
> 
> What if we do this in two steps
> 
> 1 - we require passing a protocol and no protocol still means `file` 2 - we rewrite our examples, policies and so on to include file:// 3 - we replace what no protocol means
> 
> wdyt?

By default it will behave as before. This is the sequence:
1. check for embedded policy. If it exists, just use it.
2. check for `chainloop://` protocol in `ref`. If it exists, it uses the remote loader
3. otherwise, it uses the default behaviour for `ref`, interpreting `https://`, `env://` protocols, and defaulting to filesystem if no protocol is found.

PR #1232, opened 2024-08-14
chore(policies): improve error message when policy is not found

### LINKED ISSUE #1225: policies: improve error response
When a remote policy fails to be fetched we show an error that doesn't tell me which one actually failed.

We should show which one couldn't be loaded. 

```
chainloop att add --name sbom --value ./sbom.cyclonedx.json
INF uploading sbom.cyclonedx.json - sha256:16159bb881eb4ab7eb5d8afc5350b0feeed1e31c0a268e355e74f9ccbe885e0c
ERR adding material: error applying policies to material: policy error: failed to load policy spec: loading policy spec: loading policy: rpc error: code = NotFound desc = policy not found
```

On another note, and maybe as part of #1207, we should also figure out how to extract the basic error from the grpc.

btw, this is the contract I used

```json
{
  "schemaVersion": "v1",
  "materials": [
    {
      "type": "SBOM_CYCLONEDX_JSON",
      "name": "sbom"
    }
  ],
  "policies": {
    "materials": [
      {
        "ref": "cyclonedx-licenses.yaml"
      },
      {
        "ref": "https://raw.githubusercontent.com/chainloop-dev/chainloop/main/docs/examples/policies/sbom/cyclonedx-licenses.yaml"
      },
      {
        "ref": "chainloop://cyclonedx-licenses"
      },
      {
        "ref": "chainloop://platform/cyclonedx-licenses"
      }
    ]
  }
}

```

jiparis:
This PR fixes 2 issues:
* properly wrap upstream error
* add provider and policy name to the error messsage
```
ERR applying policies to statement: policy error: loading policy spec: requesting remote policy (provider: idontexist, name: sbom-present): rpc error: code = NotFound desc = policy not found
```

Fixes #1225

### DIFF
--- a/pkg/policies/policies.go
+++ b/pkg/policies/policies.go
@@ -50,6 +50,10 @@ func (e *PolicyError) Error() string {
 	return fmt.Sprintf("policy error: %s", e.err.Error())
 }
 
+func (e *PolicyError) Unwrap() error {
+	return e.err
+}
+
 type PolicyVerifier struct {
 	schema *v1.CraftingSchema
 	logger *zerolog.Logger
--- a/pkg/policies/loader.go
+++ b/pkg/policies/loader.go
@@ -111,7 +111,7 @@ func (c *ChainloopLoader) Load(ctx context.Context, attachment *v1.PolicyAttachm
 		PolicyName: name,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("loading policy: %w", err)
+		return nil, fmt.Errorf("requesting remote policy (provider: %s, name: %s): %w", provider, name, err)
 	}
 
 	// cache result


pkg/policies/policies.go
```diff
@@ -50,6 +50,10 @@ func (e *PolicyError) Error() string {
 	return fmt.Sprintf("policy error: %s", e.err.Error())
 }
 
+func (e *PolicyError) Unwrap() error {
```
migmartri, line 53: interesting, what was the result before this?

migmartri (approved):

pkg/policies/policies.go
jiparis: It was losing the error wrapping chain, so it couldn't be recognized as a RPC error with `errors.As`.

Issue #995, opened 2024-06-19
make contract server-side validations as good as the client side ones in the CLI


migmartri:
When creating a contract through the API, if it contains a typo on the actual fields of the json, it throws an error from the buf validation, we should return a better error as in the CLI:
```
$ chainloop wf contract create --name "testing" -f 'testing/scratch_8.yml'
ERR proto: (line 1:472): unknown field "type:"
```

Interesting enough, using grpcurl produces the same error:
```
$ grpcurl -plaintext -d "$(cat 'testing/scratch_8.yml')" -H "Authorization: Bearer $TOKEN" localhost:9000 controlplane.v1.WorkflowContractService/Create
Error invoking method "controlplane.v1.WorkflowContractService/Create": error getting request data: message type workflowcontract.v1.CraftingSchema.Material has no known field named type:
```

migmartri:
```
{
  "schemaVersion": "v1",
  "materials": [
    {
      "type": "CONTAINER_IMAGE",
      "name": "skynet-control-plane",
      "output": true,
      "annotations": [
        {
          "name": "component",
          "value": "control-plane"
        },
        {
          "name": "asset"
        }
      ]
    },
    { "type:": "ARTIFACT", "name": "rootfs" },
  ]
}
```

migmartri:
https://chainloop.slack.com/archives/C077D8D33KJ/p1718738998043859

danlishka:
what are the next steps on this then?

migmartri:
I'd like to double check the possibility of improving validation error messages server side. It's very discouraging that at first this is not trivial. So next step is to look into how hard is to have custom error messages applied to the proto validations.

PR #1239, opened 2024-08-16
feat(contracts): support multi-format 

### LINKED ISSUE #1237: contract format persistence and server-side validations
Currently, contracts can be provided in the CLI in three formats, yaml, json or cue but they are marshaled to protocolbuffers and stored in that binary representation in the DB

The problem with that approach is that. 

- We do not know what the format of preference of the user
- Comments, ordering, and so on will be lost
- we are not showing back a yaml representation of the contract
- Validations are inconsistent https://github.com/chainloop-dev/chainloop/issues/995

This could be solved by storing the raw representation of the contract (yaml, json, ...) and do the marshalling serverside to run validations https://github.com/chainloop-dev/chainloop/issues/995

migmartri:
This PR allows users to manage contracts in different formats during the whole lifecycle #1237

In practice, for the user, this means that the system will preserve the contracts in the format (preserving even comments and ordering). It also improves the validation errors, moves them server-side, and improves error messaging.

### Multi-format support

For example, creating a contract pointing to a yaml file now will maintain that format
```
chainloop wf contract create --name test-in-yaml -f  ~/work/chainloop/chainloop/docs/examples/contracts/skynet/contract.yaml
```

image

the API also returns the detected format, this is a hint to help future clients of this API render the contacts

```
chainloop wf contract describe --name test-in-yaml -o json | jq .revision.rawBody.format
"FORMAT_YAML"
```

of course we can update the format at any time, let's move it to json 
image

```
chainloop wf contract describe --name test-in-yaml -o json | jq .revision.rawBody.format
"FORMAT_JSON"
```

or cue

image

```
chainloop wf contract describe --name test-in-yaml -o json | jq .revision.rawBody.format
"FORMAT_CUE"
```

### Validations

Validations have been moved to the server side and especially improved in the case of a yaml file by introducing [protoyaml-go](https://github.com/bufbuild/protoyaml-go/). Fixes https://github.com/chainloop-dev/chainloop/issues/995

See below a contract that has errors and how it indicates the line and column where the error happens.

```sh
ERR failed to validate contract: validation error: :9:11 unknown enum value "NON_EXISTING", expected one of [MATERIAL_TYPE_UNSPECIFIED STRING CONTAINER_IMAGE ARTIFACT SBOM_CYCLONEDX_JSON SBOM_SPDX_JSON JUNIT_XML ...]
   9 |   - type: NON_EXISTING
   9 | ..........^

:1:16 schema_version: value must equal `v1` (string.const)
   1 | schemaVersion: v2
   1 | ...............^

:9:11 materials[0].type: value must not be in list [0] (enum.not_in)
   9 |   - type: NON_EXISTING
   9 | ..........^

exit status 1
``` 

On the technical side, this patch.

- Adds two new database columns to `raw_contract` and `contract_format` that store the raw content and format of the contract, respectively. The `body` column that until today stored the binary representation of the proto will not be populated anymore,
- Updates API to return the raw value and format alongside the current wired version in proto. The reason for returning both is to keep compatibility with clients that expect the binary proto, i.e attestations.
- This implementation works with existing contracts. For those, we generate the json representation of the stored contract, which in practice means that nothing changes for the user.
- Update the CLI to use this raw data instead during read operations with contracts

Closes #1237


app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -321,3 +347,104 @@ func (uc *WorkflowContractUseCase) GetPolicy(providerName, policyName, token str
...
+
+		if err := protojson.Unmarshal(jsonRawData, contract); err != nil {
+			return nil, NewErrValidation(err)
+		}
+	case ContractRawFormatYAML:
+		// protoyaml allows validating the contract while unmarshalling
+		yamlOpts := protoyaml.UnmarshalOptions{Validator: validator}
+		if err := yamlOpts.Unmarshal(raw, contract); err != nil {
+			return nil, NewErrValidation(err)
+		}
```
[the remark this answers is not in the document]
migmartri: we do fail int he line 409, but let me add it inside this method too. Thanks!

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -273,14 +277,39 @@ func (r *WorkflowContractRepo) SoftDelete(ctx context.Context, id uuid.UUID) err
...
+	// Scenario 1: contracts that have been stored (and not updated) before the introduction of the raw_body field will have an empty raw_body
+	// so we will generate a json representation of the contract to populate the raw_body field in that case
+	// that way clients can always expect a raw_body field to be present
+	if len(contract.Raw) == 0 {
+		schema := &schemav1.CraftingSchema{}
+		if err := proto.Unmarshal(w.Body, schema); err != nil {
+			return nil, err
+		}
+
+		contract, err = biz.SchemaToRawContract(schema)
```
jiparis, line 298: I guess in this case we always return JSON because raw format it's still not known.

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -273,14 +277,39 @@ func (r *WorkflowContractRepo) SoftDelete(ctx context.Context, id uuid.UUID) err
...
+	}
+
+	// We have two ways of storing the contract body, the old way is the body column which contains the binary representation of the proto message
+	// and the new way which is the raw_body and raw_body_format pairs
+	// Regardless of what's stored, we want to make sure we always return the contract object that contains the raw and binary representation
+	var err error
+	// Scenario 1: contracts that have been stored (and not updated) before the introduction of the raw_body field will have an empty raw_body
+	// so we will generate a json representation of the contract to populate the raw_body field in that case
+	// that way clients can always expect a raw_body field to be present
+	if len(contract.Raw) == 0 {
```
jiparis, line 292: I know it's the same, but I would put here `len(w.RawBody)`, just to keep consistency with the `else if` below.

jiparis (approved):
Looks good. Only two minor comments.

PR #1240, opened 2024-08-19
chore(policies): validate remote policies on contract creation

### LINKED ISSUE #1222: policies: verify remote policies exist during contract crafting
Currently, policy attachments are not validated (other than plain protobuf validation). The goal is to perform a server-side validation for remote policies. This is: 1) parse contract for remote policies, 2) query providers to retrieve those policies, 3) validate they exist and are well-formed.

jiparis:
This PR adds a server-side validation on contracts with remote policies. It would look for the appropriate provider and request the policy. If the provider or policy is not found, it will raise a validation error.

For example, for this contract:
```
schemaVersion: v1
materials:
policies:
  materials:
    - ref: chainloop://doesnotexist/cyclonedx-freshness
      with:
        limit: "1"
```

```
✗ chainloop wf contract update --name embedded --contract test/my-contract.yaml
ERR invalid policies: policy provider not found
```

Fixes #1222


app/controlplane/internal/service/workflowcontract.go
```diff
@@ -109,10 +118,16 @@ func (s *WorkflowContractService) Update(ctx context.Context, req *pb.WorkflowCo
 		return nil, err
 	}
 
+	token, err := usercontext.GetRawToken(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Unauthenticated, "invalid token")
+	}
+
 	schemaWithVersion, err := s.contractUseCase.Update(ctx, currentOrg.ID, req.Name,
```
migmartri, line 126: are you passing the token on update?

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -91,11 +94,17 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
...
+		return nil, status.Error(codes.Unauthenticated, "invalid token")
+	}
+
 	// Currently supporting only v1 version
 	schema, err := s.contractUseCase.Create(ctx, &biz.WorkflowContractCreateOpts{
 		OrgID: currentOrg.ID,
 		Name:  req.Name, Description: req.Description,
-		RawSchema: req.RawContract})
+		RawSchema: req.RawContract,
+		Token:     token})
```
migmartri, line 107: I find strange passing the token here, it looks like it's going to get stored.

What about creating another method in the use-case called `ValidateContractPolicies` and call it maybe at the service level? I am not 100 sure about it either but at least this "optional" feature is not part of the core part of creating/updating a contract. WDYT?

app/controlplane/internal/service/workflowcontract.go
jiparis: Yes, it's being passed both on create and update.

app/controlplane/internal/service/workflowcontract.go
jiparis: I know what you mean. I'll try extracting the logic to its own use case method. It might be cleaner and easier to understand.

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -91,6 +94,15 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
 		return nil, err
 	}
 
+	token, err := usercontext.GetRawToken(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Unauthenticated, "invalid token")
+	}
+
+	if err = s.contractUseCase.ValidateContractPolicies(req.RawContract, token); err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid policies")
```
migmartri, line 103: this probably hides the validation error, what if you use 

`handleUseCaseError` here instead?

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -299,6 +300,47 @@ func (uc *WorkflowContractUseCase) Update(ctx context.Context, orgID, name strin
 	return c, nil
 }
 
+func (uc *WorkflowContractUseCase) ValidateContractPolicies(rawSchema []byte, token string) error {
```
migmartri, line 303: not required in this PR but it would be good to add some tests on this biz method

migmartri (approved):

app/controlplane/pkg/biz/workflowcontract.go
jiparis: I will do it.

PR #1241, opened 2024-08-19
chore: expose workflow_id in integration attachment

migmartri:


javirln (approved):
Locally tested and all workflows are being retrieved seamlessly

PR #1242, opened 2024-08-19
refactor(integration-attachments): Map all integration attachments properties

javirln:
This patch wraps the integration attachments listing into a structure that holds both a `Integration` and `IntegrationAttachment`.

### DIFF
--- a/app/controlplane/pkg/biz/integration.go
+++ b/app/controlplane/pkg/biz/integration.go
@@ -77,7 +77,7 @@ type IntegrationRepo interface {
 
 type IntegrationAttachmentRepo interface {
 	Create(ctx context.Context, integrationID, workflowID uuid.UUID, config []byte) (*IntegrationAttachment, error)
-	List(ctx context.Context, orgID, workflowID uuid.UUID) ([]*IntegrationAttachment, error)
+	List(ctx context.Context, orgID, workflowID uuid.UUID) ([]*IntegrationAndAttachment, error)
 	FindByIDInOrg(ctx context.Context, orgID, ID uuid.UUID) (*IntegrationAttachment, error)
 	SoftDelete(ctx context.Context, ID uuid.UUID) error
 }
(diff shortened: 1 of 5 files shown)


app/controlplane/pkg/data/integrationattachment.go
```diff
@@ -93,28 +93,39 @@ func (r *IntegrationAttachmentRepo) FindByIDInOrg(ctx context.Context, orgID, id
 		return nil, nil
 	}
 
-	return entIntegrationAttachmentToBiz(integration), nil
+	return entIntegrationAttachmentToBiz(integration).IntegrationAttachment, nil
```
migmartri, line 96: ?

app/controlplane/pkg/biz/integration.go
```diff
@@ -314,7 +314,7 @@ func (uc *IntegrationUseCase) Delete(ctx context.Context, orgID, integrationID s
 }
 
 // List attachments returns the list of attachments for a given organization and optionally workflow
-func (uc *IntegrationUseCase) ListAttachments(ctx context.Context, orgID, workflowID string) ([]*IntegrationAttachment, error) {
+func (uc *IntegrationUseCase) ListAttachments(ctx context.Context, orgID, workflowID string) ([]*IntegrationAndAttachment, error) {
```
migmartri, line 317: It would be nice to add some quick integration tests

app/controlplane/pkg/data/integrationattachment.go
javirln: `entIntegrationAttachmentToBiz` now returns a `IntegrationAndAttachment` so we need to extract the `IntegrationAttachment`

app/controlplane/pkg/biz/integration.go
javirln: Added!

app/controlplane/pkg/data/integrationattachment.go
```diff
@@ -93,28 +93,39 @@ func (r *IntegrationAttachmentRepo) FindByIDInOrg(ctx context.Context, orgID, id
 		return nil, nil
 	}
 
-	return entIntegrationAttachmentToBiz(integration), nil
+	return entIntegrationAndAttachmentToBiz(integration).IntegrationAttachment, nil
```
migmartri, line 96: my bad, I didn't understand the diff, it showed that it's another function

migmartri (approved):

PR #1247, opened 2024-08-20
chore(policies): Refactor policy Registry to pkg

javirln:
This patch refactors the `policies` package and moves it to `pkg`. In the process the dependency with `conf.PolicyProviders` was removed and changed to a local configuration and translation.


app/controlplane/pkg/policies/registry.go
```diff
@@ -17,16 +17,20 @@ package policies
 
 import (
 	"fmt"
-
-	conf "github.com/chainloop-dev/chainloop/app/controlplane/internal/conf/controlplane/config/v1"
 )
 
+type NewPolicyProviderConfig struct {
```
migmartri, line 22: NewRegistryConfig

app/controlplane/pkg/biz/testhelpers/database.go
```diff
@@ -208,8 +209,12 @@ func NewDataConfig(in *conf.Data) *data.NewConfig {
 	}
 }
 
-func NewPolicyProviders(in *conf.Bootstrap) []*conf.PolicyProvider {
-	return in.PolicyProviders
+func NewPolicyProviderConfig(in *conf.Bootstrap) []*policies.NewPolicyProviderConfig {
```
migmartri, line 212: `in` can be `PolicyProviders`

migmartri (approved):

PR #1250, opened 2024-08-20
feat: docker compose configuration for local testing

### LINKED ISSUE #1246: Docker compose configuration
Provide a docker-compose configuration for a full Chainloop setup, including CAS and all supporting services (database, etc). This will be useful for less restricted scenarios, or for developers and integrations to play with different configuration options before going full into production.

jiparis:
This PR adds a new docker compose file for those who want to quickly setup a local instance of Chainloop using upstream images.
This configuration sets up: control-plane, artifacts-cas, control-plane-migrations, and leverages existing devel compose configuration for postgresql, vault and dex.
I've setup service dependencies properly so that they are run in order. Interestingly, I've added a service for running DB migrations.

Caveats:
* It uses `dex` as the Dex host (this is to ensure there is connectivity between docker instances). This means that users must add a `127.0.0.1 dex` entry in `/etc/hosts`, otherwise `chainloop auth login` won't work. This might be fixed by running all services with `host` networking, but it only works in Linux and recent versions of Docker Desktop.

How to use:
```
> docker compose up -d
```

And then just configure Chainloop CLI to use the local instance:
```
> chainloop config save --insecure --control-plane localhost:9000 --artifact-cas localhost:9001
```
And use as normal (note that `--insecure` is needed in every command, so maybe creating an alias could help `alias cl="chainloop --insecure --debug"`

Note that I've also included an "optional" Minio service, that must be explicitly run with `docker compose --profile optional up`, to be able to configure a S3-compatible backend for the CAS.

Fixes #1246

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -50,9 +50,11 @@ services:
         condition: service_healthy
     command: "migrate apply --url postgres://postgres:@postgresql:5432/controlplane?sslmode=disable --dir file:///migrations"
 
-#  minio:
-#    image: quay.io/minio/minio
-#    ports:
-#      - 9002:9000
-#      - 9003:9001
-#    command: server /data --console-address ":9001"
+  minio:
+    image: quay.io/minio/minio
+    ports:
+      - 9002:9000
+      - 9003:9001
+    command: server /data --console-address ":9001"
+    profiles:
+      - optional
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,57 @@
+name: chainloop-lab
+
+include:
+  - ./devel/compose.yml
+services:
+  control-plane:
+    image: ghcr.io/chainloop-dev/chainloop/control-plane:v0.96.1
+    environment:
+      CP_DB_HOST: postgresql
+      CP_DEX_DOMAIN: http://dex:5556/dex
+      CP_VAULT_ADDRESS: http://vault:8200
+      CP_FILE_CA_CERT_PATH: /devkeys/ca.pub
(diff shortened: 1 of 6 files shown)


compose.yaml
```diff
@@ -0,0 +1,64 @@
+name: chainloop-lab
```
migmartri, line 1: I am a little bit concerned that by making this a default compose setup it might be interpreted as anything closer to production or even development. It's not suited for any of those purposes.

I am wondering if instead we should have this compose.yaml file and have it inside devel as `labs.compose.yaml` so it doesn't get executed by mistake.

migmartri (approved):
Nice! 

Two comments I have

- Maybe make it more explicit that's a labs setup, see my comment inline.
- So you'd rather not to use the host network then? What's exactly the problem? Note that there is a way to override the returning URL in the controlplane

jiparis:
> * So you'd rather not to use the host network then? What's exactly the problem? Note that there is a way to override the returning URL in the controlplane

The problem is that the OIDC domain and issuer must be exactly the same. Otherwise `go-oidc`'s `NewProvider` method would fail with `oidc: issuer did not match the issuer returned by provider, expected ...`, during chainloop boot. We could avoid this error if we explicitly set the issuer in the context (`issuerURL, skipIssuerValidation := ctx.Value(issuerURLKey).(string)`). 
I'll leave this for another PR. In the meantime, the issuer must be a URL reachable from docker, which in the case of Compose, is the service name (hence the `dex` domain).

PR #1266, opened 2024-08-22
chore(policies): deprecate `deny` and move main rule to `violations`

jiparis:
This PR just moves from `deny` to `violations`. This way the main rule represents what it actually is: a set of violations, since Chainloop policies don't deny nor allow, just attest policy evaluations.

Note: we still support the old deprecated `deny` rule.


pkg/policies/engine/rego/rego.go
```diff
@@ -67,8 +69,8 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
 	// add module
 	regoFunc := rego.ParsedModule(parsedModule)
 
-	// add query. Note that the predefined rule to look for is `deny`
-	query := rego.Query(fmt.Sprintf("%v.deny\n", parsedModule.Package.Path))
+	// add query. Note that the predefined rule to look for is `violations`
+	query := rego.Query(fmt.Sprintf("%v.%s\n", parsedModule.Package.Path, mainRule))
```
migmartri, line 73: should we have some retro-compat until all the policies have been written? 

Basically document the new format but allow `deny` if it's already set?

jiparis: sure, I've added retro compatibility.

javirln (approved):
Awesome!

migmartri (approved):

PR #1276, opened 2024-08-27
feat(workflow-contract): Add contract name to revision item

javirln:
This patch adds the name of the contract along with the rest of information on the contract revision item.


jiparis (approved):

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -45,10 +45,11 @@ type WorkflowContract struct {
...
 type WorkflowContractVersion struct {
-	ID        uuid.UUID
-	Revision  int
-	CreatedAt *time.Time
-	Schema    *Contract
+	ID           uuid.UUID
+	Revision     int
+	CreatedAt    *time.Time
+	Schema       *Contract
+	ContractName string
```
migmartri, line 52: there is no test to update?

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -45,10 +45,11 @@ type WorkflowContract struct {
 }
 
 type WorkflowContractVersion struct {
-	ID        uuid.UUID
-	Revision  int
-	CreatedAt *time.Time
-	Schema    *Contract
+	ID           uuid.UUID
```
migmartri, line 48: I belive there are two objects no? A contract and a version, why haven't we instead put in in the workflow?

It feels wrong that we are adding a property from another object into the version.

migmartri: ok, I see now, it's because of the biz/repo calls that we make.

What if we just update the `FindVersionByID` to return `WorkflowContractWithVersion` instead of `WorkflowContractVersion`?

That will make it clear what we are returning and this change will affect only that function.

app/controlplane/api/controlplane/v1/response_messages.proto
```diff
@@ -135,6 +135,9 @@ message WorkflowContractVersionItem {
 
   RawBody raw_contract = 5;
 
+  // The name of the contract used for this run
+  string contract_name = 6;
```
migmartri, line 139: Do you still need this?

migmartri: ahh yes, you do :)

PR #1281, opened 2024-08-29
feat(jsonschema): Generate jsonschema for workflowcontract proto

javirln:
This patch adds a plugin to generate json schemas based on the `proto` defines in the code. Since we are only interested in the ones about `workflowcontracts`, I've modified the Make file to get rid of those unwanted files.

### DIFF
--- a/app/controlplane/Makefile
+++ b/app/controlplane/Makefile
@@ -10,8 +10,6 @@ config: check-buf-tool
 api: check-buf-tool
 	cd ./plugins/sdk/v1/plugin/api && buf generate
 	cd ./api && buf generate
-	# Remove all json schema that are not part of the workflowcontract since we don't need them
-	find ./api/gen/jsonschema -type f ! -name 'workflowcontract.*' -exec rm {} +
 
 .PHONY: build
 # build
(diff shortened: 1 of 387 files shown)


app/controlplane/Makefile
```diff
@@ -10,6 +10,8 @@ config: check-buf-tool
 api: check-buf-tool
 	cd ./plugins/sdk/v1/plugin/api && buf generate
 	cd ./api && buf generate
+	# Remove all json schema that are not part of the workflowcontract since we don't need them
+	find ./api/gen/jsonschema -type f ! -name 'workflowcontract.*' -exec rm {} +
```
migmartri, line 14: mm, I am not sure this is correct. in the CI we might be running generation without makefile, and afaik we are not enforcing using `make` for things

What's wrong with having all the files? How many are those?

app/controlplane/api/buf.gen.yaml
```diff
@@ -19,3 +19,5 @@ plugins:
       - outputClientImpl=grpc-web # client implementation it generates
       - esModuleInterop=true # use imports as required in modern ts setups
       - useOptionals=messages # use optional TypeScript properties instead of undefined
+  - plugin: buf.build/bufbuild/protoschema-jsonschema
```
migmartri, line 22: can you make sure we ignore them here too?

https://github.com/chainloop-dev/chainloop/blob/main/.gitattributes

app/controlplane/Makefile
javirln: There are 189 generated files without the option of not generating them.

migmartri (approved):
LGTM!

PR #1284, opened 2024-08-30
feat: store FQ reference to the policy

migmartri:
This PR stores the fully qualified reference, including digest, of the policy, both for local (file://) or remote (https://, chainloop://). In the case of the remote option, the body is now not stored in the attestation.

For example, having this contract with a mix of local, remote, and provider-based policies

```yaml
schemaVersion: v1
materials:
  - name: sbom
    type: SBOM_CYCLONEDX_JSON
policies:
  attestation:
    # http url
    - ref: https://raw.githubusercontent.com/chainloop-dev/chainloop/main/docs/examples/policies/sbom/sbom-present.yaml
    # local file
    - ref: file:///Users/miguelmartinez/work/chainloop/policies-playground/cyclonedx-licenses.yaml
    # provider with protocol
    - ref: chainloop://sbom-present
  materials:
    - ref: chainloop://cyclonedx-freshness
      with:
        limit: 30
    # provider without protocol
    - ref: cyclonedx-banned-agpl
```

This is the information stored in the attestation. 

Screenshot 2024-09-01 at 11 00 33

Some implementation decisions

- I decided to add the sha256 of the raw policy to all references
- The provider-based reference gets expanded to FWDN including the provider domain, for example `chainloop://localhost:8002/cyclonedx-banned-agpl`.


app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -33,51 +34,84 @@ type PolicyProvider struct {
...
+}
+
+func (p *PolicyProvider) resolveRef(policyName, digest string) (*PolicyReference, error) {
+	// Extract hostname from the policy provider URL
+	uri, err := url.Parse(p.host)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing policy provider URL: %w", err)
+	}
+
+	if uri.Host == "" {
```
jiparis, line 105: I wonder if we should store the full URL, not only the host.

pkg/policies/policies.go
```diff
@@ -152,29 +157,37 @@ func (pv *PolicyVerifier) VerifyStatement(ctx context.Context, statement *intoto
 			return nil, NewPolicyError(err)
 		}
 
+		// We store the body in the attestation unless the policy comes from a remote provider
+		// in which case with the reference it will suffice
+		var body string
+		if !strings.HasPrefix(ref.GetName(), chainloopLoaderPrefix) {
```
jiparis, line 163: I think you can use `loader.IsProviderScheme` for this

pkg/policies/loader.go
```diff
@@ -40,62 +43,86 @@ const (
...
-		err     error
+		raw []byte
+		err error
 	)
 
 	ref := attachment.GetRef()
 	filePath, err := ensureScheme(ref, fileScheme)
 	if err != nil {
-		return nil, err
+		return nil, nil, fmt.Errorf("invalid policy reference %q", ref)
```
jiparis, line 70: For these cases, where we just use auxiliary functions in the same package, I think wrapping the error is too much duplicity (see the error from `ensureScheme`, it's exactly the same message).

pkg/policies/loader.go
```diff
@@ -200,3 +237,12 @@ func refParts(ref string) (string, string) {
 	}
 	return "", parts[0]
 }
+
+func policyReferenceResourceDescriptor(ref string, digest crv1.Hash) *v12.ResourceDescriptor {
```
jiparis, line 241: it's weird that we need to import protos from Attestation API for ResourceDescriptor. We might want to extract it in a follow up.

jiparis (approved):

app/controlplane/pkg/policies/policyprovider.go
migmartri: the reason of not storing the full URL is because to me, the implementation of how this policy is retrieved is not that relevant, in fact it might change in the future.

I was thinking of the OCI registry example, when you use an identifier such as i.e gcr.dev/repo/image, this will get translated to API calls to `v2/manifests` and `v2/blobs`. So this is the same. The client, the provider will know how to retrieve the policy. 

WDYT?

pkg/policies/loader.go
migmartri: I think you are right, will change that

pkg/policies/loader.go
migmartri: ok

app/controlplane/pkg/policies/policyprovider.go
jiparis: Yes, it makes sense 👍, thanks

PR #1289, opened 2024-09-01
refactor: support headless auth in windows

migmartri:
This patch adds support to compiling the binary for windows by adding the proper system calls during authentication.


app/cli/cmd/terminalunix.go
```diff
@@ -0,0 +1,35 @@
+//go:build darwin
```
migmartri, line 1: the key are these build tags

javirln (approved):

PR #1294, opened 2024-09-05
chore(policies): decouple evaluations representation

jiparis:
This PR decouples Policy Evaluation representation in the different layers: crafting, render and attestation API.
Apart from the decoupling advantage, this change solves the different [de]serialization problem when dealing with protobuf and structpb representations, in particular with Enums and oneOfs. 
Thanks to this change, `type` field in evaluations has been fixed.
```
          "CHAINLOOP.ATTESTATION": [
@@ -74,7 +86,7 @@
                   },
                   "name": "chainloop://localhost:8002/sbom-present"
                },
-               "type": 12
+               "type": "ATTESTATION"
             },
             {
                "description": "Verifies that the attestation explicitly references a specific Git commit",
@@ -85,8 +97,34 @@
                   },
                   "name": "chainloop://localhost:8002/source-commit"
                },
-               "type": 12
+               "type": "ATTESTATION"
             }
```
Note: I've added a fix in `extractPredicate` method to be able to read old attestations with the wrong `type` field (int)


pkg/attestation/renderer/chainloop/v02.go
```diff
@@ -38,7 +38,24 @@ type ProvenancePredicateV02 struct {
 	*ProvenancePredicateCommon
 	Materials []*intoto.ResourceDescriptor `json:"materials,omitempty"`
 	// Map materials and policies
-	PolicyEvaluations map[string][]*v1.PolicyEvaluation `json:"policy_evaluations,omitempty"`
+	PolicyEvaluations map[string][]*PolicyEvaluation `json:"policy_evaluations,omitempty"`
+}
+
+type PolicyEvaluation struct {
```
migmartri, line 44: makes sense to me

pkg/attestation/renderer/renderer.go
```diff
@@ -178,11 +178,15 @@ func addPolicyResults(statement *intoto.Statement, policyResults []*v1.PolicyEva
 		return fmt.Errorf("unmarshalling predicate: %w", err)
 	}
 
-	// insert policy evaluations
+	// insert policy evaluations for attestation
```
migmartri, line 181: I think there is smth wrong with this method, it still references `v1 "github.com/chainloop-dev/chainloop/pkg/attestation/crafter/api/attestation/v1"` and we might don't need this?

jiparis: It was needed by PolicyVerifier. I've moved all policy-related logic to v02, since that's the predicate version that understands about policies. Now `renderer.go` is clean.

pkg/attestation/crafter/materials/materials.go
```diff
@@ -209,30 +207,3 @@ func Craft(ctx context.Context, materialSchema *schemaapi.CraftingSchema_Materia
 
 	return m, nil
 }
-
-// LoadJSONBytes Extracts raw data in JSON format from different sources, i.e cue or yaml files
-func LoadJSONBytes(rawData []byte, extension string) ([]byte, error) {
```
jiparis, line 214: Moved to avoid import cycles

pkg/attestation/renderer/chainloop/chainloop.go
```diff
@@ -189,6 +193,23 @@ func extractPredicate(statement *intoto.Statement, v *ProvenancePredicateV02) er
 	return nil
 }
 
+func fixPolicyTypeField(statement *intoto.Statement) {
```
jiparis, line 196: Needed to prevent marshaling errors with old, wrong attestations with enum (number) `type` field.

javirln (approved):

migmartri (approved):

PR #1297, opened 2024-09-08
feat(policies): digest support

### LINKED ISSUE #1111: Allow to pin policies by digest
when reference a policy, it would be useful to pin them by hash, just to ensure that the policy version at schema creation time is the one used at evaluation time. 
```
schemaVersion: "v1"
policies:
  attestation:
    - ref: testdata/policies/policy_rego.yaml@sha256:fcae9e0e7313c6467a7c6632ebb5e5fab99bd39bd5eb6ee34a211353e647827a
```

migmartri:
This patch allows setting `@sha256:digest` in the policy reference which will be used for

- pull the right policy version in the case of a policy from a chainloop provider.
- client-side validate other providers such as file or HTTP

This patch also includes some minor updates on error handling to cleanup a little bit the message, I know @jiparis you were asking for this, and I think you are right.

This is now an example of a policy that digest mismatches

```
failed to load policy attachment "file:///Users/miguelmartinez/work/chainloop/policies-playground/cyclonedx-licenses.yaml@sha256:12b7d023ea5410cb4f35804f8a6cd66a6e1bdc0bedb12c65a04e6b4b4c15f352": digest mismatch: got sha256:4f07666ec9c464885bb3ec65d400a63dca2c47b4c8ac278a9365146137265093, want sha256:12b7d023ea5410cb4f35804f8a6cd66a6e1bdc0bedb12c65a04e6b4b4c15f352
```

closes #1111

### DIFF
--- a/app/controlplane/pkg/biz/workflowcontract.go
+++ b/app/controlplane/pkg/biz/workflowcontract.go
@@ -333,7 +333,7 @@ func (uc *WorkflowContractUseCase) findPolicy(att *schemav1.PolicyAttachment, to
 		provider, name := loader.ProviderParts(att.GetRef())
 		remotePolicy, err := uc.GetPolicy(provider, name, token)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get policy '%s': %w", name, err)
+			return nil, err
 		}
 		return remotePolicy.Policy, nil
 	}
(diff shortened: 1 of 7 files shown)


jiparis (approved):
Looks great!. Maybe adding some documentation after this PR.

PR #1308, opened 2024-09-12
fix(policies): Support file:// URLs with no path

### LINKED ISSUE #1307: bug: "path is empty" error with policy references like `file://mypolicy.yaml`
Policy references are parsed as URLs. For cases like `file://mypolicy.yaml`, the validation is checking for a URL path, but the whole reference is parsed as the "host", leaving the "path" empty, resulting in a validation error:
```
cl wf contract update --name keyfactor-keyfactor --contract my-contract.yaml
ERR validation error: invalid reference "file://sbom-present.yaml": path is empty
```

jiparis:
this PR adds support to local file URLs with no path like: `file://my-policy.yaml`

Fixes #1307


migmartri:
Thanks!

PR #1322, opened 2024-09-13
feat(policies): multi-kind policies

jiparis:
Support for multi-kind policies, allowing to specify execution branches (Rego scripts) depending on the material kind:
```
kind: Policy
metadata:
  name: multikind
  description: multikind policy
  annotations:
    category: SBOM
spec:
  policies:
    - kind: SBOM_SPDX_JSON
      path: testdata/spdx_licenses.rego
    - kind: SBOM_CYCLONEDX_JSON
      path: testdata/cyclomedx_licenses.rego
```


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -189,6 +189,37 @@ message Metadata {
...
+      10,
+      11
+    ]
+  }, deprecated = true];
+
+  repeated PolicySpecV2 policies = 4;
+
+  option (buf.validate.message).cel = {
+    id: "policyspec",
+    message: "either spec source or policies fields must be provided",
```
migmartri, line 217: great, thanks

pkg/policies/policies.go
```diff
@@ -122,68 +127,84 @@ func (pv *PolicyVerifier) VerifyMaterial(ctx context.Context, material *v12.Atte
...
 		if err != nil {
 			return nil, NewPolicyError(err)
 		}
 
-		// 4. verify the policy
-		ng := getPolicyEngine(spec)
-		res, err := ng.Verify(ctx, script, material, getInputArguments(policyAtt.GetWith()))
+		pv.logger.Info().Msgf("evaluating policy '%s' on attestation", policy.Metadata.Name)
+
+		violations, scriptBody, err := pv.executeScript(ctx, policy, scripts, material, policyAtt)
```
migmartri, line 160: this method signagture is strange `pv.executeScript` in singular accepts multiple scripts. Why did you decide to do this instead of looping on the method and collecting the results here?

pkg/policies/policies.go
```diff
@@ -122,68 +127,84 @@ func (pv *PolicyVerifier) VerifyMaterial(ctx context.Context, material *v12.Atte
...
+
+	for _, script := range scripts {
+		// verify the policy
+		ng := getPolicyEngine(policy)
+		res, err := ng.Verify(ctx, script, material, getInputArguments(att.GetWith()))
+		if err != nil {
+			return nil, nil, NewPolicyError(err)
+		}
+
+		body.WriteString(beginPolicySep)
```
migmartri, line 198: should we store it as an array of base64? Would it that be possible? If not that's ok.

migmartri (approved):

pkg/policies/policies.go
jiparis: Initially I thought of that, yes, that's why I accidentally kept the singular version of it. I'll extract the loop outside, I agree it would be easier to understand.

jiparis: Finally I kept the loop inside the method, just to reuse the most possible. Of course I changed the method name.

pkg/policies/policies.go
jiparis: I've created a new property `sources`

PR #1324, opened 2024-09-15
feat(dagger): support workflow creation

### LINKED ISSUE #1323: support creating workflows from dagger module
It's common to create chainloop workflows on the fly in the CI before an attestation starts. 

To create a workflow, you can use the CLI (or UI) as shown [here](https://docs.chainloop.dev/getting-started/workflow-definition#workflow-create). If you want to create those workflows programmatically from the CI before each attestation, you can use [this reusable workflow](https://github.com/chainloop-dev/labs/blob/main/.github/workflows/chainloop_onboard.yml), authenticated with an [API token](https://docs.chainloop.dev/reference/operator/api-tokens#api-tokens).

It would be great if we could provide this functionality to our Dagger module

migmartri:
extends our dagger module to support creating workflows

```
Create a new workflow

USAGE
  dagger call workflow-create [arguments]

ARGUMENTS
      --name string          Workflow name [required]
      --project string       Workflow project [required]
      --token Secret         Chainloop API token [required] (default )
      --description string
      --public               Set workflow as public so other organizations can see it
      --skip-if-exists       If the workflow already exists, skip the creation and return success
      --team string
```

closes #1323

### DIFF
--- a/dagger.json
+++ b/dagger.json
@@ -2,5 +2,5 @@
   "name": "chainloop",
   "sdk": "go",
   "source": "extras/dagger",
-  "engineVersion": "v0.12.3"
+  "engineVersion": "v0.13.0"
 }
--- a/extras/dagger/README.md
+++ b/extras/dagger/README.md
@@ -193,7 +193,6 @@ init --token env:CHAINLOOP_TOKEN --workflow-name test-from-dagger-2 attestation-
 
 Please run `dagger call with-instance -h` for more information on the available options.
 
-
 ## Documentation
 
 To learn more, please visit the Chainloop project's documentation website, https://docs.chainloop.dev where you will find a getting started guide, FAQ, examples, and more.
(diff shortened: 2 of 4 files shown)


dagger.json
```diff
@@ -2,5 +2,5 @@
   "name": "chainloop",
   "sdk": "go",
   "source": "extras/dagger",
-  "engineVersion": "v0.12.3"
+  "engineVersion": "v0.13.0"
```
migmartri, line 5: my local dagger version is 0.13, so I decided to bump it. The changes in the go.mod seems to also be a cause of `go develop`

migmartri:
I need to add support to attach the workflow to an existing contract

dagger.json
javirln: They released the 0.13 last week, we should all update to latest.

javirln (approved):

PR #1327, opened 2024-09-16
feat(policies): document multi-kind policies

jiparis:
Follow up to document multi-kind policies.
Also, all examples have been moved to multikind format

### DIFF
--- a/docs/docs/reference/policies.mdx
+++ b/docs/docs/reference/policies.mdx
@@ -32,6 +32,28 @@ spec:
   path: my-script.rego
 ```
 
+### Supporting multiple material types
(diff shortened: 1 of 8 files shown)


docs/docs/reference/policies.mdx
```diff
@@ -32,6 +32,28 @@ spec:
   path: my-script.rego
 ```
 
+### Supporting multiple material types
```
migmartri, line 35: I'd review the example above to only use this new spec format, not the previous one, which is now deprecated correct? We should also rewrite the policies we have in place to that no?

javirln (approved):

migmartri (approved):

PR #1337, opened 2024-09-18
chore(dagger): downgrade dagger engine

migmartri:
https://github.com/chainloop-dev/chainloop/pull/1324 introduced by mistake a dagger engine upgrade. We should do this upgrade separately to allow clients that do not have this version to stick to this commit hash.


javirln (approved):
LGTM

PR #1345, opened 2024-09-20
fix(controlplane): make email case insensitive

### LINKED ISSUE #1344: bug: make email address case-insensitive
email address capitalization is kept and it's affecting invitation mechanism. We should normalize email addresses in both user signing and invitation.

migmartri:
- stores email addresses for users and invitations in lowercase
- updates all email addresses in the DB to be lowercase

Fixes #1344


javirln (approved):
LGTM

jiparis (approved):

PR #1348, opened 2024-09-24
fix(jenkins): use WORKSPACE instead of AGENT_WORKDIR

### LINKED ISSUE #1347: bug: some jenkins setups do not include AGENT_WORKDIR
It seems that the variable is configuration-dependent; instead, we should rely on `WORKDIR`.

migmartri:
Fixes #1347


javirln (approved):
LGTM!

Issue #1353, opened 2024-10-02
Error during helm installation with current chart version : 1.101.0


hanygirgis:
Hello,
I'm trying to install chainloop but I'm getting this error : 

```
Error: INSTALLATION FAILED: 1 error occurred:
	* Deployment.apps "chainloop-cas" is invalid: [spec.template.spec.volumes[2].secret.secretName: Required value, spec.template.spec.containers[0].volumeMounts[2].name: Not found: "server-certs"]
```

A values file was passed, which contains a few customCAs : 

```
controlplane:
  customCAs:
    - |-
      -----BEGIN CERTIFICATE-----
      -----END CERTIFICATE-----
cas:
  customCAs:
    - |-
      -----BEGIN CERTIFICATE-----
      -----END CERTIFICATE-----
```

This is the full command-line used : 

```
$ helm install chainloop oci://ghcr.io/chainloop-dev/charts/chainloop --values ./values-noingress.yaml \
>     --set global.openshift=true \
>     --set development=true \
>     --set controlplane.auth.oidc.url=https://hostname/realms/Chainloop \
>     --set controlplane.auth.oidc.clientID=chainloop \
>     --set controlplane.auth.oidc.clientSecret=REDACTED \
>     --set controlplane.auth.oidc.externalURL=https://hostname \
>     --set controlplane.image.repository=repo/chainloop/control-plane \
>     --set controlplane.image.tag=v0.96.13 \
>     --set controlplane.migration.image.repository=repo/chainloop/control-plane-migrations \
>     --set controlplane.migration.image.tag=v0.96.13 \
>     --set cas.image.repository=repo/chainloop/artifact-cas \
>     --set cas.image.tag=v0.96.13 \
>     --set controlplane.tlsConfig.secret.name=grpc-tls-secret \
>     --set cas.tlsConfig.secret.name=grpc-tls-secret
Pulled: ghcr.io/chainloop-dev/charts/chainloop:1.101.0
Digest: sha256:0d3e7eabbd79dddb3712943c2ec0209f97594429d0c114409fb750451747ed9f
Error: INSTALLATION FAILED: 1 error occurred:
	* Deployment.apps "chainloop-cas" is invalid: [spec.template.spec.volumes[2].secret.secretName: Required value, spec.template.spec.containers[0].volumeMounts[2].name: Not found: "server-certs"]
```

migmartri:
[...]

PR #1354, opened 2024-10-02
fix(chart): Map server-certs to correct secret name

javirln:
This patch fixes a regression introduced in the chart that was taking only into account the legacy option for TLS configuration and not the new one.

Refs: #1353


jiparis (approved):

migmartri (approved):
Thanks

PR #1356, opened 2024-10-03
feat(policies): policy groups provider API

jiparis:
This PR adds a new endpoint to query remote policy groups through the providers API. To configure it, a new `url` field was added to the existing `policy_providers` key, to configure both policies and groups in a single host from now on.

```yaml
policy_providers:
  - name: chainloop
    default: true
    host: http://localhost:8002/v1/policies # deprecated, use url instead
    url: http://localhost:8002/v1 # New preferred option, used for both policies (http://localhost:8002/v1/policies) and groups (http://localhost:8002/v1/groups)
```

If the old configuration is found, the control plane will show a message like this in the logs:
```
2024-10-07T13:59:39.451+0200    WARN    {"msg": "the policy provider 'chainloop' is using a deprecated 'host' configuration with value 'http://localhost:8002/v1/policies'. Please use 'url' instead. Configuring 'url' with value 'http://localhost:8002/v1'"}
```
Given this contract:
```yaml
schemaVersion: v1
materials: []
policyGroups:
  - ref: sbom-quality
```

It will use a policy group provided by an external source:
```shell
✗ cldev wf contract update --name myproject-mywf --contract test/contract-group.yaml
INF Contract updated!
┌────────────────────────────────────────────┐
│ Contract                                   │
├──────────────────────┬─────────────────────┤
│ Name                 │ myproject-mywf      │
├──────────────────────┼─────────────────────┤
│ Description          │                     │
├──────────────────────┼─────────────────────┤
│ Associated Workflows │                     │
├──────────────────────┼─────────────────────┤
│ Revision number      │ 12                  │
├──────────────────────┼─────────────────────┤
│ Revision Created At  │ 03 Oct 24 16:12 UTC │
└──────────────────────┴─────────────────────┘
┌────────────────────────┐
│ schemaVersion: v1      │
│ materials: []          │
│ policyGroups:          │
│   - ref: sbom-quality  │
│                        │
└────────────────────────┘
```

When adding a SBOM material, it will evaluated the corresponding policies from the group (as it can be seen in  the debug messages):
```shell
➜  cldev att add --value test/cyclonedx.json
...
DBG loading policy group "sbom-quality" using *policies.ChainloopGroupLoader
DBG loading policy spec "cyclonedx-banned-licenses" using *policies.ChainloopLoader
DBG loading policy spec "cyclonedx-banned-components" using *policies.ChainloopLoader
DBG loading policy spec "cyclonedx-freshness" using *policies.ChainloopLoader
DBG loading policy spec "cyclonedx-banned-licenses" using *policies.ChainloopLoader
INF evaluating policy cyclonedx-banned-licenses against material-1727973358194536000
DBG loading policy spec "cyclonedx-banned-components" using *policies.ChainloopLoader
INF evaluating policy cyclonedx-banned-components against material-1727973358194536000
DBG loading policy spec "cyclonedx-freshness" using *policies.ChainloopLoader
INF evaluating policy cyclonedx-freshness against material-1727973358194536000
WRN found policy violations (cyclonedx-freshness) for material-1727972910786176000
WRN  - SBOM created at: 2024-07-23T15:58:37+00:00 which is too old (freshness limit set to 30 days)
WRN found policy violations (cyclonedx-freshness) for material-1727973358194536000
WRN  - SBOM created at: 2024-07-23T15:58:37+00:00 which is too old (freshness limit set to 30 days)
DBG added to state key=material-1727973358194536000
INF material kind detected kind=SBOM_CYCLONEDX_JSON
```


migmartri (changes requested):
The change of `host` will break compatibility, let's use another key

app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -53,10 +54,44 @@ func (p *PolicyProvider) Resolve(policyName string, token string) (*schemaapi.Po
 
 	// the policy name might include a digest in the form of <name>@sha256:<digest>
 	policyName, digest := policies.ExtractDigest(policyName)
+
+	var policy schemaapi.Policy
+	// old, deprecated
+	endpoint := fmt.Sprintf("%s/%s", p.host, policyName)
+	if p.uri != "" {
+		endpoint = fmt.Sprintf("%s/policies/%s", p.uri, policyName)
```
migmartri, line 62: so this is breaking compatibility or requiring the previous configuration?

wouldn't make more sense if we only have `host` in the provider struct but it's set either by the new URI object or by extracting the URI from the host?

migmartri (changes requested):

app/controlplane/pkg/policies/policyprovider.go
jiparis: I've just did it this way. If no URL is provided, it will do the best effort and extract the URL from there. If the URL is provided, it will use it.
Added a bunch of tests to ensure this behaviour is correct.

app/controlplane/pkg/policies/registry.go
```diff
@@ -17,13 +17,16 @@ package policies
...
 	"fmt"
+	"strings"
 
+	"github.com/go-kratos/kratos/v2/log"
 	"golang.org/x/exp/maps"
 )
 
 type NewRegistryConfig struct {
 	Name    string
 	Host    string
```
migmartri, line 28: indicate that Host is deprecated in a comment

app/controlplane/pkg/policies/registry_test.go
```diff
@@ -73,3 +76,41 @@ func (s *providerTestSuite) TestGetProvider() {
...
+			expectedURL: "http://myhost/v1",
+		},
+		{
+			name:        "only url",
+			url:         "http://myhost/v1",
+			expectedURL: "http://myhost/v1",
+		},
+		{
+			name:        "both",
+			host:        "http://myhost/v1/policies",
```
migmartri, line 99: I'd fail in this case.

migmartri (approved):

migmartri (approved):

Issue #1359, opened 2024-10-06
feat: add support for ZAP material type


migmartri:
We could look into supporting accepting the default zip file and look inside the json for @programname to identify the material type.

It must support both simple and full scans (not sure if there is a difference in format). And we should consider making this two material types, i.e ZAP_DAST_FULL and ZAP_DAST_SIMPLE if we see fit.

Example of full scan [zap_scan.zip](https://github.com/user-attachments/files/17268638/zap_scan.zip)

migmartri:
just researched a little bit if there is a good way of differentiating between scan times and there doesn't seem to be one. Some output from our friend


> In the JSON report generated by OWASP ZAP, there isn't a specific field labeled directly as "Full Scan" or "Simple Scan." However, you can infer the type of scan based on several fields in the report that reflect the depth, scope, and configuration of the scan.
> 
> Here are a few key elements you can check in the JSON report to help determine whether the scan was a Full Scan or a Simple Scan:
> 
> 1. "sites" Field
> Full Scan: If the "sites" field contains multiple entries (representing different URLs, endpoints, and pages within the application), this suggests that ZAP crawled extensively across the entire site.
> Simple Scan: If this field only contains one or a few entries (typically the homepage or a small number of resources), this is indicative of a Simple Scan.

migmartri:
are we uploading the whole zip file or just the json? We should do the former.

javirln:
> are we uploading the whole zip file or just the json? We should do the former.

Only the json file.

migmartri:
>  And we should consider making this two material types, i.e ZAP_DAST_FULL and ZAP_DAST_SIMPLE if we see fit.

did you look at if there is a difference between the two scans?

> Only the json file.

we should do the zip

javirln:
There is no difference between the full and baseline scan more than the number of objects analyzed, for the rest, it's identical.

Also, by default the tool does not export to .zip instead, it lets you specified the name of the files that by default are the ones matching the implementation, in this case, `report_json.json`. My take is that zip file was taken from the GitHub action: https://github.com/zaproxy/action-full-scan

migmartri:
Thanks @javirln for looking into this

PR #1360, opened 2024-10-07
feat(materials): Add support for ZAP reports

javirln:
This patch introduces support for a new material type called ZAP_DAST_ZIP, which represents a ZAP DAST report in ZIP format. The crafter will validate the incoming file as a valid ZIP archive and, using the extracted JSON report, ensure that the program used is ZAP.

Refs: #1359


app/controlplane/api/workflowcontract/v1/crafting_schema_validations.go
```diff
@@ -30,6 +30,7 @@ import (
 // detection. The order of the list is important as it defines the order of the
 // detection process. Normally from most common one to the least common one and weaker validation method.
 var CraftingMaterialInValidationOrder = []CraftingSchema_Material_MaterialType{
+	CraftingSchema_Material_ZAP_DAST_ZIP,
```
migmartri, line 33: why first? Isn't it too expensive? I'd make sure we order by less expensive to more

pkg/attestation/crafter/materials/zap.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return &ZAPCrafter{
+		backend:       backend,
+		crafterCommon: &crafterCommon{logger: l, input: materialSchema},
+	}, nil
+}
+
+// Craft will extract the ZAP JSON report from the zip file and upload it to the CAS
+func (i *ZAPCrafter) Craft(ctx context.Context, filePath string) (*api.Attestation_Material, error) {
+	archive, err := zip.OpenReader(filePath)
```
migmartri, line 65: could we check if it's a zip file?

app/controlplane/api/workflowcontract/v1/crafting_schema_validations.go
javirln: Yes, I don't know why I placed it there, it should be lower on the list.

pkg/attestation/crafter/materials/zap.go
javirln: That OpenReader checks already that the incoming file is a zip file but we can do a manual check in any case.

migmartri (commented):
ptal at my comment

PR #1361, opened 2024-10-07
fix(materials): Remove check on zip files

javirln:
This patch removes the check for zip files in ZAP material type since it will load the whole file into memory.

Refs: #1359


migmartri (commented):
Please upload the whole zip file. Thanks

migmartri (approved):

PR #1363, opened 2024-10-07
feat(materials): Add support for TwistCTL scan reports

### LINKED ISSUE #1358: feat: add support for twistctl material type

javirln:
This patch implements the detection of TwistCTL scan reports in a best effort way.

Closes #1358


migmartri (approved):
Thanks

PR #1377, opened 2024-10-09
feat(policies): allow to query organization policies and groups through providers

### LINKED ISSUE #1376: Allow organization scoped policies in policy providers
The goal is to allow to specify an organization name in the policy reference. This will enable policy providers to implement policy sharing mechanisms between organizations, given that the provider call is including a Bearer token with the organization ID.

Current policy reference looks like this:
```
ref: policy_name # (assuming default provider)
```
The proposal is to extend this format to allow to provide an org name to the reference, looking like this:
```
ref://[provider_name:][org_name/]policy_name 
```
(both provider and org are optional)

Examples:
```
ref://builtin:policy_name # "builtin" provider, no org
ref://builtin:myorg/policy_name # "builtin" provider, org-scoped policy
ref://myorg/policy_name # default provider, org-scoped policy
ref://policy_name # default provider, no org
```

jiparis:
This PR will allow namespaced policies in policy references:
```
schemaVersion: v1
materials: []
policies:
  materials:
    - ref: my-org/secrets-detection
```

Policy providers will get a new query param `organization_name` with the proper value set.
`GET https://myprovider.com/my-api/policies/secrets-detection?organization_name=my-org`

Closes #1376


app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -87,25 +87,32 @@ func (p *PolicyProvider) ResolveGroup(groupName string, token string) (*schemaap
...
 	}
 
+	query := uri.Query()
 	if digest != "" {
-		uri.RawQuery = fmt.Sprintf("digest=%s", digest)
+		query.Set("digest", digest)
 	}
 
+	if orgName != "" {
+		query.Set("org", orgName)
```
javirln, line 111: The query param is called `organization_name`

javirln (approved):
Awesome!!

Issue #1375, opened 2024-10-09
feat: make workflow project scoped


migmartri:
We want to allow people to create workflows with the same name in the scope of the projects. 

So for example we could have project-1/build-and-test or project-2/build-and-test

note: This might potentially mean a breaking change w.r.t initializing workflows.

```[tasklist]
### Phase 0
- [x] Rename project by product
- [x] Release CLI with –project support
- [x] Double-check that project-name is also DNS-1131 compatible
- [x] migrate workflow create --product flag
- [x] migrate --product in att init
- [x] update labs https://github.com/chainloop-dev/labs/pull/34
- [x] migrate gitlab
- [x] internal repos
- [ ] https://github.com/chainloop-dev/chainloop/issues/1390
```

```[tasklist]
### Phase 1
- [x] migrate dagger
- [x] Change DB constraint
- [x] Upgrade Dagger module
- [x] Update CLI accordingly to start attestation with the project
- [x] cleanup and enforce constraint
- [x] remove dagger optional project
- [x] remove optional project in CLI
```

PR #1388, opened 2024-10-11
chore: add index workflowName,project

migmartri:
We are starting to do queries based on name + project already

Ref #1375


app/controlplane/pkg/data/ent/schema/workflow.go
```diff
@@ -70,6 +70,10 @@ func (Workflow) Edges() []ent.Edge {
 
 func (Workflow) Indexes() []ent.Index {
 	return []ent.Index{
+		index.Fields("name", "project").Edges("organization").Unique().Annotations(
+			entsql.IndexWhere("deleted_at IS NULL"),
```
javirln, line 74: I love this!

javirln (approved):
LGTM

PR #1399, opened 2024-10-15
feat(contract): Introduce workflow references

javirln:
This patch modifies the response of `controlplane.v1.WorkflowContractService/Describe` to expand the content of a Workflow by providing:
- ID
- Name
- Project

Also modifies the output of `workflow contract describe --name` in the CLI to append the project name associated with the workflow and print them in separated lines.

It marks as deprecated `workflow_names` that will be removed in a later update.

Ref: #1375


app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -374,17 +374,21 @@ func contractInOrgQuery(ctx context.Context, q *ent.OrganizationQuery, orgID uui
...
 	}
 
 	c.LatestRevision = version.Revision
 	return c
 }
 
-// get the list of workflows associated with a given contract
-func getWorkflowNames(ctx context.Context, schema *ent.WorkflowContract) ([]string, error) {
+// getWorkflowNamesReferences get the list of workflows associated with a given contract
+func getWorkflowNamesReferences(ctx context.Context, schema *ent.WorkflowContract) ([]*biz.WorkflowRef, error) {
```
jiparis, line 391: should it be `getWorkflowNamesReference` --> `getWorkflowReferences`?

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -397,10 +401,14 @@ func getWorkflowNames(ctx context.Context, schema *ent.WorkflowContract) ([]stri
 		}
 	}
 
-	names := make([]string, 0, len(workflows))
```
migmartri, line 400: this will break compatibility with current clients? Do you think it would be a problem? 

The reason I am asking is that in the API you mark `names` as deprecated instead of deleting it but in practice it will not return any data does it?

So we need to make a decision wether maintain both or remove it altogether, in general I'd say return both marking it as deprecated

app/cli/cmd/workflow_contract_describe.go
```diff
@@ -100,3 +101,19 @@ func contractDescribeTableOutput(contractWithVersion *action.WorkflowContractWit
...
+		for _, w := range contract.WorkflowRefs {
+			workflows = append(workflows, fmt.Sprintf("%s/%s", w.ProjectName, w.Name))
+		}
+		// sort the workflows
+		sort.Strings(workflows)
+
+		return strings.Join(workflows, "\n")
+	}
+
+	return strings.Join(contract.Workflows, ", ")
```
migmartri, line 118: I do not understand this method, is it maintaining both ways of showing it or smth?

app/cli/cmd/workflow_contract_list.go
```diff
@@ -50,7 +50,12 @@ func contractListTableOutput(contracts []*action.WorkflowContractItem) error {
 
 	t.AppendHeader(table.Row{"Name", "Latest Revision", "Created At", "# Workflows"})
 	for _, p := range contracts {
-		t.AppendRow(table.Row{p.Name, p.LatestRevision, p.CreatedAt.Format(time.RFC822), len(p.Workflows)})
+		wfLen := len(p.Workflows)
```
migmartri, line 53: I do not think the client needs to take into account both modes, it is released together (client and server), you need to think more about previous clients.

migmartri (changes requested):
See my comments. In general it looks good except I think there is some mistmatch on retrocompatibility

- The API seems to have the old property but I am not sure if it's been set.
- This CLI code doesn't need to take into account for the old API, it's released together and we can assume (for now), that as a minimum, our users should deploy the version

migmartri (approved):
ah no, I was wrong, you create teh workflow names at the service layer, makes sense!

re: CLI compatibility, I think what you did is correct but since we are pre 1.0, we can assume that updates are required, not sure what you think

javirln:
> ah no, I was wrong, you create teh workflow names at the service layer, makes sense!

Yes, it adds a new parameter for previous compatibility at the service layer, the internals are modified to use the new structure that is used by the old structure to be populated.

> re: CLI compatibility, I think what you did is correct but since we are pre 1.0, we can assume that updates are required, not sure what you think

I can remove the fallbacks, sure, we can assume updates are required.

PR #1400, opened 2024-10-15
feat(policies): support new policy output format

### LINKED ISSUE #1321: Improve policy results format
Currently, policy evaluations are reported as a single `violations` rule result, that will contain elements in the case of non compliance. 
However, if the policy couldn't be evaluated (because of invalid input or some conditions were not met) it would still report an empty `violations` set, which is wrong, since it couldn't be evaluated.
There are three variables in policy evaluations:
* does it need to be evaluated? : we know this because of the policy selection logic (by kind and/or material name)
* was it evaluated?: this would come from the Rego itself
* violations

The proposal is to add a structured output to policies, leaving room for future improvements:
```
{
    "result": {
        "evaluated": true,
        "violations": []
    },
    "violations": [] // <-- current policy output
  }
```
How `evaluated` is generated, it's up to each policy implementation. For example, a CycloneDX policy would set it if there are `components` in the report, or a SARIF policy for trivy would set it if `runs[].tool.driver.name == "Trivy"`.

jiparis:
This PR adds support to a new `result` object in Rego output, with the goal of minimizing false positives and provide a foundation for next improvements.
Closes #1321


pkg/policies/engine/rego/rego.go
```diff
@@ -145,7 +151,51 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
+	}, nil
+}
+
+func parseResultRule(res rego.ResultSet, policy *engine.Policy) (*engine.EvaluationResult, error) {
+	result := &engine.EvaluationResult{Violations: make([]*engine.PolicyViolation, 0)}
+	for _, exp := range res {
+		for _, val := range exp.Expressions {
+			ruleResult, ok := val.Value.(map[string]any)
+			if !ok {
+				return nil, fmt.Errorf("failed to evaluate policy evaluation result: %s", val.Text)
```
javirln, line 167: Shall we create a custom error when the expected fields are not present in the script?

pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
```diff
@@ -140,6 +140,12 @@ message PolicyEvaluation {
   // material type, if any, of the evaluated policy
   workflowcontract.v1.CraftingSchema.Material.MaterialType type = 8;
 
+  // whether this evaluation was skipped or not (because of an invalid input, for example)
+  bool skipped = 13;
+
+  // Evaluation messages, intended to communicate evaluation errors (invalid input)
+  repeated string messages = 14;
```
migmartri, line 147: should we remove messages for now, it might get confusing as we discussed the other day. i.e SARIF from different tools, telling you that it was not from that tool when in fact it went through another path?

pkg/policies/engine/rego/rego.go
```diff
@@ -43,8 +43,8 @@ const (
 	// EnvironmentModePermissive allows all operations on the compiler
 	EnvironmentModePermissive EnvironmentMode = 1
 	inputArgs                                 = "args"
-	mainRule                                  = "violations"
-	deprecatedMainRule                        = "deny"
+	violationsRule                            = "violations"
+	resultRule                                = "result"
```
migmartri, line 47: nitpick

```suggestion
	rootRule                                = "result"
```

pkg/policies/engine/rego/rego.go
```diff
@@ -107,22 +107,28 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
-	if err := executeQuery(mainRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
+	if err := executeQuery(resultRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
 		return nil, err
 	}
 
 	// If res is nil, it means that the rule hasn't been found
 	if res == nil {
 		// Try with the deprecated main rule
-		if err := executeQuery(deprecatedMainRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
+		if err := executeQuery(violationsRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
```
migmartri, line 117: add a TODO here to remove it in the future?

pkg/policies/engine/rego/rego.go
```diff
@@ -43,8 +43,8 @@ const (
 	// EnvironmentModePermissive allows all operations on the compiler
 	EnvironmentModePermissive EnvironmentMode = 1
 	inputArgs                                 = "args"
-	mainRule                                  = "violations"
-	deprecatedMainRule                        = "deny"
+	violationsRule                            = "violations"
```
migmartri, line 46: I'd add a comment here saying that violations rule is deprecated in favor of results

pkg/policies/engine/rego/rego.go
```diff
@@ -107,22 +107,28 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
 	if res == nil {
 		// Try with the deprecated main rule
-		if err := executeQuery(deprecatedMainRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
+		if err := executeQuery(violationsRule, r.OperatingMode == EnvironmentModeRestrictive); err != nil {
 			return nil, err
 		}
 
 		if res == nil {
-			return nil, fmt.Errorf("failed to evaluate policy: no '%s' nor '%s' rule found", mainRule, deprecatedMainRule)
+			return nil, fmt.Errorf("failed to evaluate policy: neither '%s' nor '%s' rule found", resultRule, violationsRule)
```
migmartri, line 122: I actuallyb liked the name of the variable before, it was clearly stating the fact that was deprecated. Now it's not clear if violations is smth we should always have or not

pkg/policies/engine/rego/rego.go
```diff
@@ -107,22 +107,28 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
 		}
+
+		return parseViolationsRule(res, policy)
 	}
 
+	return parseResultRule(res, policy)
+}
+
+func parseViolationsRule(res rego.ResultSet, policy *engine.Policy) (*engine.EvaluationResult, error) {
```
migmartri, line 131: please add some context about this, why is this best effort. It's important that it's clear in the code this is a migration/compatibility path

pkg/policies/engine/rego/rego.go
```diff
@@ -145,7 +151,51 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
 
-	return violations, nil
+	return &engine.EvaluationResult{
+		Violations: violations,
+		Skipped:    false, // best effort
+		Message:    "",
+	}, nil
+}
+
+func parseResultRule(res rego.ResultSet, policy *engine.Policy) (*engine.EvaluationResult, error) {
```
migmartri, line 161: could you elaborate here in a comment the logic used to calculate skipped/result?

Unit testing this method would be useful too.

pkg/policies/engine/rego/rego.go
```diff
@@ -145,7 +151,51 @@ func (r *Rego) Verify(ctx context.Context, policy *engine.Policy, input []byte,
...
+			if !ok {
+				return nil, fmt.Errorf("failed to evaluate 'message' field in policy evaluation result: %s", val.Text)
+			}
+
+			violations, ok := ruleResult["violations"].([]any)
+			if !ok {
+				return nil, fmt.Errorf("failed to evaluate 'violations' field in policy evaluation result: %s", val.Text)
+			}
+
+			result.Skipped = skipped
```
migmartri, line 185: I do not think I understand this. Is this doing the `nor` of the different paths?

pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
jiparis: Moved to "skip_reason", only available when the policy is skipped

pkg/policies/engine/rego/rego.go
jiparis: Merge doesn't happen at this layer. It's done in the `policies` package, since it's what the crafter uses to verify materials.

pkg/policies/engine/rego/rego.go
jiparis: This result represents a single evaluation. Merge is done in `policies` package. See `evaluatePolicyAttachment` method.

PR #1414, opened 2024-10-19
feat: remove workflow-name org scoping

### LINKED ISSUE #1375: feat: make workflow project scoped

migmartri:
This patch allows users to create workflows with the same name in the same org as long as they are in different projects. 

It also enforces now the requirement of passing `project name` during attestation and updates documentation and dagger module.

```
go run main.go wf ls
┌────────┬─────────┬────────────┬────────┬─────────────┬─────────────────┬─────────────────────┐
│ NAME   │ PROJECT │ CONTRACT   │ PUBLIC │ RUNNER      │ LAST RUN STATUS │ CREATED AT          │
├────────┼─────────┼────────────┼────────┼─────────────┼─────────────────┼─────────────────────┤
│ build  │ p2      │ p2-build   │ false  │ Unspecified │ success         │ 19 Oct 24 14:29 UTC │
│ build  │ p1      │ p1-build   │ false  │ Unspecified │ initialized     │ 19 Oct 24 14:29 UTC │
```

```
go run main.go wf create --name build --project p2
ERR duplicated: workflow "build" in project "p2" already exists
```

closes #1375


app/controlplane/pkg/data/ent/migrate/migrations/20241019145455.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "workflows" table
+ALTER TABLE "workflows" ALTER COLUMN "project" SET NOT NULL;
```
javirln, line 2: We don't need a migration of previous workflows right?

javirln (approved):

app/controlplane/pkg/data/ent/schema/workflow.go
```diff
@@ -35,7 +35,7 @@ type Workflow struct {
 func (Workflow) Fields() []ent.Field {
 	return []ent.Field{
 		field.String("name").Immutable(),
-		field.String("project").Optional(),
+		field.String("project"),
```
jiparis, line 38: What are we doing with the current data? Will we migrate it?

jiparis (approved):

app/controlplane/pkg/data/ent/schema/workflow.go
migmartri: There is no need to migrate since all fields, in fact, have a project name, enforced somewhere else in the stack, not in the DB until now

app/controlplane/pkg/data/ent/migrate/migrations/20241019145455.sql
migmartri: No because they happened to be set and forced in another layer.

PR #1416, opened 2024-10-20
feat: auto-create workflow during attestation

### LINKED ISSUE #1415: feat: automatically create workflow during attestation
Instead of requiring the workflow to be created in advance, we could allow att init to do it for you.

migmartri:
It adds support to automatically, during `att init`, to create the workflow referenced by its name, project and optional contract name.

I added this feature as an additional API endpoint in the `attestation service` instead of re-using the `workflow` service because I want to make sure this feature can be authenticated, authorized, and potentially toggled in the future, using the attestation token, not a user token.

In fact, I'd say that moving forward, I'd remove the ability of the API token to create workflows by default through regular `wf create`

Closes #1415


migmartri:
Tests are failing until we merge and rebase #1414

javirln (approved):

app/cli/internal/action/attestation_init.go
```diff
@@ -70,7 +70,7 @@ func NewAttestationInit(cfg *AttestationInitOpts) (*AttestationInit, error) {
 }
 
 // returns the attestation ID
-func (action *AttestationInit) Run(ctx context.Context, contractRevision int, projectName, workflowName string) (string, error) {
+func (action *AttestationInit) Run(ctx context.Context, contractRevision int, projectName, workflowName, newWorkflowContractName string) (string, error) {
```
jiparis, line 73: This changes the current behaviour a little bit. Currently, contract is not mandatory.

migmartri: It is not mandatory here either. The name can be empty string

PR #1434, opened 2024-10-23
feat(attestation): Support GHAS report types

### LINKED ISSUE #1429: Support github code scanning alerts as a material type
Github scanning APIs provide a way to consume automatic reports for [Code scanning](https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28), [Secret scanning](https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28) and [Vulnerabilities (Dependabot)](https://docs.github.com/en/rest/dependabot/alerts?apiVersion=2022-11-28), among others.
The goal of this task is to create new material types in chainloop to provide direct support for these reports.

jiparis:
This PR adds new kinds for GHAS reports taken from [GH API](https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28) for code, secret and dependency scans.

I've intentionally kept these types out of autodetection feature in the CLI, since they accept empty reports (empty JSON arrays), and result would be ambiguous. We might want to enable autodetection for non-empty reports, but that would require changes outside the scope of this change.

Tests are pending.

Closes #1429


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -113,6 +113,14 @@ message CraftingSchema {
       BLACKDUCK_SCA_JSON = 18;
       // Twistcli scan output in json format: https://docs.prismacloud.io/en/compute-edition/30/admin-guide/tools/twistcli-scan-images
       TWISTCLI_SCAN_JSON = 19;
+
+      // GitHub Advanced Security API reports
+      // https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28
+      GHAS_CODE_SCAN = 20;
```
migmartri, line 119: could you add these types to this json that's then used to generate the docs? https://github.com/chainloop-dev/chainloop/blob/main/docs/docs/partials/materials.json

migmartri (approved):

javirln (approved):

Issue #1433, opened 2024-10-23
normalize projects and versions


migmartri:
- [x] setup new updates in the datamodel

![image](https://github.com/user-attachments/assets/88c12920-6627-45b4-90b0-9835b6e13cf0)

- [x] migrate existing data
- [x] store denormalized project
- [ ] enable versions with default sink version

PR #1436, opened 2024-10-23
feat: project normalization

migmartri:
This PR normalizes the projects column into its own entity. 

- Create new projects table associated with workflows and organizations
- Migrates the existing data into the normalized table
- Updates the DB constraints accordingly 
- For new WF, the result is also now normalized
- Supported soft-deletion

From the biz/API layers POV, there is no difference, all the changes happen in the data layer.

NOTE for reviewers

There is a lot of generated code because I've enabled the On Conflict functionality

image

Refs #1433


app/controlplane/pkg/data/workflow.go
```diff
@@ -54,9 +56,34 @@ func (r *WorkflowRepo) Create(ctx context.Context, opts *biz.WorkflowCreateOpts)
...
+	defer func() {
+		if err != nil {
+			if rbErr := tx.Rollback(); rbErr != nil {
+				r.log.Errorf("rolling back transaction: %v", rbErr)
+			}
+		}
+	}()
+
+	// Find or create project.
+	projectID, err := tx.Project.Create().SetName(opts.Project).SetOrganizationID(orgUUID).
```
migmartri, line 74: @javirln @jiparis this is the interesting bit.

To create a workflow and a project only if it does not exist

app/controlplane/pkg/data/workflowrun.go
```diff
@@ -236,7 +236,7 @@ func entWrToBizWr(wr *ent.WorkflowRun) *biz.WorkflowRun {
 	}
 
 	if wf := wr.Edges.Workflow; wf != nil {
-		w, _ := entWFToBizWF(wf, nil)
+		w, _ := entWFToBizWF(context.TODO(), wf, nil)
```
javirln, line 239: TODO?

javirln (approved):
LGTM! Just small comment

app/controlplane/pkg/data/workflowrun.go
migmartri: yeah, we do not have the context there and I thought it might not be worth updating all the code, but let me double check

app/controlplane/pkg/data/ent/schema/project.go
```diff
@@ -0,0 +1,65 @@
...
+// Project holds the schema definition for the Project entity.
+type Project struct {
+	ent.Schema
+}
+
+// Fields of the Project.
+func (Project) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).Default(uuid.New).Unique(),
+		field.String("name").Immutable().Validate(biz.ValidateIsDNS1123),
```
migmartri, line 39: @jiparis @javirln another interesting bit, to validate the name in the schema as well

app/controlplane/pkg/data/workflowrun.go
migmartri: updated, thanks for pointing it out

PR #1444, opened 2024-10-25
feat(policies): render group name in policy evaluations

jiparis:
Small PR to add the group name to policy evaluations, when evaluated in the context of a group.


migmartri (approved):
Thanks

migmartri (commented):
Is this the FQDN of the group, meaning potentially org/group_name?

jiparis:
> Is this the FQDN of the group, meaning potentially org/group_name?

Let me check if we can store the whole group reference instead of the name.

jiparis:
Closed in favour of https://github.com/chainloop-dev/chainloop/pull/1487

PR #1445, opened 2024-10-28
feat: add versions data model

migmartri:
- Adds project_versions table associated with projects and workflow runs

image

- Creates an empty version object for each project. This empty version means `unversioned`.
- Updates each existing run to join this version object
- Finds or Creates the version during workflow run creation
- exposes basic information in the API

A following PR will contain the ability of providing the --version in the command line as well as exposing the info in the CLI/API 

Ref #1433


app/controlplane/pkg/data/workflowrun.go
```diff
@@ -240,6 +277,17 @@ func entWrToBizWr(ctx context.Context, wr *ent.WorkflowRun) *biz.WorkflowRun {
 		r.Workflow = w
 	}
 
+	// Load version preloaded or otherwise query it
```
jiparis, line 280: This will potentially introduce a N+1 problem in the data layer (when listing workflow runs, for example, we will make this query multiple times). I'd suggest preload it with a inner join.

migmartri: it is, in practice, preloaded (see code above).

image

This check is to make sure the code works regardless of the preloading status

jiparis (approved):

PR #1446, opened 2024-10-28
feat(workflow): Allow to filter workflows by project reference

javirln:
This patch adds an optional parameter to be sent to the `WorkflowList` endpoint from the `WorkflowService` to only retrieve `Workflows` within a specific project that can be reference by name or id.

Ref: #1433


app/controlplane/api/controlplane/v1/workflow.proto
```diff
@@ -101,7 +101,16 @@ message WorkflowServiceDeleteRequest {
 
 message WorkflowServiceDeleteResponse {}
 
-message WorkflowServiceListRequest {}
+message WorkflowServiceListRequest {
+  optional string project_name = 1 [(buf.validate.field) = {
```
migmartri, line 105: How about supporting smth like project_ref that can be either the id or the name? 

The reason is to make this API suitable for clients that use IDs instead of names?

javirln: Sure thing will do!

app/controlplane/internal/service/workflow.go
```diff
@@ -103,13 +105,33 @@ func (s *WorkflowService) Update(ctx context.Context, req *pb.WorkflowServiceUpd
...
+			ID:   req.GetProjectReference().GetEntityId(),
+			Name: req.GetProjectReference().GetEntityName(),
+		})
+		if err != nil {
+			return nil, handleUseCaseErr(err, s.log)
+		}
+
+		// If the project is not found, return an empty response
+		if pro == nil {
+			return &pb.WorkflowServiceListResponse{}, nil
```
migmartri, line 128: return 404 instead? Why return an empty list? I am not sure what's the right way of doing this, just curious

migmartri (approved):
Thanks

app/controlplane/internal/service/workflow.go
migmartri: I can see 404 if the workflow is not found when filtering runs?

```
chainloop wf run ls --workflow foo --project bar
ERR failed to get workflow: workflow not found
```

PR #1449, opened 2024-10-28
fix(policies): Fix policy reference

### LINKED ISSUE #1448: bug: wrong policy reference in attestations
Looks like policy reference URIs are broken. This is from a recent attestation:
```
        {
...
            "digest": {
              "sha256": "c1ad64c21bac1bc5df57ace6f4d432eaaf5b8ef1f3ae28b98c6888c3f4e9ad31"
            },
            "name": "chainloop://api.app.chainloop.dev/https://api.app.chainloop.dev/v1/policies/cves-in-kev"
          },
...
        }
```

jiparis:
This PR fixes the way policy references are calculated in the CLI and does some refactor by removing protobuf dependency in this calculation. Once resolved, a descriptor is created containing the URI, Name, OrgName (for custom remote policies) and Digest, which is then used through the attestation process. Previously, a in-toto protobuf message was used instead.

Fixes #1448

### DIFF
--- a/app/controlplane/pkg/policies/policyprovider.go
+++ b/app/controlplane/pkg/policies/policyprovider.go
@@ -68,12 +68,16 @@ func (p *PolicyProvider) Resolve(policyName, orgName, token string) (*schemaapi.
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to resolve policy: %w", err)
 	}
-	ref, err := p.queryProvider(endpoint, digest, orgName, token, &policy)
+	url, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error parsing policy provider URL: %w", err)
+	}
+	providerDigest, err := p.queryProvider(url, digest, orgName, token, &policy)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to resolve policy: %w", err)
 	}
 
-	return &policy, ref, nil
+	return &policy, createRef(url, policyName, providerDigest, orgName), nil
 }
 
 // ResolveGroup calls remote provider for retrieving a policy group definition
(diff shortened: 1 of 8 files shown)


app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -114,75 +116,60 @@ func (p *PolicyProvider) queryProvider(path, digest, orgName, token string, out
...
-		return nil, fmt.Errorf("error parsing policy provider URL: %w", err)
-	}
-
-	if uri.Host == "" {
-		return nil, fmt.Errorf("invalid policy provider URL")
-	}
-
-	if path == "" || digest == "" {
-		return nil, fmt.Errorf("both path and digest are mandatory")
+func createRef(policyURL *url.URL, name, digest, orgName string) *PolicyReference {
```
migmartri, line 166: I am assuming that this method will be used also for groups?

pkg/policies/loader.go
```diff
@@ -44,20 +43,60 @@ const (
 
 // Loader defines the interface for policy loaders from contract attachments
 type Loader interface {
-	Load(context.Context, *v1.PolicyAttachment) (*v1.Policy, *v12.ResourceDescriptor, error)
+	Load(context.Context, *v1.PolicyAttachment) (*v1.Policy, *PolicyDescriptor, error)
+}
+
+// PolicyDescriptor Represents a policy reference. Used as FQ references.
+type PolicyDescriptor struct {
```
migmartri, line 50: so in the end you moved away from intoto descriptor why?

migmartri (approved):

pkg/policies/loader.go
jiparis: This is internal. The final representation in the attestation is still an in-toto descriptor, but I didn't want to get locked into any structure in the internal logic for policies. Actually I think we should apply the same to the rest of the policies logic.

migmartri: oh, I get it, and agree!

PR #1455, opened 2024-10-30
feat(policies): materials in policy groups

jiparis:
This PR adds the ability to configure required materials in policy groups, in the same way it's done through contracts. It reuses the same protobuf messages for this purpose.
Materials and policies are now configured together using the following schema:
```yaml
kind: PolicyGroup
metadata:
  name: sbom-quality
  description: This policy group applies a number of SBOM-related policies
  annotations:
    category: SBOM
spec:
  policies:
    attestation:
      - ref: file://testdata/s.yaml
    materials:
      - name: sbom
        type: SBOM_CYCLONEDX_JSON
        policies:
          - ref: file://testdata/required-sbom-tool.yaml
            with:
               tool: trivy
      - name: container
        type: CONTAINER_IMAGE
```
group attachments are configured in the contract as always:
```yaml
schemaVersion: v1
policyGroups:
  - ref: file://test/policy-group.yaml
```
During `att init`, the CLI will download all attached groups and compute the final list of required materials:
```
> chainloop --debug att init --workflow mywf --project myproject --replace
DBG Retrieving attestation definition
DBG workflow contract and metadata retrieved from the control plane
DBG loading policy group "file://test/policy-group.yaml" using *policies.FileGroupLoader
DBG adding materials from policy group 'sbom-quality'
...
┌────────────────────────────────┐
│ Materials                      │
├──────────┬─────────────────────┤
│ Name     │ container           │
│ Type     │ CONTAINER_IMAGE     │
│ Set      │ No                  │
│ Required │ Yes                 │
├──────────┼─────────────────────┤
│ Name     │ sbom                │
│ Type     │ SBOM_CYCLONEDX_JSON │
│ Set      │ No                  │
│ Required │ Yes                 │
└──────────┴─────────────────────┘
```

Warning: This is a breaking change in policy groups schemas


app/cli/internal/action/attestation_init.go
```diff
@@ -186,3 +196,47 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
+}
+
+// override existing materials and return new ones
+func overrideMaterials(group *v1.PolicyGroup, fromContract []*v1.CraftingSchema_Material, logger *zerolog.Logger) []*v1.CraftingSchema_Material {
+	toAdd := make([]*v1.CraftingSchema_Material, 0)
+	for _, groupMaterial := range group.GetSpec().GetPolicies().GetMaterials() {
+		// check if material already exists in the contract, and override it
+		found := false
+		for i, mat := range fromContract {
+			if mat.GetName() == groupMaterial.GetName() {
```
migmartri, line 228: I personally think it should be the other way around. Contracts are more specific than groups. Groups are meant to be shared, but if you have a contract you could override. 

Any thoughts on why to do it the other way around?

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -272,7 +275,7 @@ message PolicyGroup {
   }
 
   message PolicyGroupPolicies {
-    repeated PolicyAttachment materials = 1;
+    repeated CraftingSchema.Material materials = 1;
```
migmartri, line 278: this looks clean :)

migmartri (approved):
Awesome! 

See my comment about changing the override order and let me know what you think

javirln (approved):

app/cli/internal/action/attestation_init.go
jiparis: It was about enforcing the materials, but you are right, the other way around is way more flexible for users.

PR #1459, opened 2024-10-30
chore(db): delete dangling versions if project is deleted

migmartri:
The fully deletion of an organization wasn't working as expected if there were versions in a given project.

```
failed to delete org: ent: constraint failed: ERROR: update or delete on table "projects" violates foreign key constraint "project_versions_projects_versions" on table "project_versions" (SQLSTATE 23503)
```


jiparis (approved):
When is it expected to remove organizations? Should we soft-delete instead?

PR #1460, opened 2024-10-31
feat(policies): use raw response from providers if available

jiparis:
A new raw format is supported for policy providers. Instead of returning an embedded policy or policy group, it can be provided in raw format in a base64 encoded JSON, YAML or CUE:
```json
GET /policies/my-policy
{
  "digest": "sha256:...",
  "raw": {
    "format": "FORMAT_JSON",
    "body": "eyJhcGlWZXJzaW9uIjoid29ya2Zsb3djb250cmFjdC5jaGFpb..."
  }
}
```
This will provider flexibility on the handling logic and adapt to future changes in the policy schema.
The previous `data` field is still supported.


app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -150,19 +159,59 @@ func (p *PolicyProvider) queryProvider(url *url.URL, digest, orgName, token stri
...
+			return "", fmt.Errorf("error unmarshalling policy response: %w", err)
+		}
 	}
 
 	return response.Digest, nil
 }
 
+func unmarshalFromRaw(raw *RawMessage, out proto.Message) error {
+	switch raw.Format {
+	case "FORMAT_JSON":
```
javirln, line 184: Is the actual format coming in lowercase or uppercase?

jiparis: upcase, that's how protobuf does it. 
```
GET /policies/my-policy
{
  "digest": "sha256:...",
  "raw": {
    "format": "FORMAT_JSON",
    "body": "eyJhcGlWZXJzaW9uIjoid29ya2Zsb3djb250cmFjdC5jaGFpb..."
  }
}
```

app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -150,19 +159,59 @@ func (p *PolicyProvider) queryProvider(url *url.URL, digest, orgName, token stri
...
-		return "", fmt.Errorf("error unmarshalling policy response: %w", err)
+		if err := protojson.Unmarshal(jsonPolicy, out); err != nil {
+			return "", fmt.Errorf("error unmarshalling policy response: %w", err)
+		}
 	}
 
 	return response.Digest, nil
 }
 
+func unmarshalFromRaw(raw *RawMessage, out proto.Message) error {
```
migmartri, line 182: It seems that this method is becoming more common. Are you sure it doesn't exist already? I originally implemented this for contracts

app/controlplane/pkg/policies/policyprovider.go
```diff
@@ -150,19 +159,59 @@ func (p *PolicyProvider) queryProvider(url *url.URL, digest, orgName, token stri
...
-	if err != nil {
-		return "", fmt.Errorf("error marshalling policy response: %w", err)
-	}
+	// if raw message is provided, just interpret it as a base64 encoded string
+	if response.Raw != nil {
+		if err := unmarshalFromRaw(response.Raw, out); err != nil {
+			return "", fmt.Errorf("error unmarshalling policy response: %w", err)
+		}
+	} else if response.Data != nil {
+		// extract the policy payload from the query response
```
migmartri, line 168: Mark this path as deprecated

migmartri (approved):

app/controlplane/pkg/policies/policyprovider.go
jiparis: I've extracted that logic to a common package

Issue #1466, opened 2024-11-04
Include descriptions on project data model


javirln:
We have recently introduced the project data model but in order to give it more meaning, the goal of the task would be to add a description on the data model.

PR #1468, opened 2024-11-04
feat(projects): Add description column

javirln:
This patch adds a `description` column to the table `projects`.

Ref #1466


migmartri (approved):
it's missing make generate

PR #1482, opened 2024-11-06
feat(policies): arguments for policies and groups with input spec

### LINKED ISSUE #1425: Add policy parameters to the spec
Currently, parameters are implicit and there is no way to enforce nor validate policy attachments.

A first version could consist on a plain list of inputs, without type information (all strings), following the style of Github workflows, as in this example:

```
kind: Policy
metadata:
  name:  banned-components
spec:
  inputs:
      - name: author
        description: "commit author"
        required: true
        default: "myteam@acme.org"

  policies:
    - kind: SBOM_CYCLONEDX_JSON
      path: script.rego
```

jiparis:
This PR adds the ability to specify supported inputs for a given policy or group. Now, policy and groups specs have an optional `inputs` field:
```
...
spec:
  inputs:
    - name: version
      description: Supported CycloneDX version 
      required: true
  policies:
    - kind: SBOM_CYCLONEDX_JSON
      embedded: |
        package main
 ...
```
Inputs have the following properties:
* name: the input short name. It will be matched against the policy arguments
* description
* required
* default: a default value to be used when "required" is not true

Groups can also have inputs. The interesting part is that policy attachments within a group can use those inputs to interpolate arguments, using Go templating. An example:
```
kind: PolicyGroup
metadata:
  name: group-with-inputs
  description: test group
  annotations:
    category: test
spec:
  inputs:
    - name: username
      required: false
      default: "foo"
  policies:
    materials:
      - name: sbom
        type: SBOM_CYCLONEDX_JSON
        policies:
          - ref: file://testdata/policy_with_inputs.yaml
            with:
              email: "{{ .username }}@chainloop.dev"
```
Here, the argument passed to the policy will be "foo@chainloop.dev" by default, or "devel@chainloop.dev" if the "username" argument is set to "devel".

Finally, arguments passed to groups and policies are matched with the spec inputs. For example, the lack of a required input will make the attestation to fail.
Passing arguments to a group:
```
# In the contract:
materials: []
policyGroups:
  - ref: "file://group_with_inputs.yaml"
    with:
      username: "devel"
```
Closes #1425


javirln (approved):
Awesome job here!!

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -327,11 +329,22 @@ func (uc *WorkflowContractUseCase) findPolicy(att *schemav1.PolicyAttachment, to
...
 	}
 
-	// Otherwise, don't return an error, as it might consist of a local policy, not available in this context
-	return nil, nil
+	if policy != nil {
+		// validate policy arguments
+		with := att.GetWith()
+		for _, input := range policy.GetSpec().GetInputs() {
+			_, ok := with[input.GetName()]
+			if !ok && input.GetRequired() {
```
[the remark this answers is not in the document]
jiparis: if it's required, what does it mean to have a default value? does it make sense at all?

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -346,6 +359,16 @@ func (uc *WorkflowContractUseCase) findPolicyGroup(att *schemav1.PolicyGroupAtta
 			uc.logger.Warnf("policy group '%s' skipped since it's not found or it might use an old schema version", att.GetRef())
 			return nil, nil
 		}
+		if remoteGroup.PolicyGroup != nil {
+			// validate group arguments
+			with := att.GetWith()
+			for _, input := range remoteGroup.PolicyGroup.GetSpec().GetInputs() {
```
migmartri, line 365: we should extract this somehow to common logic. It's repeated

pkg/policies/policies.go
```diff
@@ -183,6 +192,70 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
...
+	// Policies without inputs in the spec
+	// TODO: Remove this in next release, once users have migrated their policies
+	if len(inputs) == 0 {
+		result = args
+	}
+
+	// Check for required inputs
+	for _, input := range inputs {
+		if _, ok := args[input.Name]; !ok {
+			if input.Required {
```
migmartri, line 207: I thought we discussed that required will work in tandem with default?

jiparis: Added the validation to either one or the other.

jiparis:
Hello, thanks for your feedback ❤️ . I've just updated the PR with:
* raise error if required == true and default value is set
* now prefix `inputs` or `.inputs` is mandatory for all interpolations
* renders "" if no value is provided (for non required inputs with no default value)

app/controlplane/pkg/biz/workflowcontract.go
jiparis: I'll do it in a separate PR, thanks

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -346,6 +359,19 @@ func (uc *WorkflowContractUseCase) findPolicyGroup(att *schemav1.PolicyGroupAtta
...
+				_, ok := with[input.GetName()]
+				if !ok && input.GetRequired() {
+					return nil, NewErrValidation(fmt.Errorf("missing required input %q for group", input.GetName()))
+				}
+				if input.GetRequired() && input.GetDefault() != "" {
```
migmartri, line 370: does the validation need to happen here?

pkg/policies/policies.go
```diff
@@ -183,6 +195,81 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
...
+	// TODO: Remove this in next release, once users have migrated their policies
+	if len(inputs) == 0 {
+		result = args
+	}
+
+	// Check for required inputs
+	for _, input := range inputs {
+		// Illegal combination
+		if input.Required && input.Default != "" {
+			return nil, fmt.Errorf("required input %s with a default value is illegal", input.Name)
```
migmartri, line 211: nitpick `illegal` sounds quite technical. Maybe `input %s can not be required and have a default at the same time.`

pkg/policies/policies.go
```diff
@@ -183,6 +195,81 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
...
+}
+
+// renders an input template using bindings in Go templating format
+func applyBinding(input string, bindings map[string]string) (string, error) {
+	if bindings == nil {
+		return input, nil
+	}
+
+	// Support both `.inputs.foo` and `inputs.foo`
+	input = inputsPrefixRegexp.ReplaceAllString(input, "{{ .inputs.")
```
migmartri, line 254: what if there is no space? `{{.inputs...)` can't it be done with an actual regex 

/$\.inputs/

migmartri (approved):
Approved modulo not sure of the quality of the replacement

app/controlplane/pkg/biz/workflowcontract.go
jiparis: Yes, this is one of the duplicated cases. This is during contract creation/update.

pkg/policies/policies.go
jiparis: That's not the regex, that's the final replacement. The regex looks like this (check top of the file) `{{\s*(inputs.)

jiparis: re: what if there is no space?. All cases are tested, no spaces, multiple spaces, ...

PR #1489, opened 2024-11-06
feat(docs): document policy groups

### LINKED ISSUE #1488: document policy groups in reference documentation
After merging https://github.com/chainloop-dev/chainloop/pull/1465, we should document the whole feature.

jiparis:
Closes #1488


javirln (approved):

docs/docs/reference/policy-groups.mdx
```diff
@@ -0,0 +1,66 @@
...
+apiVersion: workflowcontract.chainloop.dev/v1
+kind: PolicyGroup
+metadata:
+  name: sbom-quality
+  description: This policy group applies a number of SBOM-related policies
+  annotations:
+    category: SBOM
+spec:
+  inputs:
+    - name: bannedlicenses
```
migmartri, line 20: can you use the `-` in the examples? :) 

```
banned-licenses
```

docs/docs/reference/policy-groups.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: Policy Groups
```
migmartri, line 2: I'd make this part of the `policies` reference guide. Or at least make sure that somebody reading the policies part don't miss this one.

migmartri (approved):

docs/docs/reference/policy-groups.mdx
migmartri: ok, it seems it's not valid, can we use camelcase? would the validations support that? I remember seeing some dns validation

jiparis: Yes, currently it uses dns validation. We'll need to remove it and provide a custom validation instead.

jiparis: For now I'll put it in camelCase

Issue #1470, opened 2024-11-04
add pre-release version support


migmartri:
We should allow users to indicate when a version provided during attestation is pre-release

PR #1496, opened 2024-11-07
feat: clean old pre-releases

migmartri:
Since promoting a pre-release was a new feature introduced after the versions were added, we want to mark them as regular releases as an initial cleanup

ref #1470


app/controlplane/pkg/data/ent/migrate/migrations/20241107213854.sql
```diff
@@ -0,0 +1,12 @@
+-- we mark the latest version of each project as prerelease and unset the rest
```
javirln, line 1: I don't understand the query. If want we want is to mark all existing pre-releases as released, the query I don't think that does that, I guess it shall be something like this:

```sql
WITH ranked_versions AS (SELECT id,
                                ROW_NUMBER() OVER (PARTITION BY project_id ORDER BY created_at DESC) as rn
                         FROM project_versions
                         WHERE prerelease = true)
UPDATE project_versions
SET prerelease = false
WHERE (id IN (SELECT id
              FROM ranked_versions
              WHERE rn = 1));
```

javirln, line 1: I don't understand the query.
```sql
SET prerelease = (id IN (
  SELECT id
  FROM ranked_versions
  WHERE rn = 1
))
```
We are setting the value of `prerelease` to the IDs of the `ranked_version` query? Shouldn't that be on the WHERE clause to filter the values and set the prerelease to something else?

migmartri: it's the other way around, we want to remove the prerelease status from the older ones (since before we culdn't not make them prerelease). Your proposal unset pre-release to the top one.

Does it make sense?

javirln (approved):

PR #1505, opened 2024-11-12
feat(data): Adds latest_run and updated_at fields to workflow model

javirln:
This patch introduces two new columns in the `workflows` table: `latest_run` and `updated_at`. These columns are designed to track the most recent run of a workflow and the last update time, respectively.

Changes in Logic:

- When a new workflow run is created, the corresponding workflow is updated to reference this run and the `updated_at` field is set accordingly.
- The `entWrToBizWr` function in the data package has been refactored to leverage direct edge relationships, eliminating the need for additional queries to fetch the latest run for specified workflows.

Database Modifications:

- Addition of `latest_run` and `updated_at` columns in the `workflows` table.
- A query to update existing workflows so that they reference their most recent workflow run, if available.


app/controlplane/pkg/data/ent/schema/workflow.go
```diff
@@ -46,11 +46,17 @@ func (Workflow) Fields() []ent.Field {
...
+			Default(time.Now).
+			Annotations(&entsql.Annotation{
+				Default: "CURRENT_TIMESTAMP",
+			}),
 		field.Time("deleted_at").Optional(),
 		// public means that the workflow runs, attestations and materials are reachable
 		field.Bool("public").Default(false),
 		field.UUID("organization_id", uuid.UUID{}),
 		field.UUID("project_id", uuid.UUID{}),
+		field.UUID("latest_run", uuid.UUID{}).Optional().Nillable(),
```
migmartri, line 59: what does nillable mean compared to optional?

app/controlplane/pkg/data/workflow.go
```diff
@@ -241,13 +223,7 @@ func (r *WorkflowRepo) FindByID(ctx context.Context, id uuid.UUID) (*biz.Workflo
 		return nil, err
 	}
 
-	// Not efficient, we need to do a query limit = 1 grouped by workflowID
-	lastRun, err := getLastRun(ctx, workflow)
```
migmartri, line 245: nice :)

app/controlplane/pkg/data/workflow.go
```diff
@@ -317,8 +293,8 @@ func entWFToBizWF(ctx context.Context, w *ent.Workflow, r *ent.WorkflowRun) (*bi
 		wf.ContractRevisionLatest = lv.Revision
 	}
 
-	if r != nil {
-		lastRun, err := entWrToBizWr(ctx, r)
+	if latestRun := w.Edges.LatestWorkflowRun; latestRun != nil {
```
migmartri, line 296: should we query if it's not here? We do that pattern in some other places, either take it from the preloaded edge or load it?

app/controlplane/pkg/data/workflowrun.go
```diff
@@ -93,6 +93,15 @@ func (r *WorkflowRunRepo) Create(ctx context.Context, opts *biz.WorkflowRunRepoC
 		return nil, err
 	}
 
+	// Update the workflow with the last run reference
```
migmartri, line 96: I like that's in the transaction :)

migmartri (approved):
Thanks!

app/controlplane/pkg/data/ent/schema/workflow.go
javirln: - `Optional`: Indicates that the field is not required when creating or updating an entity. The field can be omitted in the input data.
- `Nillable`: Indicates that the field can hold a `nil` (or `null` in SQL) value in the database.

app/controlplane/pkg/data/workflow.go
javirln: When I removed the workflowRun option from `entWFToBizWF`, I added `WithLatestWorkflowRun()` to all the places where we were previously fetching the latest workflow run on demand. This ensures it works the same as before. My concern with always fetching the latest workflow run when not found on the edge is that it could lead to unnecessary queries since the logic doesn't always use it.

I prefer to keep it this way, and we can update the query if we need the latest workflow run in the future.

PR #1513, opened 2024-11-13
feat(ci): Add reusable workflow for GitHub release

javirln:
This patch changes the definition of the `release` workflow to be a reusable workflow. This workflow will be called once the release process done by Goreleaser is done, so we can attest the whole GitHub release by passing the tag information.


jiparis (approved):

jiparis:
Awesome, thanks!

PR #1530, opened 2024-11-18
feat(policies): add requirements to policy attachments

### LINKED ISSUE #1529: Add requirement identifiers to policy attachment
This feature request consist on being able to relate policy evaluations to organization's requirements. 
While it's up to each company to define what are those requirements in terms of SDLC and security, policy evaluations in Chainloop should correlate to them, to, for example, allowing SecOps team to have a full security overview based on their compliance requirements.
Requirement IDs can optionally be locked by using a `@version` suffix. Not that `version` in this context is up to the organization, and may consist of a semver version number, or a sha256 digest, for example.
The easiest way is to add such IDs at the policy attachment level (both in contracts and policy groups):
```
yaml
# my contract
policies:
  materials:
    - ref: artifact-signed
      requirements:
        - req-id1@v1.2.3
        - req-id2@v0.1
```
The above example can be read as: if `artifact-signed` policy succeeds, it will contribute to satisfy requirements `req-id1@v1.2.3` and `req-id2@v0.1`.

jiparis:
This PR adds requirement IDs to crafting schema and state, so that requirement IDs are stored in the policy evaluations.
Given a contract like this:
```yaml
schemaVersion: v1
materials: []
policies:
  attestation:
    - ref: source-commit
      requirements:
        - req-01@v1.0
        - req-02
```
It would generate the following attestation:
```shell
> cldev wf run describe --digest sha256:50567683ffe0993ec2fc3c9d2bf8869d49b59fd56acfd092fe53fbdcc2f7d648 --output statement
...
"policy_evaluations": {
         "CHAINLOOP.ATTESTATION": [{
               "description": "Verifies that the attestation explicitly references a specific Git commit",
               "name": "source-commit",
               "policy_reference": {
                  "annotations": {
                     "name": "source-commit",
                     "organization": ""
                  },
                  "digest": {
                     "sha256": "5f779a2c28816df613d1925c033da02b1b7346b46440fe70db09615a1780b1ba"
                  },
                  "name": "source-commit",
                  "uri": "chainloop://localhost:8002/source-commit"
               },
               "requirements": [
                  "req-01@v1.0",
                  "req-02"
               ],
               "skipped": false,
               "type": "ATTESTATION"
            }
...
```
This PR includes changes to:
* crafting schema
* crafting state
* policy evaluation logic
* attestation API responses

Fixes #1529


app/controlplane/api/controlplane/v1/response_messages.proto
```diff
@@ -135,6 +135,7 @@ message PolicyEvaluation {
   PolicyReference policy_reference = 10;
   bool skipped = 12;
   repeated string skip_reasons = 13;
+  repeated string requirements = 14;
```
javirln, line 138: They are not type? just regular strings?

jiparis: yes, they are strings, just like they come from the attestation.

javirln (approved):

Issue #1405, opened 2024-10-16
Add group name to policy evaluations


jiparis:
When a policy is included in a policy group, this context is lost when the attestation is sent. Adding a `group` field in the evaluation result would solve this.

PR #1531, opened 2024-11-18
fix: add group reference to attestation API

jiparis:
Group reference was missing in attestation API responses. This PR fixes that.
Related to #1405


app/controlplane/internal/service/attestation.go
```diff
@@ -470,9 +470,19 @@ func extractPolicyEvaluations(in map[string][]*chainloop.PolicyEvaluation) map[s
 			}
 
 			if ev.PolicyReference != nil {
+				name, _ := ev.PolicyReference.GetAnnotations().AsMap()["organization"].(string)
```
javirln, line 473: This won't work with old attestations right? Could we add a check to only add the organization name if the fields exists?

jiparis: If the key doesn't exist, or there are no annotations, `orgName` will be assigned an empty string `""`. I've tested it with references without groups and works as expected.

javirln (approved):

migmartri (approved):

Issue #1533, opened 2024-11-19
tweak attestation process performance


migmartri:
There are some low hanging fruit that could improve performance of the attestation process substantially. Let's review it.

PR #1537, opened 2024-11-19
feat: add backoff retry to all attestation commands

migmartri:
general purpose retry mechanism for `att init, add and push`

ref #1533

### DIFF
--- a/app/cli/cmd/attestation_add.go
+++ b/app/cli/cmd/attestation_add.go
@@ -16,18 +16,14 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
-	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


app/cli/cmd/attestation_add.go
```diff
@@ -90,30 +86,17 @@ func newAttestationAddCmd() *cobra.Command {
 
 			// In some cases, the attestation state is stored remotely. To control concurrency we use
 			// optimistic locking. We retry the operation if the state has changed since we last read it.
-			return backoff.RetryNotify(
```
migmartri, line 93: I extracted this logic to a common function and extended it with transient API errors

javirln (approved):

jiparis (approved):
awesome!

PR #1546, opened 2024-11-20
feat: add profiler to controlplane

migmartri:
Add optional pprof profiler endpoint on port 6060

To enable it, just provide the setting on the Helm Chart

Ref #1533


javirln (approved):
Great!

PR #1559, opened 2024-11-20
chore: add requirement validation

### LINKED ISSUE #1558: Validate requirement ID
Requirements in policy attachments will have an optional version. The proposal is to support the syntax `req_id@version`. Both requirement ID and version must be flexible enough to hold any user-defined value. 

Not allowing whitespaces seem to be a fair compromise: `^\S+(@\S+)?$`

jiparis:
This PR adds some validation to the requirement IDs, allowing only non space characters an 1 `@` only to split versions.

Closes #1558


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -168,7 +168,11 @@ message PolicyAttachment {
   map<string, string> with = 5;
 
   // List of requirements this policy contributes to satisfy
-  repeated string requirements = 6;
+  repeated string requirements = 6 [(buf.validate.field).repeated.items = {
+    string: {
+      pattern: "^[^\\s@]+(@[^\\s@]+)?$", // whitespaces not allowed, lock is optional (@)
```
migmartri, line 173: I know that I said we should force it but now I am wondering if we should be lax by default as you did. What do you think?

jiparis: Yes, but at least I would force the `@` for the version separator (or `/`, or `:`, anything), as we do in other places

jiparis: I've updated the rule to have a mandatory requirement prefix for grouping them into frameworks, groups, or similar. `myframework/myrequirement@version`

migmartri (approved):
thanks

PR #1566, opened 2024-11-22
chore: improve queries

migmartri:
remove nested querying in the marshalling onbect.


app/controlplane/cmd/__debug_bin4021906817
javirln, line 1: I think this is a leftover

javirln (approved):

jiparis (approved):

app/controlplane/cmd/__debug_bin4021906817
migmartri: thanks

Issue #1536, opened 2024-11-19
improve pagination


migmartri:
- show page information i.e `showing 10-25 out of 30`
- increase the number to 50

![image](https://github.com/user-attachments/assets/a5535d23-1019-484e-b2af-14b58a2103aa)

PR #1575, opened 2024-11-23
chore: improve init performance

migmartri:
- removes eager loading from init method
- fixes project/version loading

refs #1533 
Closes https://github.com/chainloop-dev/chainloop/issues/1536

### DIFF
--- a/app/controlplane/internal/service/attestation.go
+++ b/app/controlplane/internal/service/attestation.go
@@ -169,17 +169,25 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
 		return nil, handleUseCaseErr(err, s.log)
 	}
 
+	var orgName string
 	// Find the organization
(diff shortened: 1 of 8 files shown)


app/controlplane/internal/service/attestation.go
```diff
@@ -169,17 +169,25 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
 		return nil, handleUseCaseErr(err, s.log)
 	}
 
+	var orgName string
 	// Find the organization
-	org, err := s.orgUseCase.FindByID(ctx, robotAccount.OrgID)
-	if err != nil {
-		return nil, handleUseCaseErr(err, s.log)
+	if org := usercontext.CurrentOrg(ctx); org != nil {
```
jiparis, line 174: If all tokens (user and API token) have an org, we can safely use `requireCurrentOrg` function instead. Do chainloop still support robot account tokens?

PR #1586, opened 2024-11-24
handle rollbacks

migmartri:
Properly handle tx rollbacks thorough the app 

refs #1533


jiparis:
You might want to implement this instead https://entgo.io/docs/transactions#best-practices

jiparis (approved):

PR #1596, opened 2024-11-26
chore: improve performance store attestation

migmartri:
There were a couple of issues with storing attestations

- The go-routine was leaving some open connections, this was a regression introduced a couple of days ago.
- But these open connections were happening because of the referrer storage logic that was runnign a very long transaction. This code removes the transaction and simplifies the code.

before we were getting this

```
SELECT count(*) as total, usename AS username, state, query FROM pg_stat_activity where state = 'idle in transaction' group by (username, state, query);
 total | username  |        state        | query
-------+-----------+---------------------+-------
    50 | chainloop | idle in transaction | begin
```

refs #1533

### DIFF
--- a/app/controlplane/configs/config.devel.yaml
+++ b/app/controlplane/configs/config.devel.yaml
@@ -51,9 +51,9 @@ data:
   database:
     driver: pgx
     source: postgresql://postgres:@${DB_HOST:0.0.0.0}/controlplane
-    max_open_conns: 5
-    min_open_conns: 1
-    max_conn_idle_time: 120s
+    # max_open_conns: 5
+    # min_open_conns: 1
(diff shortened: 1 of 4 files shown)


app/controlplane/configs/config.devel.yaml
```diff
@@ -51,9 +51,9 @@ data:
   database:
     driver: pgx
     source: postgresql://postgres:@${DB_HOST:0.0.0.0}/controlplane
-    max_open_conns: 5
-    min_open_conns: 1
-    max_conn_idle_time: 120s
+    # max_open_conns: 5
```
jiparis, line 54: Is this an accidental change?

Issue #1628, opened 2024-12-08
allow to send events to nats event bus


migmartri:
To improve interoperability we want to allow the controlplane to send structured events to a nats stream.

PR #1629, opened 2024-12-08
feat: nats event publisher

migmartri:
This PR adds support publishing events through a Nats.io jetstream bus.

It includes

- configurable endpoint
- Bootstrap of publisher
- Use-case that loads user information in the context and sends it to the publisher.
- For ^^ I needed to decouple part of the middlewares to move the entities to their own packages to prevent biz->usercontext cycles

refs #1628


jiparis (approved):

danlishka (approved):
LGTM

PR #1634, opened 2024-12-09
feat(events): org created/invited

migmartri:
new set of events for org management

- orgCreate
- orgInvite
- OrgJoined
- OrgLeft

image

This PR also adds some logic to allow `event publishers` to define if the event must contain actor or not. That way the system will not send events that do not have the full context

refs #1628

### DIFF
--- a/app/controlplane/pkg/auditor/events/organization.go
+++ b/app/controlplane/pkg/auditor/events/organization.go
@@ -114,7 +114,7 @@ func (p *OrgUserInvited) ActionType() string {
 }
 
 func (p *OrgUserInvited) Description() string {
-	return fmt.Sprintf("{{ .ActorEmail }} has invited the user %s to the organization %s with role %s", p.ReceiverEmail, p.OrgName, p.Role)
+	return fmt.Sprintf("{{ .ActorEmail }} has invited %s to the organization %s with role %s", p.ReceiverEmail, p.OrgName, p.Role)
 }
 
 func (p *OrgUserInvited) ActionInfo() (json.RawMessage, error) {
--- a/app/controlplane/pkg/auditor/events/organization.go
+++ b/app/controlplane/pkg/auditor/events/organization.go
@@ -1,0 +1,126 @@
+//
+// Copyright 2024 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


app/controlplane/pkg/auditor/events/organization.go
```diff
@@ -0,0 +1,126 @@
...
+	_ auditor.LogEntry = (*OrgUserLeft)(nil)
+	_ auditor.LogEntry = (*OrgCreated)(nil)
+)
+
+const (
+	OrgType                    auditor.TargetType = "Organization"
+	userJoinedOrgActionType    string             = "UserJoined"
+	userLeftOrgActionType      string             = "UserLeft"
+	userInvitedToOrgActionType string             = "UserInvited"
+	orgCreatedActionType       string             = "OrgCreated"
```
javirln, line 38: What do you think about using the full name of the resource rather a shorter version? So `OrganizationCreated` ?

javirln (approved):
Small comment

app/controlplane/pkg/auditor/events/organization.go
migmartri: Yeah, you are probably right, let me revisit it

app/controlplane/pkg/auditor/events/organization.go
```diff
@@ -0,0 +1,126 @@
...
+func (p *OrgUserLeft) ActionType() string {
+	return userLeftOrgActionType
+}
+
+func (p *OrgUserLeft) Description() string {
+	return fmt.Sprintf("{{ .ActorEmail }} has left the organization %s", p.OrgName)
+}
+
+// user got invited to the organization
+type OrgUserInvited struct {
```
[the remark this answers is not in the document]
migmartri: it is the actor, see the description of the evetn

app/controlplane/pkg/auditor/logentry.go
```diff
@@ -55,6 +55,7 @@ type LogEntry interface {
 	TargetID() *uuid.UUID
 	// Description returns a templatable string, see the DescriptionVariables struct.
 	Description() string
+	RequiresActor() bool
```
javirln, line 58: Ah ok I haven't seen this!

jiparis (approved):

PR #1663, opened 2024-12-16
feat(events): Send events for workflow contract changes

javirln:
This patch creates a new event type `WorkflowContract` and a few types associated with it. Instruments the code base to be able to send events related to that resource.

Ref #1628


app/controlplane/pkg/auditor/events/workflowcontract.go
```diff
@@ -0,0 +1,183 @@
...
+	WorkflowContractType                       auditor.TargetType = "WorkflowContract"
+	WorkflowContractCreatedActionType          string             = "WorkflowContractCreated"
+	WorkflowContractUpdatedActionType          string             = "WorkflowContractUpdated"
+	WorkflowContractDeletedActionType          string             = "WorkflowContractDeleted"
+	WorkflowContractContractAttachedActionType string             = "WorkflowContractContractAttached"
+	WorkflowContractContractDetachedActionType string             = "WorkflowContractContractDetached"
+)
+
+// WorkflowContractBase is the base struct for workflow contract events
+type WorkflowContractBase struct {
```
migmartri, line 46: Thanks @javirln!

BTW, I am curious to see what's your take on the testing framework I put here for the generation of events, let me know what you think.

app/controlplane/pkg/auditor/events/testdata/workflowcontracts/workflow_attached_to_contract.json
```diff
@@ -0,0 +1,17 @@
+{
+  "ActionType": "WorkflowContractContractAttached",
+  "TargetType": "WorkflowContract",
+  "TargetID": "1089bb36-e27b-428b-8009-d015c8737c54",
```
migmartri, line 4: nice!

app/controlplane/pkg/biz/workflow.go
```diff
@@ -190,9 +209,48 @@ func (uc *WorkflowUseCase) Update(ctx context.Context, orgID, workflowID string,
 		return nil, NewErrNotFound("workflow")
 	}
 
+	// Dispatch events to the audit log regarding the contract if it has changed
+	// or if it has been attached or detached
+	if opts.ContractID != nil {
+		if preUpdateWorkflow.ContractID != uuid.Nil {
+			uc.dispatchContractEvent(ctx, orgID, preUpdateWorkflow.ContractID, preUpdateWorkflow.ContractName, wf.ID, wf.Name, false)
```
migmartri, line 216: ok, so this is detaching the previous contract and attaching it to a new one?

migmartri (approved):

app/controlplane/pkg/biz/workflow.go
javirln: yes, in case of a change of contract it sends the detach and attach. See for example when we create a new workflow, we send the event of a new contract attached

PR #1668, opened 2024-12-17
feat(policies): validate policy attachments against remote provider

jiparis:
This is an extension of the remote provider feature. Given that they would expose a /policies/validate endpoint that would validate a policy attachment via POST.
```
curl -d '{"policy_attachment":"{\"ref\":\"source-commit\",\"requirements\":[\"foo/bar\"]}"}' -H "Authorization: Bearer $USER_TOKEN" -H "Content-Type: application/json" "localhost:8002/v1/policies/validate"
```

Then it would be used during contract validation to validate any custom metadata in the attachment. For instance, if requirement references were validated:
```shell
> chainloop wf contract update --name myproject-mywf --contract test/my-contract.yaml
ERR validation error: invalid attachment: validation failures: [finding requirement: requirement bar not found]
```

### DIFF
--- a/app/controlplane/pkg/biz/workflowcontract.go
+++ b/app/controlplane/pkg/biz/workflowcontract.go
@@ -359,6 +359,19 @@ func (uc *WorkflowContractUseCase) ValidateContractPolicies(rawSchema []byte, to
 	return nil
 }
 
+func (uc *WorkflowContractUseCase) ValidatePolicyAttachment(providerName string, att *schemav1.PolicyAttachment, token string) error {
+	provider, err := uc.findProvider(providerName)
+	if err != nil {
+		return err
+	}
+
+	if err = provider.ValidateAttachment(att, token); err != nil {
+		return fmt.Errorf("invalid attachment: %w", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -370,6 +383,11 @@ func (uc *WorkflowContractUseCase) findAndValidatePolicy(att *schemav1.PolicyAtt
 	// [chainloop://][provider:][org_name/]name
 	if loader.IsProviderScheme(att.GetRef()) {
 		pr := loader.ProviderParts(att.GetRef())
+		// Validate attachment
+		if err := uc.ValidatePolicyAttachment(pr.Provider, att, token); err != nil {
```
migmartri, line 387: should we handle the case where the API doesn't exist in the provider backend? Maybe handling 404 would be enough?

jiparis: Right, we can add it and no-op in that case.

jiparis: Done in the provider implementation

migmartri (approved):

PR #1669, opened 2024-12-17
feat(cli): handle org selection client-side

### LINKED ISSUE #350: review multi-organization setup
Currently, the data model root is namespaced (soft multi-tenancy) by an organization entity. 

Users are connected to different organizations through memberships, and the way we decide which one if the current is by a `current` database field.    

image

To change the current organization you can use the CLI

```
$ chainloop organization ls
┌──────────────────────────────────────┬──────────────────────┬─────────┬─────────────────────┐
│ ID                                   │ ORG NAME             │ CURRENT │ JOINED AT           │
├──────────────────────────────────────┼──────────────────────┼─────────┼─────────────────────┤
│ 34970208-2f3c-4c6e-ad8a-88f2b2d2e04f │ integrations testing │ false   │ 10 Jul 23 09:23 UTC │
├──────────────────────────────────────┼──────────────────────┼─────────┼─────────────────────┤
│ 7d487b73-ba94-4d62-99a6-c54fa6ae0e2b │ chainloop core org   │ true    │ 03 Jan 23 13:25 UTC │
├──────────────────────────────────────┼──────────────────────┼─────────┼─────────────────────┤
...
```

```
$ chainloop organization set --id  7d487b73-ba94-4d62-99a6-c54fa6ae0e2b
```

The problem with this approach is that it can be very confusing for an user to see server-side persistence of these kind of configuration when it's more like a client/per-request basis. Imagine an user selecting an org in the UI and that affecting the CLI.

It will also make it more difficult for users and automation to work with multiple orgs.

Managing multiple "namespaces" (orgs in our cases) is a known pattern and we could solve it by making sure API calls always require an org ID, and the current org is stored in the client side configuration. We could make it work similarly to kubernetes, where you have a namespace set up. but it can be overridden through a global flag. 

```[tasklist]
### Tasks
- [x] Expose organizations an user belongs to through the currentUser API endpoint
- [ ] Store the current org client side, we could store the first one if there is only one and prompt the user to select one if there is more than one
- [ ] The org set command could be used to set locally the current org
- [ ] Start sending the org ID in each API call
- [x] Update the controlplane middleware that loads the current ORG to start using this parameter instead
- [ ] Drop `current` column in the memberships DB
```

migmartri:
This PR moves the selection of the current organization to the client side configuration file.

This is how it works

The first time you login, or if the config file doesn't exist, or the `organization` entry in the file doesn't exist, the system will retrieve it and do it transparently.

For example, the organization-2 is marked as default server side and it's also marked as current client-side

```
$ cl org ls
┌────────────────┬─────────┬─────────┬───────┬─────────────────────┐
│ NAME           │ CURRENT │ DEFAULT │ ROLE  │ JOINED AT           │
├────────────────┼─────────┼─────────┼───────┼─────────────────────┤
│ organization-1 │ false   │ false   │ owner │ 17 Dec 24 12:40 UTC │
├────────────────┼─────────┼─────────┼───────┼─────────────────────┤
│ organization-2 │ true    │ true    │ owner │ 17 Dec 24 12:40 UTC │
└────────────────┴─────────┴─────────┴───────┴─────────────────────┘
``` 

to switch, you'll use `org set` as before but by default, this one will only affect the local state

```
$ cl org set --name organization-1
$ cl org ls
┌────────────────┬─────────┬─────────┬───────┬─────────────────────┐
│ NAME           │ CURRENT │ DEFAULT │ ROLE  │ JOINED AT           │
├────────────────┼─────────┼─────────┼───────┼─────────────────────┤
│ organization-1 │ true    │ false   │ owner │ 17 Dec 24 12:40 UTC │
├────────────────┼─────────┼─────────┼───────┼─────────────────────┤
│ organization-2 │ false   │ true    │ owner │ 17 Dec 24 12:40 UTC │
└────────────────┴─────────┴─────────┴───────┴─────────────────────┘

$ cl org describe
┌──────────────────────────────────────────────────────────┐
│ Current Context                                          │
├─────────────────────┬────────────────────────────────────┤
│ Logged in as        │ john@chainloop.local               │
├─────────────────────┼────────────────────────────────────┤
│ Organization        │ organization-1 (role=owner)        │
├─────────────────────┼────────────────────────────────────┤
│ Default CAS Backend │  (provider=INLINE, status="valid") │
└─────────────────────┴────────────────────────────────────┘
```
you can set the default value server side by providing a `--default` flag to `cl org set`

You can also select the organization on the fly with the `--org` or `-n` flags.

```
$ cl org describe --org organization-2
┌──────────────────────────────────────────────────────────┐
│ Current Context                                          │
├─────────────────────┬────────────────────────────────────┤
│ Logged in as        │ john@chainloop.local               │
├─────────────────────┼────────────────────────────────────┤
│ Organization        │ organization-2 (role=owner)        │
├─────────────────────┼────────────────────────────────────┤
│ Default CAS Backend │  (provider=INLINE, status="valid") │
└─────────────────────┴────────────────────────────────────┘
```

I've also fixed a bug that prevented the current-context endpoint from returning org information when the client was authenticated as an user.

Fixes #350

### DIFF
--- a/app/cli/cmd/config.go
+++ b/app/cli/cmd/config.go
@@ -26,7 +26,7 @@ import (
 
 // Map of all the possible configuration options that we expect viper to handle
 var confOptions = struct {
-	authToken, controlplaneAPI, CASAPI, controlplaneCA, CASCA, insecure *confOpt
+	authToken, controlplaneAPI, CASAPI, controlplaneCA, CASCA, insecure, organization *confOpt
 }{
 	insecure: &confOpt{
 		viperKey: "api-insecure",
@@ -49,6 +49,10 @@ var confOptions = struct {
 	CASCA: &confOpt{
 		viperKey: "artifact-cas.api-ca",
 		flagName: "artifact-cas-ca",
+	},
+	organization: &confOpt{
+		viperKey: "organization",
+		flagName: "org",
 	},
 }
 
--- a/app/cli/cmd/organization_create.go
+++ b/app/cli/cmd/organization_create.go
@@ -17,9 +17,11 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 func newOrganizationCreateCmd() *cobra.Command {
@@ -32,6 +34,12 @@ func newOrganizationCreateCmd() *cobra.Command {
 			org, err := action.NewOrgCreate(actionOpts).Run(context.Background(), name)
 			if err != nil {
 				return err
+			}
+
+			// set the local state
+			viper.Set(confOptions.organization.viperKey, name)
+			if err := viper.WriteConfig(); err != nil {
+				return fmt.Errorf("writing config file: %w", err)
 			}
 
 			logger.Info().Msgf("Organization %q created!", org.Name)
--- a/app/cli/cmd/organization_list.go
+++ b/app/cli/cmd/organization_list.go
@@ -22,6 +22,7 @@ import (
 	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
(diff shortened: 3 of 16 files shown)


jiparis (approved):

jiparis:
this is great, thanks!

PR #1689, opened 2025-01-02
refactor: run attestation-level policy evaluations in crafting layer

migmartri:
Runs the attestation-related policy evaluations in the crafting layer storing it in the state.

This is a pre-requisite for #1685.

In practice nothing changes except that policy-level evaluations happen in `status` and `push` and then the renderer layer just takes the info in the state.

### DIFF
--- a/app/cli/internal/action/attestation_push.go
+++ b/app/cli/internal/action/attestation_push.go
@@ -160,6 +160,19 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
 		return nil, err
 	}
 
+	// execute policy evaluations
+	// We do not want to evaluate policies here during render since we want to do it in a separate step
+	statement, err := renderer.RenderStatement(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("rendering statement: %w", err)
+	}
+
+	// Add attestation-level policy evaluations
+	if err := crafter.EvaluateAttestationPolicies(ctx, attestationID, statement); err != nil {
+		return nil, fmt.Errorf("evaluating attestation policies: %w", err)
+	}
+
+	// render final attestation with all the evaluated policies inside
 	envelope, err := renderer.Render(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


app/cli/internal/action/attestation_push.go
```diff
@@ -160,6 +160,19 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
 		return nil, err
 	}
 
+	// execute policy evaluations
+	// We do not want to evaluate policies here during render since we want to do it in a separate step
+	statement, err := renderer.RenderStatement(ctx)
```
migmartri, line 165: I thought about also running it on att.add but I am not sure it's needed yet.

app/cli/internal/action/attestation_status.go
```diff
@@ -141,12 +141,12 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string,
 		}
 
 		// We do not want to evaluate policies here during render since we want to do it in a separate step
-		statement, err := renderer.RenderStatement(ctx, chainloop.WithSkipPolicyEvaluation(true))
```
migmartri, line 144: renderer never evaluates policies anymore

app/cli/internal/action/attestation_push.go
jiparis: Probably not, since we want to block only on "push"

jiparis (approved):
Looks great!

pkg/attestation/renderer/chainloop/v02.go
```diff
@@ -122,88 +103,9 @@ func (r *RendererV02) Statement(ctx context.Context, opts ...RenderOpt) (*intoto
...
-		}
-	}
-
 	return statement, nil
 }
 
-// addPolicyResults adds policy evaluation results to the statement. It does it by deserializing the predicate from a structpb.Struct,
-// filling PolicyEvaluations, and serializing it again to a structpb.Struct object, using JSON as an intermediate representation.
-// Note that this is needed because intoto predicates are generic structpb.Struct
-func addPolicyResults(statement *intoto.Statement, policyResults []*v1.PolicyEvaluation) error {
```
jiparis, line 156: I love this change.

PR #1691, opened 2025-01-02
feat: fail attestation process if policy violation

### LINKED ISSUE #1685: feat: option to fail attestation if there are policy evaluations

migmartri:
Closes #1685


app/cli/cmd/attestation_status.go
```diff
@@ -83,6 +84,9 @@ func attestationStatusTableOutput(status *action.AttestationStatusResult, full b
 	gt.AppendSeparator()
 	meta := status.WorkflowMeta
 	gt.AppendRow(table.Row{"Attestation ID", status.AttestationID})
+	if status.Digest != "" {
```
migmartri, line 87: I moved the digest inside since it was affecting with blocking in the case of json output.

This makes it also more consistent

app/cli/main.go
```diff
@@ -97,6 +97,9 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 			logger.Debug().Msg("GracefulErrorExit enabled (exitCode 0). If you want to disable it set --graceful-exit=false")
 			exitCode = 0
 		}
+	case errors.Is(err, cmd.ErrBlockedByPolicyViolation):
+		// default exit code for policy violations
+		exitCode = 3
```
javirln, line 102: Could we add a global constant so it's reachable? It won't make any difference but just for clarity purposes

danlishka (approved):
Awesome! I will give it a try once deployed

javirln (approved):
Small comment, LGTM!

PR #1700, opened 2025-01-07
chore(docs): Upgrade version of Docusaurus

javirln:
This patch upgrades the version of Docusaurus to version 3.7.0 in an attempt to fix https://github.com/chainloop-dev/chainloop/security/dependabot/79


javirln:
it will fix some vulnerabilities in dependencies but they will one remaining regarding express which we are not using in production: https://github.com/chainloop-dev/chainloop/security/dependabot/79

migmartri (approved):

PR #1711, opened 2025-01-08
chore(requirement): Allow organization namespaced requirements

### LINKED ISSUE #1710: Include optional organization name in requirement reference
We would like to extend the changes introduced at https://github.com/chainloop-dev/chainloop/issues/1558 to allow also scope the requirement to a specific organization.

javirln:
This patch extends the regex that validates the requirement references for the validate policy attachment so it allows an optional organization name to be added at the beginning of the reference.

Closes #1710


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -172,7 +172,7 @@ message PolicyAttachment {
   // List of requirements this policy contributes to satisfy
   repeated string requirements = 6 [(buf.validate.field).repeated.items = {
     string: {
-      pattern: "^([^\\s\\/]+\\/)([^\\s@\\/]+)(@[^\\s@]+)?$", // whitespaces not allowed, framework prefix (/) and version are optional (@)
+      pattern: "^([a-z0-9-]+\\/)?([^\\s\\/]+\\/)([^\\s@\\/]+)(@[^\\s@]+)?$", // whitespaces not allowed, organization names (/) optional, framework prefix (/) and version are optional (@)
```
migmartri, line 175: `framework prefix (/) and version are optional (@)` is confusing, is framework prefix optional?

javirln: Don't think so, I think it's just the comment, let me fix this.

Issue #1686, opened 2024-12-20
feat: show material information after adding it


migmartri:
now that status show the information about policies, we could expose such info after adding a material, so for example after adding a sbom

```
chainloop att add --name sbom-2 --value ~/Desktop/result.cdx
```

we could show

```
INF material kind detected kind=SBOM_CYCLONEDX_JSON
INF material added to attestation
┌────────────────────────────────────────────────────────────────────────────────────┐
│ Name     │ sbom-2                                                                  │
│ Type     │ SBOM_CYCLONEDX_JSON                                                     │
│ Set      │ Yes                                                                     │
│ Required │ No                                                                      │
│ Value    │ result.cdx                                                              │
│ Digest   │ sha256:925ed5a789bad4bc2658ba371530bf3b80639be02d7acd9069c0e52690508132 │
│ Policies │ ------                                                                  │
│          │ sbom-ntia:                                                              │
│          │   - missing author                                                      │
│          │   - missing supplier for 'alpine'                                       │
│          │   - missing unique identifier (PURL, CPE, SWID) for 'alpine'            │
│          │ sbom-banned-licenses: Ok                                                │
│          │ sbom-freshness: Ok                                                      │
│          │ sbom-ntia:                                                              │
│          │   - missing author                                                      │
│          │   - missing supplier for 'alpine'                                       │
│          │   - missing unique identifier (PURL, CPE, SWID) for 'alpine'            │
│          │ sbom-banned-licenses: Ok                                                │
│          │ sbom-freshness: Ok                                                      │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

migmartri:
cc/ @jiparis since we discussed this offline

Horiodino:
Hey, if no one is working on this, I would like to take a shot at it!

migmartri:
@Horiodino awesome, I'll assign it to you. Let me know if you want to do some pair programming or need any help setting the project up. You can find us in Slack [here](https://join.slack.com/t/chainloop-community/shared_invite/zt-2k34dvx3r-u85uGP_KiLC6ic5Wy4aRnQ)


In [devel](https://github.com/chainloop-dev/chainloop/tree/main/devel) you should have enough information to get you started locally.

Thank you!

migmartri:
> [@miguel](https://chainloop-community.slack.com/team/U076PU4GREZ)
[...]

PR #1714, opened 2025-01-08
refactor(cli): store material info in crafting state

migmartri:
> This PR is part of some refactoring that could simplify the implementation of #1686

The main change is that `attestation add` cmd now has available structured information of the material that was added, so then it can be rendered as part of #1686

To achieve that, we needed to update the underlying crafting state so the materials now can contain all its properties, as well as the identifier, as well as factoring out the material stateAPI->action translation code so it can be used outside of `status` command.

cc/ @Horiodino


app/cli/cmd/attestation_add.go
```diff
@@ -89,7 +89,9 @@ func newAttestationAddCmd() *cobra.Command {
 			// optimistic locking. We retry the operation if the state has changed since we last read it.
 			return runWithBackoffRetry(
 				func() error {
-					if err := a.Run(cmd.Context(), attestationID, name, value, kind, annotations); err != nil {
+					// TODO: take the material output and show render it
+					_, err := a.Run(cmd.Context(), attestationID, name, value, kind, annotations)
```
migmartri, line 93: this is the entrypoint that can help with implementing showing the material in the CLI

app/cli/internal/action/attestation_add.go
```diff
@@ -141,37 +141,43 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
...
+		mt, err = crafter.AddMaterialContractFree(ctx, attestationID, materialType, materialName, materialValue, casBackend, annotations)
 	}
 
 	if err != nil {
-		return fmt.Errorf("adding material: %w", err)
+		return nil, fmt.Errorf("adding material: %w", err)
 	}
 
-	return nil
+	// TODO: remove from this PR
```
jiparis, line 176: leftover?

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -206,7 +206,7 @@ func (m *Attestation_Material) CraftingStateToIntotoDescriptor(name string) (*in
 
 	// string materials don't have an artifact nor container, so a name is not available.
```
jiparis, line 207: This probably is not true anymore. If the ID is outside the artifact, string materials would have an ID as well, right?

jiparis (approved):

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
migmartri: not sure, since it's taking about name, ... let me double check

Issue #1715, opened 2025-01-09
Improve release pipeline to include all artifacts


javirln:
Right now the current `build_and_package` pipeline only attest the result of goreleaser but, we have more information to be attested.

The goal of the task would be to include as much as possible, container images and SBOMs at least and add policies around the contract.

PR #1716, opened 2025-01-09
feat(.github): Attest container images and generate SBOMs

javirln:
This patch modifies the `build_and_package` workflow so we are able to attest all container images with their different tags and generate a SBOM file for each of them. The result of those operations are being added as part of the attestation.

Ref: #1715

### DIFF
--- a/.github/workflows/build_and_package.yaml
+++ b/.github/workflows/build_and_package.yaml
@@ -13,9 +13,31 @@ jobs:
   test:
     uses: chainloop-dev/chainloop/.github/workflows/test.yml@main
 
+  init_attestation:
+    runs-on: ubuntu-latest
+    needs: test
+    if: github.ref_type == 'tag' # Guard to make sure we are releasing once
+    outputs:
+      attestation_id: ${{ steps.init_attestation.outputs.attestation_id }}
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Install Chainloop
+        run: |
+          curl -sfL https://docs.chainloop.dev/install.sh | bash -s
+
+      - name: Initialize Attestation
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_package.yaml
```diff
@@ -13,9 +13,42 @@ jobs:
   test:
     uses: chainloop-dev/chainloop/.github/workflows/test.yml@main
 
+  onboard_workflow:
+    name: Onboard Chainloop Workflow
+    needs: test
+    uses: chainloop-dev/labs/.github/workflows/chainloop_onboard.yml@edd4d30e954a4a456dd770f3b7c5e6871c84a661
```
migmartri, line 19: I do not think this is needed anylonger, `init` creates the workflow if needed.

.github/workflows/build_and_package.yaml
```diff
@@ -137,14 +130,104 @@ jobs:
...
+        run: |
+          curl -sfL https://docs.chainloop.dev/install.sh | bash -s
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: anchore/sbom-action@df80a981bc6edbc4e220a492d3cbe9f5547a6e75 # v0.17.9
+        with:
+          image: ${{ matrix.bundle.image }}
+          format: cyclonedx-json
+          artifact-name: ${{ matrix.bundle.artifactName }}
```
migmartri, line 200: what's the artifact name for?

.github/workflows/build_and_package.yaml
javirln: Oki!

.github/workflows/build_and_package.yaml
javirln: It's name of the artifact that will be used when adding it to the release notes. Basically the name it will have there.

PR #1723, opened 2025-01-10
feat(github): Use regular CLI steps in scorecards workflow

javirln:


.github/workflows/scorecards.yml
```diff
@@ -18,34 +18,40 @@ on:
...
       id-token: write
       contents: read
       actions: read
+    env:
+      CHAINLOOP_WORKFLOW_NAME: "chainloop-vault-scorecards"
+      CHAINLOOP_PROJECT: "chainloop"
+      CHAINLOOP_TOKEN: ${{ secrets.CHAINLOOP_TOKEN }}
 
     steps:
+      - name: Install Cosign
```
migmartri, line 37: do we need cosign anymore? This is to verify signature of the installed chainloop correct?

.github/workflows/scorecards.yml
```diff
@@ -83,13 +89,25 @@ jobs:
...
-      attestation_name: "scorecards"
-      workflow_name: "chainloop-vault-scorecards"
+      - name: Attest analysis
+        run: |
+          chainloop attestation add --name sarif-results --value results.sarif
+
+      - name: Finish and Record Attestation
+        if: ${{ success() }}
+        run: |
+          chainloop attestation status --full
```
migmartri, line 99: no need to show status anymore, push will do it for you

migmartri (approved):
Thanks

.github/workflows/scorecards.yml
javirln: Yes, we can remove it!

PR #1735, opened 2025-01-20
feat: reserve organization names

migmartri:
Reserve organization names so they can be used as FQDN references for built-in resources.

### DIFF
--- a/app/controlplane/pkg/biz/organization.go
+++ b/app/controlplane/pkg/biz/organization.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024 The Chainloop Authors.
+// Copyright 2024-2025 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


app/controlplane/pkg/biz/organization.go
```diff
@@ -131,6 +133,13 @@ var errOrgName = errors.New("org names must only contain lowercase letters, numb
 func (uc *OrganizationUseCase) doCreate(ctx context.Context, name string, opts ...CreateOpt) (*Organization, error) {
 	uc.logger.Infow("msg", "Creating organization", "name", name)
 
+	// Check if the name is in the list of protected names
+	for _, protected := range reservedOrgNames {
+		if name == protected {
+			return nil, NewErrValidation(fmt.Errorf("organization name '%s' is reserved and cannot be used", name))
```
javirln, line 139: Can you print the whole list here?

I know it might seem overkill but printing one by one can be thought as trial and error. If we return the whole list there the user has a clear idea of which cannot be used.

migmartri: I could, but I don't know what the user needs to know about these system-reserved orgs when the possibility of conflict is so low. I am not strongly opinionated though

jiparis (approved):

PR #1736, opened 2025-01-20
feat: allow to set policy blocking strategy

migmartri:
```
go run main.go org update --name foobar --block
INF Organization updated!
```

```
go run main.go org ls
┌────────────────┬─────────┬─────────┬───────┬─────────────────────────┬─────────────────────┐
│ NAME           │ CURRENT │ DEFAULT │ ROLE  │ DEFAULT POLICY STRATEGY │ JOINED AT           │
├────────────────┼─────────┼─────────┼───────┼─────────────────────────┼─────────────────────┤
│ new-org        │ false   │ false   │ owner │ ADVISORY                │ 10 Jan 25 11:58 UTC │
├────────────────┼─────────┼─────────┼───────┼─────────────────────────┼─────────────────────┤
│ foobar         │ true    │ true    │ owner │ ENFORCED                │ 22 Dec 24 08:02 UTC │
├────────────────┼─────────┼─────────┼───────┼─────────────────────────┼─────────────────────┤
│ organization-2 │ false   │ false   │ owner │ ADVISORY                │ 17 Dec 24 12:40 UTC │
└────────────────┴─────────┴─────────┴───────┴─────────────────────────┴─────────────────────┘
```

refs #1685


app/controlplane/internal/service/organization.go
```diff
@@ -59,6 +59,25 @@ func (s *OrganizationService) Create(ctx context.Context, req *pb.OrganizationSe
...
 
+func (s *OrganizationService) Update(ctx context.Context, req *pb.OrganizationServiceUpdateRequest) (*pb.OrganizationServiceUpdateResponse, error) {
+	currentUser, err := requireCurrentUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	org, err := s.orgUC.FindByName(ctx, req.Name)
+	if err != nil {
+		return nil, handleUseCaseErr(err, s.log)
```
jiparis, line 70: Before retrieving any entity, I would check that the user has access to it. I know it's done in `orgUC.Update`, but there it's too late in my opinion (not failing here would give clues that the org exist, even if it returns a 404 in that method).

migmartri: Updated, thanks!

jiparis (approved):

PR #1738, opened 2025-01-20
feat(cli): policy checks escape hatch

migmartri:
This PR adds an escape hatch on the push command that will allow the command to return exit code 0 even if there are policy violations and the policy strategy is ENFORCED.

The escape hatch can be enabled using the `--exception-bypass-policy-check` flag during `att push`, the effect would be

- command returning 0
- warning message shown to the user at the end of the command

image

- this fact is stored in the attestation alongside the rest of the information so the operator can clearly see whether the attestation was blocked, bypassed, or was meant to be enforced in the first place. 

```
"policyAttBlocked": false,
"policyBlockBypassEnabled": true,
"policyCheckBlockingStrategy": "ENFORCED",
"policyHasViolations": true,
```

refs #1685


app/cli/cmd/attestation_push.go
```diff
@@ -119,8 +127,11 @@ func newAttestationPushCmd() *cobra.Command {
 	cmd.Flags().StringVar(&signServerCAPath, "signserver-ca-path", "", "custom CA to be used for SignServer communications")
 	cmd.Flags().StringVar(&signServerAuthUser, "signserver-auth-user", "", "")
 	cmd.Flags().StringVar(&signServerAuthPass, "signserver-auth-pass", "", "")
+	cmd.Flags().BoolVar(&bypassPolicyCheck, "exception_bypass_policy_check", false, "do not fail this command on policy violations enforcement")
```
javirln, line 130: One question, why the snake case with `_` rather than `-` as we have the other flags?

migmartri: hi, it's wrong, good catch

migmartri: Fixed, please review

javirln (approved):

PR #1759, opened 2025-01-27
feat(attestation): store resolved organization

migmartri:
This PR makes sure that the built-in policy and group provider leverages organization auto-discovery returned by the server.

This means that if you for example want to get a policy `my-policy` and gets resolved to `/my-rg/foobar`, my-org is now stored.


app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -77,6 +77,7 @@ message AttestationServiceGetPolicyResponse {
 message RemotePolicyReference {
   string url = 1;
   string digest = 2;
+  string org_name = 3;
```
jiparis, line 80: If I'm not wrong, the org is already in the URL in the form of a URL parameter `org=my-org`. Is this what we need?

migmartri: That's true, changed, thanks!

jiparis (approved):

PR #1777, opened 2025-02-02
feat(verification): Store attestation bundles in Chainloop database

jiparis:
This PR adds a new endpoint to receive and store attestation bundles. For optimization and backwards compatibility they will be stored in an independent table.

The implementation is backwards compatible. When a bundle is stored it will still store the attestation in the WorkflowRuns table. Old CLIs can still use the AttestationService/Store endpoint. And new CLIs will send the bundle payload to the same endpoint, falling back to the previous behaviour if it's not accepted.

Referrers table is still being populated with the wrapped attestation, referenced by its digest. Bundle digest is not used at this point.
Next PRs will deprecate attestations against bundles.

Refs #914


app/controlplane/api/controlplane/v1/workflow_run.proto
```diff
@@ -30,6 +30,7 @@ service AttestationService {
   rpc GetContract(AttestationServiceGetContractRequest) returns (AttestationServiceGetContractResponse);
   rpc Init(AttestationServiceInitRequest) returns (AttestationServiceInitResponse);
   rpc Store(AttestationServiceStoreRequest) returns (AttestationServiceStoreResponse);
+  rpc StoreBundle(AttestationServiceStoreBundleRequest) returns (AttestationServiceStoreBundleResponse);
```
migmartri, line 33: why did you choose to have another endpoint?

migmartri (commented):
Awesome @jiparis I'll review it tomorrow. I left one question inline for now :)

app/controlplane/pkg/data/ent/schema/bundle.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"entgo.io/ent"
+	"entgo.io/ent/dialect/entsql"
+	"entgo.io/ent/schema/edge"
+	"entgo.io/ent/schema/field"
+	"github.com/google/uuid"
+)
+
+// Bundle represents attestation bundles
+type Bundle struct {
```
migmartri, line 29: have you considered to call this attestation? and one of the properties is in fact a bundle?

app/controlplane/pkg/data/ent/schema/bundle.go
```diff
@@ -0,0 +1,50 @@
+//
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri, line 2: nitpick, 2025 only since it's a new file, I think

app/controlplane/api/controlplane/v1/workflow_run.proto
jiparis: I wanted to keep them completely independent. Do you suggest using the same endpoint and message? Maybe reusing the `attestation` field for both purposes? That would work as well.

migmartri: or another field in the message called bundle and mark the other one as deprecated?

migmartri: The reason I am thinking about using the same API endpoint it is so we maintain the signature during logging, monitoring, etc

app/controlplane/pkg/data/ent/schema/bundle.go
jiparis: That would also work. I'll rename it.

app/controlplane/api/controlplane/v1/workflow_run.proto
jiparis: Adding another field to the same message won't work because old servers will refuse the request. Check the latest update, I've reused the same field for both purposes. The client will fall back if it finds an old server and send the deprecated payload instead.

jiparis: Nevermind, I've added another field and works as expected.

app/controlplane/internal/service/attestation.go
```diff
@@ -186,34 +189,75 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
+		if err := protojson.Unmarshal(req.GetBundle(), &bundle); err != nil {
+			return nil, handleUseCaseErr(err, s.log)
+		}
+		envelope = *envelopeFromBundle(&bundle)
+	} else {
+		// trying with envelope instead
+		if err := json.Unmarshal(req.GetAttestation(), &envelope); err != nil {
+			return nil, handleUseCaseErr(err, s.log)
+		}
+	}
```
migmartri, line 210: should we check that envelope is not nil and return otherwise?

app/controlplane/internal/service/attestation.go
```diff
@@ -186,34 +189,75 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
 func (s *AttestationService) Store(ctx context.Context, req *cpAPI.AttestationServiceStoreRequest) (*cpAPI.AttestationServiceStoreResponse, error) {
+	var envelope dsse.Envelope
 	robotAccount := usercontext.CurrentRobotAccount(ctx)
 	if robotAccount == nil {
 		return nil, errors.NotFound("not found", "robot account not found")
 	}
 
-	// This will make sure the provided workflowRunID belongs to the org encoded in the robot account
-	wf, err := s.findWorkflowFromTokenOrNameOrRunID(ctx, robotAccount.OrgID, "", "", req.WorkflowRunId)
+	// Try unmarshaling a bundle first, falling back to plain dsse envelopes
```
migmartri, line 198: this message is not accurate anymore.

app/controlplane/internal/service/attestation.go
```diff
@@ -186,34 +189,75 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
 
-	// Decode the envelope through json encoding but
-	// TODO: Verify the envelope signature before storing it
-	// see sigstore's dsse signer/verifier helpers
-	envelope := &dsse.Envelope{}
-	if err := json.Unmarshal(req.Attestation, envelope); err != nil {
-		return nil, handleUseCaseErr(err, s.log)
+	// save bundle after saving attestation
+	if bundle.GetDsseEnvelope() != nil {
+		_, err = s.wrUseCase.SaveBundle(ctx, req.WorkflowRunId, &bundle)
```
migmartri, line 219: I have the feeling we shoudl have used the same storeAttestation use-case but it can receive either a bundle or an attestation. 
note that you might be storing the bundle before we store the older attestation for example. Also, are all the authorization checks done in this case too? i.e can't save a bundle for a workflow you don't own?

app/cli/internal/action/attestation_push.go
```diff
@@ -209,21 +213,40 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
...
 	if err != nil {
 		return "", fmt.Errorf("encoding attestation: %w", err)
 	}
 
 	client := pb.NewAttestationServiceClient(conn)
+
+	// Store bundle next versions will perform this in a single call)
 	resp, err := client.Store(ctx, &pb.AttestationServiceStoreRequest{
-		Attestation:           encodedAttestation,
+		Bundle:                encodedBundle,
```
migmartri, line 226: so bundle is always created regardless of the signing mechanism correct?

app/cli/internal/action/attestation_push.go
```diff
@@ -209,21 +213,40 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
...
 		WorkflowRunId:         workflowRunID,
 		MarkVersionAsReleased: &markVersionAsReleased,
 	})
-
 	if err != nil {
-		return "", fmt.Errorf("contacting the control plane: %w", err)
+		if status.Code(err) == codes.InvalidArgument {
```
migmartri, line 231: is this even needed?

migmartri (approved):
Thanks. My only comment is about the need of that fallback code you added in the CLI

app/cli/internal/action/attestation_push.go
jiparis: Yes, since it holds the DSSE envelope. However, verification material is only added for keyless methods.

app/cli/internal/action/attestation_push.go
jiparis: Right, we can send both in the same request now.

migmartri (approved):

app/controlplane/internal/service/attestation.go
jiparis: Done

app/controlplane/internal/service/attestation.go
jiparis: Implemented in the same use case.

PR #1800, opened 2025-02-10
chore: dump attestation bundle for "attestation" output format

jiparis:
This PR switches from DSEE envelope to Bundle format when using `--output attestation` option in `workflow run describe` command.

### DIFF
--- a/app/cli/internal/action/workflow_run_describe.go
+++ b/app/cli/internal/action/workflow_run_describe.go
@@ -49,6 +49,7 @@ type WorkflowRunItemFull struct {
 
 type WorkflowRunAttestationItem struct {
 	Envelope    *dsse.Envelope `json:"envelope"`
+	Bundle      []byte         `json:"bundle"`
 	statement   *intoto.Statement
 	Materials   []*Material   `json:"materials,omitempty"`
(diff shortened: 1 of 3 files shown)


app/cli/internal/action/workflow_run_describe.go
```diff
@@ -49,6 +49,7 @@ type WorkflowRunItemFull struct {
 
 type WorkflowRunAttestationItem struct {
 	Envelope    *dsse.Envelope `json:"envelope"`
```
migmartri, line 51: is this property used anymore?

migmartri (approved):

app/cli/internal/action/workflow_run_describe.go
jiparis: Yes, it's still used as a fallback for old clients

migmartri: but you are the client no? I mean, this change will not apply to other clients?

jiparis: Actually, if we remove it, we won't be able to show the envelope for old servers. I'd keep it as a fallback for some time.

PR #1807, opened 2025-02-11
feat(verification): get trusted root material and allow rotation

jiparis:
This PR adds an endpoint to get the root material from the configured certificate authorities:
* Adds the possibility to have one signing authority and multiple previous CAs, useful for certificate rotation.
* Adds an endpoint (private) to get all root certificate (and chains) from the configured CAs
* Implemented for FileCA and EJBCA.

```shell
> buf curl --http2-prior-knowledge --protocol grpc  -H "Authorization: Bearer $CHAINLOOP_TOKEN" http://localhost:9000/controlplane.v1.SigningService/GetTrustedRoot
{
  "keys": {
    "2a522d9652e0933d2a1237c395bc116e012f86dffff13122da59f76e0d2abe27": {
      "certificates": [
        "-----BEGIN CERTIFICATE-----\nMI... ==\n-----END CERTIFICATE-----\n"
      ]
    },
    "e6b4419565f9ccab5c68d7a99d96722931d2d9a7d920c938fae46c45e25989d2": {
      "certificates": [
        "-----BEGIN CERTIFICATE-----\n... Q==\n-----END CERTIFICATE-----\n"
      ]
    }
  }
}
```

Chart changes:
```diff
✗ diff before.yaml after.yaml
445c445
<   generated_jws_hmac_secret: "THVVd1dBOFRrbw=="
---
>   generated_jws_hmac_secret: "R1NoTllaMkFsdQ=="
449,453c449,454
<     certificate_authority:
<       file_ca:
<         cert_path: "/ca_secrets/file_ca.cert"
<         key_path:  "/ca_secrets/file_ca.key"
<         key_pass: "foo"
---
>     certificate_authorities:
>       - issuer: true
>         file_ca:
>           cert_path: "/ca_secrets/file_ca.cert"
>           key_path:  "/ca_secrets/file_ca.key"
>           key_pass: "foo"
476c477
<       generated_jws_hmac_secret: "LuUwWA8Tko"
---
>       generated_jws_hmac_secret: "GShNYZ2Alu"
1613c1614
<         checksum/secret-config: 5befa75f8e0d765e8585f890d75547ca0c39eac815b31961619026e70b87ffd4
---
>         checksum/secret-config: fbe01695cfacff3e8e5a3954bffbae73d6b15a6452ea5c6bbd3ee39feedc35c2
```

Note that chart configuration doesn't support multiple CAs yet.


app/controlplane/cmd/wire.go
```diff
@@ -96,3 +99,16 @@ func newCASServerOptions(in *conf.Bootstrap_CASServer) *biz.CASServerDefaultOpts
 		DefaultEntryMaxSize: in.GetDefaultEntryMaxSize(),
 	}
 }
+
+func newSigningCAs(cas []*conf.CA, logger log.Logger) (*ca.CertificateAuthorities, error) {
```
migmartri, line 103: nice, no need to do it explictly

app/controlplane/configs/config.devel.yaml
```diff
@@ -19,20 +19,21 @@ server:
...
-#  ejbca_ca:
-#    server_url: "https://localhost/ejbca"
-#    key_path: "../../devel/devkeys/superadmin.key"
-#    cert_path: "../../devel/devkeys/superadmin.pem"
-#    root_ca_path: "../../devel/devkeys/ManagementCA.pem"
+certificate_authorities:
+  - file_ca:
+      cert_path: ${FILE_CA_CERT_PATH:../../devel/devkeys/ca.pub}
+      key_path: ${FILE_CA_KEY_PATH:../../devel/devkeys/ca.pem}
+      key_pass: chainloop
```
migmartri, line 26: should you set the issuer value?

deployment/chainloop/templates/controlplane/secret-config.yaml
```diff
@@ -29,24 +29,24 @@ stringData:
...
-    certificate_authority:
-      file_ca:
-        cert_path: "/ca_secrets/file_ca.cert"
-        key_path:  "/ca_secrets/file_ca.key"
-        key_pass: "{{- required "FileCA keyPass is mandatory" .keyPass }}"
+    certificate_authorities:
+      - file_ca:
+          cert_path: "/ca_secrets/file_ca.cert"
+          key_path:  "/ca_secrets/file_ca.key"
+          key_pass: "{{- required "FileCA keyPass is mandatory" .keyPass }}"
```
migmartri, line 36: should you make it issuer by default to not to break compatibility?

app/controlplane/pkg/ca/ca.go
```diff
@@ -18,12 +18,82 @@ package ca
...
+	return c.CAs
+}
+
+func (c *CertificateAuthorities) GetSignerCA() (CertificateAuthority, error) {
+	if c.SignerCA != nil {
+		return c.SignerCA, nil
+	}
+
+	// use as signer if only one is available
+	if len(c.CAs) == 1 {
```
migmartri, line 94: mm, I am not convinced about having a fallback here. I'd rather that we set it in NewCertificateAuthoritiesFromConfig

migmartri (commented):
Thanks @jiparis

app/controlplane/internal/server/grpc.go
```diff
@@ -254,7 +254,7 @@ func allowListEnabled() selector.MatchFunc {
 }
 
 func requireRobotAccountMatcher() selector.MatchFunc {
-	const requireMatcher = "controlplane.v1.AttestationService/.*|controlplane.v1.AttestationStateService/.*|controlplane.v1.SigningService/.*"
+	const requireMatcher = "controlplane.v1.AttestationService/.*|controlplane.v1.AttestationStateService/.*|controlplane.v1.SigningService/GenerateSigningCert"
```
jiparis, line 257: This is needed, since the new endpoint should be available for `workflow run describe`. This setting will make it public, but we might want to move it to its own service.

app/controlplane/cmd/wire.go
jiparis: Yes, it was not using wire, not sure why I didn't do it at that time.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -192,6 +195,8 @@ message CA {
     EJBCA ejbca_ca = 2;
   }
 
+  bool issuer = 3; // one and only one at a time when used in certificate_authorities field
```
migmartri, line 198: nitpick, document what this does, i.e will be used to indicate the signer

app/controlplane/internal/service/signing.go
```diff
@@ -53,3 +53,15 @@ func (s *SigningService) GenerateSigningCert(ctx context.Context, req *v1.Genera
 
 	return &v1.GenerateSigningCertResponse{Chain: &v1.CertificateChain{Certificates: certs}}, nil
 }
+
+func (s *SigningService) GetTrustedRoot(ctx context.Context, _ *v1.GetTrustedRootRequest) (*v1.GetTrustedRootResponse, error) {
```
migmartri, line 57: just to make sure, what happens if you are not using keyless att all?

app/controlplane/internal/server/grpc.go
migmartri: I think it's fine since it can enable offline verification

javirln (approved):

deployment/chainloop/templates/controlplane/secret-config.yaml
```diff
@@ -29,24 +29,26 @@ stringData:
...
     {{- with .Values.controlplane.keylessSigning.ejbcaCA }}
-    certificate_authority:
-      ejbca_ca:
-        cert_path: "/ca_secrets/ejbca_client.cert"
-        key_path:  "/ca_secrets/ejbca_client.key"
-        server_url: "{{- required "EJBCA server URL is mandatory" .serverURL }}"
-        certificate_profile_name: "{{- required "EJBCA certificate profile name is mandatory" .certProfileName }}"
-        end_entity_profile_name: "{{- required "EJBCA end entity profile name is mandatory" .endEntityProfileName }}"
-        certificate_authority_name: "{{- required "EJBCA certificate authority name is mandatory" .caName }}"
+    certificate_authorities:
```
migmartri, line 43: should we add the issuer option in the values.yaml file too? Or should we do it later on to support multiple CASs?

migmartri (approved):
Thanks @jiparis

app/controlplane/internal/service/signing.go
jiparis: Good point. We can choose to either return a NotImplementedErr or just an empty list. I would vote for the error, just to denote that it's not configured.

deployment/chainloop/templates/controlplane/secret-config.yaml
jiparis: Yes, I didn't want to disrupt current values.yaml. I would add it when we support multiple CAs.

app/controlplane/internal/service/signing.go
jiparis: Update: I'm returning an error, as we were doing in `GetSigningCert` method as well.

PR #1813, opened 2025-02-13
feat(verification): option to verify bundle server side in WorkflowRun/View endpoint

jiparis:
This PR introduces a new parameter (`Verify: true`) in the request to perform bundle verification on WorkfloRun/View endpoint. 
Note this is just a convenience and doesn't replace client side validation.
Refs #914


app/controlplane/pkg/biz/workflowrun.go
```diff
@@ -388,6 +395,50 @@ func (uc *WorkflowRunUseCase) GetByIDInOrg(ctx context.Context, orgID, runID str
 	return wfRun, nil
 }
 
+type VerificationResult struct {
+	Result        bool
+	FailureReason string
+}
+
+func (uc *WorkflowRunUseCase) Verify(ctx context.Context, run *WorkflowRun) (*VerificationResult, error) {
```
migmartri, line 403: what would happen if there is no bundle, i.e an old attestation?

migmartri (approved):
Thanks @jiparis

I put inline one question about knowing how this code behaves with old attestations that don't have bundle

app/controlplane/pkg/biz/workflowrun.go
jiparis: It would just skip the verification, but I've added an additional check to have full control.

migmartri (approved):

PR #1814, opened 2025-02-14
feat(verification): attestation verify command

jiparis:
This PR adds a new subcommand to the attestation command to verify a downloaded attestation bundle.
```shell
➜  chainloop wf run describe --digest sha256:0c432b7de0dde8c1aea926086e9346980c03abecd3461e00bb87413f50466037 --output attestation > att.json

➜  chainloop att verify att.json
INF attestation verified successfully
```

It also supports http(s) endpoints.
```shell
➜ chainloop att verify https://gist.githubusercontent.com/jiparis/fac2dbc4488503b14779ac590618d525/raw/d577442adec89221201c4403102dd5603eb0c803/gistfile1.txt
INF attestation verified successfully
```

### DIFF
--- a/app/cli/cmd/attestation_verify.go
+++ b/app/cli/cmd/attestation_verify.go
@@ -1,0 +1,53 @@
+//
+// Copyright 22025 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
+	"github.com/spf13/cobra"
+)
+
+func newAttestationVerifyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "verify file-or-url",
+		Short:                 "verify an attestation",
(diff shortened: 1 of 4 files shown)


app/cli/cmd/attestation_verify.go
```diff
@@ -0,0 +1,53 @@
+//
+// Copyright 22025 The Chainloop Authors.
```
migmartri, line 2: not yet :)

app/cli/cmd/attestation_verify.go
```diff
@@ -0,0 +1,53 @@
...
+		Short:                 "verify an attestation",
+		Long:                  "Verify an attestation by validating its validation material against the configured trusted root",
+		DisableFlagsInUseLine: true,
+		Example: `  # verify local attestation
+  chainloop attestation verify attestation.json
+
+  # verify an attestation stored in an https endpoint
+  chainloop attestation verify https://myrepository/attestation.json`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			res, err := action.NewAttestationVerifyAction(actionOpts).Run(cmd.Context(), args[0])
```
migmartri, line 37: note that this is the first command that we do that relies on arguments and no flags, if we want to be consistent, we should use a flag i.e `-b` or `--bundle`. 

I am not saying that using an argument is wrong, in fact I have mixed feelings about our current approach, but I am just pointing out the inconsistency. What's your take?

In the future we might want to also allow passing the root so they do not need to contact the controlplane, but that can come later.

app/cli/internal/action/attestation_verify.go
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri, line 2: 2025, this is a new file :)

migmartri (approved):
See my comment about consistency in the command, thanks!

app/cli/cmd/attestation_verify.go
jiparis: Yes, make sense, I'll implement the change.

PR #1821, opened 2025-02-17
feat(verification): support non up to spec bundles

jiparis:
some bundles were recorded as 0.3 but contained a CertificateChain, which is illegal according to specs. They can still be verified as long as the root material is taken out of band.
This PR adds support for such case. The plan is to remove it in a future release.

Refs #914


pkg/attestation/verifier/verifier.go
```diff
@@ -49,22 +49,46 @@ func VerifyBundle(ctx context.Context, bundleBytes []byte, tr *TrustedRoot) erro
 		return fmt.Errorf("invalid bundle: %w", err)
 	}
 
+	var signingCert *x509.Certificate
 	if bundle.GetVerificationMaterial() == nil || bundle.GetVerificationMaterial().GetCertificate() == nil {
-		// nothing to verify
-		return ErrMissingVerificationMaterial
-	}
+		// it's a malformed bundle (according to specs) but still verifiable
+		if bundle.GetVerificationMaterial().GetX509CertificateChain() != nil {
```
migmartri, line 55: I'd add a todo here to remove it and mark it as deprecated

migmartri (approved):

pkg/attestation/verifier/verifier.go
jiparis: Done.

PR #1825, opened 2025-02-18
feat(attestation): federated verification

### LINKED ISSUE #1817: feat: identity provider token delegation
We would like to allow our users to leverage native tokens such as Github/Gitlab tokens instead of chainloop tokens during attestation. 

This issue is about designing and implementing a delegation mechanism

migmartri:
This patch adds support for delegated, federaetd verification during the attestation process.

This means that when an API token is received during attestation, now, in addition to the built-in chainloop authenticator, we can delegate the authN/authZ to a third party service that we already trust.

This feature is configurable via the Helm Chart

```
2025-02-18T11:34:22.672+0100	INFO	{"component": "jwtMiddleware", "msg": "federated verification enabled, using URL: http://localhost:8002/machine-identity/verify-token"}
```

Demo: Attestation process using a Gitlab token, there is no difference :)

```
go run main.go att init --workflow from-gitlab --project chainloop-contrib --replace --version 1.2.3 --release=false
┌───────────────────────────┬────────────────────────────────────────────┐
│ Initialized At            │ 18 Feb 25 10:34 UTC                        │
├───────────────────────────┼────────────────────────────────────────────┤
│ Attestation ID            │ e8ef6e7d-d9be-48dd-89ca-97862ba5827f       │
│ Organization              │ foobar                                     │
│ Name                      │ from-gitlab                                │
│ Project                   │ chainloop-contrib                          │
│ Version                   │ 1.2.3 (already released)                   │
│ Contract                  │ chainloop-contrib-from-gitlab (revision 1) │
│ Policy violation strategy │ ENFORCED                                   │
└───────────────────────────┴────────────────────────────────────────────┘
```

You can also provide the --org to the att command if needed 

Closes #1817


app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -83,13 +83,25 @@ message Bootstrap {
...
   message NatsServer {
     // Connection URI
     string uri = 1 [(buf.validate.field).string.min_len = 1];
     // TODO: add authentication options
   }
 }
 
+message FederatedVerification {
+  // URL of the federated verification endpoint
+  string url = 1 [(buf.validate.field).string.min_len = 1];
```
javirln, line 100: The `(buf.validate.field).string.uri = true` couldn't work here?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -160,14 +173,40 @@ func withTokenProvider(providerKey string, opts ...TokenProviderOption) JWTOptio
...
 
+// WithFederatedProvider adds support to ask a third party service to verify the token
+// verify URL must be an API that receives a json encoded body with the following structure:
+//
+//	{
+//		"token": "<jwt token>",
+//		"org_name": "<organization name>"
+//	}
+//
+// and returns a json with the following structure:
```
javirln, line 184: It's missing the structure right?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware_test.go
```diff
@@ -126,8 +128,9 @@ func TestAttestationAPITokenProvider(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			ctx := transport.NewServerContext(context.Background(), &mockTransport{reqHeader: tc.tokenHeader})
+			logger := log.NewStdLogger(io.Discard)
```
javirln, line 131: Is this a leftover?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -191,6 +230,29 @@ func WithJWTMulti(opts ...JWTOption) middleware.Middleware {
 							continue
 						}
 
+						// If federated verification is enabled, we try to get the information remotely
+						if o.federatedVerificationURL != "" {
```
jiparis, line 234: I might be missing where the decision is made, but Is this logic called always, no matter if the token belongs to the federated provider?

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware_test.go
migmartri: No, a logger is now required by the use case

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -83,13 +83,25 @@ message Bootstrap {
   // nats server configuration for sending events
   NatsServer nats_server = 14;
 
+  // Federated verification allows to delegate the API tokens received during the attestation
+  // to a third party service for verification
+  // For example, we can forward Gitlab or Github API tokens
+  FederatedVerification federated_verification = 16;
```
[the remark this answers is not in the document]
migmartri: changed

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
migmartri: That's correct, the federated provider is the one that will figure out if it understands the provided token

jiparis: Cannot this be a performance issue?  Doesn't it have an `issuer` claim that can be used to early detect if the token is federated or not?

javirln (approved):
LGTM!

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
jiparis: Calling an external URL in every single request to check a token that might be already been processed and understood. Unless I'm missing something, of course.

PR #1875, opened 2025-03-04
chore: redirect to login URL with error message

jiparis:
When there is a callback error (ie, when SSO is delegated to Auth0), this PR redirects to the configured login URL with the provided error message from the Auth provider.

### DIFF
--- a/app/controlplane/internal/service/auth.go
+++ b/app/controlplane/internal/service/auth.go
@@ -65,7 +65,7 @@ type oauthResp struct {
 	code          int
 	err           error
 	showErrToUser bool
-	redirectURL   string
+	redirectURL   *url.URL
 }
 
 // NewOauthResp builds an oauthResp object without redirection
@@ -203,8 +203,8 @@ func (svc *AuthService) RegisterLoginHandler() http.Handler {
 // Implement http.Handler interface
 func (h oauthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if resp := h.H(h.svc, w, r); resp != nil {
-		if resp.redirectURL != "" {
-			http.Redirect(w, r, resp.redirectURL, http.StatusTemporaryRedirect)
+		if resp.redirectURL != nil {
+			http.Redirect(w, r, resp.redirectURL.String(), http.StatusTemporaryRedirect)
 			return
 		}
 		http.Error(w, resp.ErrorMessage(h.svc.log), resp.code)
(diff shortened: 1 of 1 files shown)


app/controlplane/internal/service/auth.go
```diff
@@ -258,33 +274,34 @@ func (c *upstreamOIDCclaims) preferredEmail() string {
 
 func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) *oauthResp {
 	ctx := context.Background()
-	// if OIDC provider returns an error, return it to and show it to the user
-	if desc := r.URL.Query().Get("error_description"); desc != "" {
-		return &oauthResp{http.StatusUnauthorized, errors.New(desc), true}
+	// if OIDC provider returns an error, redirect to the login page to and show it to the user
+	if desc := r.URL.Query().Get(oidcErrorParam); desc != "" {
+		redirectURL := fmt.Sprintf("%s?%s=%s", svc.AuthURLs.Login, oidcErrorParam, desc)
```
javirln, line 279: I'm not sure if we can use here `url.Parse` and build the final URL. I'm not sure if it needs to be encoded.

See this example: https://go.dev/play/p/YCTvdluws-r

jiparis: Yes, makes sense. I'll encode the parameter value.

app/controlplane/internal/service/auth.go
```diff
@@ -258,33 +274,42 @@ func (c *upstreamOIDCclaims) preferredEmail() string {
 
 func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) *oauthResp {
 	ctx := context.Background()
-	// if OIDC provider returns an error, return it to and show it to the user
-	if desc := r.URL.Query().Get("error_description"); desc != "" {
-		return &oauthResp{http.StatusUnauthorized, errors.New(desc), true}
+	// if OIDC provider returns an error, redirect to the login page to and show it to the user
+	if desc := r.URL.Query().Get(oidcErrorParam); desc != "" {
+		redirectURL, err := url.Parse(svc.AuthURLs.Login)
```
migmartri, line 279: I am not sure we should redirect always but only if the login landing page is set no? https://github.com/chainloop-dev/chainloop/blob/e07cc0fd7674d9224fa5861007237285727746ac/app/controlplane/internal/conf/controlplane/config/v1/conf.proto#L193

Otherwise this will cause a redirect loop?

PR #1876, opened 2025-03-04
chore: only redirect if login URL is overridden

jiparis:
We want to redirect only in case there is a dedicated login page, otherwise we can incur in login loops.

### DIFF
--- a/app/controlplane/internal/service/auth.go
+++ b/app/controlplane/internal/service/auth.go
@@ -155,7 +155,8 @@ func NewAuthService(userUC *biz.UserUseCase, orgUC *biz.OrganizationUseCase, mUC
 }
 
 type AuthURLs struct {
-	Login, callback string
+	Login, callback   string
+	loginIsOverridden bool
 }
 
 // urlScheme is deprecated, now it will be inferred from the serverConfig externalURL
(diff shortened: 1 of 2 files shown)


javirln (approved):

app/controlplane/internal/service/auth.go
```diff
@@ -285,7 +294,7 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) *
 		q.Set(oidcErrorParam, desc)
 		redirectURL.RawQuery = q.Encode()
 
-		return &oauthResp{http.StatusUnauthorized, errors.New(desc), true, redirectURL}
+		return &oauthResp{http.StatusUnauthorized, nil, true, redirectURL}
```
migmartri, line 297: are you sure this is correct? Is `svc.AuthURLs.Login` either the regular login or the one overriden?

jiparis: Yes, `Login` holds the overridden one when it's configured: https://github.com/chainloop-dev/chainloop/blob/7edf868b50493189f4f644d981497f579625e24a/app/controlplane/internal/service/auth.go#L182

PR #1888, opened 2025-03-11
feat(workflows): Allow to filter workflows that needs attention reg compliance

javirln:
This patch allows to filter workflows that _need attention_. It means that its compliance is not met.


migmartri:
What happened?

javirln:
> What happened?

I found a way of doing it how we thought at the begging using JSON filters

PR #1889, opened 2025-03-11
feat(workflows): Allow to filter workflows using a JSON filter

javirln:
This patch allows to filter Workflows using a JSON filter. A JSON filter is a filter that works over JSON data in a specific column. Such data shall be stored in the database.

The JSON filter is later converted into a valid SQL statement.


app/controlplane/api/jsonfilter/v1/jsonfilter.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package jsonfilter.v1;
+
+import "buf/validate/validate.proto";
+
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/jsonfilter/v1";
```
migmartri, line 7: is there any chance we could import this through buf instead?

app/controlplane/pkg/data/workflow.go
```diff
@@ -137,7 +139,7 @@ func (r *WorkflowRepo) List(ctx context.Context, orgID uuid.UUID, filter *biz.Wo
 	}
 
 	// Initialize the base query for WorkflowRun
-	baseQuery := orgScopedQuery(r.data.DB, orgID).QueryWorkflows()
+	baseQuery := orgScopedQuery(r.data.DB.Debug(), orgID).QueryWorkflows()
```
migmartri, line 142: debug

pkg/jsonfilter/jsonfilter.go
```diff
@@ -0,0 +1,150 @@
...
+	entsql "entgo.io/ent/dialect/sql"
+	"entgo.io/ent/dialect/sql/sqljson"
+)
+
+// columnNotFoundSQLState is the PostgresSQL state for column not found error.
+// Reference: https://www.postgresql.org/docs/8.2/errcodes-appendix.html
+const columnNotFoundSQLState = "42703"
+
+// columnNotExistRegex extracts column name from the error message
+var columnNotExistRegex = regexp.MustCompile(`column "([^"]+)" does not exist`)
```
migmartri, line 33: I'd not support dynamic selection of columns, that's an implementation detail to me that should not be exposed in then API

app/controlplane/api/jsonfilter/v1/jsonfilter.proto
migmartri: nevermind, this is custom code.

migmartri (commented):
Thanks @javirln, my take on this implementation

- I'd not allow dynamically supporting column selection, that's an implementation details that I do not think it should get exposed at the API layer, nevertheless I am scared that it could be a threat ventor (i.e crafting to query columns that you are not supposed to query). Instead, I'd let the data layer to decide what column the filter must be applied to.
- In the name of Yagni (which I can be quite opinionated sometimes), I'd not add any other code than the one we are going to use. This means, no more selectors than the one that we plan to use, no code is the best code.

Thanks!

jiparis (approved):

app/controlplane/api/jsonfilter/v1/jsonfilter.proto
```diff
@@ -0,0 +1,21 @@
...
+// JSONFilter represents a filter for JSON fields.
+message JSONFilter {
+  string field_path = 1 [(buf.validate.field).string.min_len = 1];
+  JSONOperator operator = 2 [(buf.validate.field).enum = {defined_only: true}];
+  optional string value = 43;
+}
+
+// JSONOperator represents supported JSON filter operators.
+enum JSONOperator {
+  JSON_OPERATOR_UNSPECIFIED = 0;
```
jiparis, line 18: Looking at the payload, I see these operators too verbose. I'd just set them to "EQ", "NEQ".

jiparis (approved):
Awesome, @javirln . Just one nit comment about the operator verbosity. 
Thanks!

app/controlplane/api/jsonfilter/v1/jsonfilter.proto
javirln: I know but I think the protobuf linter will complain if the enum values are not prefixed by the name of the enum. I believe it's also recommended [here](https://protobuf.dev/programming-guides/style/), section Enums:
```
Name collisions are a high risk when enums are defined at the top level of a file (not nested inside a message definition); in that case the siblings include enums defined in other files that set the same package, where protoc may not be able to detect the collision has occurred at code generation time.

To avoid these risks, it is strongly recommended to do one of:

    Prefix every value with the enum name (converted to UPPER_SNAKE_CASE)
    Nest the enum inside a containing message
```

jiparis: Thanks for the clarification. Then I guess the only solution is to nest `JSONOperator` inside another message, but it may not worth it or apply.

app/controlplane/pkg/data/workflow.go
```diff
@@ -223,6 +225,22 @@ func applyWorkflowFilters(wfQuery *ent.WorkflowQuery, opts *biz.WorkflowListOpts
...
+				// Build the predicates for each JSON filter
+				predicates := make([]*sql.Predicate, 0, len(opts.JSONFilters))
+				for _, filter := range opts.JSONFilters {
+					// Include the column where the filter is applied
+					filter.Column = workflow.FieldMetadata
+					jsonPredicate, _ := jsonfilter.BuildEntSelectorFromJSONFilter(filter)
+					predicates = append(predicates, jsonPredicate)
+				}
+				// Combine the predicates using OR logic
+				selector.Where(sql.Or(predicates...))
```
migmartri, line 240: interestingly, you chose to make `or` the default; it feels counterintuitive, no?

I mean if you provide multiple filters, my fist thought would be to become an `AND`. What do you think?

javirln: You might be right I was just biassed by another PR. I've changed it to merge all filters in an AND. Thanks.

PR #1895, opened 2025-03-13
fix(docs): update policy examples

### LINKED ISSUE #1882: outdated policy examples
https://github.com/chainloop-dev/chainloop/tree/main/docs/examples/policies

jiparis:
Some policies were outdated (using old Rego version and Chainloop result rule).

Fixes #1882


docs/examples/policies/chainloop-commit.yaml
```diff
@@ -17,18 +17,27 @@ kind: Policy
 metadata:
   name: chainloop-commit
 spec:
-  type: ATTESTATION
-  embedded: |
-    package main
+  policies:
+    - type: ATTESTATION
+      embedded: |
+        package main
```
migmartri, line 23: should we use the actual template we use with comments and so on?

jiparis: Yes, it makes sense. I'll add those comments

migmartri (approved):

PR #1902, opened 2025-03-13
att init: create contract on workflow creation

### LINKED ISSUE #1901: Create contract on `att init` when the workflow is created
Currently, when the workflow will be autocreated, `att init` admits a `--contract` option to specify the contract to use. The goal of this task is to also allow passing a URL (local/remote) so that it can be used as the contract for the new workflow.

jiparis:
This PR allows to reuse `--contract` flag in `att init` to specify a contract file or URL to be used if the workflow has to be created. In this case, it will first create the contract with the file contents and pass the reference to the Workflow/Create call.

Fixes #1901

### DIFF
--- a/app/cli/cmd/attestation_init.go
+++ b/app/cli/cmd/attestation_init.go
@@ -90,7 +90,7 @@ func newAttestationInitCmd() *cobra.Command {
 						ProjectName:                  projectName,
 						ProjectVersion:               projectVersion,
 						WorkflowName:                 workflowName,
-						NewWorkflowContractName:      newWorkflowcontractName,
+						NewWorkflowContractRef:       newWorkflowcontractName,
 						ProjectVersionMarkAsReleased: projectVersionRelease,
 					})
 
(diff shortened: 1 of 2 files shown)


app/cli/cmd/attestation_init.go
```diff
@@ -146,7 +146,7 @@ func newAttestationInitCmd() *cobra.Command {
 
 	cmd.Flags().StringVar(&projectName, "project", "", "name of the project of this workflow")
 	cobra.CheckErr(cmd.MarkFlagRequired("project"))
-	cmd.Flags().StringVar(&newWorkflowcontractName, "contract", "", "name of an existing contract to attach it to the auto-created workflow (it doesn't update an existing one)")
+	cmd.Flags().StringVar(&newWorkflowcontractName, "contract", "", "name of an existing contract or the path/URL to a contract file, to attach it to the auto-created workflow (it doesn't update an existing one)")
```
migmartri, line 149: change the name of the variable

migmartri (approved):

PR #1907, opened 2025-03-18
fix: Store DSSE envelope when bundle is not available

jiparis:
When using old CLI versions, only DSSE envelope is available. There is a bug introduced in #1890 that prevents it from being stored in the WorkflowRun table. This PR fixes it, so that it's backwards compatible.

### DIFF
--- a/app/controlplane/pkg/data/workflowrun.go
+++ b/app/controlplane/pkg/data/workflowrun.go
@@ -183,6 +183,10 @@ func (r *WorkflowRunRepo) FindByIDInOrg(ctx context.Context, orgID, id uuid.UUID
 func (r *WorkflowRunRepo) SaveAttestation(ctx context.Context, id uuid.UUID, att *dsse.Envelope, digest string) error {
 	q := r.data.DB.WorkflowRun.UpdateOneID(id).
 		SetAttestationDigest(digest)
+
+	// the envelope will come empty in normal attestations, since bundles are stored separately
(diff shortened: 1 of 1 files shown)


app/controlplane/pkg/data/workflowrun.go
```diff
@@ -181,9 +181,13 @@ func (r *WorkflowRunRepo) FindByIDInOrg(ctx context.Context, orgID, id uuid.UUID
...
 // SaveAttestation Saves the attestation for a workflow run in the database
 func (r *WorkflowRunRepo) SaveAttestation(ctx context.Context, id uuid.UUID, att *dsse.Envelope, digest string) error {
-	run, err := r.data.DB.WorkflowRun.UpdateOneID(id).
-		SetAttestationDigest(digest).
-		Save(ctx)
+	q := r.data.DB.WorkflowRun.UpdateOneID(id).
+		SetAttestationDigest(digest)
+	if att != nil {
+		// Set attestation when using old CLI versions
+		q.SetAttestation(att)
```
migmartri, line 188: but I thought that we wanted to store always bundle. Didn't we end up doing some sort of automatic wrapping on a bundle?

migmartri (approved):

app/controlplane/pkg/data/workflowrun.go
jiparis: We could. But even if we do it, old attestations must still be supported

PR #1918, opened 2025-03-25
chore: upgrade atlas

migmartri:
0.32 in fact didn't fix the issue

### DIFF
--- a/app/controlplane/Dockerfile.migrations
+++ b/app/controlplane/Dockerfile.migrations
@@ -1,7 +1,7 @@
 # Container image built by go-releaser that's used to run migrations against the database during deployment
 # See https://atlasgo.io/guides/deploying/image
-# arigaio/atlas:0.32.0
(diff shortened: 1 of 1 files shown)


app/controlplane/Dockerfile.migrations
```diff
@@ -1,7 +1,7 @@
 # Container image built by go-releaser that's used to run migrations against the database during deployment
 # See https://atlasgo.io/guides/deploying/image
-# arigaio/atlas:0.32.0
-FROM arigaio/atlas@sha256:a7bccf43cf2c0f09d1203096b4727e64d68d0fb924b2b91beb2a4b43eae50b7e as base
+# arigaio/atlas:latest
```
javirln, line 3: Can you exec the container image locally and print the version of atlas to write it here please?

javirln (approved):

app/controlplane/Dockerfile.migrations
migmartri: Done, good idea. Also added a comment on what I did

Issue #1920, opened 2025-03-26
dynamically show cas backend during downloads


migmartri:
It seems that the `artifact download` command does not leverage the `cas_mappings` table and hence it does not "remember" where the file is uploading, falling back to the default cas-backend.

PR #1921, opened 2025-03-26
feat: decouple cas-mappings table from attestations

migmartri:
This PR allows the cas_mapping table to be used for CAS entries coming from different sources, not just attestations. We achieve this by removing the requirement of having a `run_id` in the DB entry plus tweaking the visibility flag, which was the reason runs are used in the first place. 

Refs #1920


app/controlplane/internal/service/cascredential.go
```diff
@@ -103,7 +103,7 @@ func (s *CASCredentialsService) Get(ctx context.Context, req *pb.CASCredentialsS
 		}
 
 		// If we can't find a mapping, we'll use the default backend
-		if err != nil && !biz.IsNotFound(err) && !biz.IsErrUnauthorized(err) {
```
migmartri, line 106: I realized that this error was not useful in general since it was representing a NotFound, that's why I simplified it.

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -54,15 +54,20 @@ func NewCASMappingUseCase(repo CASMappingRepo, mRepo MembershipRepo, logger log.
 	return &CASMappingUseCase{repo, mRepo, servicelogger.ScopedHelper(logger, "cas-mapping-usecase")}
 }
 
-func (uc *CASMappingUseCase) Create(ctx context.Context, digest string, casBackendID, workflowRunID string) (*CASMapping, error) {
+// Create a mapping with an optional workflow run id
+func (uc *CASMappingUseCase) Create(ctx context.Context, digest string, casBackendID string, workflowRunID string) (*CASMapping, error) {
```
migmartri, line 58: kept the run as string, not pointer, to limit the blast radius of the change.

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -101,14 +106,28 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByUser(ctx context.Context
...
+func (uc *CASMappingUseCase) FindCASMappingForDownloadByOrg(ctx context.Context, digest string, orgs []string) (result *CASMapping, err error) {
 	if _, err := cr_v1.NewHash(digest); err != nil {
 		return nil, NewErrValidation(fmt.Errorf("invalid digest format: %w", err))
 	}
 
+	// log the result
+	defer func() {
+		if result != nil {
+			uc.logger.Infow("msg", "mapping found!", "digest", digest, "orgs", orgs, "casBackend", result.CASBackend.ID, "default", result.CASBackend.Default, "public", result.Public)
+		} else if err == nil || IsNotFound(err) {
```
migmartri, line 126: we now also log if we didn't find it, before we were logging only the successes

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -129,24 +148,19 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByOrg(ctx context.Context,
...
+		return defaultOrFirst(orgMappings), nil
 	}
 
 	// 3 - mappings that are public
 	publicMappings := filterByPublic(mappings)
 	// The user has not access to neither proprietary nor public mappings
 	if len(publicMappings) == 0 {
 		uc.logger.Warnw("msg", "digest exist but user does not have access to it", "digest", digest, "orgs", orgs)
-		return nil, NewErrUnauthorized(errors.New("unauthorized access to the artifact"))
+		return nil, NewErrNotFound("digest not found in any mapping")
```
migmartri, line 159: I can't see the reason we wanted to differentiate this from NotFound in practice.

PR #1923, opened 2025-03-27
feat(users): Store user access on user's table

javirln:
This patch adds a new column, `has_restricted_access`, to the `Users` table. While not currently in use, it is synchronized with the allowlist configured on the server.

A background goroutine runs at startup, syncing the `Users` table with the allowlist every 10 seconds.

As we have in the allowlist middleware, if the allowlist is empty, we allow all users to get in. In this scenario, it means to update all users `has_restricted_access=true` to switch it `false`.


app/controlplane/pkg/biz/useraccess_syncer.go
```diff
@@ -0,0 +1,188 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			u.logger.Infow("msg", "stopping user access sync")
+			return nil
+		case <-ticker.C:
+			u.logger.Infow("msg", "Syncing user access")
+
+			// Count the number of users with restricted access
+			usersWithRestrictedAccess, err := u.userRepo.CountUsersWithRestrictedAccess(ctx)
```
migmartri, line 61: move inside

app/controlplane/pkg/biz/useraccess_syncer.go
```diff
@@ -0,0 +1,188 @@
...
+
+		offset++
+	}
+
+	return nil
+}
+
+// updateUserAccessRestriction updates the access restriction status of a user
+func (u *UserAccessSyncerUseCase) updateUserAccessRestriction(ctx context.Context, user *User, isAllowListDeactivated bool) error {
+	allow, err := UserEmailInAllowlist(u.allowList.GetRules(), user.Email)
```
migmartri, line 132: we do not need to do it if deactivated

javirln:
Updated the PR to:
- By default `has_restricted_access` to NULL
- No sync if the allow list is empty and no users with restricted access
- Code cleanup

jiparis (approved):

migmartri (approved):

PR #1935, opened 2025-04-01
chore(ci): Include all catalogers in syft analysis

javirln:
This patch reverts the PR #1873 since the underlaying NTIA policy has been modified to allow local files.


migmartri (approved):
Thanks

PR #1952, opened 2025-04-07
Add SLSA provenance after GoReleaser

gr0:
Simple change adding the SLSA attestation action just after GoReleaser.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -44,6 +44,8 @@ jobs:
       contents: write # required for goreleaser to upload the release assets
       packages: write # to push container images
       pull-requests: write
+      id-token: write # required for SLSA provenance
+      attestations: write # required for SLSA provenance
     env:
       CHAINLOOP_TOKEN: ${{ secrets.CHAINLOOP_TOKEN }}
       ATTESTATION_ID: ${{ needs.init_attestation.outputs.attestation_id }}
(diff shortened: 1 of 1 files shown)


migmartri:
Thanks for your contribution @gr0 

It seems that DCO check is not happy, could you make sure you are committing with `-s` flag? 

More info here https://github.com/chainloop-dev/chainloop/blob/main/CONTRIBUTING.md

.github/workflows/release.yaml
```diff
@@ -100,6 +102,11 @@ jobs:
           POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
           POSTHOG_ENDPOINT: ${{ secrets.POSTHOG_ENDPOINT }}
 
+      - name: Build SLSA Attestation
+        uses: actions/attest-build-provenance@v2
+        with:
+          subject-path: dist/**/*
```
migmartri, line 108: so this is going to attest the binaries? Should we also attest the container images? It seems to be possible with this [attestor](https://github.com/actions/attest-build-provenance?tab=readme-ov-file#container-image)

The container image format can be found in the go-releaser file and are basically these three

image

.github/workflows/release.yaml
```diff
@@ -100,6 +102,11 @@ jobs:
           POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
           POSTHOG_ENDPOINT: ${{ secrets.POSTHOG_ENDPOINT }}
 
+      - name: Build SLSA Attestation
```
migmartri, line 105: btw, this is not taking the result of the generation and passing it to Chainloop as part of the attestation as a material. How can we do that?

.github/workflows/release.yaml
gr0: Yes, finally all elements should be included in attestation.

Issue #1958, opened 2025-04-10
Include repository source code in Chainloop's attestation


javirln:
Once Goreleaser finishes and pushes the GitHub release, by default the source code of the repository is also added to the release. The goal of the task would be to grab that compressed archive and include it in the Chainloop's attestation.

PR #1963, opened 2025-04-10
feat(ci): Include source code to chainloop attestation

javirln:
This patch adds the source code of the GitHub release to Chainloop's attestation.

Close: #1958

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -176,6 +176,13 @@ jobs:
         run: |
           chainloop attestation --name slsa-attestation add --value ${{ steps.slsa-attest.outputs.bundle-path }} --kind SLSA_PROVENANCE --attestation-id ${{ env.ATTESTATION_ID }}
 
+      - name: Include source code on attestation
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -176,6 +176,13 @@ jobs:
         run: |
           chainloop attestation --name slsa-attestation add --value ${{ steps.slsa-attest.outputs.bundle-path }} --kind SLSA_PROVENANCE --attestation-id ${{ env.ATTESTATION_ID }}
```
migmartri, line 177: out of curiosity why are we using remote state?

migmartri (approved):

.github/workflows/release.yaml
javirln: I can’t remember exactly, but maybe we needed it before—like if we were using a matrix in the job or something.

We could probably drop the remote state; I don’t think there’s anything really making us use it anymore or at least I cannot think of anything on top of my head right now.

PR #1967, opened 2025-04-11
chore(contract): Modify name of contract

javirln:
This patch renames the contract from `release` to `chainloop-vault-release`, which is the actual contract used by the `release` workflow in Chainloop. Previously, even though the contract was being updated, it had no effect—because no workflow was actually linked to it.


.github/workflows/contracts/chainloop-vault-release.yml
```diff
@@ -1,4 +1,4 @@
-# Contract for the build-and-package workflow
+# Contract for the release.yaml workflow
```
migmartri, line 1:
```suggestion
# Contract for the release workflow
```

migmartri (approved):

PR #1974, opened 2025-04-11
feat(attestation): update contracts on attestation init

### LINKED ISSUE #1694: confusing behavior att init --contract
when you run for example `chainloop att init --workflow my-wf --contract my-contract ...` what happens is

1 - a new workflow called `my-wf` is created (if it doesn't exist) associated with the contract `my-contract`
2 - a new run is performed

This means that if you after the first run, you provide another contract name `chainloop att init --workflow my-wf --contract contract-2`, **the first contract (`my-contract` will be used, in other words, the provided `contract-2` will be ignored**, which was reported as unexpected behavior

jiparis:
Previously the `--contract` argument in `att init` was used only in the case the workflow must be created. With this change, the contract is always updated, no matter if the workflow is new or not.
Chainloop takes care of the actual update happening only if there are changes in the contract content.

Fixes #1694


app/cli/internal/action/attestation_init.go
```diff
@@ -103,21 +103,37 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
-	if err != nil && status.Code(err) == codes.NotFound {
-		// Not found, let's see if we need to create the contract
-		if contractRef != "" {
-			// Try to find it by name
-			_, err := NewWorkflowContractDescribe(action.ActionsOpts).Run(contractRef, 0)
-			// An invalid argument might be raised if we use a file or URL in the "name" field, which must be DNS-1123
-			// TODO: validate locally before doing the query
-			if err != nil && (status.Code(err) == codes.NotFound || status.Code(err) == codes.InvalidArgument) {
-				createResp, err := NewWorkflowContractCreate(action.ActionsOpts).Run(fmt.Sprintf("%s-%s", opts.ProjectName, opts.WorkflowName), nil, contractRef)
+	// Not found, let's see if we need to create the contract
```
migmartri, line 106: wrong comment

app/cli/internal/action/attestation_init.go
```diff
@@ -103,21 +103,37 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
+	if contractRef != "" {
```
migmartri, line 107: thanks for doing now the check before the call

app/cli/internal/action/attestation_init.go
```diff
@@ -103,21 +103,37 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
+				wfContractItem = contractWithRevision.Contract
 			}
+
+			// it exists, let's update it (chainloop will validate that there is an actual change in the contract file)
+			updateResp, err := NewWorkflowContractUpdate(action.ActionsOpts).Run(wfContractItem.Name, &wfContractItem.Description, contractRef)
+			if err != nil {
+				return "", err
+			}
+
+			contractRef = updateResp.Contract.Name
```
migmartri, line 134: isn't this just contractRef?

app/cli/internal/action/attestation_init.go
```diff
@@ -103,21 +103,37 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
-			// An invalid argument might be raised if we use a file or URL in the "name" field, which must be DNS-1123
-			// TODO: validate locally before doing the query
-			if err != nil && (status.Code(err) == codes.NotFound || status.Code(err) == codes.InvalidArgument) {
-				createResp, err := NewWorkflowContractCreate(action.ActionsOpts).Run(fmt.Sprintf("%s-%s", opts.ProjectName, opts.WorkflowName), nil, contractRef)
+	// Not found, let's see if we need to create the contract
+	if contractRef != "" {
+		// Try to find it by name
+		_, err := NewWorkflowContractDescribe(action.ActionsOpts).Run(contractRef, 0)
+		// An invalid argument might be raised if we use a file or URL in the "name" field, which must be DNS-1123
+		if err != nil && (status.Code(err) == codes.NotFound || status.Code(err) == codes.InvalidArgument) {
```
migmartri, line 111: I find this hack not ideal, should we just check if it's a file or HTTP?

app/cli/internal/action/attestation_init.go
```diff
@@ -103,21 +103,37 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
 				}
-				contractRef = createResp.Name
+			case err != nil:
+				return "", err
+			default:
+				wfContractItem = contractWithRevision.Contract
 			}
+
+			// it exists, let's update it (chainloop will validate that there is an actual change in the contract file)
+			updateResp, err := NewWorkflowContractUpdate(action.ActionsOpts).Run(wfContractItem.Name, &wfContractItem.Description, contractRef)
```
migmartri, line 129: for update, we should describe the workflow, and update the default contract?

migmartri (commented):
I find this code difficult to follow.

The way I see this is

- You provide a contractRef
  - no -> done
  - yes: It can be two things
    - a file or URL
      - Load the content of the file or URL
      - Check if the workflow exists
        - yes? -> update the contract (the associated one) with new content
        - no? -> create a default contract with content
    - an existing (or not) contract name -> just pass the reference to the att init create.

app/cli/internal/action/attestation_init.go
jiparis: I've refactored it

app/cli/internal/action/attestation_init.go
jiparis: fixed.

app/cli/internal/action/attestation_init.go
jiparis: Done.

jiparis:
I've refactored it, please take a look

javirln (approved):

app/cli/internal/action/attestation_init.go
```diff
@@ -242,6 +220,96 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 	return attestationID, nil
 }
 
+// createWorkflow creates a new workflow if it doesn't exist, as well as its contract
+func (action *AttestationInit) createWorkflow(ctx context.Context, opts *AttestationInitRunOpts) (*WorkflowItem, error) {
+	// 1. find workflow if exists
+	wf, err := NewWorkflowDescribe(action.ActionsOpts).Run(ctx, opts.WorkflowName, opts.ProjectName)
+	if err != nil && status.Code(err) != codes.NotFound {
+		return nil, fmt.Errorf("error looking for workflow %q: %w", opts.WorkflowName, err)
+	}
```
migmartri, line 229: to simplify readability here I'd `return wf` if wf != nil && contractRef == ""

to indicate there is nothing to create/update

app/cli/internal/action/attestation_init.go
```diff
@@ -242,6 +220,96 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
+		if err != nil {
+			return nil, fmt.Errorf("error creating workflow %q: %w", opts.WorkflowName, err)
+		}
+	}
+
+	return wf, nil
+}
+
+// isContractReference checks if the reference points to an existent contract name (DNS-1123
+func isContractReference(ref string) bool {
```
migmartri, line 290: can't this method just check if is `valid DNS-1123`? 

also, we should have that validation error in biz

migmartri (approved):

PR #2002, opened 2025-04-28
feat(cli): Generate CLI documentation

javirln:
This patch adds a script that generates a single `.mdx` file from the CLI command documentation. Each command becomes a main section, with its subcommands included as subsections.

It's integrated in the main `make generate` command.


jiparis (approved):
awesome!

migmartri (approved):

PR #2005, opened 2025-04-29
fix(cli-gen): Modifies ent generation

javirln:


.github/workflows/test.yml
```diff
@@ -50,13 +50,17 @@ jobs:
           make -C app/controlplane migration_sync
 
       # Check that the generated ent code is up to date
-      # see https://entgo.io/docs/ci/
-      - uses: ent/contrib/ci@e38dfb6484dfbe64b8bd060fe6a219a1aa5da770 # master
-        name: "Check all ent generated code is checked in"
+      - name: "Check all ent generated code is checked in"
         if: ${{ matrix.app != 'main-module' }}
```
jiparis: I think the solution is adding here `matrix.app != 'cli'` as well. We don't use Ent in the CLI

jiparis, line 54: I think the solution is adding here `matrix.app != 'cli'` as well. We don't use Ent in the CLI

.github/workflows/test.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       # see https://entgo.io/docs/ci/
       - uses: ent/contrib/ci@e38dfb6484dfbe64b8bd060fe6a219a1aa5da770 # master
         name: "Check all ent generated code is checked in"
-        if: ${{ matrix.app != 'main-module' }}
+        if: ${{ matrix.app != 'main-module' || matrix.app != 'cli' }}
```
jiparis, line 56: it should be `&&`

javirln:
Added on this PR https://github.com/chainloop-dev/chainloop/pull/1974

PR #2008, opened 2025-04-30
Support Gitlab authentication information

gr0:
This PR introduced checking for authentication information coming from Gitlab.

### DIFF
--- a/pkg/attestation/crafter/runners/gitlabpipeline.go
+++ b/pkg/attestation/crafter/runners/gitlabpipeline.go
@@ -95,7 +95,7 @@ func (r *GitlabPipeline) Environment() RunnerEnvironment {
 		switch r.gitlabToken.RunnerEnvironment {
 		case "gitlab-hosted":
 			return Managed
-		case "self-hosted":
+		case oidc.SelfHostedRunner:
 			return SelfHosted
 		default:
 			return Unknown
--- a/pkg/attestation/crafter/runners/gitlabpipeline.go
+++ b/pkg/attestation/crafter/runners/gitlabpipeline.go
@@ -16,15 +16,29 @@
 package runners
 
 import (
+	"context"
 	"os"
 
 	schemaapi "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
+	"github.com/chainloop-dev/chainloop/pkg/attestation/crafter/runners/oidc"
 )
 
-type GitlabPipeline struct{}
+type GitlabPipeline struct {
+	gitlabToken *oidc.GitlabToken
+}
 
-func NewGitlabPipeline() *GitlabPipeline {
-	return &GitlabPipeline{}
+func NewGitlabPipeline(ctx context.Context) *GitlabPipeline {
+	client, err := oidc.NewGitlabClient(ctx)
+	if err != nil {
+		// TODO: add logging
+		return &GitlabPipeline{
+			gitlabToken: nil,
+		}
+	}
+
+	return &GitlabPipeline{
+		gitlabToken: client.Token,
+	}
 }
 
 func (r *GitlabPipeline) ID() schemaapi.CraftingSchema_Runner_RunnerType {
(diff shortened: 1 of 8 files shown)


pkg/attestation/crafter/runners/oidc/gitlab.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"os"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/rs/zerolog"
+)
+
+// GitlabTokenEnv is the environment variable name for Gitlab OIDC token.
+// #nosec G101 - This is just the name of an environment variable, not a credential
+const GitlabTokenEnv = "GITLAB_OIDC"
```
jiparis, line 29: Not sure if there is a "common" name for this env var. For instance, we use just ID_TOKEN in some of our workflows. We can have a set of most used env vars. Not important at all.

pkg/attestation/crafter/runners/gitlabpipeline.go
```diff
@@ -16,15 +16,30 @@
...
 
-type GitlabPipeline struct{}
+type GitlabPipeline struct {
+	gitlabToken *oidc.GitlabToken
+}
+
+func NewGitlabPipeline(ctx context.Context, logger *zerolog.Logger) *GitlabPipeline {
+	client, err := oidc.NewGitlabClient(ctx, logger)
+	if err != nil {
+		logger.Debug().Err(err).Msg("failed to create Gitlab OIDC client")
```
jiparis, line 34: Please wrap the former error `Msgf("failed to create Gitlab OIDC client: %w", err)`

pkg/attestation/crafter/runners/oidc/gitlab.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// retrieve the Gitlab server on which the pipeline is running, which is the provider URL
+	providerURL := os.Getenv(CIServerURLEnv)
+	logger.Debug().Str("providerURL", providerURL).Msg("retrieved provider URL")
+	if providerURL == "" {
+		return nil, fmt.Errorf("%s environment variable not set", CIServerURLEnv)
+	}
+
+	tokenContent := os.Getenv(GitlabTokenEnv)
+	logger.Debug().Str("tokenContent", tokenContent).Msg("retrieved token content")
```
jiparis, line 59: Printing the token can lead to security issues, please remove "tokenContent" from logs.

pkg/attestation/crafter/runners/oidc/gitlab.go
```diff
@@ -0,0 +1,108 @@
...
+
+func parseToken(ctx context.Context, providerURL string, tokenString string) (*GitlabToken, error) {
+	provider, err := oidc.NewProvider(ctx, providerURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to OIDC provider: %w", err)
+	}
+
+	verifier := provider.Verifier(&oidc.Config{
+		SkipClientIDCheck: true, // Skip client ID check since we're just parsing
+		SkipExpiryCheck:   true, // Skip expiry check to allow viewing expired tokens
```
jiparis, line 81: Any reason for these skips? Expired tokens are invalid and should not be accepted.

jiparis (approved):
looking good, please see comments

pkg/attestation/crafter/runners/gitlabpipeline.go
gr0: Done.

pkg/attestation/crafter/runners/oidc/gitlab.go
gr0: Mhm, very good catch - removed.

pkg/attestation/crafter/runners/oidc/gitlab.go
gr0: Good point - this is a left over from local tests with local token. Removed.

pkg/attestation/crafter/runners/oidc/gitlab.go
gr0: I'll leave it for now, because we have a different cache in GitHub. With the next integrations I will keep that in mind, maybe we can find a common place for that.

jiparis:
Thanks, ok to merge.

migmartri (commented):
Very Cool!!

PR #2016, opened 2025-05-02
Verify Gitlab OIDC token audience

gr0:
This PR adds verification of the Gitlab OIDC token.

### DIFF
--- a/pkg/attestation/crafter/runners/oidc/gitlab.go
+++ b/pkg/attestation/crafter/runners/oidc/gitlab.go
@@ -17,12 +17,11 @@ package oidc
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
+	"slices"
 
 	"github.com/coreos/go-oidc/v3/oidc"
-	"github.com/golang-jwt/jwt/v4"
 	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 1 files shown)


migmartri (approved):

pkg/attestation/crafter/runners/oidc/gitlab.go
```diff
@@ -85,6 +89,10 @@ func parseToken(ctx context.Context, providerURL string, tokenString string) (*G
 		return nil, fmt.Errorf("token verification failed: %w", err)
 	}
 
+	if slices.Compare(idToken.Audience, []string{ExpectedAudience}) != 0 {
```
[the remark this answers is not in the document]
gr0: Nice one, thank you - adjusted.

pkg/attestation/crafter/runners/oidc/gitlab.go
```diff
@@ -95,6 +100,24 @@ func parseToken(ctx context.Context, providerURL string, tokenString string) (*G
 		return nil, fmt.Errorf("failed to extract claims: %w", err)
 	}
 
+	// Convert raw token to JWT token to use VerifyAudience
+	// Parse the JWT token without validation (we already validated it with the OIDC provider)
+	parsedToken, _, err := new(jwt.Parser).ParseUnverified(tokenString, jwt.MapClaims{})
```
migmartri, line 105: do you need to parse it? At this point I am not sure `VerifyAudience` makes sense if it requires all these changes. Did you look at what the method does inside to see if it's worth it?

migmartri: can't you use `idtoken` to get the claims? You seem to be doing smth already?

```
var claims map[string]interface{}
	if err := idToken.Claims(&claims); err != nil {
		return nil, fmt.Errorf("failed to extract claims: %w", err)
	}
```

jiparis (approved):

pkg/attestation/crafter/runners/oidc/gitlab.go
jiparis: In theory, just `idToken.Audience` would return the list of audience values. So something like this would also work:
```
strings.Contains(idToken.Audience, ExpectedAudience)
```

gr0: Yes, but @migmartri wanted to move away from the simple checking ;)

migmartri: and note that `strings.Contains(idToken.Audience, ExpectedAudience)` might not be valid since it's checking a substring, so for example 

`hacking-chainloop` will patch the `chainloop` audience?

jiparis: Ok, but not that `verifier.Verify` is already doing the claim extraction for you into `idToken.Audience` field. https://github.com/coreos/go-oidc/blob/v3/oidc/verify.go#L250

migmartri: ok, up to you guys, this is the method I was talking about which we use in all our middlewares today https://pkg.go.dev/github.com/golang-jwt/jwt/v4@v4.5.2#RegisteredClaims.VerifyAudience

gr0: Let's do it like this - I'll have a look tomorrow if I can simplify the code and adjust and merge after.

gr0: I've looked through the code in the `verify.go` in `go-oidc` - I think we can compare slices, the additional verification is already done. I'll modify the code and merge.

migmartri (approved):

PR #2026, opened 2025-05-06
Remove GitHub attestation to use Chainloop native capabilities

gr0:
This PR removes GitHub native SLSA attestation and adds the `id_token: write` permission.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,7 @@ on:
 # https://github.com/ossf/scorecard/blob/7ed886f1bd917d19cb9d6ce6c10e80e81fa31c39/docs/checks.md#token-permissions
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -8,6 +8,7 @@ on:
 # https://github.com/ossf/scorecard/blob/7ed886f1bd917d19cb9d6ce6c10e80e81fa31c39/docs/checks.md#token-permissions
 permissions:
   contents: read
+  id-token: write
```
migmartri, line 11: add a comment here saying that this is for slsa attestations, and maybe a link to the docs?

migmartri (approved):
Thanks

.github/workflows/release.yaml
gr0: Done.

PR #2038, opened 2025-05-12
Configure referrer REST endpoint for private workflows

### LINKED ISSUE #2036: Make private workflows artifacts available through the /discover REST endpoint
Currently, that endpoint is only available to public workflows without authentication. Ideally, it should also allow querying for private artifacts given the proper Chainloop token.

We could choose to merge results for public and private workflows. In this case, make sure that private workflows have precedence, not to incur into vulnerabilities.

jiparis:
This PR reconfigures the /discover/{digest} endpoint to search in private workflows. Public workflows can still be queried using /discover/shared/{digest}
Pending: generate OpenAPI documentation (in next PR)

Closes #2036


migmartri (approved):
re: open-api docs, check with @javirln

migmartri:
it seems that probably some test-data has expired?

```
            	            	couldn't match JWT provider: error: code = 401 reason = UNAUTHORIZED message = JWT token has expired metadata = map[] cause = <nil>

```

javirln (approved):
Regarding OpenAPI spec. We are not generating anything regarding REST API documentation. We could potentially do so and include it also in our documentation page. See this example:
Screenshot 2025-05-13 at 09 43 15

In that example the documentation is embedding two different source of information and it's organized in two different sections.

migmartri:
please rebase.

and @javirln could you create an issue in the repo with information on how to generate swagger docs the right way similarly we've done it somewhere else?

PR #2043, opened 2025-05-13
chore: upgrade setup-go

jiparis:
Tests are taking longer because cache action has been deprecated. This upgrade fixes the issue


migmartri (approved):
❤️

PR #2050, opened 2025-05-16
chore: reset patch when bumping minor version

jiparis:
When bumping the minor version, we should reset the patch version to `0`.

### DIFF
--- a/.github/workflows/utils/bump-project-version.sh
+++ b/.github/workflows/utils/bump-project-version.sh
@@ -35,7 +35,7 @@ version=$(cat "${project_yaml}" | awk '/^projectVersion:/ {
         # Handle minor version bumps (e.g., v1.0.0 -> v1.1.0)
         split(version, ver_parts, /\./);
         ver_parts[2] = ver_parts[2] + 1;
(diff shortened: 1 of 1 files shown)


.github/workflows/utils/bump-project-version.sh
```diff
@@ -35,7 +35,7 @@ version=$(cat "${project_yaml}" | awk '/^projectVersion:/ {
         # Handle minor version bumps (e.g., v1.0.0 -> v1.1.0)
         split(version, ver_parts, /\./);
         ver_parts[2] = ver_parts[2] + 1;
-        print ver_parts[1] "." ver_parts[2] "." ver_parts[3];
+        print ver_parts[1] "." ver_parts[2] ".0";
```
migmartri, line 38: why did you decide to make this change? Isn't this happening already?

migmartri (approved):

.github/workflows/utils/bump-project-version.sh
migmartri: Thanks

PR #2055, opened 2025-05-19
feat(gitlab): try to parse token from provided creds

migmartri:
Extends the current authenticated runner feature for gitLab to also use the authentication token provided during attestation process.

This means that you'll not need to craft another OIDC token but just use whatever you are using for the attestation process, which might already be a GitLab OIDC token


gr0 (approved):
Thank you!

PR #2077, opened 2025-05-26
fix: add contract without policies for getting started

### LINKED ISSUE #2076: Getting started contract doesn't work
Running the Getting Started with the provided contract fails:
```bash
> chainloop att init --workflow build --project test-app --contract https://raw.githubusercontent.com/chainloop-dev/chainloop/refs/heads/main/docs/examples/quickstart/quickstart-contract.yaml

ERR validation error: invalid attachment: error executing policy request: Post "http://localhost:8002/v1/policies/validate": dial tcp [::1]:8002: connect: connection refused
```

The contract contains policies that are trying to get retrieved and validated with a default policy provider configured in `config.devel.yaml` (which is also inexistent)
```yaml
policy_providers:
  - name: chainloop
    default: true
    url: http://localhost:8002/v1
```

jiparis:
The contract shouldn't have any non-local policy, since policy provider is not configured by default. 
Closes #2076

### DIFF
--- a/app/controlplane/configs/config.devel.yaml
+++ b/app/controlplane/configs/config.devel.yaml
@@ -94,10 +94,11 @@ auth:
 prometheus_integration:
   - org_name: "development"
 
-policy_providers:
-  - name: chainloop
-    default: true
-    url: http://localhost:8002/v1
+# Policy providers configuration
+# policy_providers:
+#   - name: chainloop
+#     default: true
+#     url: http://localhost:8002/v1
 
 enable_profiler: true
 
--- a/docs/examples/quickstart/quickstart-contract.yaml
+++ b/docs/examples/quickstart/quickstart-contract.yaml
@@ -11,22 +11,3 @@ materials:
   - name: vulnerabilities-report
     type: SARIF
     optional: true
-policies:
-  attestation:
-    # Container with sbom checks that all containers added to the attestation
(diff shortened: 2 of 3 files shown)


app/controlplane/configs/config.devel.yaml
```diff
@@ -94,10 +94,11 @@ auth:
 prometheus_integration:
   - org_name: "development"
 
-policy_providers:
```
migmartri, line 97: Agreed with this change

docs/examples/quickstart/quickstart-contract.yaml
```diff
@@ -11,22 +11,3 @@ materials:
   - name: vulnerabilities-report
     type: SARIF
     optional: true
-policies:
```
migmartri, line 14: but not this one, note that it's being used from the quickstart in the platform docs.

migmartri (changes requested):

docs/examples/quickstart/quickstart-contract.yaml
jiparis: Then it won't work for regular OSS users. Should we then provide two different contracts and distinguish them in the docs?

jiparis: I've created another contract without policies for opensource users.

javirln (approved):

PR #2092, opened 2025-06-04
chore(api tokens): refactor to make code independent from internal dependencies

jiparis:
This PR makes the APIToken use case usable from third party code.
Additionally, it adds a new use case method to re-generate the JWT for existing API tokens.


app/controlplane/cmd/wire.go
```diff
@@ -63,14 +63,21 @@ func wireApp(*conf.Bootstrap, credentials.ReaderWriter, log.Logger, sdk.Availabl
...
+func newJWTConfig(conf *conf.Auth) *biz.APITokenJWTConfig {
+	return &biz.APITokenJWTConfig{
+		SymmetricHmacKey: conf.GeneratedJwsHmacSecret,
+	}
+}
+
+func newDataConf(in *conf.Data_Database) *pkgConf.DatabaseConfig {
+	c := &pkgConf.DatabaseConfig{Driver: in.Driver, Source: in.Source, MinOpenConns: in.MinOpenConns, MaxOpenConns: in.MaxOpenConns}
 	if in.MaxConnIdleTime != nil {
-		c.MaxConnIdleTime = in.MaxConnIdleTime.AsDuration()
```
migmartri, line 73: why of this change?

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -181,6 +187,27 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description
 	return token, nil
 }
 
+// RegenerateJWT will regenerate a new JWT for the given token. Use with caution, since old JWTs are not invalidated.
+func (uc *APITokenUseCase) RegenerateJWT(ctx context.Context, tokenID uuid.UUID, expiresAt *time.Time) (*APIToken, error) {
```
migmartri: what's the use-case for this?

migmartri (commented):
Please fix the linter errors but other than that lgtm

app/controlplane/cmd/wire.go
jiparis: Now `c.MaxConnIdleTime` (from `pkgConf.DatabaseConfig`) is a `google.protobuf.Duration` too.

app/controlplane/pkg/biz/apitoken.go
jiparis: Since JWTs are not stored anywhere, it's needed to generate a new one for each integration call, and then it's discarded. Ideally, `expiresAt` should be short to make it ephemeral.
Another option would be to revoke the existing API Token, and create a new one, but that would end up with a lot of noise in the database.

jiparis: Update: for consistency, it's now also updating `expiresAt` in database record

migmartri (approved):

Issue #2095, opened 2025-06-05
add default proxy-body-size in controlplane ingress


migmartri:
This for example is added to the CAS 

```
 nginx.ingress.kubernetes.io/proxy-body-size: "100m"
```

in the controlplane you might get 

- [ ] ```

chainloop att push i got
ERR pushing to control plane: storing attestation: rpc error: code = Unknown desc = unexpected HTTP status code received from server: 413 (Request Entity Too Large); transport: received unexpected content-type "text
```

PR #2096, opened 2025-06-05
chore(deployment): add default proxy-body-size in controlplane ingress #2095

snieguu:
Fixes https://github.com/chainloop-dev/chainloop/issues/2095


jiparis (approved):
Thanks!

migmartri:
Thanks for the contribution @snieguu! 

It seems that the commit wasn't signed, mind signing it. Note that that process is different than Signoff. Thanks!

image

migmartri:
Also, to release the chart, please bump the patch version [here](https://github.com/chainloop-dev/chainloop/blob/48bffa800fd0d42737da82580e9dbcdff3f337b7/deployment/chainloop/Chart.yaml#L10) as part of the same PR

This will trigger the CI and package a new chart :)

PR #2097, opened 2025-06-05
feat(auth): Extract more information from OIDC token

javirln:
This patch extracts two additional properties from the OIDC token: `given_name` and `family_name`. This enhancement aims to improve the user personalization experience.

To support this, two new optional columns—first_name and last_name—have been added to the users table.

The authentication service has been updated to decode these claims and use them to populate a user’s first and last name during registration. For existing users, these fields will be updated if the claims are present.

On the CLI side, the commands organization current-context and organization describe have been updated to display user information when available. The Logged in as section will now show one of the following formats:
- FirstName LastName email
- email (if name information is not available)


app/controlplane/internal/service/auth.go
```diff
@@ -260,6 +260,8 @@ type upstreamOIDCclaims struct {
 	// https://learn.microsoft.com/en-us/entra/identity/authentication/howto-authentication-use-email-signin
 	// https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference
 	PreferredUsername string `json:"preferred_username"`
+	FamilyName        string `json:"family_name"`
```
jiparis, line 263: these claims probably require the `profile` scope. Please check that it's being requested. https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference, https://developers.google.com/identity/openid-connect/openid-connect

Also, check that the `name` claim could be just enough for this purpose (as it contains the user name in human readable format)

app/controlplane/internal/service/auth.go
```diff
@@ -304,7 +306,10 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) *
 	}
 
 	// Create user if needed
-	u, err := svc.userUseCase.FindOrCreateByEmail(ctx, claims.preferredEmail())
+	u, err := svc.userUseCase.FindOrCreateByEmail(ctx, claims.preferredEmail(), &biz.FindOrCreateByEmailOpts{
+		FirstName: &claims.GivenName,
```
jiparis, line 310: Don't assume these claims will have values always. It's up to the user to give consent.

app/controlplane/internal/service/auth.go
javirln: The system is requesting already the `profile` claim [here](https://github.com/chainloop-dev/chainloop/blob/ca92064bd8ce7089a0fb07f1149f585443738300/app/controlplane/internal/oidcauthenticator/auth.go#L45).

The issue with using name is that we’d need to parse it to separate the first name and last name. That’s why I only included the individual fields instead of the combined one. I’m not sure if we want to handle the parsing, but using the separate fields seems simpler.

app/controlplane/internal/service/auth.go
javirln: Yes, that's why those fields are pointers and in the case they are set empty, there is a check in the biz layer that prevents the update of those fields. I'll ping you there.

app/controlplane/pkg/biz/user.go
```diff
@@ -134,17 +151,30 @@ func (uc *UserUseCase) FindOrCreateByEmail(ctx context.Context, email string, di
 			LoggedIn: time.Now(),
 		}, nil)
 
+		if (opts.FirstName != nil && u.FirstName != *opts.FirstName) ||
+			(opts.LastName != nil && u.LastName != *opts.LastName) {
```
javirln, line 155: Here @jiparis. This line ensures that if new first or last name information is provided and it differs from what is currently stored for the user, the user's name in the database will be updated accordingly.

By default the `FirstName` and `LastName` of all users after applying the migration will be NULL which on the struct is "", so yes, they won't get updated if they are the same as the incoming ones, only different.

app/cli/cmd/organization_describe.go
```diff
@@ -44,7 +44,11 @@ func newOrganizationDescribeCmd() *cobra.Command {
 func contextTableOutput(config *action.ConfigContextItem) error {
 	gt := newTableWriter()
 	gt.SetTitle("Current Context")
-	gt.AppendRow(table.Row{"Logged in as", config.CurrentUser.Email})
+	loggedInAs := config.CurrentUser.Email
+	if config.CurrentUser.FirstName != "" && config.CurrentUser.LastName != "" {
+		loggedInAs = fmt.Sprintf("%s %s <%s>", config.CurrentUser.FirstName, config.CurrentUser.LastName, config.CurrentUser.Email)
```
migmartri: is there any change that these values are empty?

app/controlplane/internal/service/auth.go
```diff
@@ -304,7 +306,10 @@ func callbackHandler(svc *AuthService, w http.ResponseWriter, r *http.Request) *
 	}
 
 	// Create user if needed
-	u, err := svc.userUseCase.FindOrCreateByEmail(ctx, claims.preferredEmail())
+	u, err := svc.userUseCase.FindOrCreateByEmail(ctx, claims.preferredEmail(), &biz.FindOrCreateByEmailOpts{
```
migmartri: rename to Upsert

migmartri (commented):
Also expose it and add it to 

```
chainloop org member ls
```

Issue #2088, opened 2025-06-03
chainloop attestation init returns misleading message if no org exists


migmartri:
If no org exists, we should return a different message than

```
ERR the organization you are trying to access does not exist, please run "chainloop auth login"
```

for example, what we return in other commands

```
you are not part of any organization, please run "chainloop organization create --name ORG_NAME" to create one
``` 

cc/ @jiparis

PR #2105, opened 2025-06-12
fix: change error message if no organization exists

Piskoo:
Fix for #2088

### DIFF
--- a/app/cli/main.go
+++ b/app/cli/main.go
@@ -89,7 +89,9 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 		msg = "your authentication token has expired, please run chainloop auth login again"
 	case isWrappedErr(st, jwtMiddleware.ErrMissingJwtToken):
 		msg = "authentication required, please run \"chainloop auth login\""
-	case v1.IsUserNotMemberOfOrgErrorNotInOrg(err), v1.IsUserWithNoMembershipErrorNotInOrg(err):
+	case v1.IsUserNotMemberOfOrgErrorNotInOrg(err):
+		msg = "the organization you are trying to access does not exist or you are not part of it, please run \"chainloop auth login\""
+	case v1.IsUserWithNoMembershipErrorNotInOrg(err):
 		msg = "you are not part of any organization, please run \"chainloop organization create --name ORG_NAME\" to create one"
 	case errors.As(err, &cmd.GracefulError{}):
 		// Graceful recovery if the flag is set and the received error is marked as recoverable


app/cli/main.go
```diff
@@ -90,7 +90,7 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 	case isWrappedErr(st, jwtMiddleware.ErrMissingJwtToken):
 		msg = "authentication required, please run \"chainloop auth login\""
 	case v1.IsUserNotMemberOfOrgErrorNotInOrg(err), v1.IsUserWithNoMembershipErrorNotInOrg(err):
-		msg = "the organization you are trying to access does not exist, please run \"chainloop auth login\""
+		msg = "you are not part of any organization, please run \"chainloop organization create --name ORG_NAME\" to create one"
```
migmartri, line 95: thanks for the PR! 

I think we might need to split the switch into the two cases, `v1.IsUserNotMemberOfOrgErrorNotInOrg(err),` might not mean that tit doesn't have any org but instead that it's not part of this specific org. 

Let me review the current custom error throwing, thanks!

migmartri: Hi, thanks again for providing a PR but I although I think it fixes the issue, it adds a regression where we will tell users that they do not have an organization when they just don't have access to it.

The following change, indeed fixes the issue

```diff
diff --git a/app/cli/main.go b/app/cli/main.go
index ff807fce..ca1b9540 100644
--- a/app/cli/main.go
+++ b/app/cli/main.go
@@ -89,8 +89,10 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
                msg = "your authentication token has expired, please run chainloop auth login again"
        case isWrappedErr(st, jwtMiddleware.ErrMissingJwtToken):
                msg = "authentication required, please run \"chainloop auth login\""
-       case v1.IsUserNotMemberOfOrgErrorNotInOrg(err), v1.IsUserWithNoMembershipErrorNotInOrg(err):
-               msg = "the organization you are trying to access does not exist, please run \"chainloop auth login\""
+       case v1.IsUserNotMemberOfOrgErrorNotInOrg(err):
+               msg = "the organization you are trying to access does not exist or you are not part of it, please run \"chainloop auth login\""
+       case v1.IsUserWithNoMembershipErrorNotInOrg(err):
+               msg = "you are not part of any organization, please run \"chainloop organization create --name ORG_NAME\" to create one"
        case errors.As(err, &cmd.GracefulError{}):
                // Graceful recovery if the flag is set and the received error is marked as recoverable
                if cmd.GracefulExit {
```

what I do is to make sure the error handling is split between the two cases

a) when the user doesn't have any organization
b) when the user doesn't have access to the provided organization

The original issue https://github.com/chainloop-dev/chainloop/issues/2088 was about making sure that during the attestation process, we show a) not b), and this change fixes it.

Feel free to apply those changes in your branch and let me know.

Thanks!

Piskoo: Thanks for the feedback! You're right, that was causing a misleading error. I've applied your suggested changes to handle both cases separately.

migmartri:
Congratulations for your first contribution!

PR #2111, opened 2025-06-16
feat(rbac): refactor membership data to support polymorphic memberships

jiparis:
This PR adds new database fields (MembershipType, MemberID, ResourceType and ResourceID) to support flexible RBAC schemas at different levels (organizations, projects) using different role sources (users, groups, teams...).

The data layer logic has been updated to maintain the current behaviour. A migration is also included.


app/controlplane/pkg/data/ent/schema/membership.go
```diff
@@ -49,18 +50,28 @@ func (Membership) Fields() []ent.Field {
...
+		field.UUID("member_id", uuid.UUID{}).Optional(),
+
+		field.Enum("resource_type").GoType(biz.ResourceType("")).Optional(),
+		field.UUID("resource_id", uuid.UUID{}).Optional(),
 	}
 }
 
 func (Membership) Edges() []ent.Edge {
 	return []ent.Edge{
+		// Deprecated: use polymorphic membership instead
```
migmartri, line 65: how difficult do you think it would be to kill those columns? 

I am ok if we keep it in multiple steps though

migmartri (approved):

app/controlplane/pkg/data/ent/schema/membership.go
jiparis: I think I can create an additional migration for it, but in a different version, not to break compatibility.

PR #2119, opened 2025-06-18
add attestation annotation based filter for sbom forwarding

### LINKED ISSUE #1982: allow to filter entries in dependency track fanout
Have a way to define selectors so the plugin can choose wether to send the data or not.

Piskoo:
Closes #1982  

Adds filtering of SBOM forwarding based on attestation annotations to allow selective forwarding.

Highlights
- Filter controls SBOM forwarding by matching attestation annotations.
- Added filter flag validation tests.

Testing
1. Registered Dependency-Track instance is required.
2. Attach the instance .
```
 $ cl integration attached add \
  --integration dependency-track \
  --project myproject \
  --workflow myworkflow \
  --opt projectName=projectName \
  --opt filter=forward=true
```
`filter` is optional.
- If not set, all SBOMs added to the attestation will be forwarded.
- If set, SBOMs will only be forwarded if the attestation includes the correct annotation key/value.

3. Initialize attestation and add materials.
4. Push attestation.
```
$ cl att push 
  --annotation forward=true
``` 
Include the correct --annotation to forward materials to Dependency-Track.


migmartri:
You are on fire! 🔥

migmartri:
Hi @Piskoo, it would be great if you could add some information on the description of the PR. Such as highlights of what it contains, questions, examples of reproducing the issue, etc

Thanks!

Piskoo:
> Hi @Piskoo, it would be great if you could add some information on the description of the PR. Such as highlights of what it contains, questions, examples of reproducing the issue, etc
> 
> Thanks!

Hi,  I’ve updated the PR description, let me know if something is missing.

app/controlplane/plugins/core/dependency-track/v1/README.md
```diff
@@ -82,6 +83,11 @@ See https://docs.chainloop.dev/guides/dependency-track/
       "type": "string",
       "minLength": 1,
       "description": "ID of parent project to create a new project under"
+    },
+    "filter": {
+      "type": "string",
+      "minLength": 1,
```
migmartri, line 89: Do we support multiple filters?

migmartri (commented):
Awesome, there are two things that I'd like to think about

1 - Can we support multiple filters, so they become an AND? i.e filter='role=backend,env=production' 
2 - Have we thought if we could also support material annotations? 

Annotations can be added at the attestation level but also at the material level, I am wondering if we could/should support both. 

For example, we currently support [interpolation](https://docs.chainloop.dev/guides/dependency-track#dynamically-set-the-project-name)

for example

```
    --opt projectName="project-{{ .Material.Annotations.component }}-{{ .Attestation.Annotations.version }}"

``` 

so we could extend this feature so the filter could receive `.Material.Annotations.[annotationName]` and `.Attestation.Annotations.[annotationName]`

So I could do something like

--opt 'filter=.Attestation.Annotations.environment=production` 

--opt 'filter=.Material.Annotations.latest=true` 

or a combination of both

--opt 'filter=.Attestation.Annotations.environment=production,.Material.Annotations.latest=true` 

what do you think?

cc/ @jiparis

app/controlplane/plugins/core/dependency-track/v1/README.md
Piskoo: Only one in the current version, I'll add support for multiple filters

Piskoo:
> 1 - Can we support multiple filters, so they become an AND? i.e filter='role=backend,env=production' 
> 2 - Have we thought if we could also support material annotations?

I've added support for multiple comma-separated annotations.  

For material annotations, I think it makes sense to split them into two separate options: `filter-material` and `filter-attestation`  As for how they should work together, I'm not sure if doing an AND between the two is the right approach.

app/controlplane/plugins/core/dependency-track/v1/README.md
```diff
@@ -82,6 +83,11 @@ See https://docs.chainloop.dev/guides/dependency-track/
...
+      "description": "Annotation and its value required for SBOM forwarding"
```
migmartri: should we add some examples (single and multiple filters)? maybe as part of the description? Or I do not remember if there was another way with our openapi spec generator. I'd look into other plugin examples we have to see how we provide examples.

migmartri: the message is also incorrect, since now we support multiple ones.

migmartri:
> I've added support for multiple comma-separated annotations.

Nice, just added a comment re: documentation

> For material annotations, I think it makes sense to split them into two separate options: filter-material and filter-attestation As for how they should work together, I'm not sure if doing an AND between the two is the right approach.

why would you split it in two? 

>  I'm not sure if doing an AND between the two is the right approach.

What would you suggest?

Piskoo:
> > For material annotations, I think it makes sense to split them into two separate options: filter-material and filter-attestation As for how they should work together, I'm not sure if doing an AND between the two is the right approach.
> 
> why would you split it in two?

 It feels simpler and clearer, but then sticking to the previous, more explicit convention could be the better approach for consistency.

> > I'm not sure if doing an AND between the two is the right approach.
> 
> What would you suggest?

After rethinking it, I take that back. implementing some kind of complex connection logic between them would only make things unnecessarily complicated.

migmartri:
Okay, I am not convinced about only supporting attestation level annotations during filtering because specifically this integration is about SBOM materials, not attestations, but I am not 100% sure about my approach either of adding the whole `attestation.Annotations.environment` (too noisy)

So my proposal would be to also take into account material level annotations during filtering, so for example

```
filter='env=prod' 
``` 

could try to find a match between the union of `attestation.annotations` and `material[].annotations`. In the case of both of them having the same annotation, we can just make a decision and document that for example it will take the material one.

What do you think? 

Also interested in your thoughts @jiparis @javirln

Piskoo:
> So my proposal would be to also take into account material level annotations during filtering, so for example
> 
> ```
> filter='env=prod' 
> ```
> 
> could try to find a match between the union of `attestation.annotations` and `material[].annotations`. In the case of both of them having the same annotation, we can just make a decision and document that for example it will take the material one.

 Sounds great to me!

EDIT: Material annotations can now be used to filter sboms

app/controlplane/plugins/core/dependency-track/v1/README.md
Piskoo: Added more fitting description for schema, unfortunately you cannot use `=` sign in the description(it is being used by a parser) so I could not add any examples here, but I've updated example available in cli.

migmartri:
One last thing, It would be good if you could bump the version of the plugin too

app/cli/cmd/attached_integration_add.go
```diff
@@ -28,7 +28,7 @@ func newAttachedIntegrationAttachCmd() *cobra.Command {
 		Use:     "add",
 		Aliases: []string{"attach"},
 		Short:   "Attach an existing registered integration to a workflow",
-		Example: `  chainloop integration attached add --workflow deadbeef --project my-project --integration beefdoingwell --opt projectName=MyProject --opt projectVersion=1.0.0`,
+		Example: `  chainloop integration attached add --workflow deadbeef --project my-project --integration beefdoingwell --opt projectName=MyProject --opt projectVersion=1.0.0 --opt filter=environment=prod,team=security`,
```
migmartri: nitpick. I'd add this as another example, so people don't think that filter is required.

Piskoo: Updated to 
```
  chainloop integration attached add --workflow deadbeef --project my-project --integration beefdoingwell --opt projectName=MyProject --opt projectVersion=1.0.0

  # Attach dependency track with filters: AND logic, material annotations take precedence over attestation
  chainloop integration attached add --workflow deadbeef --project my-project --integration dependency-track --opt projectName=MyProject --opt filter="environment=prod,team=security"
```

app/cli/cmd/attached_integration_add.go
```diff
@@ -28,7 +28,10 @@ func newAttachedIntegrationAttachCmd() *cobra.Command {
 		Use:     "add",
 		Aliases: []string{"attach"},
 		Short:   "Attach an existing registered integration to a workflow",
-		Example: `  chainloop integration attached add --workflow deadbeef --project my-project --integration beefdoingwell --opt projectName=MyProject --opt projectVersion=1.0.0`,
+		Example: `  chainloop integration attached add --workflow deadbeef --project my-project --integration beefdoingwell --opt projectName=MyProject --opt projectVersion=1.0.0
+
+  # Attach dependency track with filters: AND logic, material annotations take precedence over attestation
```
migmartri:
```suggestion
  # Only send SBOMs to dependency track when the annotations in the attestation or material match the filter in an AND operation. Note: material annotations take precedence over attestation ones.
```

app/controlplane/plugins/core/dependency-track/v1/extension.go
```diff
@@ -244,6 +259,36 @@ func doExecute(ctx context.Context, req *sdk.ExecutionRequest, sbom *sdk.Execute
 	return nil
 }
 
+// checks if all filter conditions are met by either:
+// 1) Material annotations (higher precedence if same key exists in both)
+// 2) Attestation annotations (fallback if key not in material)
+func matchesAllFilters(attestationAnnotations, materialAnnotations map[string]string, filter string) bool {
```
migmartri: could you add some unit tests for this method? It's not clear to me how the precedence works there :)

migmartri (commented):
Thanks

app/controlplane/plugins/core/dependency-track/v1/extension_test.go
```diff
@@ -335,3 +338,88 @@ func TestValidateAttachmentConfiguration(t *testing.T) {
...
+
+	materialAnnotations := map[string]string{
+		"environment": "staging",
+		"critical":    "true",
+	}
+
+	testCases := []struct {
+		name    string
+		filter  string
+		wantErr bool
```
migmartri: nitpick. I'd not call it wantErr, just `want`, otherwise it can be interpreted as you want to return an error in golang :)

Nice tests btw :)

migmartri:
Feel free to merge it! 🚀

PR #2122, opened 2025-06-23
chore(golang): Bump version of Golang to 1.24.4 and fix CVEs

javirln:
This patch bumps the Golang version to 1.24.4 fixing several CVEs as well as the Atlas version to v0.35.0 which also fixes a few CVEs.


migmartri (approved):
Thanks!

PR #2124, opened 2025-06-23
feat(rbac): RBAC foundation

jiparis:
This PR includes:
* RBAC logic to apply fine grained permissions when user has `role:org:member` role on the organization
* Regular roles don't get this logic applied
* Permission checks for `att init` and `att push` use cases (and all related calls for workflow and contract)
* Membership entries updated when a project is created (making the calling user owner of the project)

Note that this PR is safe to merge since no user will get that role unless the DB manually modified.


app/controlplane/pkg/authz/authz.go
```diff
@@ -153,6 +168,18 @@ var rolesMap = map[Role][]*Policy{
 		PolicyArtifactUpload,
 		// + all the policies from the viewer role inherited automatically
 	},
+	RoleOrgMember: {
```
jiparis, line 179: These are used for endpoint call validation in the Authz middleware. Fine-grained RBAC is not applied yet at this level

app/controlplane/pkg/authz/authz.go
```diff
@@ -153,6 +168,18 @@ var rolesMap = map[Role][]*Policy{
...
+		// Allowed endpoints. RBAC will be applied where needed
+		PolicyWorkflowRead,
+		PolicyWorkflowContractRead,
+		PolicyWorkflowContractCreate,
+	},
+	RoleProjectAdmin: {
```
jiparis, line 217: All these checks are applied in the `authorizeResource` service method, and are context-aware (it expects explicit membership on the workflow's project).

app/cli/documentation/cli-reference.mdx
```diff
@@ -2440,7 +2440,7 @@ Options
 ```
 -h, --help          help for update
 --id string     Membership ID
---role string   Role of the user in the organization, available admin, owner, viewer (default "viewer")
+--role string   Role of the user in the organization, available admin, owner, viewer, member (default "viewer")
```
migmartri: maybe we should not document member until we finish the feature? Or document it somehow?

app/controlplane/pkg/authz/authz.go
```diff
@@ -153,6 +170,36 @@ var rolesMap = map[Role][]*Policy{
...
+		PolicyWorkflowUpdate,
+		PolicyWorkflowDelete,
+
+		PolicyWorkflowRunList,
+	},
+	// RoleProjectAdmin: RBAC will be applied in all these
+	RoleProjectAdmin: {
+		// attestations
+
+		PolicyWorkflowRead,
```
migmartri, line 220: I do not think you need to duplicate the policies but instead make admin a superset of member and in here just add the things that it can do that others can't

app/controlplane/pkg/biz/membership.go
```diff
@@ -51,6 +51,16 @@ type MembershipRepo interface {
 	SetRole(ctx context.Context, ID uuid.UUID, role authz.Role) (*Membership, error)
 	Create(ctx context.Context, orgID, userID uuid.UUID, current bool, role authz.Role) (*Membership, error)
 	Delete(ctx context.Context, ID uuid.UUID) error
+
+	// RBAC methods
+
```
migmartri, line 56: extra line

app/controlplane/pkg/authz/membership.go
```diff
@@ -0,0 +1,49 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authz
+
+// Polymorphic membership
+
```
migmartri: extra line

app/controlplane/pkg/biz/workflow.go
```diff
@@ -72,6 +72,9 @@ type WorkflowCreateOpts struct {
 	// Public means that the associated workflow runs, attestations and materials
 	// are reachable by other users, regardless of their organization
 	Public bool
+
+	// Owner identifies the user to be marked as owner of the project
+	Owner *uuid.UUID
```
migmartri, line 77: why is it a pointer to uuid?

app/controlplane/pkg/biz/workflow.go
```diff
@@ -99,18 +102,21 @@ type WorkflowListOpts struct {
 	WorkflowRunLastStatus WorkflowRunStatus
 	// JSONFilters is the filters to apply to the JSON fields
 	JSONFilters []*jsonfilter.JSONFilter
+	//
```
migmartri: spurious comment?

app/controlplane/pkg/biz/workflow.go
```diff
@@ -99,18 +102,21 @@ type WorkflowListOpts struct {
...
+	VisibleProjectsFromRBAC []uuid.UUID
```
migmartri: I'd personally just call it ProjectIDs, since at this point the layer should not care how this data was added to the filter no?

app/controlplane/pkg/biz/workflow.go
```diff
@@ -151,6 +157,13 @@ func (uc *WorkflowUseCase) Create(ctx context.Context, opts *WorkflowCreateOpts)
 		return nil, fmt.Errorf("failed to create workflow: %w", err)
 	}
 
+	// Set project admin if a new project has been created
+	if opts.Owner != nil {
```
migmartri, line 162: why would this ever be nil?

app/controlplane/internal/service/workflow.go
```diff
@@ -57,6 +59,11 @@ func (s *WorkflowService) Create(ctx context.Context, req *pb.WorkflowServiceCre
 		return nil, err
 	}
 
+	if err = s.userHasPermissionOnProject(ctx, s.projectsUseCase, currentOrg.ID, req.GetProjectName(), authz.PolicyWorkflowCreate); !errors.IsNotFound(err) {
```
migmartri: it feels like the second argument could be omited? `s.projectsUseCase`?

app/controlplane/internal/service/workflow.go
```diff
@@ -67,6 +74,16 @@ func (s *WorkflowService) Create(ctx context.Context, req *pb.WorkflowServiceCre
 		Public:       req.GetPublic(),
 	}
 
+	// add current user as the owner of the project in case it needs to be created
+	if rbacEnabled(ctx) {
```
migmartri: why don't we do it always so we remove the `nil` edge case?

app/controlplane/internal/service/workflow.go
migmartri: I do not understand this code. 

`s.userHasPermissionOnProject` will return `nil` and hence return the whole thing?

app/controlplane/internal/service/service.go
```diff
@@ -136,6 +139,75 @@ func WithLogger(logger log.Logger) NewOpt {
...
+
+	p, err := pUC.FindProjectByReference(ctx, orgID, &biz.EntityRef{Name: pName})
+	if err != nil {
+		return handleUseCaseErr(err, s.log)
+	}
+
+	return s.authorizeResource(ctx, policy, authz.ResourceTypeProject, p.ID)
+}
+
+func (s *service) visibleProjects(ctx context.Context, policy *authz.Policy) ([]uuid.UUID, error) {
```
migmartri: why to pass a policy here? It looks coupled to me

app/controlplane/internal/service/service.go
```diff
@@ -136,6 +139,75 @@ func WithLogger(logger log.Logger) NewOpt {
...
+	if !rbacEnabled(ctx) {
+		return nil, nil
```
migmartri: this logic is confusing to me too, so visibleProjects return no projects if I am an admin of all the projects?

I'd rather make this method error `ProjectRBACNotNeeded` or similar so it's clear that we just don't need this check?

app/cli/internal/action/attestation_init.go
```diff
@@ -104,20 +104,24 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
-		if contractRef != "" {
-			// Try to find it by name
-			_, err := NewWorkflowContractDescribe(action.ActionsOpts).Run(contractRef, 0)
-			// An invalid argument might be raised if we use a file or URL in the "name" field, which must be DNS-1123
-			// TODO: validate locally before doing the query
-			if err != nil && (status.Code(err) == codes.NotFound || status.Code(err) == codes.InvalidArgument) {
-				createResp, err := NewWorkflowContractCreate(action.ActionsOpts).Run(fmt.Sprintf("%s-%s", opts.ProjectName, opts.WorkflowName), nil, contractRef)
-				if err != nil {
-					return "", err
+	if err != nil {
```
migmartri: why of this change?

migmartri (commented):
I like the direction, a couple of things that I'd improve

- Make sure the project:admin role inherits from project:viewer instead of re-defining the same policies
- Simplify/decouple the project enforcer / listing used in the service, I find it a little bit confusing, maybe we can revisit them.

Thanks!

app/controlplane/pkg/authz/authz.go
migmartri: my bad, it's role:org:admin

app/controlplane/internal/service/workflow.go
migmartri: can we move the lookup of project outside?

app/controlplane/internal/service/service.go
migmartri: remove enforcement

app/controlplane/internal/service/service.go
migmartri: let's keep it but let's document the two types of nil/empty

app/cli/internal/action/attestation_init.go
jiparis: When there is an error, the code was ignoring it, leading to inconsistent status. I'll create a task to fix it instead

app/controlplane/internal/service/service.go
jiparis: Instead of enforcing, I'll just check that there is ANY role for the project.

app/controlplane/pkg/biz/workflow.go
jiparis: It's optional, since it might be created using an API token, so no owner in that case.

app/controlplane/pkg/biz/workflow.go
jiparis: For API tokens, yes.

app/controlplane/internal/service/service.go
jiparis: I have added a comment to make it clear that `nil` and `empty slices` are handled in a different way. `nil` means no RBAC, while `empty slice` means no access.
I'll add some tests to this method.

app/controlplane/pkg/authz/authz.go
jiparis: The one above is ORGMember, and those roles are just for the middleware to allow executing those endpoints

app/controlplane/internal/service/workflow.go
jiparis: This particular case is needed because the CreateWorkflow method is also responsible of creating projects. This way, if the project doesn't exist, we get a 404, but it's still valid in this scenario, so it must be ignored.
I'll make sure it's clear in the code.

app/cli/documentation/cli-reference.mdx
migmartri: inviting doesn't seem to work 

```
 go run main.go org member invitation create --receiver sarah@chainloop.local --role member
 
 ERR validation error: role is required
```

migmartri: nor membership update

```
go run main.go org member update --id d4e96bca-ba98-4934-81dd-6fef5af5e251 --role member
```

migmartri: @jiparis this

```diff
index a7bc0db8..a3eb5cc9 100644
--- a/app/controlplane/pkg/biz/user.go
+++ b/app/controlplane/pkg/biz/user.go
@@ -254,6 +254,8 @@ func PbRoleToBiz(r pb.MembershipRole) authz.Role {
                return authz.RoleAdmin
        case pb.MembershipRole_MEMBERSHIP_ROLE_ORG_VIEWER:
                return authz.RoleViewer
+       case pb.MembershipRole_MEMBERSHIP_ROLE_ORG_MEMBER:
+               return authz.RoleOrgMember
        default:
                return ""
        }
```

migmartri:
I made it work, looking good :)

Some minor issue is that I found some inconsistency between the message we get from the middleware and potentially from the one from the new helper, for example

```
go run main.go cas-backend ls
ERR operation not allowed
```

and

```
go run main.go att init --name test-annotations --project project-sarah  --replace
Flag --name has been deprecated, please use --workflow instead
This command is will run against the organization "foobar-222"
y
ERR user not authorized
```

migmartri (approved):

migmartri:
@jiparis one thing that you didn't end up adding I think is the cache we talked about when retrieving memberships, etc. Are you still going to do it?

jiparis:
> @jiparis one thing that you didn't end up adding I think is the cache we talked about when retrieving memberships, etc. Are you still going to do it?

Yes, it's added for 10 seconds.

app/controlplane/internal/usercontext/currentorganization_middleware.go
```diff
@@ -19,15 +19,21 @@ import (
...
+	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/authz"
 	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/biz"
 	"github.com/go-kratos/kratos/v2/log"
 	"github.com/go-kratos/kratos/v2/middleware"
+	"github.com/google/uuid"
+	"github.com/hashicorp/golang-lru/v2/expirable"
 )
 
-func WithCurrentOrganizationMiddleware(userUseCase biz.UserOrgFinder, logger *log.Helper) middleware.Middleware {
+var membershipsCache = expirable.NewLRU[string, *entities.Membership](0, nil, time.Second*10)
```
migmartri: 10 seconds might be too long no? I can see that you reset on project creation during the attestation, but remember that we also create projects on demand on platform. Also, this will also cause problems once we add users to groups etc no?

We should probably just make it 1s so it lasts only one user interaction? Up to you, but we need to be careful with this and keep an eye on the cache.

also, could you add a comment why we want the cache and maybe wrap it in its own srtuct since I can see you have some functions to reset this singleton?

javirln (approved):

app/cli/documentation/cli-reference.mdx
jiparis: I've removed it from the documentation and `--help` message. We can enable later on.

app/controlplane/internal/service/service.go
```diff
@@ -136,6 +140,77 @@ func WithLogger(logger log.Logger) NewOpt {
...
+	}
+}
+
+func WithProjectUseCase(projectUseCase *biz.ProjectUseCase) NewOpt {
+	return func(s *service) {
+		s.projectUseCase = projectUseCase
+	}
+}
+
+func (s *service) authorizeResource(ctx context.Context, op *authz.Policy, resourceType authz.ResourceType, resourceID uuid.UUID) error {
```
migmartri, line 159: for another PR is ok but we should document these method

migmartri (commented):
LGTM

app/controlplane/internal/usercontext/currentorganization_middleware.go
jiparis: Yes, I needed to expire it on project creation since following operations were getting a cached version without the new project. I'll lower it down to 1 second.

app/controlplane/internal/service/service.go
jiparis: Done.

PR #2136, opened 2025-06-26
feat(rbac): apply RBAC to CAS downloads

jiparis:
In this PR:
* a new `project_id` field is added to CAS backends.
* Artifacts coming from new attestations will be stored with the project ID they come from
* `project_id` will be used to filter out candidate during CASMapping query. This applies to both CASCredential and CASRedirect services.


app/controlplane/pkg/biz/casmapping.go
```diff
@@ -163,14 +184,20 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByOrg(ctx context.Context,
...
+// Extract only the mappings associated with a list of orgs and optionally a list of projects
+func filterByOrgs(mappings []*CASMapping, orgs []uuid.UUID, projectIDs map[uuid.UUID][]uuid.UUID) ([]*CASMapping, error) {
 	result := make([]*CASMapping, 0)
 
 	for _, mapping := range mappings {
 		for _, o := range orgs {
-			if mapping.OrgID.String() == o {
-				result = append(result, mapping)
+			if mapping.OrgID == o {
+				if visibleProjects, ok := projectIDs[mapping.OrgID]; ok {
```
jiparis, line 194: Here is where the filtering happens. If there are visible projects for this organization (meaning that the user is a "member" of the organization and RBAC is enabled), it will be used to filter mappings.

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -95,26 +100,42 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByUser(ctx context.Context
...
 		return nil, fmt.Errorf("failed to list memberships: %w", err)
 	}
 
-	userOrgs := make([]string, 0, len(memberships))
+	userOrgs := make([]uuid.UUID, 0)
+	// for every org with RBAC active, the list of allowed projects
+	projectIDs := make(map[uuid.UUID][]uuid.UUID)
 	for _, m := range memberships {
-		userOrgs = append(userOrgs, m.OrganizationID.String())
+		if m.ResourceType == authz.ResourceTypeOrganization {
```
jiparis, line 113: Here the visible projects per organization are collected

app/controlplane/internal/service/attestation.go
```diff
@@ -318,7 +318,10 @@ func (s *AttestationService) storeAttestation(ctx context.Context, envelope []by
 
 		for _, ref := range references {
```
migmartri, line 319: log projectID?

app/controlplane/internal/service/cascredential.go
```diff
@@ -99,7 +100,11 @@ func (s *CASCredentialsService) Get(ctx context.Context, req *pb.CASCredentialsS
 			mapping, err = s.casMappingUC.FindCASMappingForDownloadByUser(ctx, req.Digest, currentUser.ID)
 			// otherwise, we'll try to find a mapping for the current API token associated orgs
 		} else if currentAPIToken != nil {
-			mapping, err = s.casMappingUC.FindCASMappingForDownloadByOrg(ctx, req.Digest, []string{currentOrg.ID})
+			orgID, err := uuid.Parse(currentOrg.ID)
+			if err != nil {
+				return nil, handleUseCaseErr(err, s.log)
+			}
+			mapping, err = s.casMappingUC.FindCASMappingForDownloadByOrg(ctx, req.Digest, []uuid.UUID{orgID}, nil)
```
migmartri, line 104: add a TODO to indicate that we'll pass the projectID once it's supported in API tokens

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -95,26 +100,42 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByUser(ctx context.Context
 		return nil, NewErrInvalidUUID(err)
 	}
 
-	// Load organizations for the given user
-	memberships, err := uc.membershipRepo.FindByUser(ctx, userUUID)
+	// Load ALL memberships for the given user
+	memberships, err := uc.membershipRepo.ListAllByUser(ctx, userUUID)
```
migmartri, line 104: what if it's an owner? Do we have here the whole list of projects?

app/controlplane/pkg/biz/casmapping.go
```diff
@@ -95,26 +100,42 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByUser(ctx context.Context
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to list memberships: %w", err)
 	}
 
-	userOrgs := make([]string, 0, len(memberships))
+	userOrgs := make([]uuid.UUID, 0)
+	// for every org with RBAC active, the list of allowed projects
+	projectIDs := make(map[uuid.UUID][]uuid.UUID)
```
migmartri, line 111: projectIDs by org

app/controlplane/pkg/biz/casmapping.go
jiparis: No, but RBAC is only applied when the user is "member".

Issue #2130, opened 2025-06-24
Project scoped API tokens


jiparis:
This is related to https://github.com/chainloop-dev/chainloop/issues/2121

Currently, API tokens are created at organization level, and have access to all projects (they can be used to create attestations and invoke APIs scoped to the organization).
When RBAC is enabled, API tokens should be scoped to the projects the user creating it has access to. This can be done in different ways. Here there are some ideas:
* static: storing the list of "allowed" projects in the token. This can be automatic (all projects visible to the user), or manual (allowing the user to select which projects).
```
# this command will succeed if proj1 and proj2 are visible to the user in the current org
chainloop org api-token create --name test --projects proj1,proj2
```
* dynamic: storing the "creator" of the API token in a custom claim, that can be used to narrow down every query to the user visible projects.

migmartri:
> dynamic: storing the "creator" of the API token in a custom claim, that can be used to narrow down every query to the user visible projects.

I do not think it's valid, since users might change groups while tokens will stay.

I am also curious why you think we could have multiple projects in a token and not just one?

PR #2138, opened 2025-06-26
feat(api-tokens): project level tokens CRUD

migmartri:
- Support for creating project level API tokens through a new projects service
- This service leverages new RBAC system to check that the user has permissions to that project
- The main list of tokens for now only show system tokens
- These tokens in practice do not differentiate at all from other types of tokens, the generation updates and RBAC check will come in another patch.

Refs #2130


app/controlplane/internal/service/service.go
```diff
@@ -181,19 +182,29 @@ func (s *service) authorizeResource(ctx context.Context, op *authz.Policy, resou
...
 // check authorizeResource method
-func (s *service) userHasPermissionOnProject(ctx context.Context, orgID string, pName string, policy *authz.Policy) error {
+// if it doesn't return an error, it means that the user has the permission and the project is returned
+func (s *service) userHasPermissionOnProject(ctx context.Context, orgID string, pName string, policy *authz.Policy) (*biz.Project, error) {
+	p, err := s.projectUseCase.FindProjectByReference(ctx, orgID, &biz.EntityRef{Name: pName})
+	if err != nil {
+		return nil, handleUseCaseErr(err, s.log)
+	}
+
+	if p == nil {
```
jiparis: I think this case is correctly handled in the data layer already. There is no case where `p==nil` and `err==nil`

app/controlplane/pkg/authz/authz.go
```diff
@@ -255,6 +265,11 @@ var rolesMap = map[Role][]*Policy{
 		// integrations
 		PolicyAttachedIntegrationAttach,
 		PolicyAttachedIntegrationDetach,
+
+		// Project API Token
+		PolicyProjectAPITokenList,
+		PolicyProjectAPITokenCreate,
+		PolicyProjectAPITokenRevoke,
```
jiparis, line 271: Should these two be enabled for Project Viewers? They could escalate privileges, right?

app/controlplane/pkg/data/apitoken.go
```diff
@@ -56,11 +57,11 @@ func (r *APITokenRepo) Create(ctx context.Context, name string, description *str
 		return nil, fmt.Errorf("saving APIToken: %w", err)
 	}
 
-	return entAPITokenToBiz(token), nil
+	return r.FindByID(ctx, token.ID)
```
jiparis, line 60: Why is this additional query needed?

migmartri: To make sure the resulting object contains the information about the project and org

app/controlplane/pkg/authz/authz.go
migmartri: they are not, I think Github is messing with the visualization

app/controlplane/internal/service/service.go
migmartri: mm, it didn't work for me, one sec

migmartri: You are right, the issue was that I was not sending back the project, Good catch

migmartri:
@jiparis this is ready for review

app/controlplane/pkg/authz/authz.go
jiparis: right.

jiparis (approved):

Issue #2121, opened 2025-06-18
Feature: RBAC on projects


jiparis:
We want users and groups (#2120) to be assigned to projects. 
In projects, users and groups can have either the role `Project Admin` or `Project Viewer`. Admins can manage project membership.

## Usage

CLI examples (requires the role Org Admin or Project Admin):
```
chainloop project member add --group foo --name project_name
chainloop project member add --user foo@chainloop.dev --name project_name
```

Users not existent in chainloop would receive an invitation and added automatically to the project during the acceptation callback.

## New Roles
A new role `Org Member` will serve as the feature flag to enable RBAC. Those users with org-level role as `Org Member` will only see the projects they belong to. To access project contents (including its workflows and workflow runs and attestations) users must have the project-level role `Project Admin` or `Project Viewer`.

Users be invited to a project in different ways:
* By creating a project, users become `Project Admin` in that project automatically. This could happen during the attestation flow (where projects are created automatically if they don't exist) if done with a user token.
* By invitation by an existent `Project Admin`
* By invitation by an existent `Org Admin`

Project Admins can create and update project resources, like workflows and attestations.
Project Viewers can only see them.

## Project-scoped APIs
All existent APIs that act on project resources, must enable RBAC if the org-role is `Org Member`. 
For example:
```
chainloop workflow ls
```
will only show workflows from projects the user belongs to (and has `Project Admin` or `Project Viewer` on them)

Those APIs include (the list might be incomplete)
* Attestation
* AttestationState
* Integration Attachments
* Org metrics
* Referrers
* Workflows
* WorkflowRuns

Related work:
* https://github.com/chainloop-dev/chainloop/issues/2130
* #2120

PR #2140, opened 2025-06-26
fix(RBAC): correcly apply RBAC to referrers if needed

jiparis:
Fix filtering by visible referrers in Discover endpoints. Since this endpoint search in multiple orgs, the logic must consider the user role in each of those orgs to apply RBAC on projects if needed. 
* if the user doesn't need RBAC in some Organization, the resource is visible
* if the user is "member" in all organizations, RBAC is applied by checking visible projects on each (if the resource is accessible in one of them, it's returned).

Refs #2121

### DIFF
--- a/app/controlplane/pkg/biz/referrer.go
+++ b/app/controlplane/pkg/biz/referrer.go
@@ -110,7 +110,8 @@ type GetFromRootFilters struct {
 	RootKind *string
 	// Wether to filter by visibility or not
 	Public *bool
-	// ProjectIDs stores visible projects by org for the requesting user
+	// ProjectIDs stores visible projects by org for the requesting user.
+	// If an org entry doesn't exist, it means that RBAC is not applied, hence all projects in that org are visible
 	ProjectIDs map[OrgID][]ProjectID
 }
 


migmartri:
> if the user is "member" in all organizations, RBAC is applied by checking visible projects on each (if the resource is accessible in one of them, it's returned).

I do not follow this logic, could you elaborate?

app/controlplane/pkg/biz/membership.go
```diff
@@ -344,3 +344,32 @@ func (uc *MembershipUseCase) SetProjectOwner(ctx context.Context, projectID, use
...
+func getOrgsAndRBACInfoForUser(ctx context.Context, userID uuid.UUID, mRepo MembershipRepo, pRepo ProjectsRepo) ([]uuid.UUID, map[uuid.UUID][]uuid.UUID, error) {
+	// Load ALL memberships for the given user
+	memberships, err := mRepo.ListAllByUser(ctx, userID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to list memberships: %w", err)
+	}
+
+	userOrgs := make([]uuid.UUID, 0)
+	// This map holds the list of project IDs by org with RBAC active (user is org "member")
+	projectIDs := make(map[uuid.UUID][]uuid.UUID)
```
migmartri, line 357: the last list of []uuid.UUID means projects to filter in that org no? If no values it means all projects?

This is smth that's still confusing for me.

app/controlplane/pkg/biz/referrer.go
```diff
@@ -99,17 +98,17 @@ type StoredReferrer struct {
...
 }
 
 type GetFromRootFilters struct {
 	// RootKind is the kind of the root referrer, i.e ATTESTATION
 	RootKind *string
 	// Wether to filter by visibility or not
 	Public *bool
-	// If not nil, it will be used to filter the result by project
-	ProjectIDs []uuid.UUID
+	// ProjectIDs stores visible projects by org for the requesting user
```
migmartri: indicate that if the map contains nil means all the projects

migmartri (approved):
I still struggle with this a little bit. I do not understand how you can do a multi-dymension query between orgs that include all the projects, and orgs that include only some.

app/controlplane/pkg/data/referrer.go
```diff
@@ -213,6 +210,28 @@ func (r *ReferrerRepo) doGet(ctx context.Context, root *ent.Referrer, allowedOrg
 	// add additional information related to the workflows
 	hydrateWorkflowsInfo(root, res)
 
+	// check that, if RBAC is required, the user has visibility on the artifact in at least 1 org/project
+	var found bool
+	for _, oid := range res.OrgIDs {
+		if visibleProjects, ok := projectIDs[oid]; ok {
```
jiparis: Move logic to a method

app/controlplane/pkg/biz/membership.go
jiparis: Yes, if the map entry is missing, it means all projects and RBAC is not needed.

PR #2143, opened 2025-06-26
feat(RBAC): enable project-scoped API tokens

### LINKED ISSUE #2130: Project scoped API tokens

migmartri:
This PR continues with project-scoped API tokens https://github.com/chainloop-dev/chainloop/issues/2130

- Stores additional information in the token such as the projectID, projectName if applicable
- Updates the new rbac primitives to take into account this data during both filtering and enforcement.
- System-level API tokens work the same way. The new Project-level API tokens are the ones that go through the RBAC check in the service layer. 

I tested it locally performing both attestations and regular operations with admin tokens or scoped tokens, for example this is what you get when you try to perform an attestation with a token from another project.

```
chainloop att init --name test-annotations --project project-sarah  --replace --token $CHAINLOOP_TOKEN

ERR operation not allowed: This auth token is valid only with the project "project-john"
```

closes #2130


app/controlplane/pkg/biz/apitoken.go
```diff
@@ -181,18 +181,38 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description
 		return nil, fmt.Errorf("finding organization: %w", err)
 	}
 
+	// If a project is provided, we store it in the token
+	var projectID *uuid.UUID
+	if options.project != nil {
+		projectID = ToPtr(options.project.ID)
```
jiparis, line 187: doesn't `&options.project.ID` work?

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -181,18 +181,38 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description
...
+	}
+
 	// NOTE: the expiration time is stored just for reference, it's also encoded in the JWT
 	// We store it since Chainloop will not have access to the JWT to check the expiration once created
-	token, err := uc.apiTokenRepo.Create(ctx, name, description, expiresAt, orgUUID, options.projectID)
```
jiparis, line 186: So we already had tokens with projects before this change?

app/controlplane/pkg/jwt/apitoken/apitoken.go
```diff
@@ -66,30 +67,69 @@ func NewBuilder(opts ...NewOpt) (*Builder, error) {
...
+
+	if opts.OrgName == "" {
+		return "", errors.New("orgName is required")
+	}
+
+	if opts.KeyID == uuid.Nil {
+		return "", errors.New("keyID is required")
+	}
+
+	if opts.KeyName == "" {
```
jiparis, line 98: This wasn't required before. Make sure this doesn't break compatibility with old tokens

jiparis (approved):
Thanks, I like it. Basically "RBAC" for API tokens will be enabled only if the token has a "ProjectID" claim, right?

app/controlplane/internal/usercontext/apitoken_middleware.go
```diff
@@ -186,7 +186,15 @@ func setCurrentOrgAndAPIToken(ctx context.Context, apiTokenUC *biz.APITokenUseCa
...
 	// Set the current organization and API-Token in the context
 	ctx = entities.WithCurrentOrg(ctx, &entities.Org{Name: org.Name, ID: org.ID, CreatedAt: org.CreatedAt})
-	ctx = entities.WithCurrentAPIToken(ctx, &entities.APIToken{ID: token.ID.String(), CreatedAt: token.CreatedAt, Token: token.JWT})
+
+	ctx = entities.WithCurrentAPIToken(ctx, &entities.APIToken{
+		ID:          token.ID.String(),
+		Name:        token.Name,
+		CreatedAt:   token.CreatedAt,
+		Token:       token.JWT,
+		ProjectID:   token.ProjectID,
```
jiparis, line 203: If the token has a ProjectID, is this change checking that the ProjectID in the database is the same?

app/controlplane/pkg/biz/apitoken.go
migmartri: it does, I was just using one way of doing it tbh, a little bit pointless it's true

app/controlplane/pkg/biz/apitoken.go
migmartri: Yes, the other PR created it in the DB, but not in the JWT

app/controlplane/pkg/jwt/apitoken/apitoken.go
migmartri: In theory this is for new tokens, so I think we should be fine

app/controlplane/internal/usercontext/apitoken_middleware.go
migmartri: mm, it is not, let me add it

migmartri:
> Thanks, I like it. Basically "RBAC" for API tokens will be enabled only if the token has a "ProjectID" claim, right?

That's correct

app/controlplane/internal/usercontext/apitoken_middleware.go
migmartri: Good catch, I was using the projectID from the DB when in reality we need to use the one from the token. 

Now I check that both match

PR #2144, opened 2025-06-27
feat(rbac): sync only known resource policies

jiparis:
This small patch adds some flexibility to the RBAC policies syncer so that it doesn't delete policies for "unknown" resources. This allows other sources for policies to coexist seamlessly.
The Enforcer can now be configured with a custom set of role mappings and managed resources.

Refs #2121


migmartri (approved):

jiparis:
Added some tests for collisions

PR #2145, opened 2025-06-27
feat(attestation): allow arbitrary annotations in materials

### LINKED ISSUE #1976: allow to add arbitrary annotations
Currently, we only allow adding annotations defined in the contract. Similarly to contractless attestations we should allow adding arbitrary annotations

Piskoo:
This PR implements material annotation support matching the existing attestation annotation behavior:
```
a) we do not allow overriding values that come from the contract
b) we allow adding annotations that are not defined in the contract
```

Closes #1976

### DIFF
--- a/pkg/attestation/crafter/crafter.go
+++ b/pkg/attestation/crafter/crafter.go
@@ -579,7 +579,7 @@ func (c *Crafter) addMaterial(ctx context.Context, m *schemaapi.CraftingSchema_M
 
 		// NOTE: we do not allow overriding values that come from the contract
 		if existingVal, existsInContract := mt.Annotations[kr]; existsInContract && existingVal != "" {
-			c.Logger.Info().Str("key", m.Name).Str("annotation", kr).Msg("annotation can't be changed, skipping")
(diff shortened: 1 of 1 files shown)


pkg/attestation/crafter/crafter.go
```diff
@@ -571,18 +571,19 @@ func (c *Crafter) addMaterial(ctx context.Context, m *schemaapi.CraftingSchema_M
...
-			mt.Annotations[kr] = vr
-		} else {
-			// NOTE: we do not allow overriding values that come from the contract
+		if mt.Annotations == nil {
+			mt.Annotations = make(map[string]string)
+		}
+
+		// NOTE: we do not allow overriding values that come from the contract
+		if existingVal, existsInContract := mt.Annotations[kr]; existsInContract && existingVal != "" {
 			c.Logger.Info().Str("key", m.Name).Str("annotation", kr).Msg("annotation can't be changed, skipping")
```
migmartri: nitpick

```suggestion
			c.Logger.Info().Str("key", m.Name).Str("annotation", kr).Msg("annotation value is set in the contract, can not be overridden, skipping")
```

PR #2146, opened 2025-06-27
feat(attestation): add warning when outdated contract is used

### LINKED ISSUE #1910: show a warning ot the user when they are pinned to an old version of the contract
In order to communicate that there is a new version of the contract we could show a warning message during the attestation process

Piskoo:
Adds warning when outdated contract revisions are used during attestation:
```
WRN Newer contract revision available - latest revision: 3
```
Closes #1910

### DIFF
--- a/app/cli/internal/action/attestation_init.go
+++ b/app/cli/internal/action/attestation_init.go
@@ -132,6 +132,10 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 	}
 	workflow := workflowsResp.GetResult()
 
+	if err := action.warnIfOutdatedContract(workflow.ContractRevisionLatest, int32(opts.ContractRevision)); err != nil {
+		return "", err
+	}
+
 	// 2 - Get contract
 	contractResp, err := client.GetContract(ctx, &pb.AttestationServiceGetContractRequest{
 		ContractRevision: int32(opts.ContractRevision),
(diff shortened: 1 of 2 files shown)


app/cli/internal/action/attestation_init.go
```diff
@@ -326,3 +326,35 @@ func groupMaterialToCraftingSchemaMaterial(gm *v1.PolicyGroup_Material, group *v
 		Optional: gm.Optional,
 	}, nil
 }
+
+// Shows warning if newer contract revision exists
+func (action *AttestationInit) WarnIfOutdatedContract(
```
migmartri: it might be easier if we extend the `init` API response to include the latest version if it doesn't do it already

migmartri (changes requested):
Ok, in fact I think I am wrong, there is an easier way that does not neither update the API endpoint nor doing an additional request.

The workflow response here should contain the number of the latest revision of the contract available for that workflow https://github.com/chainloop-dev/chainloop/blob/729d68da044de916a65119dd797835fd6af601cb/app/cli/internal/action/attestation_init.go#L125

and in that action, you also have the "desired" contract revision provided by the user (note that if you do not provide it, you'll get the default value 0 which means latest, so it's optional)

https://github.com/chainloop-dev/chainloop/blob/729d68da044de916a65119dd797835fd6af601cb/app/cli/cmd/attestation_init.go#L90

So basically the code shoudl just check if the value is provided and smaller that the latest one that comes from the workflow API

What do you think?

Piskoo:
> Ok, in fact I think I am wrong, there is an easier way that does not neither update the API endpoint nor doing an additional request.
> 
> The workflow response here should contain the number of the latest revision of the contract available for that workflow
> 
> https://github.com/chainloop-dev/chainloop/blob/729d68da044de916a65119dd797835fd6af601cb/app/cli/internal/action/attestation_init.go#L125
> 
> and in that action, you also have the "desired" contract revision provided by the user (note that if you do not provide it, you'll get the default value 0 which means latest, so it's optional)
> 
> https://github.com/chainloop-dev/chainloop/blob/729d68da044de916a65119dd797835fd6af601cb/app/cli/cmd/attestation_init.go#L90
> 
> So basically the code shoudl just check if the value is provided and smaller that the latest one that comes from the workflow API
> 
> What do you think?

You are right, I've updated the code.

app/cli/internal/action/attestation_init.go
```diff
@@ -326,3 +330,16 @@ func groupMaterialToCraftingSchemaMaterial(gm *v1.PolicyGroup_Material, group *v
 		Optional: gm.Optional,
 	}, nil
 }
+
+// Shows warning if newer contract revision exists
+func (action *AttestationInit) warnIfOutdatedContract(latestRevision, providedRevision int32) error {
+	if action.dryRun || action.useRemoteState || providedRevision == 0 {
```
migmartri: why if remote state is set we should ignore it?

app/cli/internal/action/attestation_init.go
```diff
@@ -132,6 +132,10 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 	}
 	workflow := workflowsResp.GetResult()
 
+	if err := action.warnIfOutdatedContract(workflow.ContractRevisionLatest, int32(opts.ContractRevision)); err != nil {
```
migmartri: it feels like this method adds an unnecessary level of indirection. For example it can never return an error.

app/cli/internal/action/attestation_init.go
Piskoo: We actually shouldn't ignore it, since it does not really matter whether the attestation is stored or where it is stored.

app/cli/internal/action/attestation_init.go
```diff
@@ -132,6 +132,12 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 	}
 	workflow := workflowsResp.GetResult()
 
+	// Show warning if newer contract revision exists
+	if opts.ContractRevision > 0 && int32(opts.ContractRevision) < workflow.ContractRevisionLatest {
+		action.Logger.Warn().
+			Msgf("Newer contract revision available - latest revision: %d", workflow.ContractRevisionLatest)
```
migmartri:
```suggestion
			Msgf("Newer contract revision available: %d, pinned version: %d", workflow.ContractRevisionLatest, opts.ContractRevision)
```

app/cli/internal/action/attestation_init.go
Piskoo: Moved remaining logic to `Run` function.

Issue #2127, opened 2025-06-24
feat: store authentication identifier in attestation


migmartri:
This task involves storing in the attestation information what kind of authentication was used and its identifier.

Today, an attestation process can be authenticated via user token, API tokens, or federated tokens (OIDC token in GitLab)

I could think of

|type|identifier|
|--|--|
|USER|UserID|
|API_TOKEN|TokenID|
|FEDERATED_GITLAB_TOKEN|tokenID|

note that this information should be inside the claims of the JWT tokens. For example, see below an example of a user token

![Image](https://github.com/user-attachments/assets/0550130a-719e-455e-990e-6cf76bade132)

or a Chainloop API token

![Image](https://github.com/user-attachments/assets/3e58180a-6534-4577-a8b4-ba84c1293530)

Once we have this information, it could be rendered in the attestation json file maybe in a `auth` json message in the predicate, for example

```
{
  "predicate": {
    "auth": {
      "kind": "USER"
      "id": "deadbeef"
    } 
}
```

What do you think?

migmartri:
cc/ @jiparis @javirln

PR #2148, opened 2025-06-30
feat(attestation): support federated tokens in authentication identifier

Piskoo:
Refs #2139 and #2127 

This PR adds support for `AUTH_TYPE_FEDERATED` tokens as authentication identifiers during attestation process.

1. **Store the issuer** (e.g., `gitlab.com`) as the token ID  
2. **Handle GitLab-style tokens** (single-string `"aud":"chainloop"`)


app/cli/internal/token/token.go
```diff
@@ -63,24 +65,26 @@ func Parse(token string) (*ParsedToken, error) {
...
 
-	// Get the audience claim
-	val, ok := claims["aud"]
-	if !ok || val == nil {
+	// Handle both string and array audience formats
+	var audience string
+	switch aud := claims["aud"].(type) {
+	case string:
+		audience = aud
+	case []interface{}:
```
migmartri, line 74: why is this needed?

Piskoo: Both are valid according to jwt standard:
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3
The token that you have posted in #2139 also contains single string:
image

migmartri: ok, can you add comments explaining the logic behind the fallback, new parsing, etc?

Thanks!

app/cli/internal/token/token.go
```diff
@@ -63,24 +65,38 @@ func Parse(token string) (*ParsedToken, error) {
...
+	if audience == "" {
 		return nil, nil
 	}
 
-	// Initialize parsedToken
 	pToken := &ParsedToken{}
 
-	// Determine the type of token based on the audience.
-	switch aud[0].(string) {
+	// Determines token type and id based on audience:
```
migmartri, line 88: to me the comment is more about explainign why we think that having the `chainloop` audience is enough to decide that's a federated token.

It's a non-obvious decision, it works at least in GITLAB but because we do a check down the line, if we change the audience requirement this will stop working.

Is there any way of doing this code more defensive? Meaning, making sure that we know that it is a Gitlab token, regardless of the audience?

migmartri (approved):

app/cli/internal/token/token.go
Piskoo: We could use some of the custom claims that are always included in gitlab tokens - https://docs.gitlab.com/ci/secrets/id_token_authentication/

Piskoo:
I've added a check based on gitlab specific claims, the number of required claims is not backed by anything, it's just a suggestion since there might be some breaking changes to gitlab token claims in future versions. Let me know what you think!

PR #2164, opened 2025-07-01
chore: include project ID in workflow list endpoint

jiparis:
This PR adds a missing project ID in workflow list endpoint


app/controlplane/api/controlplane/v1/response_messages.proto
```diff
@@ -28,6 +28,7 @@ message WorkflowItem {
   string id = 1;
   string name = 2;
   string project = 3;
```
migmartri, line 30: and what's project?

migmartri (approved):

app/controlplane/api/controlplane/v1/response_messages.proto
jiparis: It's the project name.

migmartri: and why do we need the ID now?

jiparis: memberships use IDs instead of names. The Id would be used to match permissions in platform

migmartri: ok, so it's not a bug but a feature requirement? Thanks for the info

jiparis: Yes, it's a feature for RBAC. Workflows are already filtered by visibility, but the ID would be useful to check exact permissions (read or write) on the workflow.

PR #2170, opened 2025-07-02
feat(projects): Allow to include groups to projects

javirln:
This patch updates the logic for managing project memberships to support both users and groups. The `ProjectMembershipReference` now accepts a generic `IdentityReference`, allowing the underlying logic to determine whether the reference is a user or a group and handle it accordingly.

The logic responsible for retrieving a user’s memberships has also been changed to include the groups the user belongs to, merging all memberships into a single slice. Additionally, the `authorizeResource` method has been revised: instead of failing immediately when no matching role is found for a given resource type and ID, it now checks all roles associated with that resource and authorizes the request if any role matches.

Audit logs have been added to track when groups are added to projects, and the integration tests have been updated to reflect these changes.

And finally some refactors to reduce redundancy

### DIFF
--- a/app/controlplane/internal/service/service.go
+++ b/app/controlplane/internal/service/service.go
@@ -168,6 +168,8 @@ func WithGroupUseCase(groupUseCase *biz.GroupUseCase) NewOpt {
 // For example: `s.authorizeResource(ctx, authz.PolicyAttachedIntegrationDetach, authz.ResourceTypeProject, projectUUID);`
 // checks if the user has a role in the project that allows to detach integrations on it.
 // This method is available to every service that embeds `service`
+// It goes through all the memberships of the user, direct memberships and indirect memberships (Groups)
+// and checks if the user has any role that allows the operation on the resourceType and resourceID.
 func (s *service) authorizeResource(ctx context.Context, op *authz.Policy, resourceType authz.ResourceType, resourceID uuid.UUID) error {
 	if !rbacEnabled(ctx) {
 		return nil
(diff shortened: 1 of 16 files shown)


app/controlplane/pkg/data/membership.go
```diff
@@ -248,7 +250,7 @@ func (r *MembershipRepo) ListAllByUser(ctx context.Context, userID uuid.UUID) ([
 	mm, err := r.data.DB.Membership.Query().Where(
 		membership.MembershipTypeEQ(authz.MembershipTypeUser),
 		membership.MemberID(userID),
-	).All(ctx)
+	).WithOrganization().All(ctx)
```
jiparis: This Edge is deprecated, we cannot use it anymore. Note that now we have a polymorphic table, organizations are now identified by the `resourceID` field when `resourceType` is `organization`.

jiparis (changes requested):

app/controlplane/pkg/data/membership.go
jiparis: We are removing that edge in RBAC cleanup follow ups

javirln: Sure thing, I'll change it, it should be easy

javirln: Changed

app/controlplane/pkg/data/project.go
```diff
@@ -136,22 +135,42 @@ func (r *ProjectRepo) ListMembers(ctx context.Context, orgID uuid.UUID, projectI
...
-				return nil, 0, biz.NewErrNotFound("user")
+		var mems *biz.ProjectMembership
+
+		if m.MembershipType == authz.MembershipTypeUser {
+			// Fetch the user details for user memberships
+			u, uErr := r.data.DB.User.Get(ctx, m.MemberID)
+			if uErr != nil {
+				if ent.IsNotFound(uErr) {
+					r.log.Warnf("user not found for membership: %s", m.MemberID)
+					continue
```
jiparis: Not sure about continuing here. Not finding the user would mean that the system is very broken. I would rather fail so that the error is reported to sentry. What do you think?

app/controlplane/pkg/data/project.go
```diff
@@ -136,22 +135,42 @@ func (r *ProjectRepo) ListMembers(ctx context.Context, orgID uuid.UUID, projectI
...
 			}
-			return nil, 0, fmt.Errorf("failed to find user: %w", uErr)
+			mems = entProjectMembershipToBiz(m, u, nil)
+		} else if m.MembershipType == authz.MembershipTypeGroup {
+			// Fetch the group details for group memberships
+			g, gErr := r.data.DB.Group.Get(ctx, m.MemberID)
+			if gErr != nil {
+				if ent.IsNotFound(gErr) {
+					r.log.Warnf("group not found for membership: %s", m.MemberID)
+					continue
```
jiparis: same here

app/controlplane/pkg/data/project.go
javirln: Yes, I will fail here and in the next one

app/controlplane/internal/service/service.go
```diff
@@ -168,6 +168,8 @@ func WithGroupUseCase(groupUseCase *biz.GroupUseCase) NewOpt {
 // For example: `s.authorizeResource(ctx, authz.PolicyAttachedIntegrationDetach, authz.ResourceTypeProject, projectUUID);`
 // checks if the user has a role in the project that allows to detach integrations on it.
 // This method is available to every service that embeds `service`
+// It goes through all the memberships of the user, direct memberships and indirect memberships (Groups)
```
jiparis, line 171: @javirln Check also `visibleProjects` method. We might need to add groups memberhips there as well.

jiparis (changes requested):
Requesting changes for `visibleProjects` change.

app/controlplane/internal/service/service.go
javirln: `entities.CurrentMembership(ctx)` contains the memberships of the user and all of her groups. When looping through the Resources, it already takes into account the groups' roles

PR #2171, opened 2025-07-02
feat(groups): Include total number of members in a group

javirln:
Small patch that includes the total number of members a group has.


app/controlplane/pkg/data/group.go
```diff
@@ -86,7 +86,13 @@ func (g GroupRepo) List(ctx context.Context, orgID uuid.UUID, filterOpts *biz.Li
 
 	bizGroups := make([]*biz.Group, 0, len(groups))
 	for _, entGroup := range groups {
-		bizGroups = append(bizGroups, entGroupToBiz(entGroup))
+		memberCount, countErr := g.data.DB.GroupMembership.Query().
```
migmartri: can't this be an actual column in the group that gets incremented/decremented on addition/removal of members?

migmartri: this is done like this in workflow or runs, one of those

javirln: Changed

PR #2178, opened 2025-07-02
chore(rbac): Set organization ID in all memberships

jiparis:
Initially, the field was meant to be deprecated, but it's worth maintaining it for future queries for org memberships, that include all org resource memberships.


migmartri (approved):

migmartri:
I believe that @javirln added some queries already related to project memberships, would this simplify it?

PR #2152, opened 2025-06-30
feat(projects): Allow to include users to projects

javirln:
This patch introduces support for managing users within a project. Key changes include:

- New API endpoints:
  - ListMembers to list all members is a project
  - AddMember to add users to a project
  - RemoveMember to remove users from a project
- Audit logging for membership changes
- Extended test coverage to include the new endpoints
- Service-level refactor to use a generic `IdentityReference` for project-related requests

Permissions matrix:
- Admins and Owners have full access to group and membership operations
- Only Admins, Owners, and Project Admins can add or remove members

Pending:
- Allow to add groups to projects

PR #2182, opened 2025-07-03
feat: store last-used-at in the API token

### LINKED ISSUE #2169: feat: store last-used-at in the API token
It would be great to store and expose the last time an API token was used for authentication

Piskoo:
### Changes
 - Added last_used_at field to API tokens
 - Updated during auth via middleware
 - Included in API token response 
### Testing
  ```
  $ cl organization api-token ls
```
```
┌──────┬─────────────┬─────────────────────┬────────────┬────────────┬─────────────────────┐
│ NAME │ DESCRIPTION │ CREATED AT          │ EXPIRES AT │ REVOKED AT │ LAST USED AT        │
├──────┼─────────────┼─────────────────────┼────────────┼────────────┼─────────────────────┤
│ test │             │ 24 Jun 25 11:45 UTC │            │            │ 03 Jul 25 06:49 UTC │
└──────┴─────────────┴─────────────────────┴────────────┴────────────┴─────────────────────┘
```

Closes #2169

### CODE REVIEW

File: app/controlplane/pkg/biz/apitoken.go [hunk 1]
```diff
@@ -396,3 +398,25 @@ func (suc *APITokenSyncerUseCase) SyncPolicies() error {
...
 	return nil
 }
+
+func (uc *APITokenUseCase) UpdateLastUsed(ctx context.Context, tokenID string) error {
+	id, err := uuid.Parse(tokenID)
+	if err != nil {
+		return NewErrInvalidUUID(err)
+	}
+
+	token, err := uc.apiTokenRepo.FindByID(ctx, id)
```
migmartri: You might not need to search for the token first, calling update should fail with not found (see my comment in data)

app/controlplane/pkg/data/apitoken.go
```diff
@@ -153,6 +153,15 @@ func (r *APITokenRepo) UpdateExpiration(ctx context.Context, id uuid.UUID, expir
 	return nil
 }
 
+func (r *APITokenRepo) UpdateLastUsed(ctx context.Context, id uuid.UUID, lastUsedAt time.Time) error {
+	err := r.data.DB.APIToken.UpdateOneID(id).SetLastUsedAt(lastUsedAt).Exec(ctx)
```
migmartri: you might want to handle not found here, see how `func (r *APITokenRepo) Revoke` does it.

Once you do it, you'll not need to find the token in the biz layer anymore before updating it.

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -396,3 +398,25 @@ func (suc *APITokenSyncerUseCase) SyncPolicies() error {
...
+	if err != nil {
+		return fmt.Errorf("finding token: %w", err)
+	}
+
+	if token.RevokedAt != nil {
```
migmartri: this technically should not happen but it's ok as defensive programming.

app/controlplane/internal/server/grpc.go
```diff
@@ -231,6 +231,8 @@ func craftMiddleware(opts *Opts) []middleware.Middleware {
 			usercontext.WithAttestationContextFromUser(opts.UserUseCase, opts.MembershipUseCase, logHelper),
 			// 2.d - Set its robot account from federated delegation
 			usercontext.WithAttestationContextFromFederatedInfo(opts.OrganizationUseCase, logHelper),
+			// 3 - Update API Token last usage
+			usercontext.WithAPITokenUsageUpdater(opts.APITokenUseCase, logHelper),
```
migmartri, line 237: there is another branch of middlewares at the top. The one that you updated is the one focused on the attestations flow

you might want to add it after this too 


```	// 2.a - Set its API token and organization as alternative to the user
			usercontext.WithCurrentAPITokenAndOrgMiddleware(opts.APITokenUseCase, opts.OrganizationUseCase, logHelper),```

File: app/controlplane/pkg/biz/apitoken.go [hunk 5]
```diff
@@ -396,3 +398,25 @@ func (suc *APITokenSyncerUseCase) SyncPolicies() error {
 
 	return nil
 }
+
+func (uc *APITokenUseCase) UpdateLastUsed(ctx context.Context, tokenID string) error {
```
migmartri: nitpick. I'd call it, `UpdateLastUsedAt`

migmartri: could you add some integrations tests? ther should be apitoken_integration_test.go or similar. 

Thanks!

app/controlplane/pkg/biz/apitoken_integration_test.go
```diff
@@ -374,3 +374,45 @@ func (s *apiTokenTestSuite) SetupTest() {
...
 	require.NoError(s.T(), err)
 }
+
+func (s *apiTokenTestSuite) TestUpdateLastUsedAt() {
+	ctx := context.Background()
+
+	s.Run("update last used at", func() {
+		token, err := s.APIToken.Create(ctx, randomName(), nil, nil, s.org.ID)
+		require.NoError(s.T(), err)
+		assert.Nil(s.T(), token.LastUsedAt)
```
migmartri: since you  are using a test suite you can rename all these `assert.XYZ(s.T(), ...)` for `s.XYZ`, for example `		assert.Nil(s.T(), token.LastUsedAt)` => 		s.Nil(token.LastUsedAt)`

app/controlplane/pkg/data/apitoken.go
```diff
@@ -153,6 +153,21 @@ func (r *APITokenRepo) UpdateExpiration(ctx context.Context, id uuid.UUID, expir
 	return nil
 }
 
+func (r *APITokenRepo) UpdateLastUsedAt(ctx context.Context, id uuid.UUID, lastUsedAt time.Time) error {
+	err := r.data.DB.APIToken.UpdateOneID(id).
+		Where(apitoken.RevokedAtIsNil()).
+		SetLastUsedAt(lastUsedAt).
+		Exec(ctx)
+
```
migmartri: spurious extra line

app/cli/cmd/organization_apitoken_create.go
```diff
@@ -87,8 +87,13 @@ func apiTokenListTableOutput(tokens []*action.APITokenItem) error {
...
 
 		if p.RevokedAt != nil {
-			fmt.Println("revoked at", p.RevokedAt.Format(time.RFC822))
 			r = append(r, p.RevokedAt.Format(time.RFC822))
+		} else {
+			r = append(r, "")
+		}
+
+		if p.LastUsedAt != nil {
+			r = append(r, p.LastUsedAt.Format(time.RFC822))
```
migmartri, line 96: I think that @javirln added some formatTime helper, could you confirm?

app/controlplane/internal/server/grpc.go
migmartri: have you tried it for both a regular command and an attestation command?

migmartri (approved):

Piskoo:
>have you tried it for both a regular command and an attestation command?

Yes, no issues there

app/cli/cmd/organization_apitoken_create.go
Piskoo: Couldn't find it

migmartri:
Checked it locally, it works nicely, thanks!

app/cli/cmd/organization_apitoken_create.go
javirln: Yes, it was removed as part of this PR https://github.com/chainloop-dev/chainloop/pull/2152/files#diff-fcfe63513358e7cdc42f51f5fd80cfe7ea85ad24742b9b03c24eba452bf62cf0L147

The code looked like this:

```golang
func formatTime(timeStr string) string {
	if timeStr == "" {
		return ""
	}

	t, err := time.Parse(time.RFC3339, timeStr)
	if err != nil {
		return timeStr
	}

	return t.Format(time.RFC822)
}
```

migmartri: thanks! maybe we can add it back

PR #2191, opened 2025-07-03
chore: apply project filters in referrers API with API tokens

jiparis:
Follow up to apply project visibility on API tokens.


migmartri (approved):
That's it?

jiparis:
> That's it?

Yes, this case is simple since it only applies to one org. The filter has only one entry. The rest of logic for filtering is already in place, luckily.

PR #2192, opened 2025-07-03
feat(groups): Invite not found members to organization if added to group

javirln:
This patch updates the logic for adding members to a group. If the specified user doesn’t exist within the organization, an invitation email is sent. Once the invitation is accepted, the user will be added to the group. Support for adding projects will be introduced in a future PR.

It also adds a new endpoint to list the pending invitations of members to a group.


migmartri:
let's add a way to list pending requests.

app/controlplane/pkg/biz/group.go
```diff
@@ -355,86 +396,151 @@ func (uc *GroupUseCase) AddMemberToGroup(ctx context.Context, orgID uuid.UUID, o
...
+	if err != nil && !IsNotFound(err) {
+		return fmt.Errorf("failed to check requester's group membership: %w", err)
+	}
 
-		// If not a maintainer of this group, deny access
-		if requesterGroupMembership == nil || requesterGroupMembership.Role != authz.RoleGroupMaintainer {
-			return nil, NewErrValidationStr("requester does not have permission to add members to this group")
-		}
+	// If not a maintainer of this group, deny access
+	if requesterGroupMembership == nil || requesterGroupMembership.Role != authz.RoleGroupMaintainer {
```
jiparis, line 470: Weren't we using the enforcer here?

jiparis (approved):

app/controlplane/pkg/biz/group.go
javirln: on the service layer, here we check it again just in case

Issue #2190, opened 2025-07-03
project-scoped contracts


migmartri:
members should be able to manage their own contracts in their projects

![Image](https://github.com/user-attachments/assets/9cc98b60-2f3a-4832-8822-80b98d5b7a52)

PR #2193, opened 2025-07-03
refactor: move attestation initialization server side

### LINKED ISSUE #1694: confusing behavior att init --contract

migmartri:
This PR moves the attestation initialization process to the server this includes:

If the workflow doesn't exist

- Creation of the associated contract by providing some content
- Connecting the new workflow to an existing contract referenced by its name
- Creation of an empty default contract if needed 

If the workflow exists

- [NEW] We allow updating the contract already associated with the contract (fixes #1694) 
- Do not allow the user to change the contract

This changes is an improvement since it makes the process fully compatible with federated authentication as well a pre-requisite for #2190


app/cli/cmd/workflow_create.go
```diff
@@ -51,6 +51,11 @@ func newWorkflowCreateCmd() *cobra.Command {
 				// Try to find it by name
 				c, err := action.NewWorkflowContractDescribe(actionOpts).Run(contractRef, 0)
 				if err != nil || c == nil {
+					_, err := action.LoadFileOrURL(contractRef)
```
migmartri, line 54: this is just to improve this other command error message, not specifically about this problem.

migmartri:
marked it as draft to enable server side contract validation

PR #2206, opened 2025-07-05
fix(group/project): Fixes invitation query

javirln:
This patch fixes the queries that look for invitations since they were not taking into account the project and group respectively.


migmartri:
thanks @javirln for the fix, but let's try to not hotfix in this repository unless strictly necessary, since it affects the scorecards score.

Do you need to perform a release of this change? This feat is not in prod yet correct?

PR #2211, opened 2025-07-07
feat(groups): Allow to update the status of a maintainer in a group

javirln:
This patch includes a new endpoint to update the membership status of a member in a group. In order to update the status the request must be either admin/owner of the organization or be a current maintainer of the group.

Audit logs and tests have been updated.


app/controlplane/api/controlplane/v1/group.proto
```diff
@@ -216,3 +218,16 @@ message GroupMember {
   // Timestamp when the group membership was last modified
   google.protobuf.Timestamp updated_at = 4;
 }
+
+// GroupServiceUpdateMemberMaintainerStatusRequest contains the information needed to update a member's maintainer status
+message GroupServiceUpdateMemberMaintainerStatusRequest {
+  // IdentityReference is used to specify the group by either its ID or name
+  IdentityReference group_reference = 1 [(buf.validate.field).required = true];
+  // The user whose maintainer status is to be updated
+  string user_email = 2 [(buf.validate.field).string.email = true];
```
migmartri: should we instead use the user_id? email makes sense if you don't know if they exist in the system, but that's not the case correct?

javirln: Done

app/controlplane/api/controlplane/v1/group.proto
```diff
@@ -216,3 +218,16 @@ message GroupMember {
   // Timestamp when the group membership was last modified
   google.protobuf.Timestamp updated_at = 4;
 }
+
+// GroupServiceUpdateMemberMaintainerStatusRequest contains the information needed to update a member's maintainer status
+message GroupServiceUpdateMemberMaintainerStatusRequest {
+  // IdentityReference is used to specify the group by either its ID or name
+  IdentityReference group_reference = 1 [(buf.validate.field).required = true];
+  // The user whose maintainer status is to be updated
+  IdentityReference user_reference = 2 [(buf.validate.field).required = true];
```
migmartri: user reference? would you ever pass a name? I have the feeling we are overcomplicating the API sometimes.

javirln: The name in that case would be the email, it can happen if the API is directly used

migmartri: but isn't that wrong/non intuitive? why do we want to add the option to provide an email address?

javirln: I'll change it to use the user id to unblock this, we can come back.

jiparis: Even the group, I would say in this context it will be referenced only by ID. Not important in any case.

jiparis (approved):

app/controlplane/api/controlplane/v1/group.proto
javirln: The group had a reference for historical reason and I was thinking about those reasons to add the user_reference as well, just in case they come back not to change the API once again.

Anyway, let's use the id

PR #2212, opened 2025-07-07
feat(group/project): Only org admin/owners can invite new users

javirln:
When adding members that are not on the system, only admin and owners are the one able to invite new users, not group maintainers nor group admins.

### DIFF
--- a/app/controlplane/pkg/authz/authz.go
+++ b/app/controlplane/pkg/authz/authz.go
@@ -35,22 +35,23 @@ const (
 
 	// Resources
 
-	ResourceWorkflowContract      = "workflow_contract"
-	ResourceCASArtifact           = "cas_artifact"
-	ResourceCASBackend            = "cas_backend"
-	ResourceReferrer              = "referrer"
-	ResourceAvailableIntegration  = "integration_available"
-	ResourceRegisteredIntegration = "integration_registered"
-	ResourceAttachedIntegration   = "integration_attached"
-	ResourceOrgMetric             = "metrics_org"
-	ResourceRobotAccount          = "robot_account"
-	ResourceWorkflowRun           = "workflow_run"
-	ResourceWorkflow              = "workflow"
-	Organization                  = "organization"
-	ResourceGroup                 = "group"
-	ResourceGroupMembership       = "group_membership"
-	ResourceProjectAPIToken       = "project_api_token"
-	ResourceProjectMembership     = "project_membership"
+	ResourceWorkflowContract        = "workflow_contract"
+	ResourceCASArtifact             = "cas_artifact"
+	ResourceCASBackend              = "cas_backend"
+	ResourceReferrer                = "referrer"
+	ResourceAvailableIntegration    = "integration_available"
+	ResourceRegisteredIntegration   = "integration_registered"
+	ResourceAttachedIntegration     = "integration_attached"
+	ResourceOrgMetric               = "metrics_org"
+	ResourceRobotAccount            = "robot_account"
+	ResourceWorkflowRun             = "workflow_run"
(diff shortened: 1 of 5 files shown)


app/controlplane/pkg/biz/group.go
```diff
@@ -487,6 +487,17 @@ func (uc *GroupUseCase) handleNonExistingUser(ctx context.Context, orgID, groupI
...
 	}
 
+	// Check if the requester is an admin or owner of the organization
+	requesterMembership, err := uc.membershipRepo.FindByOrgAndUser(ctx, orgID, opts.RequesterID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check requester's role: %w", err)
+	}
+
+	// Only send an invitation if the requester is an admin or owner of the organization
+	if requesterMembership.Role != authz.RoleAdmin && requesterMembership.Role != authz.RoleOwner {
```
migmartri: what about using the enforcer? I'd say that in general, any hardcoded role check should be removed?

migmartri (approved):
It's ok for now but I think we need to update the code to use enforcer if possible

app/controlplane/pkg/biz/group.go
javirln: I don’t see how we can handle this with the enforcer, since there’s no existing policy for inviting users to a group or project. We could create one, but we’d still need to assign it to a role—most likely admin or owner.

What I mean is that this is a special case where the role is already checked: we allow group or project admins to add existing users with a specific role, but sending invitations should require being an admin or owner.

migmartri: can't you ask if the `requesterMembership.Role` can do invitations? 

> roup or project. We could create one, but we’d still need to assign it to a role—most likely admin or owner.

and yes, we could do that, we have done that for some other policies we needed to check manually using the enforcer inline in the code, for example during download/upload I believe.

javirln: > can't you ask if the requesterMembership.Role can do invitations?

There is no policy for organization invitations.

I'll create one and do the changes.

javirln:
@migmartri added the changes with the custom policy, let me know if that's what you were thinking about. cc @jiparis

app/controlplane/pkg/authz/authz.go
```diff
@@ -200,6 +203,8 @@ var RolesMap = map[Role][]*Policy{
 		// We do a manual check in the artifact upload endpoint
 		// so we need the actual policy in place skipping it is not enough
 		PolicyArtifactUpload,
+		// We manually check this policy to be able to know if the user can invite users to the system
+		PolicyOrganizationInvitationsCreate,
```
jiparis, line 207: In general, we don't use the enforcer for Admins/Owners. If we do this, then we'll need to do it for every policy.

jiparis: Nevermind, I see this case is a bit different

jiparis (approved):

app/controlplane/pkg/authz/authz.go
javirln: Let's then agree how we want to use this kind of behavior

migmartri: +1

To me, any authorization ask should be run through an `enforce(resource, policy)` call if possible. If this requires adding the policies to the list I'd add it.

The fact that today we bypass admin/owner in the middleware is due to laziness on my part that didn't do the full exercise of adding all the policies needed

javirln: No worries Migue, since we need to review the permissions matrix for groups again, I'll take it into account to the changes and refactor what we have to use the enforcer whenever needed

PR #2213, opened 2025-07-07
fix(group): Allow organization members to get group details

javirln:
This pull request adds a new authorization policy for the `Get` method in the `GroupService` API. This change ensures that the `PolicyGroupRead` policy is applied when accessing group details.

Authorization updates:

* [`app/controlplane/pkg/authz/authz.go`](diffhunk://#diff-72141455c72a8057aa8d7219c28c15d4afdd2cc10459d355626022b3787ecaa3R371): Added the `PolicyGroupRead` policy to the `ServerOperationsMap` for the `/controlplane.v1.GroupService/Get` endpoint, enabling proper authorization checks for retrieving group details.

### DIFF
--- a/app/controlplane/pkg/authz/authz.go
+++ b/app/controlplane/pkg/authz/authz.go
@@ -368,6 +368,7 @@ var ServerOperationsMap = map[string][]*Policy{
 	"/controlplane.v1.OrganizationService/ListMemberships": {PolicyOrganizationListMemberships},
 	// Groups
 	"/controlplane.v1.GroupService/List": {PolicyGroupList},
+	"/controlplane.v1.GroupService/Get":  {PolicyGroupRead},
 	// Group Memberships
 	"/controlplane.v1.GroupService/ListMembers": {PolicyGroupListMemberships},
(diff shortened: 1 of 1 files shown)


app/controlplane/pkg/authz/authz.go
```diff
@@ -368,6 +368,7 @@ var ServerOperationsMap = map[string][]*Policy{
 	"/controlplane.v1.OrganizationService/ListMemberships": {PolicyOrganizationListMemberships},
 	// Groups
 	"/controlplane.v1.GroupService/List": {PolicyGroupList},
+	"/controlplane.v1.GroupService/Get":  {PolicyGroupRead},
```
jiparis, line 371: I think `PolicyGroupRead` needs to be added to Group Maintainers, and add the logic in the service `Get` to check maintainer policies, right?

jiparis: Or another option, adding it to the OrgViewer, as they can be maintainers as well.

jiparis: Keeping viewers outside the scope for now.

Issue #2209, opened 2025-07-07
plugin installation feedback


migmartri:
- `-f, --file string       URL of the plugin to download (required)` Why is it called file and then it's an URL?
- `--filename string   Custom filename to save the plugin as (optional)` Why does the filename matter for the user?

![Image](https://github.com/user-attachments/assets/ee67d3c5-b4e2-4c3f-8b72-814febf7757f)

PR #2215, opened 2025-07-07
Remove the --filename flag and update the -f description

gr0:
This PR solves https://github.com/chainloop-dev/chainloop/issues/2209 and includes:

- removal of the `--filename` flag
- updated `-f` flag description

### DIFF
--- a/app/cli/cmd/plugins.go
+++ b/app/cli/cmd/plugins.go
@@ -245,7 +245,6 @@ func newPluginDescribeCmd() *cobra.Command {
 
 func newPluginInstallCmd() *cobra.Command {
 	var file string
-	var filename string
 	var location string
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


app/cli/cmd/plugins.go
```diff
@@ -272,8 +270,7 @@ func newPluginInstallCmd() *cobra.Command {
 	}
 
 	// Common flags
-	cmd.Flags().StringVarP(&file, "file", "f", "", "URL of the plugin to download (required)")
-	cmd.Flags().StringVarP(&filename, "filename", "", "", "Custom filename to save the plugin as (optional)")
+	cmd.Flags().StringVarP(&file, "file", "f", "", "URL or path to the plugin to install (required)")
```
migmartri:
```suggestion
	cmd.Flags().StringVarP(&file, "file", "f", "", "URL or path to the plugin to install")
```
you might want to leave the `requirement` behavior to be exposed by the tool itself.

migmartri (approved):

app/cli/cmd/plugins.go
gr0: Great point, thank you!

Issue #2210, opened 2025-07-07
failed to load plugin exec format error


migmartri:
after building the plugin for my architecture (I believe)

```
make build
mkdir -p bin/ && go build -ldflags\
	  "-X github.com/chainloop-dev/platform/backend/internal/biz.Version=\
	  -X main.Version="\
	   -o ./bin/chainloop-plugin ./cmd/...
```

and successfully adding it to the CLI

```
chainloop config plugin install -f bin/chainloop-plugin
Plugin installed successfully to: /Users/miguelmartinez/Library/Application Support/chainloop/plugins/chainloop-plugin
```

I get this error

```
chainloop config plugin ls
2025-07-07T11:57:01.235+0200 [WARN]  plugin: plugin failed to exit gracefully
failed to load plugin: failed to create RPC client: fork/exec /Users/miguelmartinez/Library/Application Support/chainloop/plugins/chainloop-plugin: exec format error
No plugins installed
```

What's the problem?

Also, should we make this error more human friendly? Something like `the plugin xyz couldn't be loaded` and then also put a debug entry in the logs with the actual error?

migmartri:
I am using Mac OS Arm

migmartri:
In addition to make it more human friendly, IMO a crash in the plugin should not break the CLI completely

gr0:
> In addition to make it more human friendly, IMO a crash in the plugin should not break the CLI completely

We are not breaking the CLI - when the plugin fails to be loaded, which is where we check for compatibility, we just continue with the execution. The error reporting is not very good right now - I'm adjusting it.

PR #2216, opened 2025-07-07
Improve error reporting from plugins loading

gr0:
This PR adds improvement error reporting for CLI plugins - https://github.com/chainloop-dev/chainloop/issues/2210. 

I would use a logger, but the problem is that in order for the plugins mechanism to be able to expand the Cobra command I'm loading the plugins before the logger is actually initialized, so before the flags are passed to the CLI command. That means at that point I don't know the user wants a debug, info or any other log level. 

If we agree on the change the response from the CLI looks as follows:

```sh
2025-07-07T23:18:29.823+0200 [WARN]  plugin: plugin failed to exit gracefully
failed to load plugin: /Users/gro/Library/Application Support/chainloop/plugins/chainloop - reason: failed to create RPC client: Unrecognized remote plugin message: Chainloop Command Line Interface
No plugins installed
```

It is definitely not ideal, but way better compared to the full log. If you have other ideas @migmartri I'm listening :)

### DIFF
--- a/app/cli/pkg/plugins/manager.go
+++ b/app/cli/pkg/plugins/manager.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-plugin"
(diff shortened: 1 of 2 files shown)


app/cli/pkg/plugins/manager.go
```diff
@@ -62,7 +63,8 @@ func (m *Manager) LoadPlugins(ctx context.Context) error {
 	for _, plugin := range plugins {
 		// Load the plugin - if there is an error just skip it - we can think of a better strategy later
 		if err := m.loadPlugin(ctx, plugin); err != nil {
-			fmt.Printf("failed to load plugin: %s\n", err)
+			fmt.Printf("failed to load plugin: %s - reason: %s\n", plugin, getFirstErrorLine(err))
```
migmartri: I believe this is not correct, this is poluting stdout, affecting whatever next command output comes in.

I'd use the logger in debug mode (that should send to stderr correct?), and then we can figure out how to add levels?

gr0: Sure thing - changed.

migmartri (approved):

PR #2217, opened 2025-07-08
feat(group/project): Include invitation ID on list of pending invitations

javirln:
In order to be able to revoke invitations for groups and project we need to include the invitation ID. This patch modifies the endpoints adding the missing field.


migmartri (approved):
Thanks

PR #2224, opened 2025-07-09
feat: add default version when workflow is created

migmartri:
This PR makes sure there is a default version for any new project created through a workflow flow.

before this, only during attestations the project would have had a version

### DIFF
--- a/app/controlplane/pkg/data/projectversion.go
+++ b/app/controlplane/pkg/data/projectversion.go
@@ -109,6 +109,15 @@ func createProjectVersionWithTx(ctx context.Context, tx *ent.Tx, projectID uuid.
 		Save(ctx)
 }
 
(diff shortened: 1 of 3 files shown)


app/controlplane/pkg/data/projectversion.go
```diff
@@ -109,6 +109,15 @@ func createProjectVersionWithTx(ctx context.Context, tx *ent.Tx, projectID uuid.
 		Save(ctx)
 }
 
+func findProjectVersionWithTx(ctx context.Context, tx *ent.Tx, projectID uuid.UUID, version string) (*ent.ProjectVersion, error) {
```
jiparis: It doesnt need to be `ent.Tx`, just use `ent.Client` and this method will be valid for bot transactional and not transactional clients.

jiparis (approved):
thanks, please check my comment.

PR #2226, opened 2025-07-09
feat(group): Prevent creation of groups with reserved words

javirln:
This patch prevents the creation or update of groups with reserved names. Any group name starting with Group- or group- is considered reserved and will be rejected.

### DIFF
--- a/app/controlplane/pkg/biz/group.go
+++ b/app/controlplane/pkg/biz/group.go
@@ -18,6 +18,7 @@ package biz
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/auditor/events"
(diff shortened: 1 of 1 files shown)


javirln:
I can move the validation further down into the DB schema. Just to consider the case of using the repository directly. cc @jiparis @migmartri

app/controlplane/pkg/biz/group.go
```diff
@@ -824,3 +834,11 @@ func (uc *GroupUseCase) ValidateGroupIdentifier(ctx context.Context, orgID uuid.
 
 	return groups[0].ID, nil
 }
+
+// isReservedGroupName checks if the provided group name is a reserved name.
+// Reserved names are used by the system for default roles and cannot be used for custom groups.
+// Any name starting with "Org-" (case insensitive) is considered reserved.
+func isReservedGroupName(name string) bool {
+	// Check if the name starts with "org-" (case insensitive)
+	return strings.HasPrefix(strings.ToLower(name), "org-")
```
jiparis, line 843: I would move this to a configuration setting and make it optional. What do you think?

javirln: @migmartri commented to have it hardcoded

javirln:
Closing this as we are not going to reserve any name

PR #2234, opened 2025-07-10
fix(projects): Always load organization on memberships

javirln:
This patch eagerly loads the organization of the membership, so it's available and can be used to check roles in the service and business layers.

### DIFF
--- a/app/controlplane/pkg/data/membership.go
+++ b/app/controlplane/pkg/data/membership.go
@@ -284,7 +284,7 @@ func (r *MembershipRepo) ListGroupMembershipsByUser(ctx context.Context, userID 
 		groupRoleMemberships, err := r.data.DB.Membership.Query().Where(
 			membership.MembershipTypeEQ(authz.MembershipTypeGroup),
 			membership.MemberIDIn(groupIDs...),
-		).All(ctx)
+		).WithOrganization().All(ctx)
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


jiparis (approved):
Thanks.
I'd rather just get the Organization ID from the foreign key instead of an additional query. But accepting since this fixes the issue.

javirln:
> Thanks. I'd rather just get the Organization ID from the foreign key instead of an additional query. But accepting since this fixes the issue.

That is how we are doing it on the direct roles of the users. We can do instead:
```
diff --git a/app/controlplane/pkg/data/membership.go b/app/controlplane/pkg/data/membership.go
index ef78cec0..aa89b7ca 100644
--- a/app/controlplane/pkg/data/membership.go
+++ b/app/controlplane/pkg/data/membership.go
@@ -284,7 +284,9 @@ func (r *MembershipRepo) ListGroupMembershipsByUser(ctx context.Context, userID
 		groupRoleMemberships, err := r.data.DB.Membership.Query().Where(
 			membership.MembershipTypeEQ(authz.MembershipTypeGroup),
 			membership.MemberIDIn(groupIDs...),
-		).WithOrganization().All(ctx)
+		).WithOrganization(func(q *ent.OrganizationQuery) {
+			q.Select(organization.FieldID)
+		}).All(ctx)
 
 		if err != nil {
 			return nil, fmt.Errorf("failed to query group role memberships: %w", err)

```

But that will break the mapping for the rest of the queries

PR #2240, opened 2025-07-10
WIP fix(membership): Removing a user from the org should cleanup resources

javirln:


migmartri:
Just from top of my mind, would it be easier and more localized to extend

```
// Delete deletes a membership by ID.
func (r *MembershipRepo) Delete(ctx context.Context, id uuid.UUID) error {
	return r.data.DB.Membership.DeleteOneID(id).Exec(ctx)
}
```

and in there make sure we delete not only that membership but all the related memberships + the group memberships? all of that from there?

javirln:
> Just from top of my mind, would it be easier and more localized to extend
> 
> ```
> // Delete deletes a membership by ID.
> func (r *MembershipRepo) Delete(ctx context.Context, id uuid.UUID) error {
> 	return r.data.DB.Membership.DeleteOneID(id).Exec(ctx)
> }
> ```
> 
> and in there make sure we delete not only that membership but all the related memberships + the group memberships? all of that from there?

Yes, that's why I marked as WIP and draft because I was trying to check where else we were cleaning up resources. It is not all centralized in a single place.

app/controlplane/pkg/data/membership.go
```diff
@@ -240,8 +242,67 @@ func (r *MembershipRepo) SetRole(ctx context.Context, membershipID uuid.UUID, ro
...
+				membership.HasOrganizationWith(organization.ID(orgID)),
+			).Exec(ctx); err != nil {
+				return fmt.Errorf("failed to delete related memberships: %w", err)
+			}
+
+			// Remove the user from all groups in the organization by soft-deleting group memberships
+			now := time.Now()
+
+			// Soft delete all group memberships for this user in this organization
+			if _, err := tx.GroupMembership.Update().Where(
```
migmartri, line 283: do we need to do anything about invitations?

javirln: Hmm I don't think so because you can only remove members that are already part of the organization meaning the invitation would be redeemed no?

migmartri: including the group ones?

javirln: They are part of the same table just with an additional context but are managed in the same way

migmartri (approved):
Thanks

PR #2248, opened 2025-07-11
feat(groups): Allow to list projects a group is member of

javirln:
This patch introduces an endpoint to retrieve all projects a group is a member of. If a requester is provided, the results are filtered based on the projects the user has visibility into.


app/controlplane/internal/service/group.go
```diff
@@ -479,6 +479,69 @@ func (g *GroupService) UpdateMemberMaintainerStatus(ctx context.Context, req *pb
...
+		return nil, errors.BadRequest("invalid", "invalid user ID")
+	}
+
+	// Initialize the options for getting projects
+	groupOpts := &biz.ListProjectsByGroupOpts{
+		IdentityReference: &biz.IdentityReference{
+			ID:   id,
+			Name: name,
+		},
+		RequesterID: &requesterUUID,
```
jiparis: Just pass `g.visibleProjects(ctx)` . That will add an array of projects to filter by, or `nil` if no filter is needed.

app/controlplane/api/controlplane/v1/group.proto
```diff
@@ -236,3 +238,35 @@ message GroupServiceUpdateMemberMaintainerStatusRequest {
...
+// ProjectInfo represents detailed information about a project that a group is a member of
+message ProjectInfo {
+  // Unique identifier of the project
+  string id = 1;
+  // Name of the project
+  string name = 2;
+  // Description of the project
+  string description = 3;
+  // Role of the group in the project (admin or viewer)
+  string role = 4;
```
migmartri: could we make this an enum?

app/controlplane/api/controlplane/v1/project.proto
```diff
@@ -66,6 +66,8 @@ message ProjectMember {
   google.protobuf.Timestamp created_at = 4;
   // Timestamp when the project membership was last modified
   google.protobuf.Timestamp updated_at = 5;
+  // The ID of latest project version this member is associated with
+  string latest_project_version_id = 6;
```
migmartri, line 70: why do we need this?

app/controlplane/internal/service/group.go
```diff
@@ -479,6 +479,69 @@ func (g *GroupService) UpdateMemberMaintainerStatus(ctx context.Context, req *pb
...
+	}
+
+	// Parse requesterID (current user)
+	requesterUUID, err := uuid.Parse(currentUser.ID)
+	if err != nil {
+		return nil, errors.BadRequest("invalid", "invalid user ID")
+	}
+
+	// Initialize the options for getting projects
+	groupOpts := &biz.ListProjectsByGroupOpts{
```
migmartri, line 502: why aren't we loading here the list of projects the user has access to with visibleProjects as we do in other places?

I believe @jiparis mentioned smth similar in another PR, about where to add the authorization logic.

app/controlplane/pkg/biz/group.go
```diff
@@ -853,3 +866,46 @@ func (uc *GroupUseCase) ValidateGroupIdentifier(ctx context.Context, orgID uuid.
...
+	}
+
+	pgOpts := pagination.NewDefaultOffsetPaginationOpts()
+	if paginationOpts != nil {
+		pgOpts = paginationOpts
+	}
+
+	var requesterMemberships []*Membership
+	// If requesterID is provided, list all memberships for the requester
+	if opts.RequesterID != nil {
```
jiparis: no need of this. get `visibleProjects` from the service and use it to create the filter.

migmartri (commented):
S

app/controlplane/pkg/authz/authz.go
```diff
@@ -393,7 +393,8 @@ var ServerOperationsMap = map[string][]*Policy{
 	"/controlplane.v1.GroupService/List": {PolicyGroupList},
 	"/controlplane.v1.GroupService/Get":  {PolicyGroupRead},
 	// Group Memberships
-	"/controlplane.v1.GroupService/ListMembers": {PolicyGroupListMemberships},
+	"/controlplane.v1.GroupService/ListMembers":  {PolicyGroupListMemberships},
+	"/controlplane.v1.GroupService/ListProjects": {PolicyGroupListMemberships},
```
migmartri: is this permission correct? Sholdn't we add a new one and assign it properly?

app/controlplane/internal/service/group.go
jiparis: Yes, that's what I mention in the other comment. We created that method just for this purpose. @javirln just call `visibleProjects` and you'll get the filtered list already. A `nil` list means no filter at all. So the code in Data is valid.

app/controlplane/pkg/data/group.go
```diff
@@ -573,6 +577,122 @@ func (g GroupRepo) UpdateMemberMaintainerStatus(ctx context.Context, orgID uuid.
...
+	membershipQuery := g.data.DB.Membership.Query().
+		Where(
+			membership.MemberIDEQ(groupID),
+			membership.MembershipTypeEQ(authz.MembershipTypeGroup),
+			membership.ResourceTypeEQ(authz.ResourceTypeProject),
+			membership.HasOrganizationWith(organization.IDEQ(orgID)),
+		)
+
+	// If visibleProjectIDs is provided, filter memberships to include only those projects
+	if len(visibleProjectIDs) > 0 {
```
jiparis: Instead, `if visibleProjectIDs != nil` when you start passing it from `visibleProjects` (`nil` means "RBAC not applied.")

app/controlplane/api/controlplane/v1/project.proto
javirln: It has been requested internally

app/controlplane/pkg/authz/authz.go
jiparis: not sure. Keep in mind this is just for the middleware to allow the call. Since visible projects are filtered already at the service level, we might not need another permission.

app/controlplane/internal/service/group.go
migmartri: I added a couple of these yesterday, this is an example https://github.com/chainloop-dev/chainloop/blob/f75ad1d449c98c8ab7eab80f4233e33387b296a0/app/controlplane/internal/service/workflowcontract.go#L51

app/controlplane/pkg/authz/authz.go
javirln: I've created another one just for the sake of having it separated but I don't think we need it

app/controlplane/pkg/data/group.go
```diff
@@ -573,6 +577,122 @@ func (g GroupRepo) UpdateMemberMaintainerStatus(ctx context.Context, orgID uuid.
...
 }
 
+// ListProjectsByGroup retrieves a list of projects that a group is a member of with pagination.
+func (g GroupRepo) ListProjectsByGroup(ctx context.Context, orgID uuid.UUID, groupID uuid.UUID, visibleProjectIDs []uuid.UUID, paginationOpts *pagination.OffsetPaginationOpts) ([]*biz.GroupProjectInfo, int, error) {
+	if paginationOpts == nil {
+		paginationOpts = pagination.NewDefaultOffsetPaginationOpts()
+	}
+
+	// Check the group exists in the organization
+	existingGroup, err := g.data.DB.Group.Query().
```
jiparis: This query was already done by the caller, right?

javirln: Yes, but I need the group to construct the result struct, and it’s possible the caller hasn’t verified its existence. Either way, it’s required to build the result

jiparis: I'm curious why it needs the group since this is just listing projects. The group is already known, because it's what originated the call, right?

javirln: I've just removed it.

javirln: We might not need it for now

app/controlplane/internal/service/group.go
```diff
@@ -479,6 +479,58 @@ func (g *GroupService) UpdateMemberMaintainerStatus(ctx context.Context, req *pb
...
+		return nil, errors.BadRequest("invalid", fmt.Sprintf("invalid group reference: %s", err.Error()))
+	}
+
+	// Initialize the options for getting projects
+	groupOpts := &biz.ListProjectsByGroupOpts{
+		IdentityReference: &biz.IdentityReference{
+			ID:   id,
+			Name: name,
+		},
+		VisibleProjectsIDs: g.visibleProjects(ctx),
```
migmartri: I'd rather call this FilterByProject since it can be empty and that means that the filter is empty, not that there are not projects

```suggestion
		VisibleProjectsIDs: g.visibleProjects(ctx),
```

jiparis (approved):

PR #2251, opened 2025-07-14
chore: org viewers cannot become project admins

jiparis:
Org Viewers only have read-only permissions and don't participate in RBAC, so it's not consistent with ProjectAdmin role, which can manage resources.

### DIFF
--- a/app/controlplane/cmd/wire.go
+++ b/app/controlplane/cmd/wire.go
@@ -100,7 +100,6 @@ func serviceOpts(l log.Logger, enforcer *authz.Enforcer, pUC *biz.ProjectUseCase
 		service.WithLogger(l),
 		service.WithEnforcer(enforcer),
 		service.WithProjectUseCase(pUC),
-		service.WithGroupUseCase(gUC),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


app/controlplane/cmd/wire.go
```diff
@@ -100,7 +100,6 @@ func serviceOpts(l log.Logger, enforcer *authz.Enforcer, pUC *biz.ProjectUseCase
 		service.WithLogger(l),
 		service.WithEnforcer(enforcer),
 		service.WithProjectUseCase(pUC),
-		service.WithGroupUseCase(gUC),
```
jiparis: I've removed this since this use case was only used by the group service

app/controlplane/internal/service/group.go
```diff
@@ -531,6 +534,74 @@ func (g *GroupService) ListProjects(ctx context.Context, req *pb.GroupServiceLis
 	}, nil
 }
 
+// userHasPermissionToAddGroupMember checks if the user has permission to add members to a group
+func (g *GroupService) userHasPermissionToAddGroupMember(ctx context.Context, orgID string, groupIdentifier *pb.IdentityReference) error {
```
jiparis: Moved here from `service.go`

app/controlplane/internal/service/service.go
```diff
@@ -253,74 +240,6 @@ func (s *service) userHasPermissionOnProject(ctx context.Context, orgID string,
 	return p, nil
 }
 
-// userHasPermissionToAddGroupMember checks if the user has permission to add members to a group
```
jiparis: moved to group.go for readability

PR #2252, opened 2025-07-14
fix(group): Created at when listing project for groups

javirln:
This patch modifies the `created_at` of the `ProjectInfo` so it contains when the membership between the group and the project was created not when the project was created.


app/controlplane/api/controlplane/v1/group.proto
```diff
@@ -267,6 +267,6 @@ message ProjectInfo {
   ProjectMemberRole role = 4;
   // The latest version ID of the project, if available
   optional string latest_version_id = 5;
-  // Timestamp when the project was created
+  // Timestamp when the membership was created
   google.protobuf.Timestamp created_at = 6;
```
migmartri: this is strange, projectInfo is a generic message no? We might need something else. Like ProjectMembership?

migmartri (approved):

app/controlplane/api/controlplane/v1/group.proto
javirln: We have `ProjectMember`. 

https://github.com/chainloop-dev/chainloop/blob/d36ba512ab8f890ff15c4c24f6fee5481f667078/app/controlplane/api/controlplane/v1/project.proto#L55

That represents the members of a Project. This is the other way around, the projects a group is member of. This `ProjectInfo` represents detailed information about a project that a group is a member of. We could call it `GroupProjectMembership` for example

migmartri: Yeah, probably call it smth less confusing, unless this is nested inside the response, is it?

PR #2256, opened 2025-07-15
fix(group): Deletion of groups handle correctly

javirln:
This patch fixes issues with group removal, where related resources were being left behind. Deleting a group now performs the following actions:
- Marks the group as deleted
- Marks all user group memberships as deleted
- Removes all group-related entries from the membership table, including users with the maintainer role and group memberships in projects
- Marks all pending invitations associated with the group as deleted, if any

It also creates a migration that fixes all current dangling groups if any in the system by performing the same steps as above.

### DIFF
--- a/app/controlplane/pkg/data/group.go
+++ b/app/controlplane/pkg/data/group.go
@@ -369,20 +369,68 @@ func (g GroupRepo) Update(ctx context.Context, orgID uuid.UUID, groupID uuid.UUI
 }
 
 // SoftDelete soft-deletes a group by setting the DeletedAt field to the current time.
+// It also marks all group memberships as deleted and removes any pending invitations related to the group.
 func (g GroupRepo) SoftDelete(ctx context.Context, orgID uuid.UUID, groupID uuid.UUID) error {
-	// Softly delete the group by setting the DeletedAt field
-	_, err := g.data.DB.Group.UpdateOneID(groupID).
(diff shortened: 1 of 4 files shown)


migmartri (approved):
Thanks

app/controlplane/pkg/data/group.go
```diff
@@ -369,20 +369,68 @@ func (g GroupRepo) Update(ctx context.Context, orgID uuid.UUID, groupID uuid.UUI
...
+		}
+
+		// Delete all memberships where this group is either the member or the resource
+		_, err = tx.Membership.Delete().Where(
+			membership.HasOrganizationWith(organization.ID(orgID)),
+			membership.Or(
+				membership.MemberID(groupID),
+				membership.And(
+					membership.ResourceID(groupID),
+					membership.ResourceTypeEQ(authz.ResourceTypeGroup),
```
jiparis, line 408: is this possible? a membership with type = group? Group memberships have their own table, right?

jiparis: replying myself, the maintainer role is stored here, thanks!

javirln: Yes the maintainer one

jiparis (approved):
Thanks. 
As a side note, not for now, I would remove completely the `deleted_at` from group_membership table, as we don't have it in other "membership" tables.

PR #2257, opened 2025-07-15
fix(group): Calculate member count as a whole

javirln:
This patch updates the way group membership counts are maintained. Instead of incrementing or decrementing the counter on each insertion or deletion, the total number of members is now recalculated after every transaction and the count is updated accordingly.

### DIFF
--- a/app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
+++ b/app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
@@ -9,5 +9,4 @@ SET "member_count" = (
     FROM "group_memberships" gm
     WHERE gm."group_id" = g."id"
       AND gm."deleted_at" IS NULL
-)
-WHERE 1=1; -- Apply to all groups
+); -- Apply to all groups
--- a/app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
+++ b/app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
@@ -1,0 +1,13 @@
+-- Fix group member counts by calculating the correct number of members for each group
+-- This migration addresses any inconsistencies in the member_count field of the groups table
+-- by counting the actual number of non-deleted memberships for each group.
+
+-- Update the member_count in groups table based on a count of active memberships
+UPDATE "groups" g
+SET "member_count" = (
+    SELECT COUNT(*)
+    FROM "group_memberships" gm
+    WHERE gm."group_id" = g."id"
+      AND gm."deleted_at" IS NULL
+)
+WHERE 1=1; -- Apply to all groups
--- a/app/controlplane/pkg/data/group.go
+++ b/app/controlplane/pkg/data/group.go
@@ -20,13 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/data/ent/projectversion"
-
-	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/data/ent/project"
-
-	"entgo.io/ent/dialect/sql/sqljson"
-
-	"entgo.io/ent/dialect/sql"
 	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/authz"
 	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/biz"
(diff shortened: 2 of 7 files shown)


app/controlplane/pkg/data/group.go
```diff
@@ -231,13 +228,12 @@ func (g GroupRepo) Create(ctx context.Context, orgID uuid.UUID, opts *biz.Create
 			SetDescription(opts.Description).
 			SetOrganizationID(orgID)
 
-		// Only set the member count and add member if userID is provided
+		// Set initial member count to 0, we'll update it after transaction
+		builder = builder.SetMemberCount(0)
```
migmartri: why isn't this a defalt in the DB?

app/controlplane/pkg/data/membership.go
```diff
@@ -279,6 +284,21 @@ func (r *MembershipRepo) Delete(ctx context.Context, id uuid.UUID) error {
 			// Remove the user from all groups in the organization by soft-deleting group memberships
 			now := time.Now()
 
+			// Find all group IDs where this user is a member in this organization
+			groupMemberships, grpMemErr := tx.GroupMembership.Query().Where(
```
migmartri, line 288: what's this change about?

app/controlplane/pkg/data/group.go
javirln: It's the default, it was a just to make it explicit. I can remove it if it causes confusion

app/controlplane/pkg/data/membership.go
javirln: After a user is removed from an organization, we need to update the member count for all groups they were part of. This change handles that by retrieving all associated groups and recalculating their member counts

migmartri: but I don't understand who changing from setting -1 to recalculating adds this amount of logic. Or is it that we didn't do it properly before?

migmartri:
please add a migration to fix existing groups, thanks

app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
```diff
@@ -0,0 +1,13 @@
...
+
+-- Update the member_count in groups table based on a count of active memberships
+UPDATE "groups" g
+SET "member_count" = (
+    SELECT COUNT(*)
+    FROM "group_memberships" gm
+    WHERE gm."group_id" = g."id"
+      AND gm."deleted_at" IS NULL
+)
+WHERE 1=1; -- Apply to all groups
```
migmartri: to non-deleted groups?

javirln: No, that means count all memberships that are not marked as deleted for the group, aka, active memberships

migmartri: but should we update only non-deleted groups?

app/controlplane/pkg/data/membership.go
javirln: The previous logic was missing a condition where it was decrementing all memberships even if it was marked as deleted. I've reverted the changes and used the old logic to take into account that part.

app/controlplane/pkg/data/membership.go
```diff
@@ -289,12 +289,12 @@ func (r *MembershipRepo) Delete(ctx context.Context, id uuid.UUID) error {
...
 			// Decrement the member count of each group this user was a member of
-			// We don't need a separate check for groups the user was a maintainer of
-			// because all memberships were already deleted above
+			// Only decrement if member_count > 0 to avoid negative values
 			if _, err := tx.Group.Update().
 				Where(
-					group.HasMembersWith(user.ID(userID)),
+					group.HasGroupUsersWith(groupmembership.UserID(userID), groupmembership.DeletedAtIsNil()),
 					group.HasOrganizationWith(organization.ID(orgID)),
+					group.MemberCountGT(0),
```
migmartri: I think this is wrong, it's not guarantee that the computation is correct does it?

migmartri (changes requested):

app/controlplane/pkg/data/ent/migrate/migrations/20250715100956.sql
migmartri: let's remove the where please

PR #2261, opened 2025-07-15
feat(rbac): org viewer cannot become group maintainer

jiparis:
This PR adds an additional validation. Org Viewers are a readonly role, so they cannot be maintainers.


jiparis:
closing as it will be handled client side.

PR #2269, opened 2025-07-17
fix(group): Only org admin/owners and group maintainers can list group members

javirln:
This patch restricts access to the ListMembers endpoint so that only organization owners, admins, or maintainers of the specific group can view its members.


app/controlplane/internal/service/group.go
```diff
@@ -531,6 +550,79 @@ func (g *GroupService) ListProjects(ctx context.Context, req *pb.GroupServiceLis
...
+	return g.userHasPermissionOnGroupMembershipsWithPolicy(ctx, orgID, groupIdentifier, authz.PolicyGroupRemoveMemberships)
+}
+
+// userHasPermissionToListPendingGroupInvitations checks if the user has permission to list pending group invitations
+func (g *GroupService) userHasPermissionToListPendingGroupInvitations(ctx context.Context, orgID string, groupIdentifier *pb.IdentityReference) error {
+	return g.userHasPermissionOnGroupMembershipsWithPolicy(ctx, orgID, groupIdentifier, authz.PolicyGroupListPendingInvitations)
+}
+
+// userHasPermissionToListGroupMember checks if the user has permission to list group members
+func (g *GroupService) userHasPermissionToListGroupMember(ctx context.Context, orgID string, groupIdentifier *pb.IdentityReference) error {
```
javirln, line 569: Moved all these methods from servier to `GroupService` and this is the new one that checks the policy for listing members in a group

app/controlplane/pkg/biz/group.go
```diff
@@ -633,33 +643,10 @@ func (uc *GroupUseCase) RemoveMemberFromGroup(ctx context.Context, orgID uuid.UU
 		return NewErrNotFound("group")
 	}
 
+	// Validate requester permissions only if RequesterID is provided
 	if opts.RequesterID != uuid.Nil {
-		// Check if the requester is part of the organization
```
jiparis, line 637: Was this code duplicated?

app/controlplane/internal/service/group.go
```diff
@@ -233,17 +236,33 @@ func (g *GroupService) ListMembers(ctx context.Context, req *pb.GroupServiceList
...
+	if err != nil {
+		return nil, errors.BadRequest("invalid", "invalid user ID")
+	}
+
 	// Initialize the options for listing members
 	opts := &biz.ListMembersOpts{
 		IdentityReference: &biz.IdentityReference{},
 		Maintainers:       req.Maintainers,
 		MemberEmail:       req.MemberEmail,
+		RequesterID:       requesterUUID,
```
jiparis, line 265: Is this just for testing? or is it needed for any use case?

app/controlplane/pkg/biz/group.go
```diff
@@ -230,6 +233,13 @@ func (uc *GroupUseCase) ListMembers(ctx context.Context, orgID uuid.UUID, opts *
 		return nil, 0, err
 	}
 
+	// Validate requester permissions only if RequesterID is provided
+	if opts.RequesterID != uuid.Nil {
```
migmartri, line 237: what's this requester part? When is it meant to not to exist?

app/controlplane/internal/service/group.go
javirln: In the same way we have `AddMembers` and `RemoveMembers, I pass the requester to check the permissions on biz, only if present and to test it, yes. In this way we allow third party code to call this code without the problem of forcing a requester.

app/controlplane/pkg/biz/group.go
javirln: When the calling this method from a third party code. For example, we have another service calling `AddMembers` and `RemoveMembers` where we don't check the requester because it's a system request.

It's basically here to check permissions a part from the service

app/controlplane/pkg/biz/group.go
javirln: Yes, this code can be found in `validateRequesterPermissions` that's why I removed it. I must have left it here after some refactoring.

javirln:
> Do you think we should also restrict the GroupsService/GetProjects endpoint?

@jiparis was that covered in the permission matrix? We originally had this endpoint, `ListMembers` open for all members because we thought it was ok. I don't think we have talked about the Projects in the scope of groups though

PR #2274, opened 2025-07-18
feat(policy): add policy develop eval

### LINKED ISSUE #1108: CLI command to evaluate policies
it would be useful to have something like `chainloop policy eval` that receives a material and a policy (file, URL) and returns a report with potential violations. I would be used in manual control gates

Piskoo:
Closes #1108


app/cli/internal/policydevel/eval.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+type EvalResult struct {
+	NoPolicies  bool
+	Skipped     bool
+	SkipReasons []string
+	Violations  []string
+}
+
+func Evaluate(opts *EvalOptions, logger zerolog.Logger) (*EvalResult, error) {
```
migmartri, line 46: we also run attestation-type policies. Their evaluation happen on `att push`. Does this cover this case?

Piskoo: Currently only material policies are evaluated, should I add a way choose between material or attestation metadata evaluation?

migmartri: Why do you need to chooose? Wouldn't providing `kind` ATTESTATION would be enough? cc/ @jiparis

Piskoo: ~~Found the problem, I'll make it work that way if ```kind``` ATTESTATION is detected/passed~~

Piskoo: Went through the existing code again, everything should be working for ```ATTESTATION``` in the current implementation.

jiparis: Yes, it works like any other material type, it should be supported already.

app/cli/internal/policydevel/eval.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	result := &EvalResult{
+		NoPolicies:  true,
+		Skipped:     false,
+		SkipReasons: []string{},
+		Violations:  []string{},
+	}
```
javirln: I think the `ignore` flag is missign

jiparis: Yes, @Piskoo there is a "ignore" flag in evaluation results, that is used for the engine to ignore the policy execution (as it never existed). It's used in cases where the decision logic is in the Rego itself. Not too important for this command though.

app/cli/cmd/policy_develop_eval.go
```diff
@@ -0,0 +1,107 @@
...
+				logger.Info().Msg("policy evaluation skipped")
+				for _, reason := range result.SkipReasons {
+					logger.Info().Msgf("%s", reason)
+				}
+
+			case len(result.Violations) > 0:
+				for _, violation := range result.Violations {
+					logger.Error().Msgf("- %s", violation)
+				}
+				logger.Error().Msg("policy evaluation failed")
```
jiparis: I think this should be `Info()` as well. Unless we know what's the user expectation, having Violations is a valid outcome, not an error.

app/cli/internal/policydevel/eval.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, e := range evs {
+		for _, v := range e.Violations {
+			result.Violations = append(result.Violations, fmt.Sprintf("%s: %s", v.Subject, v.Message))
+		}
+	}
+
+	return result, nil
+}
+
+func craftMaterial(materialPath, materialKind string, logger *zerolog.Logger) (*v12.Attestation_Material, error) {
```
jiparis, line 116: I like we are using real crafted materials, instead of a simulation.
I'd create a follow up task to extract this functionality in the CLI and use it everywhere instead of duplicating the logic (like the kind autodetect loop, for example).

app/cli/internal/policydevel/eval.go
Piskoo: It's being handled by ```NoPolicies```, I will rename it for consistency

PR #2277, opened 2025-07-18
feat(policy): add default and user regal config to lint

Piskoo:
This PR adds a default configuration for the Regal linter and introduces a flag to specify custom configuration.

Usage
```
chainloop policy develop lint --policy policy.yaml --regal config.cfg 
```

If no config was passed or config is malformed then default config is used as fallback.

Default config:
```
rules:
  idiomatic:
    directory-package-mismatch:
      level: ignore
    no-defined-entrypoint:
      level: warning
  style:
    todo-comment:
      level: ignore
    line-length:
      level: ignore
    rule-length:
      level: ignore
```

Part of #1108


migmartri:
How does regal work today in terms of config, does it auto-detect it? Maybe we can do the same? i.e if it finds a regal config file uses it? Just thinking

jiparis:
> How does regal work today in terms of config, does it auto-detect it? Maybe we can do the same? i.e if it finds a regal config file uses it? Just thinking

If no file provided via command line, it tries to find a default configuration in .regal and in parent directories. But everything happens in the CLI/cobra setup.
https://docs.styra.com/regal/configuration

I think having the command line option is enough for us, but at some point we could provide the same experience. Not important at all.

javirln:
Also agree the `default configuration + flag to custom config` is more than enough. I doubt we will change it often. Maybe I would rename the flag from `regal` to `regal-config` to make it explicit, but not a big deal.

Piskoo:
> Also agree the `default configuration + flag to custom config` is more than enough. I doubt we will change it often. Maybe I would rename the flag from `regal` to `regal-config` to make it explicit, but not a big deal.

Renamed the flag to ```regal-config```

> > How does regal work today in terms of config, does it auto-detect it? Maybe we can do the same? i.e if it finds a regal config file uses it? Just thinking
> 
> If no file provided via command line, it tries to find a default configuration in .regal and in parent directories. But everything happens in the CLI/cobra setup. https://docs.styra.com/regal/configuration
> 
> I think having the command line option is enough for us, but at some point we could provide the same experience. Not important at all.

I could add this as another config lookup step for example if no config file was provided implicitly via ```regal-config``` flag, then search that attempts to find either the .regal directory or .regal.yaml  would be performed, any only then if no cfg file was discovered we would fallback to the default config.

app/cli/internal/policydevel/.regal.yaml
```diff
@@ -0,0 +1,14 @@
+# Default config
+rules:
```
migmartri, line 3: I'd keep the credits for the original author, i.e spacelift I think?

migmartri (approved):

PR #2278, opened 2025-07-21
feat(events): Record on audit logs when user membership has been changed

javirln:
This patch adds a new event to the User type that records when the membership between the user and an organization has been changed.


jiparis (approved):
Nice! thanks

PR #2281, opened 2025-07-23
feat(rbac): allow role inheritance and additional roles

jiparis:
This PR includes some DB changes:
* a new field for membership to allow role inheritance from hierarchy or resources
* some constraints to: only allow one explicit role, and only allow one inherited role from the same parent


app/controlplane/pkg/data/ent/schema/membership.go
```diff
@@ -56,21 +56,34 @@ func (Membership) Fields() []ent.Field {
...
 
 func (Membership) Edges() []ent.Edge {
 	return []ent.Edge{
-		// Deprecated: use polymorphic membership instead
 		edge.From("organization", Organization.Type).Ref("memberships").Unique(),
-		// Deprecated: use polymorphic membership instead
 		edge.From("user", User.Type).Ref("memberships").Unique(),
+
+		// inheritance
+		edge.To("children", Membership.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
```
migmartri: I believe that ent has a native way of defining this ina. single edge https://entgo.io/docs/schema-edges/#o2m-same-type

jiparis: Nice!, I'll look into it.

migmartri (approved):
Cool

PR #2289, opened 2025-07-25
Add policy name when computing policy arguments

gr0:
This PR adds information regarding policy name when computing policy arguments.

### DIFF
--- a/pkg/policies/policies.go
+++ b/pkg/policies/policies.go
@@ -263,7 +263,7 @@ func ComputeArguments(name string, inputs []*v1.PolicyInput, args map[string]str
 			return input.Name == k
 		})
 		if !expected {
-			logger.Warn().Msgf("argument %q will be ignored by policy %s", k, name)
+			logger.Warn().Msgf("argument %q not defined in policy %q spec, ignoring it", k, name)
 			continue
 		}
 		value, err := templates.ApplyBinding(v, bindings)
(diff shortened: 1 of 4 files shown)


pkg/policies/policies.go
```diff
@@ -140,7 +140,7 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
 		pv.logger.Debug().Msgf("evaluating policy %s against attestation", policy.Metadata.Name)
 	}
 
-	args, err := ComputeArguments(policy.GetSpec().GetInputs(), attachment.GetWith(), opts.bindings, pv.logger)
+	args, err := ComputeArguments(policy.Metadata.Name, policy.GetSpec().GetInputs(), attachment.GetWith(), opts.bindings, pv.logger)
```
migmartri: why here you don't use the getters but above you do? i.e `GetMedata()`

pkg/policies/policies.go
```diff
@@ -263,7 +263,7 @@ func ComputeArguments(inputs []*v1.PolicyInput, args map[string]string, bindings
 			return input.Name == k
 		})
 		if !expected {
-			logger.Warn().Msgf("argument %q will be ignored", k)
+			logger.Warn().Msgf("argument %q will be ignored by policy %s", k, name)
```
migmartri:
```suggestion
			logger.Warn().Msgf("argument %q not defined in policy %q spec, ignoring it", k, name)
```

migmartri (approved):

pkg/policies/policies.go
gr0: Good point, unified.

PR #2290, opened 2025-07-26
feat: option to restrict org creation

### LINKED ISSUE #2270: configure who can create orgnanizations

migmartri:
This PR adds a new setting in the controlplane config to restrict the creation of orgs to instance administrators, it includes

- configuration setting
- new role `org:instance:admin` associated with a new org creation policy
- enforcement logic (if the setting is enabled) in the org creation service handler
- exposure of this feature being enabled in the infoz endpoint.

This is the error you get if you try to create an org and it's disabled and you are not admin

```
ERR creation of organizations is restricted to instance admins
exit status 1
```

Closes #2270


app/controlplane/internal/service/organization.go
```diff
@@ -158,3 +169,23 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
...
 }
+
+func (s *OrganizationService) canCreateOrganization(ctx context.Context) (bool, error) {
+	// Restricted org creation is disabled, allow creation
+	if !s.enforcer.RestrictOrgCreation {
+		return true, nil
+	}
+
+	m := entities.CurrentMembership(ctx)
+	for _, rm := range m.Resources {
```
jiparis, line 180: Maybe here you can restrict to memberships with `ResourceType == "instance"` just to avoid unnecessary enforcer calls.

jiparis (approved):
Nice, just a nit comment.

app/controlplane/internal/service/organization.go
migmartri: done, thanks

PR #2299, opened 2025-07-30
feat(invitations): Expand invitation context to external metadata

javirln:
This pull request introduces changes to improve the handling of organization invitations, particularly in the context of joining groups and projects. The updates include modifying the `OrgInvitationContext` structure to use pointers for UUID fields, adding support for external metadata, and enhancing test coverage to handle edge cases like nil UUIDs.

Before, since the groupID and projectID were not pointers, they were being stored in the DB as their nil representation of UUUID the `0000-0000...` by adding a pointer, the are not store at all. The UUID != uuid.Nil check is there for already stored data.

### DIFF
--- a/app/controlplane/pkg/biz/orginvitation.go
+++ b/app/controlplane/pkg/biz/orginvitation.go
@@ -59,11 +59,11 @@ type OrgInvitation struct {
 // OrgInvitationContext is used to pass additional context when accepting an invitation
 type OrgInvitationContext struct {
 	// GroupIDToJoin is the ID of the group to join when accepting the invitation
-	GroupIDToJoin uuid.UUID `json:"group_id_to_join,omitempty"`
+	GroupIDToJoin *uuid.UUID `json:"group_id_to_join,omitempty"`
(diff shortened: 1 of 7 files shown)


app/controlplane/pkg/biz/orginvitation.go
```diff
@@ -338,20 +341,20 @@ func (uc *OrgInvitationUseCase) FindByID(ctx context.Context, invitationID strin
 // processGroupMembership adds a user to a group if the invitation context contains a group to join
 func (uc *OrgInvitationUseCase) processGroupMembership(ctx context.Context, invitation *OrgInvitation, orgUUID uuid.UUID, userUUID uuid.UUID) error {
 	// Skip if there's no group to join in the invitation context
-	if invitation.Context == nil || invitation.Context.GroupIDToJoin == uuid.Nil {
+	if invitation.Context == nil || invitation.Context.GroupIDToJoin == nil || *invitation.Context.GroupIDToJoin == uuid.Nil {
```
jiparis, line 344: much better!

jiparis (approved):

PR #2300, opened 2025-07-31
chore(nats): Send audit logs events to specific subjects

javirln:
This pull request refines the `AuditLogPublisher` in `nats.go` to improve the flexibility and granularity of audit log publishing. The changes introduce a new subject naming convention and update the publishing logic to dynamically construct specific subject names based on event types.

### Enhancements to audit log publishing:

* **Introduced `baseSubjectName` constant**: Added a new constant, `baseSubjectName`, to define the base subject for audit log publishing. This enables a more structured subject naming pattern for publishers.
* **Dynamic subject construction in `Publish` method**: Updated the `Publish` method to dynamically construct specific subject names using the format `audit.<target_type>.<action_type>`. This allows events to be published to more granular subjects based on their type and action.

### DIFF
--- a/app/controlplane/pkg/auditor/nats.go
+++ b/app/controlplane/pkg/auditor/nats.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	streamName = "chainloop-audit"
-	// subjectName is the base subject for audit logs for the consumer to subscribe to.
+	// subjectName is the base subject for the stream to listen to.
 	subjectName = "audit.>"
 	// baseSubjectName is the base subject for audit logs for the publisher to publish to.
 	// The pattern for the specific subjects is "audit.<target_type>.<action_type>"


app/controlplane/pkg/auditor/nats.go
```diff
@@ -27,8 +28,12 @@ import (
 )
 
 const (
-	streamName  = "chainloop-audit"
+	streamName = "chainloop-audit"
+	// subjectName is the base subject for audit logs for the consumer to subscribe to.
```
jiparis: Is `subjectName` used yet?

javirln: I will check if it's being used by another service. I think I saw something that's why I left it but I'll double check it

migmartri:
Is this backwards compatible with existing consumers or do we need to create new ones?

javirln:
> Is this backwards compatible with existing consumers or do we need to create new ones?

We have tested it and all consumers we have are subscribed to `audit.>` so they are still receiving events!

migmartri:
Thanks Javi

app/controlplane/pkg/auditor/nats.go
jiparis: It's used during stream creation. Not sure if it's a mandatory argument, but we can keep it anyways.

javirln: I’ve checked, and we do need the `subjectName` here. This function creates or updates a stream when Chainloop starts. If no subject is specified, the default will be the stream name, as explained [here](https://docs.nats.io/nats-concepts/jetstream/streams#subjects). This would cause issues with other services that rely on specific subjects.

```golang
if _, err := js.CreateOrUpdateStream(ctx, jetstream.StreamConfig{
    Name: "chainloop-audit",
    // No Subjects specified
}); err != nil {
    return nil, fmt.Errorf("failed to create stream: %w", err)
}
```
The stream would only listen to messages published to the subject `chainloop-audit` (same as the stream name), NOT to all subjects.

PR #2304, opened 2025-08-04
fix(groups): remove Group->user edge

jiparis:
This PR removes the Group<->User edge, since there is no way to navigate the edge with DeletedAt=nil condition. By removing the edge, using the join table is required and we can specify the condition clearly.

This issue was making the `List` repository method to return deleted items, causing additional exceptions.


app/controlplane/pkg/data/group.go
```diff
@@ -72,8 +72,20 @@ func (g GroupRepo) List(ctx context.Context, orgID uuid.UUID, filterOpts *biz.Li
 		predicates = append(predicates, group.DescriptionContains(filterOpts.Description))
 	}
 
+	if filterOpts.UserID != nil {
```
migmartri: I'd not make this part of the `OR` clause, but instead part of the base query if provided

app/controlplane/pkg/data/group.go
```diff
@@ -243,17 +248,12 @@ func (g GroupRepo) Create(ctx context.Context, orgID uuid.UUID, opts *biz.Create
...
 		// Only add memberships if userID is not nil
 		if opts.UserID != nil {
-			// Update the group-user member to set it's a group maintainer
-			if _, grUerr := tx.GroupMembership.Update().
-				Where(
-					groupmembership.GroupIDEQ(gr.ID),
-					groupmembership.UserIDEQ(*opts.UserID),
-				).
+			// Set user as group maintainer
+			if _, grUerr := tx.GroupMembership.Create().
```
migmartri: why of this difference? Update vs Create?

app/controlplane/pkg/data/group.go
jiparis: I don't mind changing it, but that changes the method semantics.

app/controlplane/pkg/data/group.go
jiparis: The previous approach used the Create query to directly create the related membership items, and then editing it to update the maintainer role. This approach is not possible after removing the `users` edge.

app/controlplane/pkg/data/group.go
migmartri: ok, I'll leave it to you, to me is about making sure we do not leak any personal information by mistake, and userID to me is not part of a regular filter I think.

migmartri (approved):

PR #2305, opened 2025-08-05
feat(policy): allow generic material kinds

Piskoo:
Allows policy definition for ```ARTIFACT, STRING, EVIDENCE``` kinds and fixes issue with policy devel eval where  these kinds could not be tested before.
It should be noted in custom policy documentation that those kinds are now valid.


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -280,15 +280,8 @@ message PolicySpecV2 {
   }
 
   // if set, it will match any material supported by Chainloop
-  // except those not having a direct schema (STRING, ARTIFACT, EVIDENCE), since their format cannot be guessed by the crafter.
```
migmartri, line 283: I am not sure if ARTIFACT should be also permitted since ARTIFACT is at the same level but opposite than EVIDENCE.

migmartri (approved):

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
Piskoo: We want to shift responsibility of handling ```EVIDENCE``` to the user because there's no way to verify schema at this point, but at the same time there's no way to validate what the ```ARTIFACT``` is, in that way they are similar.

migmartri: @jiparis just give it a review, I am ok either way, thanks

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -280,15 +280,8 @@ message PolicySpecV2 {
...
-  // CONTAINER, HELM_CHART are also excluded, but we might implement custom policies for them in the future.
-  CraftingSchema.Material.MaterialType kind = 3 [(buf.validate.field).enum = {
-    not_in: [
-      1,
-      3,
-      11
-    ]
-  }];
+  // CONTAINER, HELM_CHART are excluded, but we might implement custom policies for them in the future.
```
jiparis: This comment is not valid anymore. You can remove it (we support policies for both)

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
jiparis: This can be a long discussion. I think that we are giving Evidence and Artifact a semantic that is orthogonal to material kinds. For example, following that convention, a CycloneDX material can be and Evidence at the same time.

Since we are not yet in that stage of the discussion (but I think we will soon), I would just allow EVIDENCE at this point (and maybe STRING), clearly stating in the docs that Artifact is not supported because it's usually an arbitrary format (a binary, a tar.gz ...), whereas Evidences are only supported if they have a JSON format.

jiparis (approved):
Thanks @Piskoo . Please take a look at the comments.

PR #2306, opened 2025-08-05
feat(policy): allow arguments with comma character

### LINKED ISSUE #1979: allow policy arguments with comma character
Current behaviour interprets them as CSV, converting them to multivalued arguments. 

There should be a way to escape the `,` and skip the CSV conversion. For example:
```yaml
policies:
  materials:
    - ref: my-policy
      with:
        foo: bar\, baz
```

Piskoo:
This PR adds a way to escape ```,``` character with ```\``` in policy arguments.

Closes #1979

### DIFF
--- a/pkg/policies/policies.go
+++ b/pkg/policies/policies.go
@@ -393,8 +393,8 @@ func getInputArguments(inputs map[string]string) map[string]any {
 			args[k] = value
 		}
 
-		// Single string, let's check for CSV
-		lines = strings.Split(s, ",")
+		// Single string, let's check for CSV and escaped commas `\,`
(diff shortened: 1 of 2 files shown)


migmartri:
thanks @Piskoo, in which cases do we need scaped commas?

Piskoo:
> thanks @Piskoo, in which cases do we need scaped commas?

 There might be a specific case where commas are included in the value fields in materials, but I'm not confident what the exact use case was @jiparis.

jiparis:
We don't have any case in our builtin policies, but this might be useful for custom policies.

PR #2308, opened 2025-08-05
feat(policy): change devel commands return to stdout

Piskoo:
This PR changes policy evaluation and linting output to support proper redirection (e.g., ```chainloop policy devel eval {flags} > results.json```).

### DIFF
--- a/app/cli/cmd/output.go
+++ b/app/cli/cmd/output.go
@@ -54,7 +54,8 @@ type tabulatedData interface {
 		*action.APITokenItem |
 		[]*action.APITokenItem |
 		*action.AttestationStatusMaterial |
-		*action.ListMembershipResult
+		*action.ListMembershipResult |
+		*action.PolicyEvalResult
 }
 
 var ErrOutputFormatNotImplemented = errors.New("format not implemented")
--- a/app/cli/cmd/policy_develop_lint.go
+++ b/app/cli/cmd/policy_develop_lint.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/chainloop-dev/chainloop/app/cli/internal/action"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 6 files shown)


app/cli/cmd/policy_develop_lint.go
```diff
@@ -68,3 +65,17 @@ func newPolicyDevelopLintCmd() *cobra.Command {
 	cmd.Flags().StringVar(&regalConfig, "regal-config", "", "Path to custom regal config (Default: https://github.com/chainloop-dev/chainloop/tree/main/app/cli/internal/policydevel/.regal.yaml)")
 	return cmd
 }
+
+func encodeResults(result *action.PolicyLintResult) error {
+	if result == nil {
+		return nil
+	}
+
+	fmt.Fprintf(os.Stdout, "Found %d issues:\n", len(result.Errors))
```
migmartri: fmt.Sprintf

migmartri (changes requested):

app/cli/cmd/output.go
```diff
@@ -54,7 +54,8 @@ type tabulatedData interface {
 		*action.APITokenItem |
 		[]*action.APITokenItem |
 		*action.AttestationStatusMaterial |
-		*action.ListMembershipResult
+		*action.ListMembershipResult |
+		*action.PolicyEvalResult
```
migmartri, line 58: does this mean that now we support json output?

Piskoo: Correct, the text version didn't follow any structure so it would cause problems later on

migmartri (approved):

app/cli/cmd/policy_develop_lint.go
```diff
@@ -55,11 +55,7 @@ func newPolicyDevelopLintCmd() *cobra.Command {
 				return nil
 			}
 
-			// Print all validation errors
-			for _, err := range result.Errors {
-				logger.Error().Msg(err)
-			}
-			return fmt.Errorf("policy validation failed with %d issues", len(result.Errors))
+			return encodeResult(result)
```
migmartri, line 58: what's the output when we do not have any violations? Do we return an empty violations array?

Piskoo: Yes, here's the output
```
[
   {
      "violations": [],
      "skip_reasons": [],
      "skipped": false
   }
]
```

migmartri: awesome, please merge :)

PR #2312, opened 2025-08-06
fix(policy): add referenced script traversal

### LINKED ISSUE #2310: Confusing linting behavior
pointing to a specific new policy says
```
chainloop policy devel lint -p ./chainloop-policy.yaml
INF policy is valid!
```

while the policy in fact is invalid if I don't explicitly pass the policy

```
chainloop policy devel lint
Found 1 issues:
  1. /tmp/test-policy/chainloop-policy.rego:1: File should be formatted with `opa fmt`
ERR policy validation failed with 1 issues
```

To me, both commands should return the same, there is only one policy in my directory.

Piskoo:
This PR changes the way policy file is linted. 
- when *.yaml file is passed all referenced *.rego files will also be linted. 
- when directory is passed or default (cwd) is used it looks for a ```policy.yaml``` file, if not found linting fails. (before it would lint all valid *.yaml and *.rego files)

Closes #2310

### DIFF
--- a/app/cli/cmd/policy_develop_lint.go
+++ b/app/cli/cmd/policy_develop_lint.go
@@ -59,7 +59,7 @@ func newPolicyDevelopLintCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&policyPath, "policy", "p", ".", "Path to policy directory")
+	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to policy file")
 	cmd.Flags().BoolVar(&format, "format", false, "Auto-format file with opa fmt")
 	cmd.Flags().StringVar(&regalConfig, "regal-config", "", "Path to custom regal config (Default: https://github.com/chainloop-dev/chainloop/tree/main/app/cli/internal/policydevel/.regal.yaml)")
 	return cmd
--- a/app/cli/internal/policydevel/lint.go
+++ b/app/cli/internal/policydevel/lint.go
@@ -109,7 +109,7 @@ func Lookup(absPath, config string, format bool) (*PolicyToLint, error) {
 	}
 
 	// Load referenced rego files from all YAML files
-	if err := policy.loadReferencedRegoFiles(); err != nil {
+	if err := policy.loadReferencedRegoFiles(filepath.Dir(resolvedPath)); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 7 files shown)


app/cli/internal/policydevel/lint.go
```diff
@@ -96,15 +96,25 @@ func Lookup(absPath, config string, format bool) (*PolicyToLint, error) {
 	}
 
 	if fileInfo.IsDir() {
-		if err := scanDirectory(policy, absPath); err != nil {
-			return nil, err
+		// If it's a directory, look for policy.yaml
```
migmartri: I'd remove the option of passing a dir. You always pass a file path with a default

migmartri: what I mean that it looks strange that the default is a directory when we could make it a file path relative to cwd?

I believe we might have some similar code in the CLI that handles this case, default with relative path or explicit absolute path. Golang might have a helper for that too.

app/cli/internal/policydevel/lint.go
```diff
@@ -113,31 +121,32 @@ func Lookup(absPath, config string, format bool) (*PolicyToLint, error) {
...
-	files, err := os.ReadDir(dirPath)
-	if err != nil {
-		return fmt.Errorf("reading directory: %w", err)
-	}
-
-	var foundValidFile bool
-	for _, file := range files {
-		if file.IsDir() {
+// Loads referenced rego files from all YAML files in the policy
+func loadReferencedRegoFiles(policy *PolicyToLint) error {
```
jiparis: very nit: the argument could be a receiver instead `func (policy *PolicyToLint) loadReferenceRegoFiles() {`

jiparis (approved):

app/cli/internal/policydevel/lint.go
```diff
@@ -113,35 +121,36 @@ func Lookup(absPath, config string, format bool) (*PolicyToLint, error) {
...
-	if err != nil {
-		return fmt.Errorf("reading directory: %w", err)
-	}
-
-	var foundValidFile bool
-	for _, file := range files {
-		if file.IsDir() {
+// Loads referenced rego files from all YAML files in the policy
+func (p *PolicyToLint) loadReferencedRegoFiles() error {
+	seen := make(map[string]struct{})
```
migmartri, line 127: are duplicates a common thing? I do not remember a case where we had two identical rego files in two different paths

app/cli/cmd/policy_develop_lint.go
```diff
@@ -59,7 +59,7 @@ func newPolicyDevelopLintCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&policyPath, "policy", "p", ".", "Path to policy directory")
+	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to policy file")
```
migmartri, line 62: should we do the same approach in eval?

Also, should we change init so the default created file is `policy.yaml`?

migmartri: actually `eval` works like this it seems, it doesn't load a default though. 

the `init` might need update too

Piskoo: I thought about it, I'll update it in this PR, although some docs changes will be needed as well

Issue #2267, opened 2025-07-16
feat: allowed domains for `http.send` in Rego engine


jiparis:
Currently, [http.send](https://www.openpolicyagent.org/docs/policy-reference/builtins/http) domains in Rego engine are locked down for security reasons. We might want to make it configurable. A way would be a server side setting that gets downloaded on `attestation init`.

- [X] backend
- [ ] Docs
- [ ] #2335

migmartri:
Agreed, this could be a instance configuration setting

PR #2314, opened 2025-08-06
feat: allow configuring rego engine

migmartri:
Prepares the rego engine code to support customization, including additional domains 

refs #2267

### DIFF
--- a/pkg/policies/engine/rego/rego.go
+++ b/pkg/policies/engine/rego/rego.go
@@ -56,12 +56,6 @@ type newEngineOptions struct {
 	allowedNetworkDomains []string
 }
 
-func WithBaseAllowedNetworkDomains(domains ...string) EngineOption {
-	return func(e *newEngineOptions) {
-		e.allowedNetworkDomains = domains
-	}
-}
-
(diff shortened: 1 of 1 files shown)


pkg/policies/engine/rego/rego.go
```diff
@@ -27,11 +27,64 @@ import (
...
+		e.allowedNetworkDomains = domains
+	}
+}
+
+type newEngineOptions struct {
+	operatingMode         EnvironmentMode
+	allowedNetworkDomains []string
+}
+
+func WithBaseAllowedNetworkDomains(domains ...string) EngineOption {
```
jiparis: Is this one duplicated?

migmartri:
Once we merge this, we might want to add support of this configuration option in the `eval` command cc/ @Piskoo 

This is going to be used during the attestation process once #2315 is merged, but the `eval` command should support it too, even if it's explicit as flags

javirln (approved):

pkg/policies/engine/rego/rego.go
migmartri: Yeah, thanks cursor 😮‍💨

PR #2315, opened 2025-08-06
feat: support storage and use of custom rego engine hostnames

### LINKED ISSUE #2267: feat: allowed domains for `http.send` in Rego engine

migmartri:
This PR allows to store, expose and use custom hostnames that can be used inside policies. If provided, they will be appended to the existing ones.

This PR includes

- Org update API and CLI now to set the allowedHostnames
- It's exposed in the `org describe` and `att init` response.
- The result is stored in the crafting state and injected in the rego engine 

closes #2267


app/cli/internal/action/attestation_init.go
```diff
@@ -197,14 +198,21 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 			return "", err
 		}
 
-		workflowRun := runResp.GetResult().GetWorkflowRun()
+		result := runResp.GetResult()
```
migmartri, line 201: some minor refactoring

app/cli/internal/action/action.go
```diff
@@ -81,7 +81,7 @@ func newCrafter(stateOpts *newCrafterStateOpts, conn *grpc.ClientConn, opts ...c
 			attestationStatePath = path
 		}
 
-		c.Logger.Debug().Str("path", attestationStatePath).Msg("using local state")
+		c.Logger.Debug().Str("path", fmt.Sprintf("file:%s", attestationStatePath)).Msg("using local state")
```
migmartri, line 84: this is a fix in the log entry

app/cli/internal/action/attestation_init.go
```diff
@@ -197,14 +198,21 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
...
-		timestampAuthorityURL = runResp.GetResult().GetSigningOptions().GetTimestampAuthorityUrl()
-		signingCAName = runResp.GetResult().GetSigningOptions().GetSigningCa()
-		if v := workflowMeta.Version; v != nil {
-			workflowMeta.Version.Prerelease = runResp.GetResult().GetWorkflowRun().Version.GetPrerelease()
+		workflowMeta.Organization = result.GetOrganization()
+		blockOnPolicyViolation = result.GetBlockOnPolicyViolation()
+		policiesAllowedHostnames = result.GetPoliciesAllowedHostnames()
+
+		signingOpts := result.GetSigningOptions()
+		if signingOpts != nil {
```
jiparis: This change is actually not needed, right? proto getters are always safe and work for `nil` receivers, so it's safe to do `GetSigningOptions().GetTimestampAuthorityUrl()` even if SigningOptions is `nil`

app/cli/internal/action/membership_list.go
```diff
@@ -32,6 +32,7 @@ type OrgItem struct {
 	ID, Name                        string
 	CreatedAt                       *time.Time
 	PolicyViolationBlockingStrategy string
+	PolicyAllowedHostnames          []string `json:"policyAllowedHostnames,omitempty"`
```
jiparis, line 35: accidental annotation?

app/cli/internal/action/org_update.go
```diff
@@ -30,14 +30,24 @@ func NewOrgUpdate(cfg *ActionsOpts) *OrgUpdate {
 }
 
 type NewOrgUpdateOpts struct {
-	BlockOnPolicyViolation *bool
+	BlockOnPolicyViolation   *bool
+	PoliciesAllowedHostnames *[]string
```
jiparis, line 34: This looks a bit weird. Slices are pointers already and are `nil` if no initialized. No need to create a pointer to a pointer. Just keep in mind that `len([]string(nil))` is 0

app/cli/internal/action/org_update.go
```diff
@@ -30,14 +30,24 @@ func NewOrgUpdate(cfg *ActionsOpts) *OrgUpdate {
...
-		Name: name, BlockOnPolicyViolation: opts.BlockOnPolicyViolation,
-	})
+
+	payload := &pb.OrganizationServiceUpdateRequest{
+		Name:                   name,
+		BlockOnPolicyViolation: opts.BlockOnPolicyViolation,
+	}
+
+	if opts.PoliciesAllowedHostnames != nil {
+		payload.PoliciesAllowedHostnames = *opts.PoliciesAllowedHostnames
```
jiparis, line 46: This is what I mean, `*variable` shouldn't be needed for slices, since they are pointers anyways.

app/cli/internal/action/attestation_init.go
migmartri: let me double check it

app/cli/internal/action/membership_list.go
migmartri: nope really, I wanted to hide this option if the default is set

app/cli/internal/action/org_update.go
migmartri: checking, thanks

migmartri: updated the pointer situation to leverage instead the duality of empty slice vs nil slice.

app/cli/internal/action/org_update.go
migmartri: this is about making sure we can detect when we want to update the value vs empty it. In any case I've implemented the same by using slices

PR #2316, opened 2025-08-07
feat(policy): update output handling for lint and eval

### LINKED ISSUE #2283: feat(policy): update output handling for lint and eval
Update how eval error are split when returned from regal. Allow output flag in lint.

Piskoo:
This PR makes some changes to how output from policy devel tools is displayed:
- lint works with the output flag while maintaining exit errors
- indentation in yaml policy files uses 2 spaces after it was formated with ```--format``` flag
- eval violations returned from regal are now split correctly and displayed one by one
- eval violation concerning file formatting points to ```--format``` flag instead of ```opa fmt```

Closes #2283

### DIFF
--- a/app/cli/internal/policydevel/lint.go
+++ b/app/cli/internal/policydevel/lint.go
@@ -29,7 +29,7 @@ import (
 	v1 "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
 	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/unmarshal"
 	"github.com/chainloop-dev/chainloop/pkg/resourceloader"
-	opaAst "github.com/open-policy-agent/opa/ast"
+	opaAst "github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/format"
 	"github.com/styrainc/regal/pkg/config"
 	"github.com/styrainc/regal/pkg/linter"


migmartri:
Is this solving by any chance #2320 ?

app/cli/internal/policydevel/lint.go
```diff
@@ -192,10 +193,14 @@ func (p *PolicyToLint) validateYAMLFile(file *File) {
 		outYAML, err := protoyaml.Marshal(&policy)
 		if err != nil {
 			p.AddError(file.Path, fmt.Sprintf("failed to marshal updated YAML: %v", err), 0)
-		} else if err := os.WriteFile(file.Path, outYAML, 0600); err != nil {
-			p.AddError(file.Path, fmt.Sprintf("failed to save updated file: %v", err), 0)
 		} else {
-			file.Content = outYAML
+			// Force 2-space indentation, protoyaml.Marshal uses 4 spaces by default and lacks
+			outYAML = bytes.ReplaceAll(outYAML, []byte("    "), []byte("  "))
```
jiparis: Isn't there are better way to do this? I've seen protoyaml has an `Indent` option. This should work:
```
outYaml, err := protoyaml.MarshalOptions{Indent:2}.Marshal(&policy)
```

app/cli/internal/policydevel/lint.go
```diff
@@ -354,38 +359,77 @@ func (p *PolicyToLint) loadDefaultConfig() (*config.Config, error) {
...
 	errorStr := rawErr.Error()
+
+	// Replace "opa fmt" with "--format" in error messages
+	errorStr = strings.ReplaceAll(errorStr, "`opa fmt`", "`--format`")
 	// Regex matches file path, line number and error message like: /path/file:line: message
-	errorRegex := regexp.MustCompile(`^` + regexp.QuoteMeta(path) + `:(\d+):\s*(.+)$`)
+	re := regexp.MustCompile(`^(.*\.rego):(\d+):\s*(.+)$`)
+
+	// Handle the special case: "1 error occurred:"
+	if strings.Contains(errorStr, "error occurred:") && !strings.Contains(errorStr, "\n") {
```
jiparis: Can you add a comment with an example of what this does?

jiparis (changes requested):
Thanks, please check my comments.

app/cli/internal/policydevel/lint.go
Piskoo: Thanks!

app/cli/internal/policydevel/lint.go
```diff
@@ -281,7 +283,7 @@ func (p *PolicyToLint) checkResultStructure(content, path string, keys []string)
 func (p *PolicyToLint) runRegalLinter(filePath, content string) {
 	inputModules, err := rules.InputFromText(filePath, content)
 	if err != nil {
-		p.AddError(filePath, fmt.Sprintf("failed to prepare for linting: %v", err), 0)
+		p.processRegalViolation(fmt.Errorf("%s: %w", filePath, err), filePath)
```
jiparis: Here, the call is creating a custom error just to format a string with `filePath` and `err`. Later, in the function, it just gets the string with `error.Error()` and parses it. Why don't just pass `filePath` and `err` as arguments directly to the method? Why `fmt.Error` is needed?
```
p.processRegalViolation(err, filePath)
```

Piskoo: Fair point, it is being used again for violations and the message passed there follows different format, but wrapping it into an error just to unwrap it instantly is unnecessary, fixed, thanks

app/cli/internal/policydevel/lint.go
```diff
@@ -354,38 +356,77 @@ func (p *PolicyToLint) loadDefaultConfig() (*config.Config, error) {
 	return &cfg, nil
 }
 
-// Splits grouped errors into individual errors
-func (p *PolicyToLint) processRegalViolation(rawErr error, path string) {
-	if rawErr == nil {
+// Formats regal violation errors
+func (p *PolicyToLint) processRegalViolation(errorStr, path string) {
```
jiparis: I've been tying this locally, and I've found that you already have all the details in the `err` object coming from opa. Check this out https://github.com/open-policy-agent/opa/blob/main/v1/ast/errors.go, which is the type of the `err` object, at least the one returned by `InputFromText`.. You can iterate through all errors and get what you need without building nor parsing any string.

jiparis: And for `report.Violations`, everything is already there as well. I'd say that we shouldn't parse a string representation of an error, if the structured data is already there. In this case, we can just iterate throught Violations and add them directly with `p.AddError`.

jiparis:
Thanks @Piskoo I've been testing the latest changes and everything looks good now. Much cleaner, thanks!!

PR #2324, opened 2025-08-07
feat(events): send user and token name as actorName

migmartri:


jiparis:
Why is this needed?

migmartri:
> Why is this needed?

to improve the audit logs with information about the user that did something, not just the email

jiparis (approved):

Issue #2332, opened 2025-08-08
improve lint error message


migmartri:
```
chainloop policy develop lint
ERR linting failed: loading policy: failed to resolve policy file: loading resource: loading URL: unrecognized scheme: policy.yaml
```

this error is not user friendly

migmartri:
and `eval` error message

```
ERR evaluating policy: failed to execute policy : failed to evaluate policy: test-http:36: eval_builtin_error: http.send: unallowed host: app.chainloop.dev
```

I'd say that the whole `ERR evaluating policy: failed to execute policy : failed to evaluate policy:` is not needed.

migmartri:
See for example

```
 chainloop wf contract update -f asdasd.yaml --name foo
ERR open asdasd.yaml: no such file or directory
```

PR #2341, opened 2025-08-11
feat(policy): policy eval should fail if no execution path is met

### LINKED ISSUE #2334: policy eval should fail if no execution path is met
For example, if I have this policy

```
kind: Policy
metadata:
    name: test-http
    description: HTTP policy that validates Chainloop platform version against expected version 1.2.3
spec:
    policies:
        - path: test-http.rego
          kind: SBOM_CYCLONEDX_JSON
```

and I run it like this

```
chainloop policy develop eval --kind EVIDENCE --material ./empty.json -p test-http.yaml
```

I get 

```
chainloop policy develop eval --kind EVIDENCE --material ./empty.json -p test-http.yaml
[
   {
      "violations": [],
      "skip_reasons": [],
      "skipped": false,
      "ignored": true
   }
]
```

with no indication if it was executed or not, in fact it was not.

As a minimum I'd make the policy fail indicating that no execution branch matched the criteria, and ideally we should show logging information that tells the user which path it was taken, even the script executed, 

cc/ @jiparis

Piskoo:
This PR adds error when no evaluations on provided material were performed.

## Example
```
kind: Policy
metadata:
  name: policy
  description: Chainloop validation policy
spec:
  policies:
    - embedded: |
        package main

        import rego.v1

        ################################
        # Common section do NOT change #
        ################################

        result := {
          "skipped": skipped,
          "violations": violations,
          "skip_reason": skip_reason,
          "ignore": ignore,
        }

        default skip_reason := ""

        skip_reason := m if {
          not valid_input
          m := "invalid input"
        }

        default skipped := true

        skipped := false if valid_input

        default ignore := false

        ########################################
        # EO Common section, custom code below #
        ########################################
        # Validates if the input is valid and can be understood by this policy
        valid_input := true

        # insert code here

        # If the input is valid, check for any policy violation here
        # default violations := []

        violations contains msg if {
          valid_input
          msg := "test"
        }
      kind: EVIDENCE
    - embedded: |
        package main

        import rego.v1

        ################################
        # Common section do NOT change #
        ################################

        result := {
          "skipped": skipped,
          "violations": violations,
          "skip_reason": skip_reason,
          "ignore": ignore,
        }

        default skip_reason := ""

        skip_reason := m if {
          not valid_input
          m := "invalid input"
        }

        default skipped := true

        skipped := false if valid_input

        default ignore := false

        ########################################
        # EO Common section, custom code below #
        ########################################
        # Validates if the input is valid and can be understood by this policy
        valid_input := true

        # insert code here

        # If the input is valid, check for any policy violation here
        # default violations := []

        violations contains msg if {
          valid_input
          msg := "test"
        }
      kind: STRING
```
For policy with no defined path for kind `SBOM_CYCLONEDX_JSON` when validated against valid cyclonedx json file it outputs:
```
ERR evaluating policy: no execution branch matched for kind SBOM_CYCLONEDX_JSON
exit status 1
```

Closes #2334

### DIFF
--- a/app/cli/cmd/policy_develop_eval.go
+++ b/app/cli/cmd/policy_develop_eval.go
@@ -63,6 +63,13 @@ evaluates the policy against the provided material or attestation.`,
 				return newGracefulError(err)
 			}
 
+			// Check if any result was ignored
+			for _, res := range result {
+				if res.Ignored {
+					return fmt.Errorf("policy evaluation failed: no execution branch matched")
+				}
+			}
+
 			return encodeJSON(result)
 		},
 	}
--- a/app/cli/cmd/policy_develop_eval.go
+++ b/app/cli/cmd/policy_develop_eval.go
@@ -60,14 +60,7 @@ evaluates the policy against the provided material or attestation.`,
 
 			result, err := policyEval.Run()
 			if err != nil {
-				return newGracefulError(err)
-			}
-
-			// Check if any result was ignored
-			for _, res := range result {
-				if res.Ignored {
-					return fmt.Errorf("policy evaluation failed: no execution branch matched")
-				}
+				return err
 			}
 
 			return encodeJSON(result)
--- a/app/cli/internal/policydevel/eval.go
+++ b/app/cli/internal/policydevel/eval.go
@@ -48,13 +48,13 @@ func Evaluate(opts *EvalOptions, logger zerolog.Logger) ([]*EvalResult, error) {
 	// 1. Create crafting schema
 	schema, err := createCraftingSchema(opts.PolicyPath, opts.Inputs)
 	if err != nil {
-		return nil, fmt.Errorf("creating crafting schema: %w", err)
+		return nil, err
 	}
 
 	// 2. Craft material with annotations
 	material, err := craftMaterial(opts.MaterialPath, opts.MaterialKind, &logger)
 	if err != nil {
-		return nil, fmt.Errorf("material crafting: %w", err)
+		return nil, err
 	}
 	material.Annotations = opts.Annotations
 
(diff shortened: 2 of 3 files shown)


app/cli/cmd/policy_develop_eval.go
```diff
@@ -63,6 +63,13 @@ evaluates the policy against the provided material or attestation.`,
 				return newGracefulError(err)
 			}
 
+			// Check if any result was ignored
+			for _, res := range result {
+				if res.Ignored {
```
migmartri: but does this also include if it was ignored on purpose? I mean, would ignore mean only that no script was evaluated?

cc/ @jiparis

migmartri: also do you think we could be more user friendly by including the kind, and the available kinds in the error output for example?

I'd like to make this tool super user friendly that would mean in general\

- removing technical boilerplate #2332 and error wrapping in general
- making errors self-explanatory
- showing useful if necessary

jiparis: Ignored means: The script was evaluated, but it must be ignored because it doesn't apply, so it will count as it if wasn't evaluated at all. 
Note that this is different to `skipped`, where the evaluation still counts.
Ignored is for some cases where an input file has multiple purposes (CycloneDX for SBOMs and also for vulnerabilities report, for example)

So in this case, what we want to check whether `result` is empty.

Piskoo: Updated to return error only when no eval results were returned. Included material kind in the error msg. Available kinds require some more work I can create another issue to address that.

migmartri: can you please write the description of the PR with some examples of outputs so we can understand better what it solves? Thanks!

Piskoo: Updated the description

PR #2343, opened 2025-08-11
feat(policy): allow empty kind in policy

### LINKED ISSUE #2337: policies eval, potential behavior change
If I have a policy with a path without a kind, it seems that's not being executed

```
kind: Policy
metadata:
    name: test-http
    description: HTTP policy that validates Chainloop platform version against expected version 1.2.3
spec:
    policies:
        - path: test-http.rego
```

This below means not executed, see https://github.com/chainloop-dev/chainloop/issues/2334
```
chainloop policy develop eval --kind EVIDENCE --material ./empty.json -p test-http.yaml
[
   {
      "violations": [],
      "skip_reasons": [],
      "skipped": false,
      "ignored": true
   }
]
```

If I set the kind, the execution works, 

```
kind: Policy
metadata:
    name: test-http
    description: HTTP policy that validates Chainloop platform version against expected version 1.2.3
spec:
    policies:
        - path: test-http.rego
          kind: EVIDENCE

```

```
# This error is expected, it means that's being executed
chainloop policy develop eval --kind EVIDENCE --material ./empty.json -p test-http.yaml
ERR evaluating policy: failed to execute policy : failed to evaluate policy: test-http:36: eval_builtin_error: http.send: unallowed host: app.chainloop.dev
```

I remember that not setting the `kind` meant any kind, have we changed this?  cc/ @jiparis

Piskoo:
This PR adds material selector handler to policy devel eval. If no kind was specified in the policy then the policy will always be applied to the given material.

Closes #2337

### DIFF
--- a/app/cli/internal/policydevel/eval_test.go
+++ b/app/cli/internal/policydevel/eval_test.go
@@ -184,5 +184,4 @@ func TestEvaluate(t *testing.T) {
 		require.True(t, foundWithoutKind, "expected violation 'without kind'")
 		require.False(t, foundEvidenceKid, "did not expect violation 'evidence kind'")
 	})
-
 }


app/cli/internal/policydevel/eval.go
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/chainloop-dev/chainloop/pkg/attestation/crafter/materials"
 )
 
+const (
+	DefaultMaterialName = "material"
```
migmartri: do you need to expose this constant? 

```suggestion
	defaultMaterialName = "material"
```

migmartri (commented):
How does this behave when the policies have kinds? does it override it?

migmartri:
Can you rebase main please? I just merged your other change to run some tests

migmartri (approved):
just checked this and it seems to work as expected

jiparis (approved):

PR #2348, opened 2025-08-11
feat: add project create event

migmartri:


app/controlplane/configs/config.devel.yaml
```diff
@@ -16,11 +16,11 @@ server:
...
     #   private_key: "../../devel/devkeys/selfsigned/controlplane.key"
 
-#nats_server:
-# uri: nats://0.0.0.0:4222
+# nats_server:
+#   uri: nats://0.0.0.0:4222
 
 # Restrict organization creation to role:instance:admin
-restrict_org_creation: true
+restrict_org_creation: false
```
migmartri, line 23: I am fixing this on purpose

app/controlplane/pkg/biz/auditor.go
```diff
@@ -73,12 +73,14 @@ func (uc *AuditorUseCase) Dispatch(ctx context.Context, entry auditor.LogEntry,
 
 	payload, err := auditor.GenerateAuditEvent(entry, opts...)
 	if err != nil {
+		uc.log.Error("failed to generate audit event", "error", err)
```
migmartri, line 76: and the system was not logging auditing errors

PR #2355, opened 2025-08-13
fix(policy-devel): inputs don't accept multiple values

### LINKED ISSUE #2351: bug(eval): policy inputs don't accept multiple values
I was trying `policy eval` with input `--input "licenses=AGPL,GPL"` trying to pass a multivalued input to the policy, but it's only passing the first one.
I've noticed that it's using `StringSliceVar` which in practice recognizes it as two inputs: `licenses=AGPL` and `GPL`. The code in `parseKeyValue` discards the second one.

Piskoo:
This PR fixes issue with passing arrays in `--input` flag.

### Example
By passing `--input licenses=AGPL,GPL` in the eval command values can now be accessed with `input.args.licenses[0]` and `input.args.licenses[1]` in the rego script.

Closes #2351

### DIFF
--- a/app/cli/cmd/policy_develop_eval.go
+++ b/app/cli/cmd/policy_develop_eval.go
@@ -72,7 +72,7 @@ evaluates the policy against the provided material or attestation.`,
 	cmd.Flags().StringVar(&kind, "kind", "", fmt.Sprintf("Kind of the material: %q", schemaapi.ListAvailableMaterialKind()))
 	cmd.Flags().StringSliceVar(&annotations, "annotation", []string{}, "Key-value pairs of material annotations (key=value)")
 	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to custom policy file")
-	cmd.Flags().StringSliceVar(&inputs, "input", []string{}, "Key-value pairs of policy inputs (key=value)")
+	cmd.Flags().StringArrayVar(&inputs, "input", []string{}, "Key-value pairs of policy inputs (key=value)")
 	cmd.Flags().StringSliceVar(&allowedHostnames, "allowed-hostnames", []string{}, "Additional hostnames allowed for http.send requests in policies")
 
 	return cmd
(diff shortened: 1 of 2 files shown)


app/cli/cmd/policy_develop_eval.go
```diff
@@ -72,7 +72,7 @@ evaluates the policy against the provided material or attestation.`,
 	cmd.Flags().StringVar(&kind, "kind", "", fmt.Sprintf("Kind of the material: %q", schemaapi.ListAvailableMaterialKind()))
 	cmd.Flags().StringSliceVar(&annotations, "annotation", []string{}, "Key-value pairs of material annotations (key=value)")
 	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to custom policy file")
-	cmd.Flags().StringSliceVar(&inputs, "input", []string{}, "Key-value pairs of policy inputs (key=value)")
```
migmartri, line 75: mind explaining why the previous one didn't work? Thanks!

Piskoo: The issue was caused by how Cobra treats commas for `StringSliceVar`. Here Cobra does preprocessing where values are split with `,` into separate strings, where in `StringArrayVar` strings are kept as is.

migmartri: so in that case we should use String no? `StringSliceVar` allows us to do 

```
--input foo=bar --input bar=baz
```

is what I wrote correct?

Piskoo: `StringArrayVar` also allows 
```
--input foo=bar --input bar=baz
```
but also allows 
```
--input foo=bar,baz
```
to be treated as multi value, but has a trade-off of not being able to declare
```
--input foo=bar,bar=baz
```
because currently that becomes
```
[]string{"foo=bar,bar", "baz"}
```
Both `StringSliceVar` and `StringArrayVar` have drawbacks that can be fixed with custom parsing logic

migmartri: right, but my question is, do we need multi input or regular strings are enough? I thought that we were asking to provide comma separated values and the engine was doing smth, I guess that not, that we want to have arrays.

Piskoo: We want multi input, the previous change related to commas was for escaping comma separated values, before there was no way to pass  e.g. 
```
with:
  foo: bar, baz
```
to get 
```
"foo": "bar, baz"
```
now `\,` can be used to get that output. But in this case using `StringArrayVar` makes it so the input is preprocessed by Cobra before it is passed to engine

PR #2357, opened 2025-08-13
feat(policy-devel): extend eval command to show raw evaluation of messages

Piskoo:


Piskoo:
`raw_results` is missing requested `script` and `result` due to limitations of naming rego scripts. data.<package_name> returned by opa eval is used for verification, otherwise the output scripts could be labeled by their filename for non embedded type, but that causes problem for embedded scripts since there's no name attached to them.

app/cli/cmd/policy_develop_eval.go
```diff
@@ -74,6 +76,7 @@ evaluates the policy against the provided material or attestation.`,
 	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to custom policy file")
 	cmd.Flags().StringSliceVar(&inputs, "input", []string{}, "Key-value pairs of policy inputs (key=value)")
 	cmd.Flags().StringSliceVar(&allowedHostnames, "allowed-hostnames", []string{}, "Additional hostnames allowed for http.send requests in policies")
+	cmd.Flags().BoolVarP(&debug, "debug", "", false, "Enable debug/verbose output and logging mode")
```
migmartri: We could probably be more explicit here, currently, the description seems like a generic debug mode, when in reality it does more than that.

app/cli/internal/action/policy_develop_eval.go
```diff
@@ -26,13 +26,15 @@ type PolicyEvalOpts struct {
...
 	AllowedHostnames []string
+	Debug            bool
 }
 
 type PolicyEvalResult struct {
-	Violations  []string `json:"violations"`
-	SkipReasons []string `json:"skip_reasons"`
-	Skipped     bool     `json:"skipped"`
-	Ignored     bool     `json:"ignored,omitempty"`
+	Violations  []string                 `json:"violations"`
```
migmartri: in your example in the description it seems that the json keys where capitalized, aren't they using these annotations?

pkg/policies/engine/rego/rego.go
```diff
@@ -147,39 +147,49 @@ func (r *Engine) Verify(ctx context.Context, policy *engine.Policy, input []byte
...
+	}
+
+	rawData := &engine.RawData{
+		Input:  decodedInput,
+		Output: regoResultSetToRawResults(res),
+	}
+
 	// Try the main rule first
-	if err := executeQuery(mainRule, r.operatingMode == EnvironmentModeRestrictive); err != nil {
+	if err := executeQuery(fmt.Sprintf("%v.%s\n", parsedModule.Package.Path, mainRule), r.operatingMode == EnvironmentModeRestrictive); err != nil {
```
migmartri: `fmt.Sprintf("%v.%s\n", parsedModule.Package.Path, mainRule` looks likke it could be in a function

pkg/policies/engine/rego/rego.go
```diff
@@ -147,39 +147,49 @@ func (r *Engine) Verify(ctx context.Context, policy *engine.Policy, input []byte
...
+		if err := executeQuery(fmt.Sprintf("%v.%s\n", parsedModule.Package.Path, deprecatedRule), r.operatingMode == EnvironmentModeRestrictive); err != nil {
 			return nil, err
 		}
 
 		if res == nil {
 			return nil, fmt.Errorf("failed to evaluate policy: neither '%s' nor '%s' rule found", mainRule, deprecatedRule)
 		}
 
-		return parseViolationsRule(res, policy)
+		return parseViolationsRule(res, policy, rawData)
```
migmartri, line 205: should we make this optional? Meaning that you can initialize the engine to return raw data too but optionally?

Piskoo: Makes sense due to only policy devel eval making use of it, thanks

app/cli/internal/policydevel/eval.go
```diff
@@ -82,41 +93,61 @@ func createCraftingSchema(policyPath string, inputs map[string]string) (*v1.Craf
...
 }
 
-func verifyMaterial(schema *v1.CraftingSchema, material *v12.Attestation_Material, materialPath string, allowedHostnames []string, logger *zerolog.Logger) ([]*EvalResult, error) {
+func verifyMaterial(schema *v1.CraftingSchema, material *v12.Attestation_Material, materialPath string, debug bool, allowedHostnames []string, logger *zerolog.Logger) (*EvalSummary, error) {
 	var opts []policies.PolicyVerifierOption
 	if len(allowedHostnames) > 0 {
 		opts = append(opts, policies.WithAllowedHostnames(allowedHostnames...))
 	}
+
+	opts = append(opts, policies.WithIncludeRawData(debug))
```
migmartri, line 102: nice

app/cli/internal/policydevel/eval.go
```diff
@@ -82,41 +93,61 @@ func createCraftingSchema(policyPath string, inputs map[string]string) (*v1.Craf
...
-	// no evaluations were returned
 	if len(policyEvs) == 0 {
 		return nil, fmt.Errorf("no execution branch matched for kind %s", material.MaterialType.String())
 	}
 
-	results := make([]*EvalResult, 0, len(policyEvs))
-	for _, policyEv := range policyEvs {
-		result := &EvalResult{
+	// Only one evaluation expected for a single policy attachment
+	policyEv := policyEvs[0]
```
migmartri, line 115: you sure about this? Can't an evaluation contain info from multiple paths?

migmartri (changes requested):
double check my comment about number of evaluations. Thanks!

app/cli/internal/policydevel/eval.go
Piskoo: Results from a single policy attachment, evaluated against a single material, are merged into one evaluation result. So results from each path in the policy will be merged in the end.

migmartri: nitpick. ok, should we protect the code?

pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
```diff
@@ -271,6 +271,8 @@ message PolicyEvaluation {
     string uri = 3 [(buf.validate.field).string.min_len = 1];
     string org_name = 4;
   }
+
+  repeated google.protobuf.Struct raw_results = 19;
```
migmartri: add a comment explainign what this is please

pkg/policies/engine/engine.go
```diff
@@ -30,6 +30,11 @@ type EvaluationResult struct {
 	Skipped    bool
 	SkipReason string
 	Ignore     bool
+	RawData    *RawData
+}
+type RawData struct {
+	Input  interface{}
```
migmartri: check json rawmessage

migmartri (approved):
Awesome

pkg/policies/engine/engine.go
Piskoo: Updated types for RawData in the engine and also in proto file

PR #2364, opened 2025-08-18
fix(build): include go package licenses in sbom fixes#2363

kaysavps:


.github/workflows/release.yaml
```diff
@@ -104,6 +104,8 @@ jobs:
 
       - name: Generate SBOMs, upload to release and attest
         id: attest_goreleaser
+        env:
+          SYFT_GOLANG_SEARCH_REMOTE_LICENSES: "true"
```
migmartri, line 108: great, do you mind adding a comment indicating why this option is needed?

kaysavps:
According to syft documentation, this setting gets the license info from the network using by default https://proxy.golang.org. 

```
golang:     
    # search for go package licences by retrieving the package from a network proxy
    #  SYFT_GOLANG_SEARCH_REMOTE_LICENSES env var    
    search-remote-licenses: false     
    # remote proxy to use when retrieving go packages from the network, 
    # if unset this defaults to $GOPROXY followed by https://proxy.golang.org    
    # SYFT_GOLANG_PROXY env var    
    proxy: ""
```

.github/workflows/release.yaml
kaysavps: added in the main thread

migmartri: a comment in the yaml code so we know why the option is in the yaml

migmartri:
> According to syft documentation, this setting gets the license info from the network using by default https://proxy.golang.org.
> 
> ```
> golang:     
>     # search for go package licences by retrieving the package from a network proxy
>     #  SYFT_GOLANG_SEARCH_REMOTE_LICENSES env var    
>     search-remote-licenses: false     
>     # remote proxy to use when retrieving go packages from the network, 
>     # if unset this defaults to $GOPROXY followed by https://proxy.golang.org    
>     # SYFT_GOLANG_PROXY env var    
>     proxy: ""
> ```

thanks. It might be more useful to have a minor comment in the code itself.

PR #2368, opened 2025-08-19
feat(projects): When adding members check all memberships

javirln:
This pull request updates the membership query logic in the `ProjectRepo` to ensure both direct and indirect project memberships are included in results. Previously, only top-level (direct) memberships were considered.

Membership query logic improvement:

* Modified the membership query in `project.go` to include both direct (where `ParentID` is nil) and indirect (where `ParentID` is not nil) memberships, instead of filtering only for top-level memberships.


app/controlplane/pkg/data/project.go
```diff
@@ -337,7 +337,11 @@ func (r *ProjectRepo) queryMembership(orgID uuid.UUID, projectID uuid.UUID, memb
 			membership.MemberID(memberID),
 			membership.ResourceTypeEQ(authz.ResourceTypeProject),
 			membership.ResourceID(projectID),
-			membership.ParentIDIsNil(), // Only top-level memberships
+			// include both direct (parent nil) and indirect (parent not nil) memberships
```
migmartri: are you sure this is not going to break anything? Why did we have `membership.ParentIDIsNil(),` before?

app/controlplane/pkg/data/project.go
```diff
@@ -337,7 +337,11 @@ func (r *ProjectRepo) queryMembership(orgID uuid.UUID, projectID uuid.UUID, memb
...
+			membership.Or(
+				membership.ParentIDIsNil(),
```
migmartri: what does this or mean?

migmartri (changes requested):
please review my comments

app/controlplane/pkg/data/project.go
migmartri: that does this OR do?

app/controlplane/pkg/biz/project.go
```diff
@@ -43,8 +43,12 @@ type ProjectsRepo interface {
...
 	UpdateMemberRoleInProject(ctx context.Context, orgID uuid.UUID, projectID uuid.UUID, memberID uuid.UUID, membershipType authz.MembershipType, newRole authz.Role) (*ProjectMembership, error)
-	// FindProjectMembershipByProjectAndID finds a project membership by project ID and member ID (user or group).
-	FindProjectMembershipByProjectAndID(ctx context.Context, orgID uuid.UUID, projectID uuid.UUID, memberID uuid.UUID, membershipType authz.MembershipType) (*ProjectMembership, error)
+	// FindProjectMembershipDirect finds a project membership by project ID and member ID (user or group).
+	// It only considers direct memberships (non-inherited).
+	FindProjectMembershipDirect(ctx context.Context, orgID uuid.UUID, projectID uuid.UUID, memberID uuid.UUID, membershipType authz.MembershipType) (*ProjectMembership, error)
+	// FindProjectMembershipEffective finds a project membership by project ID and member ID (user or group).
+	// considering inherited (effective) memberships as well.
+	FindProjectMembershipEffective(ctx context.Context, orgID uuid.UUID, projectID uuid.UUID, memberID uuid.UUID, membershipType authz.MembershipType) (*ProjectMembership, error)
```
migmartri: what it if has more than one? Do we aggregate them?

For example I can be both project admin and viewer no?

app/controlplane/pkg/data/project.go
```diff
@@ -326,8 +317,24 @@ func (r *ProjectRepo) UpdateMemberRoleInProject(ctx context.Context, orgID uuid.
...
+			membership.MemberID(memberID),
+			membership.ResourceTypeEQ(authz.ResourceTypeProject),
+			membership.ResourceID(projectID),
+			// only direct memberships (parent is nil)
+			membership.ParentIDIsNil(),
+		).WithOrganization()
+}
+
+// buildEffectiveMembershipQuery constructs a query that considers direct and inherited memberships
+func (r *ProjectRepo) buildEffectiveMembershipQuery(orgID uuid.UUID, projectID uuid.UUID, memberID uuid.UUID, membershipType authz.MembershipType) *ent.MembershipQuery {
```
migmartri: we could probably merge these two methods, since the only difference is parent ID is true

javirln:
Updated the PR, I've simplified things by just creating a method that checks for the existence of all kind of memberships, direct or inherited. The old code remains untouched.

app/controlplane/pkg/biz/project.go
```diff
@@ -342,12 +345,13 @@ func (uc *ProjectUseCase) addUserToProject(ctx context.Context, orgID uuid.UUID,
 
 	userUUID := uuid.MustParse(userMembership.User.ID)
 
-	// Check if the user is already a member of the project
-	existingMembership, err := uc.projectsRepository.FindProjectMembershipByProjectAndID(ctx, orgID, projectID, userUUID, authz.MembershipTypeUser)
-	if err != nil && !IsNotFound(err) {
+	// Check if the user is already a member of the project (including inherited memberships)
+	membershipExists, err := uc.projectsRepository.ExistsProjectMembershipEffective(ctx, orgID, projectID, userUUID, authz.MembershipTypeUser)
```
jiparis, line 349: Do we want this? This will prevent setting a custom role at the project level (inherited as viewer, local as Admin). This was one of our main use cases if I'm not wrong.

javirln: If we take into account the role when checking if the membership exists, we could have both, inherited as viewer and local as admin, that's an easy change.

javirln:
We are going to reevaluate the behavior because we have some doubts on how it should work.

PR #2370, opened 2025-08-20
fix(project-version): bootstrap missing default project versions 

Piskoo:
This PR fixes issue with missing project versions.

### DIFF
--- a/app/controlplane/pkg/data/workflow.go
+++ b/app/controlplane/pkg/data/workflow.go
@@ -104,7 +104,7 @@ func (r *WorkflowRepo) Create(ctx context.Context, opts *biz.WorkflowCreateOpts)
 				return fmt.Errorf("finding project version: %w", err)
 			}
 
-			if _, err := createProjectVersionWithTx(ctx, tx, projectID, "", false); err != nil {
+			if _, err := createProjectVersionWithTx(ctx, tx, projectID, "", true); err != nil {
 				return fmt.Errorf("creating project version: %w", err)
 			}
 		}


app/controlplane/pkg/data/ent/migrate/migrations/20250820090420.sql
```diff
@@ -0,0 +1,17 @@
+-- Create default version for projects that don't have any versions
+INSERT INTO "project_versions" ("id", "project_id", "version", "prerelease", "latest", "created_at", "updated_at")
+SELECT 
+    gen_random_uuid() as id,
+    p.id as project_id,
+    '' as version,
+    false as prerelease,
```
migmartri: look at what's the default value we set when we create it programatically, I believe it's prerelease=true

Piskoo: It's `false` according to platform biz logic

app/controlplane/pkg/data/ent/migrate/migrations/20250820090420.sql
```diff
@@ -0,0 +1,17 @@
...
+    true as latest,
+    CURRENT_TIMESTAMP as created_at,
```
migmartri: use the project timestamp

app/controlplane/pkg/data/ent/migrate/migrations/20250820090420.sql
migmartri: and this is `true`, please update the biz logic due to inconsistency

migmartri (approved):

PR #2373, opened 2025-08-20
feat(policy-devel): allow rego print statement output 

Piskoo:
This PR adds new rego engine option to handle rego `print` statements output and enables it for  `policy devel eval`.

### DIFF
--- a/app/cli/cmd/policy_develop_eval.go
+++ b/app/cli/cmd/policy_develop_eval.go
@@ -33,6 +33,7 @@ func newPolicyDevelopEvalCmd() *cobra.Command {
 		inputs           []string
 		allowedHostnames []string
 		debug            bool
+		enablePrint      bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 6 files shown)


app/cli/cmd/policy_develop_eval.go
```diff
@@ -77,6 +79,7 @@ evaluates the policy against the provided material or attestation.`,
 	cmd.Flags().StringArrayVar(&inputs, "input", []string{}, "Key-value pairs of policy inputs (key=value)")
 	cmd.Flags().StringSliceVar(&allowedHostnames, "allowed-hostnames", []string{}, "Additional hostnames allowed for http.send requests in policies")
 	cmd.Flags().BoolVarP(&debug, "debug", "", false, "Include detailed evaluation inputs/outputs in JSON output and enable verbose logging")
+	cmd.Flags().BoolVar(&enablePrint, "enable-print", false, "Enable print statements in Rego policies")
```
migmartri: why does it need to be a explicit flag?  Could it be enabled by default? or with debug?

Piskoo: Debug output is already pretty cluttered, so I thought making it an explicit flag for when you only want the print behavior was a good idea. That said, I can see how making it the default might also be fine

migmartri (approved):

PR #2375, opened 2025-08-20
fix: set latest version

migmartri:
Makes sure that projects that doesn't have a version marked as latest, are properly set


gr0 (approved):

Piskoo (approved):

javirln (approved):
LGTM

PR #2378, opened 2025-08-21
fix(examples): update example policies and test script

### LINKED ISSUE #2376: Update example policy test script to align with new eval output format
The test script in `docs/examples/policies` is currently out of date and will fail due to recent output changes in `policy devel eval`.

Piskoo:
This PR fixes issue with `_testutils.sh` functions returning exit codes in lint and eval functions. Older example policies got linted with `policy devel lint --format` to make them consistent with other examples. One of the tests in freshness policy was fixed since it was time depended. Regex is used to find result fields in the output so no additional changes were needed to adjust the script to the new `eval` output.

Closes #2376

### DIFF
--- a/docs/examples/policies/_testutils.sh
+++ b/docs/examples/policies/_testutils.sh
@@ -90,11 +90,11 @@ test_policy_lint() {
     
     if output=$($CHAINLOOP_BIN policy develop lint --policy "$policy_file" 2>&1); then
         echo -e "${GREEN}✓ PASSED${NC}"
-        ((TESTS_PASSED++))
+        ((TESTS_PASSED++)) || true
     else
         echo -e "${RED}✗ FAILED${NC}"
         echo "Output: $output"
-        ((TESTS_FAILED++))
+        ((TESTS_FAILED++)) || true
     fi
     echo ""
 }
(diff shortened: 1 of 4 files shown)


docs/examples/policies/_testutils.sh
```diff
@@ -90,11 +90,11 @@ test_policy_lint() {
     
     if output=$($CHAINLOOP_BIN policy develop lint --policy "$policy_file" 2>&1); then
         echo -e "${GREEN}✓ PASSED${NC}"
-        ((TESTS_PASSED++))
+        ((TESTS_PASSED++)) || true
```
migmartri, line 93: Hi! could you give some color on why we need this?

thanks!

Piskoo: Apparently `((TESTS_PASSED++))` uses post increment so here it returns 0 which causes the script to exit ([bash expr](https://stackoverflow.com/questions/69950459/bash-weirdness-exit-codes-of-arithmetic-expressions-in-relation-to-conditionals)). `((++TESTS_PASSED))` would fix that as well but as defensive measure I've used `|| true`
since from what I understood from other tests that use this utils, it is meant to return exit code with `test_summary` .

migmartri: but is this issue related to your change? Why wasn't it failing before? Out of curiosity what's the root cause

Piskoo: No, it was failing before, I'm not sure which released version caused it to not work, unless it wasn't working since the beginning

PR #2380, opened 2025-08-21
feat(policy): remove support for deprecated violations rule

### LINKED ISSUE #2371: Remove support for deprecated violations rule from rego engine
Remove support for the deprecated violations rule in the Rego engine to simplify policy evaluation and rely solely on the result rule.

Piskoo:
This PR removes deprecated rule handling from rego engine and updates testdata that used deprecated rule to return result rule.

Closes #2371

### DIFF
--- a/pkg/policies/engine/rego/rego.go
+++ b/pkg/policies/engine/rego/rego.go
@@ -189,6 +189,10 @@ func (r *Engine) Verify(ctx context.Context, policy *engine.Policy, input []byte
 		return nil, err
 	}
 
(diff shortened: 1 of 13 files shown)


migmartri:
I'd like a review from @jiparis on this to see if we are comfortable with making this change already

jiparis:
Thanks @Piskoo. I think we can perform this removal. All our policies and examples have been updated. However, it's a breaking change, let's make sure we state it properly in the release notes.

PR #2383, opened 2025-08-21
feat(policy-devel): drop support for yaml and embedded rego formatting

### LINKED ISSUE #2379: Chainloop policy devel lint --format should not reformat multi-line YAML descriptions
Currently, the Rego linter (or chainloop policy devel lint --format) automatically converts multi-line YAML description fields into a single-line string with \n characters.

**Example:**

```
#Original YAML:

- description: |
    This policy performs several checks:

    - Step 1: Validate input format
    - Step 2: Check required fields
    - Step 3: Apply optional rules
```

```
# YAML after running policy devel lint --format

description: "This policy performs several checks:\n\n- Step 1: Validate input format\n- Step 2: Check required fields\n- Step 3: Apply optional rules"

```

**Problem:** Makes the YAML harder to read and maintain.

**Proposed improvement:** Preserve multi-line description fields as-is in YAML format and avoid converting them into a single string with \n.

Piskoo:
This PR removes support for yaml and embedded rego formatting due issues caused by yaml libraries limitations.

Closes #2379

### DIFF
--- a/app/cli/cmd/policy_develop_lint.go
+++ b/app/cli/cmd/policy_develop_lint.go
@@ -65,7 +65,7 @@ func newPolicyDevelopLintCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&policyPath, "policy", "p", "policy.yaml", "Path to policy file")
-	cmd.Flags().BoolVar(&format, "format", false, "Auto-format file with opa fmt")
+	cmd.Flags().BoolVar(&format, "format", false, "Auto-format standalone policy rego files with opa fmt (embedded policies not supported)")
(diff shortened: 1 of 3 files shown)


migmartri:
@Piskoo do we still check that the policy is a valid policy spec? Meaning that we can marshal it?

Piskoo:
@migmartri Only for eval, the only operation that lint does on yaml files is lookup for rego script paths

migmartri:
> @migmartri Only for eval, the only operation that lint does on yaml files is lookup for rego script paths

there might be value on still checking that it's a valid policy spec no? Before we used to run even proto validations on the spec no?

Piskoo:
> > @migmartri Only for eval, the only operation that lint does on yaml files is lookup for rego script paths
> 
> there might be value on still checking that it's a valid policy spec no? Before we used to run even proto validations on the spec no?

Correct, took a look and validations are still ran on policy spec but errors are ignored. I'll fix it in another PR

PR #2413, opened 2025-09-02
feat(controlplane): Remove org requirement middleware check

javirln:
Since CAS Backend validation is now handled asynchronously, we can remove the middleware that previously enforced CAS Backend checks during certain API operations.


migmartri:
I don't think we are ready to do this, this means that the user will not get updated status each 30 minutes no?

jiparis:
> I don't think we are ready to do this, this means that the user will not get updated status each 30 minutes no?

I agree. Let's wait until we have other ways to notify via CLI.

PR #2420, opened 2025-09-05
chore(api): Only check CAS Backend status on needed endpoints

javirln:
This PR splits the `CheckOrgRequirement` middleware in two.

**Middleware changes**

- `ValidateCASBackend`: Middleware in charge of checking the organization has configured a valid CAS backend and the credentials work. It does not block if the status of the middleware is invalid.
- `BlockIfCASBackendNotValid`: Checks the default middleware in the organization is valid and blocks the request if is not valid.

This give us flexibility since there are some endpoints where we still want to check the status of the CAS backend without blocking the request if they are not valid and others where we want to directly block the request because they are needed.

**Endpoint scope changes**
`BlockIfCASBackendNotValid` applies to only the following endpoints:
```
controlplane.v1.AttestationService.GetUploadCreds
controlplane.v1.AttestationService.Init
controlplane.v1.AttestationService.Store
controlplane.v1.CASCredentialsService.Get
```

**API token handling**
The middleware branch for API token authentication has also been updated. It now includes `ValidateCASBackend` and `BlockIfCASBackendNotValid` only when the operation requires a CAS backend. For example, with a failing CAS backend:
- **Before**: a request to list workflows using an API token would fail.
- **Now**: since workflow listing does not require a CAS backend, the request will succeed.

**Others**
- `GetUploadCreds`: It will run the validation on the CAS Backend of the workflow associated instead the default CAS Backend (it could have changed since att init was done)
- `Init`, `Store`: We cannot store even an attestation that is empty, so we validate them in those cases as well.


app/controlplane/internal/service/attestation.go
```diff
@@ -284,6 +289,12 @@ func (s *AttestationService) storeAttestation(ctx context.Context, envelope []by
 
 	// If we have an external CAS backend, we will push there the attestation
 	if !casBackend.Inline {
+		if casBackend.ValidationStatus != biz.CASBackendValidationOK {
```
jiparis, line 293: I understand this is done here instead of the middleware because the CAS backend used in the attestation might not be the default one, right? Can you add a comment so that we don't forget?

app/controlplane/internal/service/attestation.go
```diff
@@ -438,6 +449,16 @@ func (s *AttestationService) GetUploadCreds(ctx context.Context, req *cpAPI.Atte
 	}
 
 	backend := wRun.CASBackends[0]
+
+	// Check the status of the backend
+	if backend.ValidationStatus != biz.CASBackendValidationOK {
```
jiparis, line 455: same here

app/controlplane/internal/service/attestation.go
javirln: Yes so on the middleware we check the default one and here we might be using something different. I'll add the comments in both places!

PR #2422, opened 2025-09-08
fix: change order query conditions

migmartri:
Changing order so the DB index work as expected


jiparis:
Please check https://github.com/chainloop-dev/chainloop/pull/2423 for the fix.

migmartri:
not needed. The order of the conditions doesn't make a difference

javirln:
I had it understood that the order matter on indexes: https://neon.com/postgresql/postgresql-indexes/postgresql-multicolumn-indexes

migmartri:
> I had it understood that the order matter on indexes: https://neon.com/postgresql/postgresql-indexes/postgresql-multicolumn-indexes

That's correct, the order in the indexes matter, but not in the queries, which is what I changed

PR #2423, opened 2025-09-08
fix(queries): fix index for expiration job

jiparis:
This query is not ordered, so the index cannot be used directly. Instead, Postgresql was partially using a different index (`workflowrun_state_finished_at`), and then filtering by `creating_at`, adding an additional performance penalization.
```
Limit  (cost=0.29..6.07 rows=1 width=250) (actual time=0.133..0.134 rows=0 loops=1)
  ->  Index Scan using workflowrun_state_finished_at on workflow_runs  (cost=0.29..6.07 rows=1 width=250) (actual time=0.131..0.132 rows=0 loops=1)
        Index Cond: ((state)::text = 'initialized'::text)
        Filter: (created_at <= '2025-06-25 11:42:12.793517+00'::timestamp with time zone)
Planning Time: 1.043 ms
Execution Time: 0.245 ms
```

With the fixed index, the query is fully optimized:
```
Limit  (cost=0.29..8.31 rows=1 width=249) (actual time=0.056..0.056 rows=0 loops=1)
  ->  Index Scan using workflowrun_state_created_at on workflow_runs  (cost=0.29..8.31 rows=1 width=249) (actual time=0.055..0.055 rows=0 loops=1)
        Index Cond: (((state)::text = 'initialized'::text) AND (created_at <= '2025-06-25 11:42:12.793517+00'::timestamp with time zone))
Planning Time: 0.674 ms
Execution Time: 0.099 ms

```


app/controlplane/pkg/data/ent/migrate/migrations/20250908160222.sql
```diff
@@ -0,0 +1,4 @@
+-- Drop index "workflowrun_state_created_at" from table: "workflow_runs"
+DROP INDEX "workflowrun_state_created_at";
```
migmartri, line 4: change it to concurrently, see the annotation and the CONCURRENTLY stanza 
for example
```sql
-- atlas:txmode none

DROP INDEX workflowrun_attestation_digest;
CREATE INDEX CONCURRENTLY "workflowrun_attestation_digest" ON "workflow_runs" ("attestation_digest");
```

jiparis:
> Dont forget
> 
> ```
> -- atlas:txmode none
> ```

Thanks, done.

javirln (approved):

PR #2425, opened 2025-09-09
build(contract): skip empty license verification for the SBOM component type file

kaysavps:
This passes the input to skip components with type "file" when executing the sbom-with-licenses policy. This will prevent unnecessary flagging as components with no license. These types of components are files installed as part of another component as a library and no additional license validation is needed.

### DIFF
--- a/.github/workflows/contracts/chainloop-vault-release.yml
+++ b/.github/workflows/contracts/chainloop-vault-release.yml
@@ -16,6 +16,7 @@ policyGroups:
   - ref: sbom-quality
     with:
       bannedLicenses: GPL, AGPL
+      skippedTypes: file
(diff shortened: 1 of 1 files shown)


.github/workflows/contracts/chainloop-vault-release.yml
```diff
@@ -16,6 +16,7 @@ policyGroups:
   - ref: sbom-quality
     with:
       bannedLicenses: GPL, AGPL
+      skippedTypes: file
       bannedComponents: log4j@2.14.1
```
jiparis, line 20: This bannedComponents was just a demo, we might want to remove it.

jiparis (approved):

PR #2431, opened 2025-09-11
feat(plugin): allow subcommands and platform configuration

Piskoo:
### Summary
- Enable plugins to register subcommands under existing CLI commands 
- Provide configuration for backend API access


migmartri (approved):
Great, Thanks! 

Next time, I'd split the PR into multiple ones so they focus on one task only. That way it's easier to revert, extend, etc.

Thanks

PR #2441, opened 2025-09-19
feat(integrations): refactor FanOut plugins

jiparis:
This PR refactors the plugins SDK to extract the specifics for FanOut plugins to its own logic, keeping a base `Integration` interface for future improvements. 
* New `Integration` interface and `IntegrationBase` struct for generic integrations. Integrations can be of multiple kinds (fanout, notification)
* Refactor FanOut interface and `FanOutIntegration` struct to use the new plugins primitives, and moved all FanOut-specific logic to these interfaces.
* Make all interfaces (execution requests) generic to support the FanOut case as well as any other future case. Changes in README are not needed as the main constructors (`NewFanOut`) have not changed.

The external interfaces have not changed, APIs and database tables remain untouched.

Before:
image

After:
image


app/controlplane/plugins/sdk/v1/integrations.go
```diff
@@ -0,0 +1,362 @@
+//
```
jiparis, line 1: Here you can find the generic `Integration` logic

app/controlplane/internal/dispatcher/dispatcher_test.go
```diff
@@ -53,6 +53,10 @@ func (s *dispatcherTestSuite) TestLoadInputsEnvelope() {
 	envelope, err := testEnvelope("testdata/attestation.json")
 	require.NoError(s.T(), err)
 
+	s.ociIntegrationBackend.(*mockedSDK.FanOut).On("IsSubscribedTo", "CONTAINER_IMAGE").Return(true)
```
migmartri, line 56: why do we need this mocks now?

app/controlplane/pkg/biz/integration_test.go
```diff
@@ -39,10 +39,23 @@ func (s *testSuite) TestCreate() {
 
 	// Mocked integration that will return both generic configuration and credentials
 	integration := integrationMocks.NewFanOut(s.T())
+	type schema struct {
+		FirstName string `json:"firstName"`
```
migmartri, line 43: what's firstName for?

app/controlplane/plugins/sdk/v1/fanout.go
```diff
@@ -254,6 +207,20 @@ func validateInputs(c *FanOutIntegration) error {
 	return nil
 }
 
+// Methods to be implemented by the specific integration
+
+func (i *FanOutIntegration) Register(_ context.Context, _ *RegistrationRequest) (*RegistrationResponse, error) {
```
migmartri, line 212: nice.

migmartri (approved):
nice!

app/controlplane/internal/dispatcher/dispatcher_test.go
jiparis: Because these used to be properties, but now they are methods in the FanOut interface. The new implementation calls them to get the same values.

app/controlplane/pkg/biz/integration_test.go
jiparis: `s.configStruct` used the tests has this property. It would fail with `validation error: additionalProperties 'firstName' not allowed` if it's not specified in the schema.

PR #2443, opened 2025-09-22
fix(cli):  error handling

### LINKED ISSUE #2434: parsing authentication shows confusing error
For example

```
chainloop att init --name testing --project new-project  --replace --version 1.2
Flag --name has been deprecated, please use --workflow instead
This command is will run against the organization "miguel-test"
y
ERR failed to load issuer: authorization error: issuer not found: cp.chainloop
```

this should have shown that the token is expired.

Piskoo:
This PR grcp errors not being properly caught and replaced with more user friendly error in CLI.

Closes #2434

### DIFF
--- a/app/cli/main.go
+++ b/app/cli/main.go
@@ -65,13 +65,13 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 		var gs grpcstatus
 		if errors.As(err, &gs) {
 			knownCodes := []codes.Code{
-				codes.AlreadyExists, codes.InvalidArgument, codes.NotFound, codes.PermissionDenied,
+				codes.AlreadyExists, codes.InvalidArgument, codes.NotFound, codes.PermissionDenied, codes.Unauthenticated,
 			}
 
-			grpcStatus := gs.GRPCStatus()
+			st = gs.GRPCStatus()
 			for _, code := range knownCodes {
 				if st.Code() == code {
-					msg = grpcStatus.Message()
+					msg = st.Message()
 					break
 				}
 			}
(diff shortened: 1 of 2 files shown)


app/cli/main.go
```diff
@@ -108,5 +109,5 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 
 func isWrappedErr(grpcStatus *status.Status, err *errors.Error) bool {
 	target := errors.FromError(grpcStatus.Err())
-	return target.Code == err.Code && err.Message == target.Message
+	return target.Code == err.Code && strings.Contains(target.Message, err.Message)
```
migmartri: mm, this looks like a little bit of a hack that could cause false positives. Do you know when and why this started failing? Was it because we are now wrapping when we should not?

migmartri:
> Example
Before it was checking for equality between JWT token has expired and rpc message JWT token has expired metadata = map[] cause = <nil>

So what's now? 

Will be be reverted to what originally was which was handled locally as I mentioned in the issue?

Piskoo:
I'll work as it did before, seems like part of the output was missing in the description and I didn't notice. In general the problem is that grpc status returns message that looks like full error instead of only returning the message part:
```
rpc error: code = Unauthenticated desc = couldn't match JWT provider: error: code = 401 reason = UNAUTHORIZED message = JWT token has expired metadata = map[] cause = <nil>
```
Which then before was matched against ```JWT token has expired ```. It resulted in inequality so the error was not replaced in the end. ~Not sure when and why is started, but it's probably related to `grpc-go` being updated at some point.~ Went through Kratos and grpc and no changes were made to related files. Some backend change must have changed it

app/cli/main.go
jiparis: The error in backend might have changed. It was a hack already anyways, since it was comparing error strings in client side, instead of just the error code or a typed error object.

migmartri:
This is the actual fix https://github.com/chainloop-dev/chainloop/pull/2450 so I'd suggest closing this one.

PR #2454, opened 2025-10-01
feat(controlplane): add last_run_at timestamp to project versions

migmartri:
## Summary
- Add `last_run_at` timestamp field to track when the last workflow run occurred for each project version
- The existing `workflow_run_count` counter continues to work as before
- Automatically updates timestamp when new workflow runs are created

## Changes
- Add `last_run_at` field to `ProjectVersion` Ent schema
- Update `WorkflowRunRepo.Create()` to set timestamp on new runs
- Add `LastRunAt` field to `biz.ProjectVersion` struct
- Update data conversion layer to include new field
- Create batched backfill migration for existing data

## Migration Strategy
- Schema migration adds nullable `last_run_at` column
- Backfill migration processes project versions in batches of 50 to avoid performance impact
- Uses the most recent workflow run timestamp (finished_at or created_at) for each version

## Test Plan
- [x] Unit tests pass
- [x] Build succeeds
- [x] Migration linting passes
- [x] Generated Ent code verified


javirln (approved):
LGMT, this is only about `last_run_at` not the counter right?

migmartri:
It already exists :)

PR #2457, opened 2025-10-06
feat(policy): add organization field to metadata

### LINKED ISSUE #2456: Extend policy and policy group metadata with organization field
Add `organization` field to metadata

...
metadata:
  name: name
  organization: target-org-name
  description: description
...

Piskoo:
This PR adds optional organization field to metadata.

Closes #2456


jiparis:
What's the purpose of this field? is it like a namespace so that it can be pushed to the right organization instead of the "current" one?

jiparis (approved):

Piskoo:
> is it like a namespace so that it can be pushed to the right organization instead of the "current" one?

correct

migmartri (approved):

PR #2460, opened 2025-10-07
fix(projectversion): Populate project versions that have missing releasedAt timestamp

javirln:
This PR introduces a SQL migration to populate any existing project version where its `released_at` was not set.


migmartri (approved):

app/controlplane/pkg/data/ent/migrate/migrations/20251007154203.sql
```diff
@@ -0,0 +1,51 @@
...
+    -- (those that are released but don't have released_at set)
+    SELECT COUNT(*)
+    INTO total_remaining
+    FROM project_versions
+    WHERE prerelease = false
+      AND released_at IS NULL
+      AND deleted_at IS NULL;
+
+    -- Loop until no more rows remain
+    WHILE total_remaining > 0 LOOP
```
jiparis, line 21: A weird approach, instead of just a simple `update ... set ... where` unless I'm missing something.

migmartri: batching?

javirln: Because we want to do it in batches to not block all rows.

migmartri: I have a question, are we sure this is not blocking anyways? Note that atlas has a transaciton mode enabled by default, should we disable it i.e `tx-mode:none` like we do with concurrent indexes?

PR #2465, opened 2025-10-14
feat(crafting-schema): add organization resolver

Piskoo:
This PR updates current organization middleware with handler for in schema defined organization.

### DIFF
--- a/app/controlplane/internal/usercontext/currentorganization_middleware.go
+++ b/app/controlplane/internal/usercontext/currentorganization_middleware.go
@@ -17,7 +17,6 @@ package usercontext
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -200,20 +199,14 @@ type ResourceBase struct {
 
 // Extracts organization from request with raw contract data
 func extractOrg(req interface{}) (string, error) {
-	// Get raw data as base64 string
-	rawSchema, err := getRawData(req)
+	// Get raw data
+	rawData, err := getRawData(req)
 	if err != nil {
 		return "", err
 	}
 
-	if rawSchema == "" {
+	if len(rawData) == 0 {
 		return "", nil
-	}
-
-	// Decode base64 raw schema
-	rawData, err := base64.StdEncoding.DecodeString(rawSchema)
-	if err != nil {
-		return "", err
 	}
 
 	// Identify format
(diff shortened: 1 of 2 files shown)


app/controlplane/internal/usercontext/currentorganization_middleware.go
```diff
@@ -166,3 +176,81 @@ func setCurrentOrganizationFromDB(ctx context.Context, user *entities.User, user
...
+		return "", err
+	}
+
+	// Unmarshal to extract organization
+	var resourceBase ResourceBase
+	if err := json.Unmarshal(jsonData, &resourceBase); err != nil {
+		return "", err
+	}
+
+	return resourceBase.Metadata.Organization, nil
```
migmartri, line 231: what if it doesn't have `metadata` as is an old format?

app/controlplane/internal/usercontext/currentorganization_middleware.go
```diff
@@ -166,3 +176,81 @@ func setCurrentOrganizationFromDB(ctx context.Context, user *entities.User, user
...
+// Extracts raw data
+func getRawData(req interface{}) (string, error) {
+	// Check if the request implements RequestWithRawContract
+	if rawContractReq, ok := req.(RequestWithRawContract); ok {
+		rawBytes := rawContractReq.GetRawContract()
+		if len(rawBytes) == 0 {
+			return "", nil
+		}
+		// Convert bytes to base64 string for consistent handling
+		return base64.StdEncoding.EncodeToString(rawBytes), nil
```
migmartri: why return strings and not bytes?

Piskoo: Removed, it was there to make it consistent with platform

app/controlplane/internal/usercontext/currentorganization_middleware.go
Piskoo: Updated unmarshalling to not return an error when old schema is used

app/controlplane/internal/usercontext/currentorganization_middleware.go
```diff
@@ -166,3 +175,72 @@ func setCurrentOrganizationFromDB(ctx context.Context, user *entities.User, user
...
+}
+
+type RequestWithRawContract interface {
+	GetRawContract() []byte
+}
+
+// Extracts raw data
+func getRawData(req interface{}) ([]byte, error) {
+	// Check if the request implements RequestWithRawContract
+	if rawContractReq, ok := req.(RequestWithRawContract); ok {
```
jiparis, line 241: This is superspecific to contracts. I think we can deprecate the field `raw_contract` in the proto definition and call it `raw_resource`, for example, just to make it generic enough so that we can reuse this interface for other resources (policies, etc). Would it be possible?

jiparis: I think we are using `raw_schema` in other places.

Piskoo: Naming is all over the place, policies use `raw_schema`, policy groups `raw_group`, contracts `raw_contract`. Chainloop OSS will be only responsible for contracts so I think there's no need to add another field to proto def

jiparis (approved):

PR #2467, opened 2025-10-14
feat(crafting-schema): add schema v2 support to cli

Piskoo:
## Summary

This PR adds support for contract schema v2 in CLI `workflow contract` `create`, `update`, and `apply` operations.

Note: The default contract body for `create` will be updated once v2 schema support is available everywhere.

### Changes
- CLI can now create and update contracts using v2 schema format:
```
  apiVersion: ...
  kind: ...
  metadata:
    name: ...
  spec:
    ...
```
- Contract name can now be extracted from the v2 field `metadata.name`, removing the need for the `--name` flag.
- Validation ensures only one name source is provided (either `--name` flag or `metadata.name` in the contract).

### DIFF
--- a/app/controlplane/internal/service/workflowcontract.go
+++ b/app/controlplane/internal/service/workflowcontract.go
@@ -324,7 +324,7 @@ func validateAndExtractName(rawContract []byte, explicitName string) (string, er
 	if explicitName != "" && metadataName != "" {
 		// If both are provided but different, that's a conflict
 		if explicitName != metadataName {
-			return "", errors.BadRequest("invalid", "conflicting names: explicit name and metadata.name are different. Please provide only one.")
+			return "", errors.BadRequest("invalid", "conflicting names: explicit name and metadata.name are different")
 		}
 		// Names match, return it
 		return explicitName, nil
--- a/app/cli/pkg/action/util.go
+++ b/app/cli/pkg/action/util.go
@@ -64,19 +64,19 @@ func ValidateAndExtractName(explicitName, filePath string) (string, error) {
 	if filePath != "" {
 		content, err = LoadFileOrURL(filePath)
 		if err != nil {
-			return "", fmt.Errorf("failed to load file: %w", err)
+			return "", fmt.Errorf("load file: %w", err)
 		}
 	}
 
 	// Extract name from v2 metadata (if present)
 	metadataName, err := extractNameFromMetadata(content)
 	if err != nil {
-		return "", fmt.Errorf("failed to parse content: %w", err)
+		return "", fmt.Errorf("parse content: %w", err)
 	}
 
 	// Both provided - ambiguous
 	if explicitName != "" && metadataName != "" {
-		return "", fmt.Errorf("conflicting names: explicit name (%q) and metadata.name (%q) both provided. Please provide only one", explicitName, metadataName)
+		return "", fmt.Errorf("conflicting names: explicit name (%q) and metadata.name (%q) both provided", explicitName, metadataName)
 	}
 
 	// Neither provided - missing required name


app/cli/cmd/workflow_contract_create.go
```diff
@@ -43,9 +43,6 @@ func newWorkflowContractCreateCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&name, "name", "", "contract name")
-	err := cmd.MarkFlagRequired("name")
-	cobra.CheckErr(err)
```
migmartri, line 47: missing validation

app/controlplane/api/controlplane/v1/workflow_contract.proto
```diff
@@ -53,25 +47,41 @@ message WorkflowContractServiceCreateRequest {
 
   // You might need to specify a project reference if you want/need to create a contract scoped to a project
   IdentityReference project_reference = 5;
+
+  option (buf.validate.message).cel = {
+    id: "name.dns-1122"
+    message: "must contain only lowercase letters, numbers, and hyphens."
+    expression: "this.name == '' || this.name.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')"
+  };
+  option (buf.validate.message).cel = {
```
migmartri: this validation doesnt mean much in legacy format

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -358,12 +357,12 @@ func entContractVersionToBizContractVersion(w *ent.WorkflowContractVersion) (*bi
...
 		// but we also want to keep the Body field populated for backward compatibility
 	} else if len(w.Body) == 0 {
-		schema := &schemav1.CraftingSchema{}
-		err := unmarshal.FromRaw(w.RawBody, w.RawBodyFormat, schema, false)
+		parsedContract, err := biz.UnmarshalAndValidateRawContract(w.RawBody, w.RawBodyFormat)
 		if err != nil {
 			return nil, fmt.Errorf("failed to unmarshal raw body: %w", err)
 		}
-		contract.Schema = schema
+		contract.Schema = parsedContract.Schema
```
migmartri, line 365: This might be broken, because Schema is necessary for compatibility reasons.

migmartri: Old CLIs must keep working with new contracts

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -358,12 +357,12 @@ func entContractVersionToBizContractVersion(w *ent.WorkflowContractVersion) (*bi
...
+		contract.Schemav2 = parsedContract.Schemav2
```
migmartri: This second option we don't want it.

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -295,3 +307,28 @@ func (s *WorkflowContractService) checkContractAccess(ctx context.Context, contr
 
 	return nil
 }
+
+// extractContractName extracts the contract name from v2 metadata or returns the fallback name
+func extractContractName(rawContract []byte, name string) (string, error) {
```
migmartri: validate that exactly one item is present

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -388,20 +403,47 @@ func (uc *WorkflowContractUseCase) ValidateContractPolicies(rawSchema []byte, to
 		return NewErrValidation(err)
 	}
 
-	for _, att := range c.Schema.GetPolicies().GetAttestation() {
-		if _, err := uc.findAndValidatePolicy(att, token); err != nil {
-			return NewErrValidation(err)
+	switch {
+	case c.isV1Schema():
```
migmartri, line 407: mark as deprecated

app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -601,22 +644,40 @@ func (uc *WorkflowContractUseCase) findProvider(providerName string) (*policies.
...
+		if err := v1Contract.ValidatePolicyAttachments(); err != nil {
+			return nil, NewErrValidation(fmt.Errorf("policy attachment validation failed: %w", err))
+		}
+		return &Contract{Raw: raw, Format: format, Schema: v1Contract}, nil
 	}
 
-	return &Contract{Raw: raw, Format: format, Schema: contract}, nil
+	// Both parsing attempts failed
+	// Best effort: provide errors for both schemas
+	return nil, NewErrValidation(fmt.Errorf("contract validation failed:\n  v2 Contract format error: %w\n  v1 CraftingSchema format error: %w", v2Err, v1Err))
```
jiparis, line 680: I think fmt.Errorf only allows 1 wrapped error, so in practice this will only take the later. Instead, use `multierror`, there are some usages already in the code.

jiparis: Also, with multierror you don't need to use different error vars, just do something like:
```
var multiErr *multierror.Error
err = ...
if err != nil {
  multiErr = multiErr.Append(err)
}
err = ...
if err != nil {
  multiErr = multiErr.Append(err)
}
```

jiparis: Nevermind, I said nothing,  I just saw that `fmt.Errorf` allows multiple wrapper errors.

app/controlplane/pkg/data/workflowcontract.go
```diff
@@ -357,13 +356,13 @@ func entContractVersionToBizContractVersion(w *ent.WorkflowContractVersion) (*bi
 		}
 		// Scenario 2: contracts that have been updated after the introduction of the raw_body field will have the raw_body field populated
 		// but we also want to keep the Body field populated for backward compatibility
+		// Note: UnmarshalAndValidateRawContract ensures both Schema (v1) and Schemav2 (v2) are populated for v2 contracts
 	} else if len(w.Body) == 0 {
-		schema := &schemav1.CraftingSchema{}
-		err := unmarshal.FromRaw(w.RawBody, w.RawBodyFormat, schema, false)
+		parsedContract, err := biz.UnmarshalAndValidateRawContract(w.RawBody, w.RawBodyFormat)
```
jiparis, line 361: This looks awkward. Data layer shouldn't call biz layer logic.

Piskoo: https://github.com/chainloop-dev/chainloop/blob/main/app/controlplane/pkg/data/workflowcontract.go#L354
That is already the case currently

jiparis: I see, let's address it in another PR then.

PR #2469, opened 2025-10-15
feat(cli): extend workflow contract with apply

Piskoo:
This PR adds apply command that combines create and update functionality into one.

Example
```
# Create/Update v1 contract
cl wf contract apply --contract contract-schema.yaml --name contract-test
```

### DIFF
--- a/app/cli/cmd/workflow_contract_apply.go
+++ b/app/cli/cmd/workflow_contract_apply.go
@@ -23,8 +23,6 @@ import (
 
 func newWorkflowContractApplyCmd() *cobra.Command {
 	var contractPath, name, description, projectName string
-	var contractName string
-	var rawContract []byte
 
 	cmd := &cobra.Command{
 		Use:   "apply",
@@ -39,7 +37,7 @@ or update it if it already exists.`,
 				desc = &description
 			}
 
-			res, err := action.NewWorkflowContractApply(ActionOpts).Run(cmd.Context(), rawContract, contractName, desc, projectName)
+			res, err := action.NewWorkflowContractApply(ActionOpts).Run(cmd.Context(), name, contractPath, desc, projectName)
 			if err != nil {
 				return err
 			}
--- a/app/cli/pkg/action/workflow_contract_apply.go
+++ b/app/cli/pkg/action/workflow_contract_apply.go
@@ -30,7 +30,7 @@ func NewWorkflowContractApply(cfg *ActionsOpts) *WorkflowContractApply {
 	return &WorkflowContractApply{cfg}
 }
 
-func (action *WorkflowContractApply) Run(ctx context.Context, rawContract []byte, contractName string, description *string, projectName string) (*WorkflowContractItem, error) {
+func (action *WorkflowContractApply) Run(ctx context.Context, contractName string, contractPath string, description *string, projectName string) (*WorkflowContractItem, error) {
 	client := pb.NewWorkflowContractServiceClient(action.cfg.CPConnection)
 
 	// Try to describe the specific contract first to determine if we should create or update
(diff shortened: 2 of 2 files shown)


app/cli/cmd/workflow_contract_apply.go
```diff
@@ -0,0 +1,67 @@
...
+		Short: "Apply a contract (create or update)",
+		Long: `Apply a contract from a file. This command will create the contract if it doesn't exist,
+or update it if it already exists.`,
+		Example: `  # Apply a contract from file
+  chainloop workflow contract apply --contract my-contract.yaml
+
+  # Apply to a specific project
+  chainloop workflow contract apply --contract my-contract.yaml --project my-project`,
+		PreRunE: func(_ *cobra.Command, _ []string) error {
+			if filePath == "" && name == "" {
```
migmartri: it would make more sense if this validates that either the explicit name or a name in the spec is available, this new helper can probably be made generically as you did in the middleware

app/cli/pkg/action/workflow_contract_apply.go
```diff
@@ -0,0 +1,131 @@
+//
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri: 2025, it's a new file

app/cli/pkg/action/workflow_contract_apply.go
```diff
@@ -0,0 +1,131 @@
...
+	res, err := client.Create(context.Background(), createReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new contract '%s': %w", contractName, err)
+	}
+
+	return pbWorkflowContractItemToAction(res.Result), nil
+}
+
+// extractContractNameFromContent tries to extract the contract name from the contract file content
+func extractContractNameFromRawSchema(content []byte) (string, error) {
```
migmartri: this function could be made generically by marshalling to anything with `meta` and run from `cmd` as part of the validation as mentioned above.

app/cli/pkg/action/workflow_contract_apply.go
```diff
@@ -0,0 +1,131 @@
...
+
+type WorkflowContractApply struct {
+	cfg *ActionsOpts
+}
+
+func NewWorkflowContractApply(cfg *ActionsOpts) *WorkflowContractApply {
+	return &WorkflowContractApply{cfg}
+}
+
+func (action *WorkflowContractApply) Run(filePath, name string, description *string, projectName string) (*WorkflowContractItem, error) {
```
migmartri: can you add ctx please?

app/cli/pkg/action/workflow_contract_apply.go
```diff
@@ -0,0 +1,131 @@
...
+		updateReq := &pb.WorkflowContractServiceUpdateRequest{
+			Name:        contractName,
+			RawContract: rawContract,
+		}
+
+		if description != nil {
+			updateReq.Description = description
+		}
+
+		res, err := client.Update(context.Background(), updateReq)
```
migmartri: and use context here? 

In general let's pass context downstream

migmartri (commented):
Thanks

app/cli/cmd/workflow_contract_apply.go
```diff
@@ -0,0 +1,71 @@
...
+				return err
+			}
+
+			logger.Info().Msg("Contract applied!")
+			return output.EncodeOutput(flagOutputFormat, res, contractItemTableOutput)
+		},
+	}
+
+	cmd.Flags().StringVarP(&filePath, "contract", "f", "", "workflow contract file path (optional)")
+	cmd.Flags().StringVar(&name, "name", "", "contract name (required if no contract file provided)")
```
jiparis: This doesn't match with the command description nor the examples (`Apply a contract from a file. This command will create the contract if it doesn't exist`). How would it work then? I would expect that the file was mandatory.

Piskoo:
I've decided to remove support for schema v2 from this PR, all schema v2 related code will be a part of https://github.com/chainloop-dev/chainloop/pull/2467. Here we have `wf contract apply` based on `create` and `update` that handles old schema. 
cc/ @jiparis @migmartri

PR #2484, opened 2025-10-27
feat(contract): reject attestation level policies in materials

Piskoo:
This PR adds validation to contract creation and update, that prevents assigning policies that contain execution path for kind `ATTESTATION` on material level.

## Example:
For `sbom-present` policy
```
kind: Policy
metadata:
  name: sbom-present
  description: desc
spec:
  policies:
    - kind: ATTESTATION
      path: sbom-present.rego
``` 

### Valid contract
```
schema_version: v1
policies:
  attestation:
    - ref: sbom-present
```
Passes, the contract is created/updated

### Invalid contract
```
schema_version: v1
policies:
  materials:
    - ref: sbom-present
```
Returns
```
ERR validation error: attestation policy "sbom-present" cannot be attached to materials
exit status 1
```

Already existing contracts are unaffected, invalid contract will also fail if raw contract is given during `att init` in `--contract` flag.

### DIFF
--- a/app/controlplane/pkg/biz/workflowcontract.go
+++ b/app/controlplane/pkg/biz/workflowcontract.go
@@ -394,8 +394,15 @@ func (uc *WorkflowContractUseCase) ValidateContractPolicies(rawSchema []byte, to
 		}
 	}
 	for _, att := range c.Schema.GetPolicies().GetMaterials() {
-		if _, err := uc.findAndValidatePolicy(att, token); err != nil {
+		policy, err := uc.findAndValidatePolicy(att, token)
+		if err != nil {
 			return NewErrValidation(err)
+		}
+		// Validate that policies attached to materials do not have kind ATTESTATION
+		if policy != nil {
+			if err := validatePolicyIsNotAttestationKind(policy); err != nil {
+				return NewErrValidation(err)
+			}
 		}
 	}
 	for _, gatt := range c.Schema.GetPolicyGroups() {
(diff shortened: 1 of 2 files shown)


app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -640,6 +647,25 @@ func SchemaToRawContract(contract *schemav1.CraftingSchema) (*Contract, error) {
...
+func validatePolicyIsNotAttestationKind(policy *schemav1.Policy) error {
+	policies := policy.GetSpec().GetPolicies()
+	if len(policies) == 0 {
+		// Legacy format or no policies defined
+		return nil
+	}
+
+	// Check if any policy has kind ATTESTATION - this is not allowed for material-level policies
+	for _, policySpec := range policies {
+		if policySpec.GetKind() == schemav1.CraftingSchema_Material_ATTESTATION {
```
migmartri, line 661: we need to review this because we support providing ATTESTATION as material type. But maybe the key here is to make sure it's not the only option. cc/ @jiparis

jiparis: ATTESTATION policies can be designed to run against an attestation material, or against the attestation itself (or both), but I don't think we have enough information at this point to know it, even if it's the only execution path. 

In my opinion, we have two options:
* change the semantics so that ATTESTATION materials are not allowed anymore (what this PR is implementing).
* or some kind of metadata in the policy that allow policy authors to specify how the policy should be run.

Piskoo: We could also "double check" if the users' intention was to apply attestation policy to a material, to do that we could require a selector

migmartri: What policies do we have that use an attestation as the material?

Issue #2488, opened 2025-10-28
Unable to attest Helm chart with empty values file


JeffResc:
I am attempting to attest a Helm chart with a values file consisting entirely of a YAML comment and no actual values

```yaml
# values.yaml
```

While this is valid with Helm, attempting to attest the chart with the Chainloop CLI results in an error
```
ERR adding material: crafting material: invalid values.yaml file: failed to unmarshal YAML file: EOF
```

Piskoo:
This will be fixed in the next release. Thanks for reporting!

PR #2491, opened 2025-10-29
fix(helm-chart): allow empty values

Piskoo:
Fixed an issue where attesting Helm charts with empty `values.yaml` files would fail with an EOF error, even though this is valid according to Helm specifications.

The validation logic now distinguishes between `Chart.yaml` (requires content) and `values.yaml` (can be empty).
Closes https://github.com/chainloop-dev/chainloop/issues/2488


jiparis:
Thanks for the quick fix!

PR #2494, opened 2025-10-29
feat(crafting-schema): update default contract

Piskoo:
This PR updates default contract schema used during workflow creation when no schema is given and contract creation when no schema is given to schema v2.

### DIFF
--- a/app/controlplane/pkg/biz/workflowcontract.go
+++ b/app/controlplane/pkg/biz/workflowcontract.go
@@ -210,8 +210,8 @@ type WorkflowContractCreateOpts struct {
 	AddUniquePrefix bool
 }
 
-// CreateDefaultContract creates a new default contract with the specified name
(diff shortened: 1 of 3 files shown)


app/controlplane/pkg/biz/workflowcontract.go
```diff
@@ -209,9 +210,29 @@ type WorkflowContractCreateOpts struct {
 	AddUniquePrefix bool
 }
 
-// EmptyDefaultContract is the default contract that will be created if no contract is provided
-var EmptyDefaultContract = &Contract{
-	Raw: []byte("schemaVersion: v1"), Format: unmarshal.RawFormatYAML,
+// CreateDefaultContract creates a new default contract with the specified name
+func CreateDefaultContract(name string) (*Contract, error) {
```
migmartri: you probably don't need to expose this function

migmartri (approved):
Thanks

PR #2481, opened 2025-10-24
feat(crafter): support multiple tools in chainloop metadata

Piskoo:
This PR adds support for multiple tools extraction to metadata, while maintaining backwards compatibility.

Example
```
# legacy keys (first found tool)
chainloop.material.tool.name: Hub
chainloop.material.tool.version: 2025.4.2
# new key (all found tools)
chainloop.material.tools: [
  "Hub@2025.4.2",
  "cyclonedx-core-java@5.0.5"
]
```

PR #2502, opened 2025-10-30
fix(attestation): material annotations output

Piskoo:
Fixed regression introduced in https://github.com/chainloop-dev/chainloop/pull/2481 where legacy annotations were filtered out in `att add` and `att status`.

### DIFF
--- a/app/cli/cmd/attestation_add.go
+++ b/app/cli/cmd/attestation_add.go
@@ -29,7 +29,6 @@ import (
 	"github.com/chainloop-dev/chainloop/app/cli/cmd/output"
 	"github.com/chainloop-dev/chainloop/app/cli/pkg/action"
 	schemaapi "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
-	"github.com/chainloop-dev/chainloop/pkg/attestation/crafter/materials"
 	"github.com/chainloop-dev/chainloop/pkg/resourceloader"
 )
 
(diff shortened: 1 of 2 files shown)


Piskoo:
Should we introduce new annotation for materials that use only one tool as well?
cc/ @javirln @migmartri @jiparis

migmartri:
Sorry I don't have enough context, what happened, what was lost?

Piskoo:
> Sorry I don't have enough context, what happened, what was lost?

I've added new ```chainloop.material.tools``` annotation for cdx and spdx files and filtered out ```chainloop.material.tools.Version/Name``` annotations in `att init/status` output, but crafter for some other material types also extracts tool used to the annotation, therefore the issue and my follow up question.

app/cli/cmd/attestation_add.go
```diff
@@ -195,9 +194,6 @@ func displayMaterialInfo(status *action.AttestationStatusMaterial, policyEvaluat
 	if len(status.Material.Annotations) > 0 {
 		mt.AppendRow(table.Row{"Annotations", "------"})
 		for _, a := range status.Material.Annotations {
-			if materials.IsLegacyAnnotation(a.Name) {
```
jiparis, line 198: Please also remove this method, since it's not used anymore.

jiparis:
> Should we introduce new annotation for materials that use only one tool as well? cc/ @javirln @migmartri @jiparis

Do you mean in the command output? If the output is going to be duplicated, I don't think so.

PR #2504, opened 2025-10-31
feat: add endpoint parameter to azure-blob cas-backend

### LINKED ISSUE #2495: Support for Azure Government Cloud in azure-blob CAS backend
### Summary
Currently, the `chainloop cas-backend add azure-blob` command only works with the default commercial Azure environment (`*.blob.core.windows.net`).  
When attempting to register an Azure Blob backend in the **Azure Government Cloud** (`*.blob.core.usgovcloudapi.net`), Chainloop fails with DNS resolution errors:

```
failed to create Blob storage Container: 
Put "https://<account>.blob.core.windows.net/...": 
dial tcp: lookup <account>.blob.core.windows.net: no such host
```

This happens because Chainloop always assumes the public Azure cloud endpoint and does not provide a way to specify an alternate endpoint or cloud environment.

### Steps to Reproduce
```bash
chainloop cas-backend add azure-blob \
  --client-id <client-id> \
  --client-secret <secret> \
  --tenant <tenant-id> \
  --storage-account <gov-storage-account> \
  --container chainloop-cas \
  --name azure-blob-usgov
```

**Result:**  
Fails with DNS error because it tries to reach `.blob.core.windows.net` instead of `.blob.core.usgovcloudapi.net`.

### Expected Behavior
Allow specifying Azure sovereign cloud environments, at minimum, the **Azure Government** endpoint (`core.usgovcloudapi.net`), so that users with US Gov storage accounts can register them as valid CAS backends.

### Proposed Solution
Add one of the following:
- A new flag, e.g.:
  - `--endpoint-suffix core.usgovcloudapi.net`
  - or `--cloud azuregovernment`

JeffResc:
Adds `--endpoint` option to `chainloop cas-backend add azure-blob` command.

For example:
```
./chainloop cas-backend add azure-blob \
  --client-id [CLIENT_ID] \
  --client-secret [CLIENT_SECRET] \
  --container [CONTAINER] \
  --storage-account [STORAGE_ACCOUNT] \
  --tenant [TENANT] \
  --name az-blob \
  --endpoint blob.core.usgovcloudapi.net
```

Resolves #2495

### DIFF
--- a/app/cli/cmd/casbackend_add_azureblob.go
+++ b/app/cli/cmd/casbackend_add_azureblob.go
@@ -102,7 +102,7 @@ func newCASBackendAddAzureBlobStorageCmd() *cobra.Command {
 
 	cmd.Flags().StringVar(&container, "container", "chainloop", "Storage Container Name")
 
-	cmd.Flags().StringVar(&endpoint, "endpoint", "", "Azure Blob endpoint suffix (e.g., blob.core.windows.net, blob.core.usgovcloudapi.net)")
+	cmd.Flags().StringVar(&endpoint, "endpoint", "", "Custom Azure Blob endpoint suffix (e.g., blob.core.usgovcloudapi.net), if not provided, the public Azure cloud endpoint will be used.")
 
 	return cmd
 }
--- a/app/cli/documentation/cli-reference.mdx
+++ b/app/cli/documentation/cli-reference.mdx
@@ -713,7 +713,7 @@ Options
 --client-id string         Service Principal Client ID
 --client-secret string     Service Principal Client Secret
 --container string         Storage Container Name (default "chainloop")
---endpoint string          Azure Blob endpoint suffix (e.g., blob.core.windows.net, blob.core.usgovcloudapi.net)
+--endpoint string          Custom Azure Blob endpoint suffix (e.g., blob.core.usgovcloudapi.net), if not provided, the public Azure cloud endpoint will be used.
 -h, --help                     help for azure-blob
 --storage-account string   Storage Account Name
 --tenant string            Active Directory Tenant ID


migmartri:
Awesome, we'll take a look at it shortly, note that some automated tests are failing.

Just to confirm, did you manage to make this feature work end-to-end locally? 

Thanks!

jiparis:
Thanks for the contribution @JeffResc, it looks great. For the CI job to pass, you might need to run `go generate ./...` in `app/cli` folder to generate the CLI documentation.

JeffResc:
> Just to confirm, did you manage to make this feature work end-to-end locally?

Yes I did, I was able to successfully add the cas-backend and push attestations to it

> For the CI job to pass, you might need to run go generate ./... in app/cli folder to generate the CLI documentation.

Done, hopefully that resolves the CI errors

app/cli/cmd/casbackend_add_azureblob.go
```diff
@@ -97,5 +102,7 @@ func newCASBackendAddAzureBlobStorageCmd() *cobra.Command {
 
 	cmd.Flags().StringVar(&container, "container", "chainloop", "Storage Container Name")
 
+	cmd.Flags().StringVar(&endpoint, "endpoint", "", "Azure Blob endpoint suffix (e.g., blob.core.windows.net, blob.core.usgovcloudapi.net)")
```
migmartri: nitpick.

Isn't `blob.core.windows.net` in fact the default and doesn't need to be provided? If so, the example provided in the flag can be somewhat misleading. 

Maybe updating the example to `Custom Azure Blob endpoint suffix (e.g., blob.core.usgovcloudapi.net), if not provided, the public Azure cloud endpoint will be used.`

JeffResc: Yes, `blob.core.windows.net` is the default and will be used if the flag is not provided, I just wanted to give more than one example, but I can see how this can cause confusion. I'll update the description!

PR #2507, opened 2025-11-01
feat: add support for container images from local oci layout

### LINKED ISSUE #1653: can I attest. a local, non-pushed container image?
```
docker build -t my-new-image .
```

```
chainloop att add --value my-new-image --kind CONTAINER_IMAGE
ERR adding material: crafting material: GET https://index.docker.io/v2/library/my-new-image/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/my-new-image Type:repository]]
```

JeffResc:
### Summary

Enables attestation of container images stored in local [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) directories, allowing users to attest images without requiring a remote registry or network access.

### Motivation

Users need to attest container images in several scenarios where remote registries are unavailable or impractical:

1. **Air-gapped environments**: Secure or isolated networks without internet access where images cannot be pulled from or pushed to external registries
2. **Registry-less deployments**: Environments where images are distributed via OCI layout bundles rather than traditional registries

### Implementation Overview

Auto-detection: The implementation automatically detects OCI layout directories by checking for the presence of an `oci-layout` file (per the [OCI Image Layout Specification v1.0.0](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)).

**Digest selector syntax**: Uses the `@digest` suffix format to select specific images from multi-image layouts, following the convention established by Skopeo and other OCI tooling (e.g., `path/to/layout@digest`).

**Safety**: When multiple images exist in a layout without a digest selector, the implementation returns an error listing available digests to prevent accidental wrong image selection.

### Usage

```bash
# Single image layout (automatic)
chainloop attestation add \
  --name my-app \
  --value /path/to/oci-layout \
  --kind CONTAINER_IMAGE

# Multi-image layout (requires digest selector)
chainloop attestation add \
  --name my-app \
  --value /path/to/oci-layout@sha256:9a7ef86e19... \
  --kind CONTAINER_IMAGE
```

### Implementation Details

- Extracts image digest from OCI layout index
- Reads repository name and tag from [OCI image annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md) (`org.opencontainers.image.ref.name`, `io.containerd.image.name`)
- Detects Cosign and Notary signatures within the layout

### References

- [OCI Image Layout Specification](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
- [OCI Image Annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md)
- Implementation uses [go-containerregistry/pkg/v1/layout](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/layout)

### Related
Resolves #1653

### DIFF
--- a/pkg/attestation/crafter/materials/oci_image_test.go
+++ b/pkg/attestation/crafter/materials/oci_image_test.go
@@ -72,19 +72,19 @@ func TestOCIImageCraft_Layout(t *testing.T) {
 		{
 			name:       "crane - single image with annotations",
 			layoutPath: "testdata/oci-layouts/crane",
-			wantName:   "oci-layout",
+			wantName:   "oci-layout:unknown",
 			wantDigest: "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400",
 		},
 		{
 			name:       "skopeo - single image with tag annotation",
 			layoutPath: "testdata/oci-layouts/skopeo",
-			wantName:   "v1.51.0",
+			wantName:   "oci-layout:v1.51.0",
 			wantDigest: "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400",
 		},
 		{
 			name:       "skopeo-alt - alternative format",
 			layoutPath: "testdata/oci-layouts/skopeo-alt",
-			wantName:   "v1.51.0",
+			wantName:   "oci-layout:v1.51.0",
 			wantDigest: "sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88",
 		},
 		{
@@ -142,21 +142,21 @@ func TestOCIImageCraft_LayoutWithDigestSelector(t *testing.T) {
 			name:           "oras - select first image by digest",
 			layoutPath:     "testdata/oci-layouts/oras",
 			digestSelector: "sha256:b1747c197a0ab3cb89e109f60a3c5d4ede6946e447fd468fa82d85fa94c6c6e5",
-			wantName:       "oci-layout",
+			wantName:       "oci-layout:unknown",
 			wantDigest:     "sha256:b1747c197a0ab3cb89e109f60a3c5d4ede6946e447fd468fa82d85fa94c6c6e5",
 		},
 		{
 			name:           "oras - select second image by digest",
 			layoutPath:     "testdata/oci-layouts/oras",
 			digestSelector: "sha256:f333056ac987169b2a121c16d06112d88ec3d7cb50b098bb17b0f14b0c52f6f3",
-			wantName:       "oci-layout",
+			wantName:       "oci-layout:unknown",
 			wantDigest:     "sha256:f333056ac987169b2a121c16d06112d88ec3d7cb50b098bb17b0f14b0c52f6f3",
 		},
 		{
 			name:           "zarf - select specific image from bundle",
 			layoutPath:     "testdata/oci-layouts/zarf",
 			digestSelector: "sha256:e8ac056f7b9b44b07935fe23b8383e5e550d479dc5c6261941e76449a8f7e926",
-			wantName:       "ghcr.io/chainloop-dev/chainloop/artifact-cas:v1.51.0",
+			wantName:       "oci-layout:ghcr.io/chainloop-dev/chainloop/artifact-cas:v1.51.0",
 			wantDigest:     "sha256:e8ac056f7b9b44b07935fe23b8383e5e550d479dc5c6261941e76449a8f7e926",
 		},
 		{


pkg/attestation/crafter/materials/oci_image_test.go
```diff
@@ -0,0 +1,295 @@
+//
+// Copyright 2024 The Chainloop Authors.
```
migmartri: 2025 :)

migmartri:
This is a great contribution @JeffResc, thanks for in fact finding the automatically closed PR.

I am trying to wrap my head around this feature, because as you might know, we have some kinds of pieces of evidence that are attested by reference (i.e `OCI CONTAINER IMAGE`, or `OCI Helm Chart`), and others uploaded to the content addressable storage and referenced by the uploaded digest (i.e Helm Chart tarball, tarball, SBOM, etc). My first question is whether this local image should be just referenced or uploaded to CAS? 

If we continue in the route of referencing it by just reference, like what you implemented. Shouldn't the result of the attestation process be the same regardless of pointing to a remote OCI image or to the local layout? 

For example, I am performing an attestation that

1 - Adds a remote image

```
 ./bin/chainloop att add --name remote-image --value ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0
INF material kind detected kind=CONTAINER_IMAGE
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ remote-image                                                            │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0                   │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

then I pull the same image with skopeo and use the new local layout feature

```sh
skopeo copy --multi-arch all docker://ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0  oci:chainloop-controlplane:v1.51.0
Getting image list signatures
Copying 2 images generated from 2 images in the list
...
```

and add it to the attestation

```
./bin/chainloop att add --name local-image --value ./chainloop-controlplane
INF material kind detected kind=CONTAINER_IMAGE
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ local-image                                                             │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ v1.51.0                                                                 │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```
 
with the final result as

```
./bin/chainloop att push
INF push completed
│ Initialized At            │ 01 Nov 25 12:19 UTC                                                     │
│ Attestation ID            │ b99cb649-691c-4756-b8ba-69919856bde8                                    │
│ Digest                    │ sha256:9bc1cd5bc014fb1474bda5a4b7bf366e4322a085af3e66fb6dc723bafa5de3f4 │
│ Organization              │ miguel-test                                                             │
│ Name                      │ container-image-test-local                                              │
│ Project                   │ myproject                                                               │
│ Version                   │ v1.51.0 (prerelease)                                                    │
│ Contract                  │ myproject-container-image-test-local (revision 1)                       │
│ Timestamp Authority       │ http://timestamp.digicert.com                                           │
┌────────────────────────────────────────────────────────────────────────────────────┐
│ Materials                                                                          │
├──────────┬─────────────────────────────────────────────────────────────────────────┤
│ Name     │ local-image                                                             │
│ Type     │ CONTAINER_IMAGE                                                         │
│ Set      │ Yes                                                                     │
│ Required │ No                                                                      │
│ Value    │ v1.51.0                                                                 │
│ Digest   │ sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88 │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Name     │ remote-image                                                            │
│ Type     │ CONTAINER_IMAGE                                                         │
│ Set      │ Yes                                                                     │
│ Required │ No                                                                      │
│ Value    │ ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0                   │
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

where we can see that the remote URL is lost and the digests are different. Are these by design? 

I am not that familiar with the layout spec, so that's why I am asking if at least the digest can be preserved so whenever the image gets pushed, the internal chainloop referrer API will be able to connect the dots.

Thanks again!

JeffResc:
Hey @migmartri 👋

The behavior you're seeing comes down to a slight nuance in the OCI spec. With the current implementation via OCI registry (`chainloop att add --kind CONTAINER_IMAGE --value ghcr.io/chainloop-dev/chainloop/artifact-cas: v1.51.0`), we're attesting a manifest list digest (or image index). This is more or less a manifest of all images with that tag, specifically pointers to the different architectures that are available whereas the `skopeo copy` command you're running is actually pulling the images for both amd64 and arm64.

~~I'm unsure where Skopeo is getting that `sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88` digest, as it doesn't seem to actually exist in the remote OCI registry:~~ See [next comment](https://github.com/chainloop-dev/chainloop/pull/2507#issuecomment-3476548487) for details
```
❯ crane pull --format=oci ghcr.io/chainloop-dev/chainloop/control-plane@sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88 /tmp/sha-test  
Error: GET https://ghcr.io/v2/chainloop-dev/chainloop/control-plane/manifests/sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88: MANIFEST_UNKNOWN: manifest unknown
```

I tested this with crane, which by default pulls the manifest list digest that Chainloop pulls from remote OCI registries and the resulting digest is the same as from the remote OCI registry:
```
❯ crane pull --format=oci ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0 /tmp/local-oci   
❯ ./chainloop att add --name local-oci --kind CONTAINER_IMAGE --value /tmp/local-oci   
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ local-oci                                                               │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ oci-layout                                                              │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
❯ ./chainloop att add --name remote-oci --kind CONTAINER_IMAGE --value ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ remote-oci                                                              │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0                   │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

Additionally, I was able to complete the same test with oras:
```
❯ oras copy ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0 --to-oci-layout control-plane-oras
Copied [registry] ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0 => [oci-layout] control-plane-oras
Digest: sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400
❯ ./chainloop att add --kind CONTAINER_IMAGE --name remote-oci-image --value ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ remote-oci-image                                                        │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0                   │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
❯ ./chainloop att add --kind CONTAINER_IMAGE --name local-oci-oras --value control-plane-oras@sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ local-oci-oras                                                          │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ oci-layout                                                              │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

JeffResc:
After some additional digging, it looks like Skopeo was automatically converting from Docker OCI format to generic OCI image format, resulting in the same image contents, but a different digest. Here's the correct Skopeo command that preserves the image digest:
```
❯ skopeo copy --multi-arch all --format v2s2 docker://ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0 oci:chainloop_control_plane:v1.51.0
Getting image list signatures
Copying 2 images generated from 2 images in list
Copying image sha256:f333056ac987169b2a121c16d06112d88ec3d7cb50b098bb17b0f14b0c52f6f3 (1/2)
Getting image source signatures
Copying blob b6229e75e504 done   |
Copying blob b3fd6151a200 done   |
Copying blob 902df98899e3 done   |
Copying blob b89f16251cb9 done   |
Copying blob 6f4bd196d069 done   |
Copying blob 91d9b18ed9ef done   |
Copying config 5bf4815b17 done   |
Writing manifest to image destination
Copying image sha256:b1747c197a0ab3cb89e109f60a3c5d4ede6946e447fd468fa82d85fa94c6c6e5 (2/2)
Getting image source signatures
Copying blob b89f16251cb9 skipped: already exists
Copying blob b6229e75e504 skipped: already exists
Copying blob 91d9b18ed9ef skipped: already exists
Copying blob 902df98899e3 skipped: already exists
Copying blob 6f4bd196d069 skipped: already exists
Copying blob b3fd6151a200 skipped: already exists
Copying config 845f696ea6 done   |
Writing manifest to image destination
Writing manifest list to image destination
Storing list signatures
❯ ./chainloop att add --kind CONTAINER_IMAGE --value chainloop_control_plane
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ material-1762015274092271000                                            │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ v1.51.0                                                                 │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
❯ ./chainloop att add --kind CONTAINER_IMAGE --value ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ material-1762015283906368000                                            │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ CONTAINER_IMAGE                                                         │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0                   │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400 │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
```

migmartri:
Thanks, that's very useful. 

How does the value get obtained? 

I can see that ORAS produced `oci-layout`, whereas with skopeo, it seems to be the tag of the layout set during the copy.

image

I am wondering how we should indicate that this is a local layout. I think we should probably adopt a similar FQDN to skopeo `oci:path:tag. An image tag in a directory compliant with "Open Container Image Layout Specification" at path.` But open to other suggestions, wdyt?

JeffResc:
The current display name comes from the `org.opencontainers.image.ref.name` annotation in the `index.json` file.

Of the three tools, skopeo seems to be the only one setting this value, while oras and crane don't have any annotation or way to get the original image name, so it defaults to `oci-layout`. I'm open to implementing `oci:path:tag` like skopeo does, but its ultimately down to the tool to provide some sort of annotation to be able to tell what the image is.

```
❯ cat skopeo/index.json | jq '.manifests[].annotations'
{
  "org.opencontainers.image.ref.name": "v1.51.0"
}
```

```
❯ cat crane/index.json | jq '.manifests[].annotations'
null
```

```
❯ cat oras/index.json | jq '.manifests[].annotations'
null
null
null
```

migmartri:
> The current display name comes from the `org.opencontainers.image.ref.name` annotation in the `index.json` file.
> 
> Of the three tools, skopeo seems to be the only one setting this value, while oras and crane don't have any annotation or way to get the original image name, so it defaults to `oci-layout`. I'm open to implementing `oci:path:tag` like skopeo does, but its ultimately down to the tool to provide some sort of annotation to be able to tell what the image is.
> 
> ```
> ❯ cat skopeo/index.json | jq '.manifests[].annotations'
> {
>   "org.opencontainers.image.ref.name": "v1.51.0"
> }
> ```
> 
> ```
> ❯ cat crane/index.json | jq '.manifests[].annotations'
> null
> ```
> 
> ```
> ❯ cat oras/index.json | jq '.manifests[].annotations'
> null
> null
> null
> ```

Thanks, Jeff, for again looking into this so thoroughly. One of the core (not written down though :() principles of Chainloop is to take the fragmentation generated by different practices and tools, and introduce our normalization layer and opinion. The higher you go in the stack, the more normalized the data, and hence the easier it is to evaluate. 

So my gut feeling tells me that we should do the same here. To have an opinion on what an identifier for this kind of image should look like. It might be as simple as `oci:[identifier from the tool, or default]` or `oci-layout:[identifier of the tool or default]`

In any case, I am open, so I'd be happy to get your take on this.

## Digest mismatch 

Ideally, we should ensure that the digest of the image, as attested using the layout, matches once the image is pushed, but it seems that's not within our control, correct? We have no way of knowing what kind of OCI registry the image will be pushed to, and therefore, we cannot determine whether it will support the OCI spec or Docker V2 spec by default. Am I understanding this correctly? 

How could we improve this? Is this a documentation issue? Should we enforce a specific spec type (to my point above, probably not)

pkg/attestation/crafter/materials/oci_image.go
```diff
@@ -281,3 +293,188 @@ func (i *OCIImageCrafter) isLatestTag(ref name.Reference, currentDigest string)
...
+				break
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("digest %s not found in OCI layout", digestSelector)
+		}
+		i.logger.Debug().Str("digest", digestSelector).Msg("selected image by digest")
+	} else {
+		// If multiple images exist, require explicit selection to avoid mistakes
+		if len(indexManifest.Manifests) > 1 {
```
migmartri: This might be refactored more as a guard instead of as if/else, meaning checking the condition at the top, potentially setting the first manifest by default, etc.

pkg/attestation/crafter/materials/oci_image.go
```diff
@@ -281,3 +293,188 @@ func (i *OCIImageCrafter) isLatestTag(ref name.Reference, currentDigest string)
...
+		}
+		// Only one image, safe to use it
+		manifest = indexManifest.Manifests[0]
+		i.logger.Debug().Msg("using only image in layout")
+	}
+
+	digest := manifest.Digest.String()
+
+	// Extract repository name from annotations if available
+	repoName := "oci-layout"
```
migmartri: nitpick, use a constant

pkg/attestation/crafter/materials/oci_image.go
```diff
@@ -281,3 +293,188 @@ func (i *OCIImageCrafter) isLatestTag(ref name.Reference, currentDigest string)
...
+	if manifest.Annotations != nil {
+		if name, ok := manifest.Annotations["org.opencontainers.image.ref.name"]; ok {
```
migmartri: so is there any instance where we could get the original remote OCI repoName i.e `chainloop-dev/controlplane-image` where the image was pulled from?

pkg/attestation/crafter/materials/oci_image_test.go
```diff
@@ -0,0 +1,295 @@
+//
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri: 2025 this time since this file has been created this year. We should probably document this in the contribution guide, sorry.

pkg/attestation/crafter/materials/oci_image_test.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+type imageSpec struct {
+	name string
+	tag  string
+}
+
+// createTestOCILayoutMultiple creates an OCI layout with multiple images for testing
+func createTestOCILayoutMultiple(t *testing.T, specs []imageSpec) (string, []string) {
```
migmartri: alternatively you could add image layouts to `testdata` and read it from there, that's what I believe we do for most other material types, have an offline, actual piece of evidence. 

If you go that route, you could even add different copies from different tools, if you see value.

JeffResc:
> To have an opinion on what an identifier for this kind of image should look like. It might be as simple as `oci:[identifier from the tool, or default]` or `oci-layout:[identifier of the tool or default]`

I think this is a great idea and definitely worth normalizing the data being attested. The problem is, most of the popular tools don't include any additional information in the oci layout directory to add any additional information. The only that seems to be supported by every tool is including the digest information and mediaType. Of the tested tools, only skopeo and zarf include any annotation and the annotation provided by skopeo may not be too helpful as its just whatever information the user provided in the copy command.

So I am onboard with the idea, I'm just not sure what information we have available besides the digest, which is already recorded in the attestation.

In this case, does it make sense to do `oci-layout:[org.opencontainers.image.ref.name if provided otherwise default]`?

These are the various `index.json` files produced by popular tools that create oci layout directories, as you'll see, there's not much info to work with here:
<details>
  <summary>crane - index.json</summary>

  ```json
  {
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
      {
        "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
        "size": 743,
        "digest": "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400"
      }
    ]
  }
````

</details>

<details>
  <summary>oras - index.json</summary>

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:b1747c197a0ab3cb89e109f60a3c5d4ede6946e447fd468fa82d85fa94c6c6e5",
      "size": 1578,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:f333056ac987169b2a121c16d06112d88ec3d7cb50b098bb17b0f14b0c52f6f3",
      "size": 1578,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
      "digest": "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400",
      "size": 743
    }
  ]
}
```

</details>

<details>
  <summary>skopeo - index.json</summary>

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
      "digest": "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400",
      "size": 743,
      "annotations": {
        "org.opencontainers.image.ref.name": "v1.51.0"
      }
    }
  ]
}
```

</details>

<details>
  <summary>zarf - index.json</summary>

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:8cb1d2d12cb680d67998ae917462779ba35ff6969ddc09fc6cdf825889bcd357",
      "size": 737,
      "annotations": {
        "org.opencontainers.image.base.name": "ghcr.io/chainloop-dev/chainloop/control-plane-migrations:v1.51.0",
        "org.opencontainers.image.ref.name": "ghcr.io/chainloop-dev/chainloop/control-plane-migrations:v1.51.0"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:e8ac056f7b9b44b07935fe23b8383e5e550d479dc5c6261941e76449a8f7e926",
      "size": 739,
      "annotations": {
        "org.opencontainers.image.base.name": "ghcr.io/chainloop-dev/chainloop/artifact-cas:v1.51.0",
        "org.opencontainers.image.ref.name": "ghcr.io/chainloop-dev/chainloop/artifact-cas:v1.51.0"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:f333056ac987169b2a121c16d06112d88ec3d7cb50b098bb17b0f14b0c52f6f3",
      "size": 1578,
      "annotations": {
        "org.opencontainers.image.base.name": "ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0",
        "org.opencontainers.image.ref.name": "ghcr.io/chainloop-dev/chainloop/control-plane:v1.51.0"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```

</details>

> Ideally, we should ensure that the digest of the image, as attested using the layout, matches once the image is pushed, but it seems that's not within our control, correct? We have no way of knowing what kind of OCI registry the image will be pushed to, and therefore, we cannot determine whether it will support the OCI spec or Docker V2 spec by default. Am I understanding this correctly?

Yes, that's correct. You can see here these two manifests that were pulled by skopeo, one with `--format v2s2` and one without. You'll notice the difference in mediaType and digest even though its the same underlying image:
```
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88",
      "size": 493,
      "annotations": {
        "org.opencontainers.image.ref.name": "v1.51.0"
      }
    }
  ]
}
```
```
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
      "digest": "sha256:fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400",
      "size": 743,
      "annotations": {
        "org.opencontainers.image.ref.name": "v1.51.0"
      }
    }
  ]
}
```

> Should we enforce a specific spec type (to my point above, probably not)

We could enforce the mediaType, but technically both are valid and I don't want to limit the use cases. If you want the digest to exactly match the registry, you'd have to do this when you copy from the registry to the local oci layout with the tool, I don't think it's possible to get the registry digest after the fact. In this case, the best we may be able to do is provide some documentation on how this all works and what commands to use with the popular tools to get the digest to match the one in the registry.

pkg/attestation/crafter/materials/oci_image.go
JeffResc: Unfortunately, it doesn't seem so unless the tool provided `org.opencontainers.image.ref.name`. See [this comment](https://github.com/chainloop-dev/chainloop/pull/2507#issuecomment-3478700992) for more details, but I can continue to research

migmartri:
> > To have an opinion on what an identifier for this kind of image should look like. It might be as simple as `oci:[identifier from the tool, or default]` or `oci-layout:[identifier of the tool or default]`
> 
> I think this is a great idea and definitely worth normalizing the data being attested. The problem is, most of the popular tools don't include any additional information in the oci layout directory to add any additional information. The only that seems to be supported by every tool is including the digest information and mediaType. Of the tested tools, only skopeo and zarf include any annotation and the annotation provided by skopeo may not be too helpful as its just whatever information the user provided in the copy command.
> 
> So I am onboard with the idea, I'm just not sure what information we have available besides the digest, which is already recorded in the attestation.
> 
> In this case, does it make sense to do `oci-layout:[org.opencontainers.image.ref.name if provided otherwise default]`?

I'd say yes, the most important part to me tbh is the "protocol part" so it doesn't get confused with existing remote OCI reference or eventually other local reference formats (if any)

in fact, the original request https://github.com/chainloop-dev/chainloop/issues/1653 was about attesting an image just after building it, how would you do it in this case? Does this mean that the local image somehow needs to be converted to layout? I am asking because it's likely we'll need to document it.

pkg/attestation/crafter/materials/oci_image_test.go
```diff
@@ -0,0 +1,217 @@
...
+		name       string
+		layoutPath string
+		wantErr    string
+		wantDigest string
+		wantName   string
+		wantTag    string
+	}{
+		{
+			name:       "crane - single image with annotations",
+			layoutPath: "testdata/oci-layouts/crane",
```
migmartri, line 74: nice :)

migmartri (approved):
LGTM, modulo my comment re: making sure we have the protocol set :)

Thanks again, and once merged we can perform a release of Chainloop

JeffResc:
> in fact, the original request https://github.com/chainloop-dev/chainloop/issues/1653 was about attesting an image just after building it, how would you do it in this case? Does this mean that the local image somehow needs to be converted to layout? I am asking because it's likely we'll need to document it.

Unfortunately the Docker CLI doesn't have native support for OCI, but you can use something like Skopeo to convert from a Docker archive to oci-layout
```
❯ docker save rancher/k3s:v1.32.5-k3s1 -o k3s.tar
❯ skopeo copy docker-archive:k3s.tar oci:./k3s-oci:v1.32.5-k3s1
Getting image source signatures
Copying blob 1f8508c67980 done   |
Copying config ec995785b7 done   |
Writing manifest to image destination
❯ cat index.json | jq
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:43f14ea005c1b7346fcaff2a46a16639d74d55245eaf05137ba1668b9815df3d",
      "size": 407,
      "annotations": {
        "org.opencontainers.image.ref.name": "v1.32.5-k3s1"
      }
    }
  ]
}
```

Now it might be outside of the scope of this PR, but due to this type of limitation, does it make sense for BOTH the `--value` of the attestation and the shown attested value to adopt something like Skopeo's prefix system (i.e., `docker-archive:`, `oci-layout:`, `docker:`, `docker-daemon:`)? That way, in the future, there could be support added for those other types such as attesting directly from a docker-archive or docker-daemon.

JeffResc:
> I'd say yes, the most important part to me tbh is the "protocol part" so it doesn't get confused with existing remote OCI reference or eventually other local reference formats (if any)

I just updated the logic to add `oci-layout:` as the prefix, then try for the value of `org.opencontainers.image.ref.name`, then `org.opencontainers.image.base.name`. If neither annotation is found, the value is currently left as `oci-layout:`, but I could update it to include some default value like `oci-layout:unknown` if its not able to find a suitable annotation

migmartri:
> Now it might be outside of the scope of this PR, but due to this type of limitation, does it make sense for BOTH the --value of the attestation and the shown attested value to adopt something like Skopeo's prefix system (i.e., docker-archive:, oci-layout:, docker:, docker-daemon:)? 

You mean that you will need to provide the prefix in the `--value` ? Now that you mention it, that's what we do in the Helm Chart, where you need ot provide `oci:` if you want to do a remote one. 

https://github.com/chainloop-dev/chainloop/blob/8bcddcf01439d8ae6c4908726e8d1e0583f654c1/pkg/attestation/crafter/materials/helmchart.go#L70-L83

We might want to do the same here just for the sake of consistency, that said, the reason we needed to add this in the chart is because we couldn't always do a remote `OCI check`, it's very expensive, so in some way I like your approach more, more user friendly fwiw.

pkg/attestation/crafter/materials/oci_image.go
```diff
@@ -281,3 +295,204 @@ func (i *OCIImageCrafter) isLatestTag(ref name.Reference, currentDigest string)
...
+	return i.buildMaterialFromManifest(layoutPath, manifest, indexManifest.Manifests)
+}
+
+// buildMaterialFromManifest constructs the attestation material from a manifest descriptor.
+func (i *OCIImageCrafter) buildMaterialFromManifest(layoutPath string, manifest v1.Descriptor, allManifests []v1.Descriptor) (*api.Attestation_Material, error) {
+
+	digest := manifest.Digest.String()
+
+	// Extract repository name from annotations if available
+	repoName := ociLayoutRepoName + ":"
```
migmartri, line 397: as you mentioned before, I'd personally add `unknown` as fallback, that way we always have two parts.

JeffResc:
> We might want to do the same here just for the sake of consistency, that said, the reason we needed to add this in the chart is because we couldn't always do a remote `OCI check`, it's very expensive, so in some way I like your approach more, more user friendly fwiw.

Got it, I was more thinking for future expansion if there was ever a need to support something such as `docker-daemon:` or `docker-archive:` instead of needing to convert to oci layout or push to oci first, but if you think this is more user friendly, I'm open to leaving the current implementation

migmartri:
@jiparis it would be great if you could take a look. Thanks!

jiparis:
Thanks @JeffResc I'm been trying this locally and works like a charm. Since it's already merged, I will create a task for further improvements. Mostly, I'd like us to discuss if we should leverage Chainloop annotations for adding metadata when it's available.
For example, this is the metadata we extract while crafting a remote OCI image, that it's included in the final attestation:
```json
      {
        "annotations": {
          "chainloop.material.image.is_latest_tag": true,
          "chainloop.material.image.tag": "latest",
          "chainloop.material.name": "my-image",
          "chainloop.material.signature": "...",
          "chainloop.material.signature.digest": "sha256:ebc5cd84e4895662e4b7a21b5f4d2e2656f1558e469a1ececea3804976b7b8d6",
          "chainloop.material.signature.provider": "cosign",
          "chainloop.material.type": "CONTAINER_IMAGE"
        },
        "digest": {
          "sha256": "fa6d9058c3d65a33ff565c0e35172f2d99e76fbf8358d91ffaa2208eff2be400"
        },
        "name": "ghcr.io/chainloop-dev/chainloop/control-plane"
      }
```
However, with this implementation for OCI layout, this is what we get:
```json
    {
        "annotations": {
          "chainloop.material.image.is_latest_tag": false,
          "chainloop.material.name": "material-1762196633776738000",
          "chainloop.material.type": "CONTAINER_IMAGE"
        },
        "digest": {
          "sha256": "a5303ef28a4bd9b6e06aa92c07831dd151ac64172695971226bdba4a11fc1b88"
        },
        "name": "oci-layout:v1.51.0"
      },
```
It would be interesting if we can identify some metadata like `chainloop.material.image.tag` or even create a new one for the format. Something like `"chainloop.material.image.format": "oci-layout"`

But as I said, this is secondary, I'll create a follow-up issue for this. Thanks!

PR #2528, opened 2025-11-07
add `skip_upload` capability to contract

### LINKED ISSUE #2522: Support attesting large artifacts by reference without CAS upload
Currently all material types (except `CONTAINER_IMAGE`) upload artifacts to CAS. This creates challenges for users who need to attest very large artifacts (multi-GB artifacts) due to storage costs, upload time, and bandwidth limitations.

Introduce a new material type `ARTIFACT_REF` that calculates the digest locally but doesn't upload to CAS. Users explicitly specify `--kind ARTIFACT_REF` when adding materials.

```bash
chainloop att add --name my-artifact --value /path/to/large-artifact.tar --kind ARTIFACT_REF
```

JeffResc:
### Description

This PR introduces the ability to control whether materials are uploaded to CAS directly from workflow contracts. This is useful for scenarios where artifacts are large and already stored elsewhere, or when you only need to attest to their existence without storing them.

Contracts can now specify `skip_upload: true` on individual materials to skip uploading them to CAS while still recording their metadata (digest and filename) in the attestation.

### Example Usage
```
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: my-workflow
spec:
  materials:
    - name: large-binary
      type: ARTIFACT
      skip_upload: true  # Only record metadata, don't upload to CAS
    - name: sbom
      type: SBOM_CYCLONEDX_JSON
      # skip_upload defaults to false - normal upload behavior
```

### Features
**Contract-controlled**: Platform administrators define upload behavior in contracts
**Metadata preservation**: Even when upload is skipped, digest and filename are recorded in attestations
**Backward compatible**: Defaults to false (upload enabled) - existing contracts work unchanged
**Flexible**: Can be set per-material, allowing fine-grained control within a single workflow

### Use Cases
- Large artifacts (multi-GB files) where CAS storage isn't practical
- Artifacts already stored in external systems
- Materials where attestation metadata is sufficient without full content storage
- Reducing storage costs for high-volume workflows

### Technical Details

- Added `skip_upload` boolean field to Material proto definition
- Updated material crafter to honor skip setting from contract
- When skipped, materials have `UploadedToCas = false` and only metadata is captured
- `CONTAINER_IMAGE` and `STRING` types remain unaffected (they never upload to CAS)

### Related Issues
Resolves #2522

### DIFF
--- a/pkg/attestation/crafter/materials/cyclonedxjson_test.go
+++ b/pkg/attestation/crafter/materials/cyclonedxjson_test.go
@@ -224,3 +224,70 @@ func TestCyclonedxJSONCraft(t *testing.T) {
 		})
 	}
 }
+
+func TestCycloneDXJSONCraft_SkipUpload(t *testing.T) {
+	testCases := []struct {
+		name       string
+		skipUpload bool
+		wantUpload bool
+	}{
+		{
+			name:       "upload enabled (default)",
+			skipUpload: false,
+			wantUpload: true,
+		},
+		{
+			name:       "upload skipped via contract",
+			skipUpload: true,
+			wantUpload: false,
+		},
+	}
+
+	ast := assert.New(t)
+	l := zerolog.Nop()
+	filePath := "./testdata/sbom.cyclonedx.json"
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			schema := &contractAPI.CraftingSchema_Material{
+				Name:       "test",
+				Type:       contractAPI.CraftingSchema_Material_SBOM_CYCLONEDX_JSON,
+				SkipUpload: tc.skipUpload,
+			}
+
+			// Mock uploader - only expect upload call if not skipped
+			uploader := mUploader.NewUploader(t)
+			if tc.wantUpload {
+				uploader.On("UploadFile", context.TODO(), filePath).
+					Return(&casclient.UpDownStatus{
+						Digest:   "deadbeef",
+						Filename: "sbom.cyclonedx.json",
+					}, nil)
+			}
+
+			backend := &casclient.CASBackend{Uploader: uploader}
+			crafter, err := materials.NewCyclonedxJSONCrafter(schema, backend, &l)
+			require.NoError(t, err)
+
+			got, err := crafter.Craft(context.TODO(), filePath)
+			require.NoError(t, err)
+
+			ast.Equal(contractAPI.CraftingSchema_Material_SBOM_CYCLONEDX_JSON.String(), got.MaterialType.String())
+
+			// Verify upload behavior matches expectation
+			if tc.wantUpload {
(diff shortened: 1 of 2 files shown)


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -92,6 +92,9 @@ message CraftingSchema {
     // List of annotations that can be used to add metadata to the material
     // this metadata can be used later on by the integrations engine to filter and interpolate data
     repeated Annotation annotations = 5 [deprecated = true];
+    // If true, skip uploading the material to CAS (only record metadata)
+    // Defaults to false (material will be uploaded)
+    bool skip_upload = 6 [deprecated = true];
```
jiparis: I guess this deprecation is not intended

JeffResc: Nice catch, yes I will address that

app/cli/cmd/attestation_add.go
```diff
@@ -191,9 +191,9 @@ func displayMaterialInfo(status *action.AttestationStatusMaterial, policyEvaluat
 		mt.AppendRow(table.Row{"Digest", status.Hash})
 	}
 
-	if len(status.Material.Annotations) > 0 {
```
migmartri, line 194: thanks for this cleanup :)

app/cli/pkg/action/attestation_reset.go
```diff
@@ -81,9 +81,10 @@ func (action *AttestationReset) Run(ctx context.Context, attestationID, trigger,
 }
 
 func parseTrigger(in string) pb.AttestationServiceCancelRequest_TriggerType {
-	if in == AttestationResetTriggerFailed {
+	switch in {
+	case AttestationResetTriggerFailed:
 		return pb.AttestationServiceCancelRequest_TRIGGER_TYPE_FAILURE
-	} else if in == AttestationResetTriggerCancelled {
```
migmartri, line 86: woah, this wasn't great :)

migmartri (approved):
LGTM in general.

My only comment would be if you could add a testCase in at least one of the material types (i.e `TestSARIFCraft`) we have today to make sure this new path is covered?

JeffResc:
> My only comment would be if you could add a testCase in at least one of the material types (i.e `TestSARIFCraft`) we have today to make sure this new path is covered?

Yep, just added one for SARIF and CycloneDX

migmartri:
LGTM!

migmartri:
@JeffResc it would be nice if you could help us add some docs about this feature, maybe [here](https://github.com/chainloop-dev/chainloop/tree/main/docs)

That could talk about the two modes of container image attestation, by reference and offline.

and how to make it offline using some of the tools you put here in the comments

What do you think?

Issue #2486, opened 2025-10-28
Review casbin token policies synchronization


jiparis:
We have noticed a random issue with API Tokens in the casbin synchronization of pods after a deployment. In some cases, token policies are not synced and some pods deny access to some resources, whereas others do. For example, when creating a workflow from the CLI, sometimes we can see this in our logs, followed by a `operation not allowed` error in the console:
```
2025-10-28 16:31:00	{"level":"info","ts":1761669060.986087,"msg":"[authZ] checking authorization","sub":"api-token:d9ed4f84-1d30-457e-b868-9fb3bb489686","operation":"/controlplane.v1.WorkflowService/Create"}
2025-10-28 16:31:00	{"level":"info","ts":1761669060.9901664,"msg":"[authZ] policy not found","sub":"api-token:d9ed4f84-1d30-457e-b868-9fb3bb489686","operation":"/controlplane.v1.WorkflowService/Create","resource":"workflow","action":"create"}
```

In these cases, a pod restart (k8s delete pod) fixes the issue.

The initialization code might be running into some race condition.

migmartri:
The current implementation of Casbin has a couple of problems. First, we store different entries for each API token created in the system in a Casbin database. This means the table is growing for each token and ultimately loading memory. The second problem is that for all pods to have the Casbin enforcer up-to-date, it requires a Postgres watcher, which has some issues as described in the issue description.

So my proposal is to move away from having different policies for API tokens and instead store the policies in the API tokens table itself and move to an enforcement mechanism based on the API token. This would be a policy based on checking a policy against existing policies or growth of the API token at runtime.

PR #2529, opened 2025-11-08
refactor: migrate Casbin to in-memory enforcer with database-stored token policies

migmartri:
## Summary

Refactored the authorization system to improve performance and simplify architecture by moving from database-stored Casbin policies to an in-memory enforcer for role-based access control while storing API token policies directly in the database.

## Benefits

- **Improved Performance**: Eliminates database queries for role policy lookups during authorization checks
- **Simplified Architecture**: Clear separation between user RBAC (in-memory) and token ACL (database-stored)
- **Reduced Dependencies**: Removes PostgreSQL adapter requirement for Casbin
- **Flexible Token Management**: Token permissions stored alongside token metadata for easier management
- **Better Maintainability**: Explicit dual enforcement model that's easier to reason about

## Technical Changes

- Casbin now uses in-memory adapter for static role policies
- API token ACL policies stored in new `policies` JSONB field in `api_tokens` table
- Added `EnforceWithPolicies` method for token-based authorization checks
- Database migration populates existing tokens with default policies
- Updated authorization middleware to route users and tokens to appropriate enforcement methods
- Removed database adapter dependencies and related sync logic

## Testing

- All existing tests passing
- Updated test infrastructure to use mockery v3
- Added test coverage for new `EnforceWithPolicies` method

refs https://github.com/chainloop-dev/chainloop/issues/2486

### DIFF
--- a/app/controlplane/cmd/main.go
+++ b/app/controlplane/cmd/main.go
@@ -64,7 +64,7 @@ func init() {
 }
 
 func newApp(logger log.Logger, gs *grpc.Server, hs *http.Server, ms *server.HTTPMetricsServer, profilerSvc *server.HTTPProfilerServer,
-	expirer *biz.WorkflowRunExpirerUseCase, plugins sdk.AvailablePlugins, tokenSync *biz.APITokenSyncerUseCase,
+	expirer *biz.WorkflowRunExpirerUseCase, plugins sdk.AvailablePlugins,
 	userAccessSyncer *biz.UserAccessSyncerUseCase, casBackendChecker *biz.CASBackendChecker, cfg *conf.Bootstrap) *app {
 	servers := []transport.Server{gs, hs, ms}
 	if cfg.EnableProfiler {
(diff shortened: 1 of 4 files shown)


app/controlplane/cmd/main.go
```diff
@@ -152,14 +152,6 @@ func main() {
 	app.runsExpirer.Run(ctx, &biz.WorkflowRunExpirerOpts{CheckInterval: 1 * time.Minute, ExpirationWindow: 1 * time.Hour})
 
 	// Since policies management is not enabled yet but instead is based on a hardcoded list of permissions
```
jiparis: I believe this comment also needs to be removed

app/controlplane/pkg/authz/enforcer.go
```diff
@@ -215,11 +171,6 @@ func doSync(e *Enforcer, c *Config) error {
 		resource := p[1]
 		action := p[2]
 
-		// if it's not a managed resource, skip deletion
-		if !slices.Contains(conf.ManagedResources, resource) {
```
jiparis, line 219: This is still needed if/when we reuse this logic in other places, not to remove core policies. Luckily, `authz.RolesMap` is a public var, not a constant, so we can add additional values there.

migmartri: it doesn't apply in practice

PR #2537, opened 2025-11-12
feat(cli): Expose Chainloop CLI logger

javirln:
This patch exposes the Chainloop's CLI logger so other extensions of the CLI can use the same logger in their commands and operations.


javirln:
Closing the PR since the logger can be found on the `ActionOpts`.

PR #2552, opened 2025-11-15
feat(policies): allow custom builtin functions in Rego policies

jiparis:
This PR adds extension capabilities to the Rego engine by using the builtins mechanism from OPA. Registering a new function is done through a call to `builtins.Register` with the builtin definition:

```go
func RegisterHelloBuiltin() error {
	return Register(&ast.Builtin{
		Name: helloBuiltinName,
		Decl: types.NewFunction(
			types.Args(
				types.Named("name", types.S), // Digest to fetch
			),
			types.Named("response", types.A), // Response as object
		),
	}, getHelloImpl)
}
```
Implementation signature must be a `topdown.BuiltinFunc`. Check provided example and tests.

This call must be done before using the Rego engine, so ideally it should be done during package initialization (`init()`) or app initialization before instantiating Rego.

### DIFF
--- a/pkg/policies/engine/rego/rego_test.go
+++ b/pkg/policies/engine/rego/rego_test.go
@@ -490,7 +490,7 @@ violations contains msg if {
 
 	t.Run("permissive builtin fails in restrictive mode", func(t *testing.T) {
 		_, err := r.Verify(context.TODO(), policy, []byte(`{"kind": "test"}`), nil)
-		// Should fail because chainloop.http_with_auth is not available in restrictive mode
+		// Should fail because test.dangerous_func is not available in restrictive mode
 		assert.Error(t, err)
 		assert.Contains(t, err.Error(), "undefined function")
 	})


app/cli/internal/policydevel/eval.go
```diff
@@ -165,9 +165,6 @@ func verifyMaterial(pol *v1.Policies, material *v12.Attestation_Material, materi
 }
 
 func craftMaterial(materialPath, materialKind string, logger *zerolog.Logger) (*v12.Attestation_Material, error) {
-	if fileNotExists(materialPath) {
```
jiparis, line 168: to be able to devel `CONTAINER_IMAGE` policies.

CLAUDE.md
```diff
@@ -246,6 +246,91 @@ Workflow Contracts define the structure and requirements for CI/CD attestations.
 - **Authentication**: JWT tokens generated by Control Plane, validated by CAS
 - **Streaming**: Uses gRPC bytestream protocol for efficient file transfers
 
+### Policy Engine Extension
+
+The OPA/Rego policy engine supports custom built-in functions written in Go.
```
migmartri: Nice

btw, I think we are abusing of CLAUDE.md, note that thisis used in the context of each call. Should we instead create a skill? https://code.claude.com/docs/en/skills

That will inject only the description of the skill, so for example if we mention about adding a builtin the skill will get kicked off and then all this detail makes sense.

pkg/policies/engine/rego/builtins/errors.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri: 2025

pkg/policies/engine/rego/builtins/registry.go
```diff
@@ -0,0 +1,179 @@
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri: remember 2025, it's a new file

pkg/policies/engine/rego/builtins/registry.go
```diff
@@ -0,0 +1,179 @@
...
+
+// NewRegistry creates a new built-in function registry
+func NewRegistry() *Registry {
+	return &Registry{
+		builtins: make(map[string]*BuiltinDef),
+	}
+}
+
+// Register adds a built-in function to the registry
+func (r *Registry) Register(def *BuiltinDef) error {
```
migmartri: so this is the meat of the PR? Some sort of custom loading mechamism? 

I am curious, why do we need to create this loader, it sounds to me that it should come from rego/opa already?

pkg/policies/engine/rego/rego.go
```diff
@@ -93,13 +103,30 @@ func NewEngine(opts ...EngineOption) *Engine {
...
 	}
 
-	return &Engine{
+	// Use global registry if none provided
+	registry := options.builtinRegistry
+	if registry == nil {
+		registry = builtins.NewRegistry()
+		// Register all built-ins from the global registry
+		for _, def := range builtins.All() {
+			_ = registry.Register(def)
```
migmartri: should we handle then error?

pkg/policies/engine/rego/builtins/registry.go
jiparis: Yeah, it could be done directly if we directly call `ast.RegisterBuiltin`. The idea was to have our own definition that holds some additional properties, like the permission level. But that can be accomplished as well with Rego Decl properties (eg categories property). Let me refactor this to simplify it a little bit.

CLAUDE.md
jiparis: Done 😄

Issue #2551, opened 2025-11-14
the latest changes from casbin is causing API-tokens to fail in some operations


migmartri:
The latest changes #2529  that include moving the enforcement API tokens to in-memory are causing different parts of the code calling the .enforce function to fail. This is not affecting attestations but only regular operations like download or upload.

PR #2553, opened 2025-11-15
fix(authZ): extract enforcer logic

migmartri:
This PR fixes https://github.com/chainloop-dev/chainloop/issues/2551 by refactoring the `Enforcer` method to make sure it is always used in all parts of the application, not just the middleware.

### DIFF
--- a/app/controlplane/pkg/authz/enforcer.go
+++ b/app/controlplane/pkg/authz/enforcer.go
@@ -41,13 +41,13 @@ type Config struct {
 	RolesMap map[Role][]*Policy
 }
 
-type Enforcer struct {
+type CasbinEnforcer struct {
 	*casbin.Enforcer
 
 	config *Config
 }
 
-func (e *Enforcer) Enforce(sub string, p *Policy) (bool, error) {
+func (e *CasbinEnforcer) Enforce(sub string, p *Policy) (bool, error) {
 	// This enforcer does not support API token subjects
 	// this is due to the fact that API tokens are not stored in casbin yet
 	// To use them, make sure you use the AuthzUseCase.Enforce method instead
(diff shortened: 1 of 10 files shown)


app/controlplane/pkg/biz/authz.go
```diff
@@ -0,0 +1,86 @@
+//
+// Copyright 2025 The Chainloop Authors.
```
migmartri, line 2: This is the main change.

It's basically a custom enforcer that knows not only about casbin but also about API-Tokens custom ACLs.

I created it in the `biz` layer since that way it can be used without the risk of package cycles.

I tried to do smth similar but by extending the current `pkg/authz` but that one is connected already referenced by both `biz` and `data`

migmartri:
@jiparis ptal at this PR, it fixes an important bug

app/controlplane/pkg/authz/enforcer.go
```diff
@@ -38,38 +38,29 @@ func (t *SubjectAPIToken) String() string {
 var modelFile []byte
 
 type Config struct {
-	RolesMap            map[Role][]*Policy
-	RestrictOrgCreation bool
+	RolesMap map[Role][]*Policy
 }
 
 type Enforcer struct {
```
jiparis: I think we should rename this one to `CasBinEnforcer` so that they are easily distinguishable.

app/controlplane/pkg/biz/authz.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/chainloop-dev/chainloop/app/controlplane/pkg/authz"
+	"github.com/go-kratos/kratos/v2/log"
+	"github.com/google/uuid"
+)
+
+type AuthzUseCase struct {
+	log *log.Helper
+	// actual CASBIN enforcer
+	enforcer            *authz.Enforcer
```
jiparis: and this variable too: => `casBinEnforcer`

app/controlplane/pkg/authz/enforcer.go
migmartri: Done

PR #2558, opened 2025-11-17
feat(policies): Chainloop discover custom builtin

jiparis:
This PR extends the CLI to support access to the Discover endpoint as a custom Rego builtin function. This is an example of extending the engine using the extension mechanism provided in https://github.com/chainloop-dev/chainloop/pull/2552

To use it in policies:
```yaml
kind: Policy
metadata:
  name: policy-builtins
spec:
  policies:
    - kind: CONTAINER_IMAGE
      embedded: |
        package main
        import rego.v1
        
        result := {"violations": violations}
        
        violations contains msg if {
          digest := sprintf("sha256:%s",[input.chainloop_metadata.digest.sha256])
          discovered := chainloop.discover(digest, "")
          
          some ref in discovered.references
          ref.kind == "ATTESTATION"
          ref.metadata.hasPolicyViolations == "true"
        
          msg:= sprintf("attestation with digest %s contains policy violations [name: %s, project: %s, org: %s]", [ref.digest, ref.metadata.name, ref.metadata.project, ref.metadata.organization])
        }
```
```shell
> go run app/cli/main.go policy devel eval --material ghcr.io/chainloop-dev/chainloop/control-plane:v1.57.0-amd64 --policy test/policy-builtins.yaml
{
   "result": {
      "violations": [
         "attestation with digest sha256:5d2ae3589cd46277fb83b1e9a3eb4c58dfb0c1e042729520119ed0533ef7cd6d contains policy violations [name: release, project: chainloop, org: read-only-demo]"
      ],
      "skip_reasons": [],
      "skipped": false
   }
}
```

### DIFF
--- a/pkg/policies/engine/rego/builtins/discover.go
+++ b/pkg/policies/engine/rego/builtins/discover.go
@@ -1,0 +1,101 @@
+//
+// Copyright 2025 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builtins
+
+import (
+	"errors"
+	"fmt"
+
+	v1 "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1"
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/topdown"
+	"github.com/open-policy-agent/opa/v1/types"
+	"google.golang.org/grpc"
+)
+
+const discoverBuiltinName = "chainloop.discover"
+
+// RegisterDiscoverBuiltin is used to register chainloop's Discover endpoint as a builtin Rego function with signature:
+//
+//	chainloop.discover(digest, kind)
+//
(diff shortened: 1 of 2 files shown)


pkg/policies/engine/rego/builtins/discover.go
```diff
@@ -0,0 +1,101 @@
...
+//
+//	  some ref in discovered.references
+//	  ref.kind == "ATTESTATION"
+//	  ref.metadata.hasPolicyViolations == "true"
+//
+//	  msg:= sprintf("attestation with digest %s contains policy violations [name: %s, project: %s, org: %s]", [ref.digest, ref.metadata.name, ref.metadata.project, ref.metadata.organization])
+//	}
+//
+// ```
+func RegisterDiscoverBuiltin(conn *grpc.ClientConn) error {
```
migmartri, line 50: one thing it would be nice to do (in the future) is to generate documentation of these built-in. We do smth for integrations already, like creating a markdown that then we can sync somewhere.

migmartri, line 50: I think that in the future the input should evolve to a struct, so we can have controlplaneConnection, CAS, and platform connections if needed.

pkg/policies/engine/rego/builtins/discover.go
```diff
@@ -0,0 +1,101 @@
...
+				types.Named("digest", types.S).Description("digest of the artifact to discover"),
+				types.Named("kind", types.S).Description("optional filter by kind to disambiguate"),
+			),
+			types.Named("response", types.A).Description("response object as in the `chainloop discover` CLI output"),
+		),
+		Nondeterministic: true,
+	}, getDiscoverImpl(conn))
+}
+
+func getDiscoverImpl(conn *grpc.ClientConn) topdown.BuiltinFunc {
```
migmartri, line 65: it would be nice to have logger support, so when running `devel` with --debug we can see some info

migmartri (approved):
Very cool!

PR #2564, opened 2025-11-18
feat: add skip field to policy group attachments

### LINKED ISSUE #2557: support a way to explicitly disable policies in a group
Currently, when you reference a policy group, all the policies inside are created. My proposal is to have an attribute in the attachment of the policy groups in the contract that allows you to explicitly disable policies you want to skip. 

For example, having this contract that runs all the policies in the quality group

```
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: example-contract
spec:
  policyGroups:
    - ref: file://groups/sbom-quality-group.yaml
      with:
        bannedComponents: log4j@2.14.1
```

We could have

```
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: example-contract
spec:
  policyGroups:
    - ref: file://groups/sbom-quality-group.yaml
      with:
        bannedComponents: log4j@2.14.1
        bannedLicenses: AGPL-1.0-only, AGPL-1.0-or-later, AGPL-3.0-only, AGPL-3.0-or-later
      skip:
        - sbom-present
        - my-other-policy
```

migmartri:
## Summary

Add support for explicitly disabling specific policies within a policy group by specifying their metadata names in a skip list.

Users can now selectively exclude policies from evaluation without modifying the policy group itself by adding a `skip` field to policy group attachments in workflow contracts.

## Implementation

- Added `skip` field to `PolicyGroupAttachment` protobuf message
- Implemented policy name resolution for both embedded and referenced policies
- Filtered skipped policies in material and attestation evaluation paths
- Added validation with warnings for non-existent policy names
- Comprehensive test coverage for all skip scenarios

## Usage

```yaml
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: example-contract
spec:
  policyGroups:
    - ref: file://groups/sbom-quality-group.yaml
      with:
        bannedComponents: log4j@2.14.1
      skip:
        - sbom-present
        - my-other-policy
```

## Behavior

- Policies are matched by their `metadata.name` field
- Works for both material and attestation policies
- Unknown policy names in skip list generate warnings but allow execution to continue
- Empty skip list has no effect on evaluation

Closes #2557

### DIFF
--- a/pkg/policies/policy_groups_test.go
+++ b/pkg/policies/policy_groups_test.go
@@ -170,7 +170,7 @@ func (s *groupsTestSuite) TestRequiredPoliciesForMaterial() {
 			}
 			group, _, err := new(FileGroupLoader).Load(context.TODO(), groupAtt)
 			s.Require().NoError(err)
-			atts, err := v.requiredPoliciesForMaterial(context.TODO(), material, group, groupAtt, nil)
+			atts, err := v.requiredPoliciesForMaterial(context.TODO(), material, group, nil)
 			s.Require().NoError(err)
 			s.Len(atts, tc.expected)
 		})


migmartri:
@jiparis ready to review, tested it locally :)

app/controlplane/configs/config.devel.yaml
```diff
@@ -98,10 +98,10 @@ prometheus_integration:
   - org_name: "development"
 
 # Policy providers configuration
-policy_providers:
-  - name: chainloop
-    default: true
-    url: http://localhost:8002/v1
+# policy_providers:
```
migmartri, line 101: this was added by mistake in another PR

pkg/policies/policy_groups.go
```diff
@@ -125,7 +131,25 @@ func (pgv *PolicyGroupVerifier) VerifyStatement(ctx context.Context, statement *
 		if err != nil {
 			return nil, NewPolicyError(err)
 		}
+
+		// Validate skip list and log warnings for unknown policy names
+		if err := pgv.validateSkipList(ctx, group, groupAtt); err != nil {
```
jiparis: this call is overloading the verification step, since it's downloading all policies just to get their names. Verification is also downloading them in the same call. 
I'd suggest skipping the validation at this point, since it has no effect. It could probably be moved to `att init` or contract command line options.
Another option / improvement would be to preload all policies in advance so that they can be used for the validation and for the verification.

Same comment for `VerifyMaterial`

migmartri: good catch, I think you are right, not worth the tradeoff at runtime indeed

pkg/policies/policy_groups.go
```diff
@@ -282,3 +307,25 @@ func (pgv *PolicyGroupVerifier) shouldApplyPolicy(ctx context.Context, policyAtt
 
 	return false, nil
 }
+
+// getPolicyName extracts the metadata.name from a PolicyAttachment
+// It handles both embedded and referenced policies by loading the policy spec when needed
+func (pgv *PolicyGroupVerifier) getPolicyName(ctx context.Context, attachment *v1.PolicyAttachment) (string, error) {
+	// Case 1: Embedded policy - direct access
+	if embedded := attachment.GetEmbedded(); embedded != nil {
```
jiparis: `loadPolicySpec` already contains the logic for embedded policies. This first case block wouldn't be needed.

pkg/policies/policy_groups.go
```diff
@@ -125,7 +125,20 @@ func (pgv *PolicyGroupVerifier) VerifyStatement(ctx context.Context, statement *
 		if err != nil {
 			return nil, NewPolicyError(err)
 		}
+
 		for _, attachment := range group.GetSpec().GetPolicies().GetAttestation() {
+			// Check if policy should be skipped
+			policyName, err := pgv.getPolicyName(ctx, attachment)
```
jiparis: Another option is to include this skip logic in `evaluatePolicyAttachment` (called below), which it's already calling `loadPolicySpec` to load the policy. We can just pass it the list of skips, and maybe return a SkipError or similar to mark it as skipped.

jiparis: Or we can add a policy cache to loadPolicySpec (it can be done in a different PR)

pkg/policies/policy_groups.go
migmartri: makes sense, thanks!

PR #2575, opened 2025-11-20
feat(devel): enable vault persistence in dev mode

### LINKED ISSUE #2574: how to run vault in development mode but with persistence
We use Vault in development mode during development using Docker Compose, and also when you run the cluster in development mode. The problem is that even though it is unsealed, it doesn't have any persistence. Any restart of the pod loses the data. Is there any way to run a Vault container unsealed and with persistence?

migmartri:
It enables persistence in Vault when ran in development mode both in local and in k8s, note: in development mode

### Docker Compose (`devel/`)
- Switch Vault from dev mode to server mode with file storage backend
- Add `vault-config.hcl` for server configuration
- Add `vault-init.sh` script to handle initialization, unsealing, and token creation
- Mount persistent volume for Vault data
- Automatically unseal Vault on startup
- Create 'notasecret' token for development convenience
- Enable KV v2 secrets engine

### Helm Chart (`deployment/chainloop/`)
- Add init script via `vault.extraDeploy` in parent chart values
- Configure file storage backend (overrides subchart defaults)
- Auto-initialize and unseal Vault on pod start
- Validate init.txt content to handle corrupted files
- Create 'notasecret' token if it doesn't exist
- Enable KV v2 secrets engine (required by controlplane)
- All configuration in parent chart (no subchart modifications)

## Benefits
- Data persists across container/pod restarts
- Maintains development convenience (auto-unseal, known token)
- No manual intervention required after restart
- Consistent behavior between Docker Compose and Helm deployments

Fixes #2574"

### DIFF
--- a/deployment/chainloop/templates/_helpers.tpl
+++ b/deployment/chainloop/templates/_helpers.tpl
@@ -74,11 +74,7 @@ secretPrefix: {{ required "secret prefix required" .secretPrefix | quote }}
 vault:
   {{- if and $.Values.development (or (not .vault) not .vault.address) }}
   address: {{ printf "http://%s-server:8200" (include "chainloop.vault.fullname" $) | quote }}
-  {{- if $tokenEnvVar }}
-  token: {{ $tokenEnvVar | quote }}
-  {{- else }}
-  token: "notasecret"
-  {{- end }}
+  token: {{ default "notasecret" $tokenEnvVar | quote }}
 {{- else if (required "vault backend selected but configuration not provided" .vault ) }}
   address: {{ required "vault address required" .vault.address | quote }}
   token: {{ required "vault token required" .vault.token | quote }}


deployment/chainloop/templates/_helpers.tpl
```diff
@@ -77,7 +77,7 @@ vault:
   {{- if $tokenEnvVar }}
   token: {{ $tokenEnvVar | quote }}
   {{- else }}
-  {{- required "VAULT_DEV_ROOT_TOKEN_ID environment variable is required when development mode is enabled" (index $.Values.vault.server.extraEnvVars "VAULT_DEV_ROOT_TOKEN_ID") }}
+  token: "notasecret"
```
jiparis: If I understand this correctly, this could be considered a dangerous default. I'd keep the error message.

migmartri: this is for development mode.

PR #2576, opened 2025-11-21
chore(policies): improve error message in policy eval when no branches were executed

jiparis:
When policy evaluation result is empty, it might happen that the material didn't match, or that the policy logic (Rego) has set the `ignore` flag to `true` (because of missing input parameters or any other criteria).

```
ERR no execution branch matched, or all of them were ignored, for kind SBOM_CYCLONEDX_JSON
```

### DIFF
--- a/app/cli/documentation/cli-reference.mdx
+++ b/app/cli/documentation/cli-reference.mdx
@@ -2869,7 +2869,6 @@ Options
 ```
 --allowed-hostnames strings   Additional hostnames allowed for http.send requests in policies
 --annotation strings          Key-value pairs of material annotations (key=value)
---debug                       Include detailed evaluation inputs/outputs in JSON output and enable verbose logging
 -h, --help                        help for eval
 --input stringArray           Key-value pairs of policy inputs (key=value)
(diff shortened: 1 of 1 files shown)


app/cli/internal/policydevel/eval.go
```diff
@@ -120,7 +120,7 @@ func verifyMaterial(pol *v1.Policies, material *v12.Attestation_Material, materi
 	}
 
 	if len(policyEvs) == 0 || policyEvs[0] == nil {
-		return nil, fmt.Errorf("no execution branch matched for kind %s", material.MaterialType.String())
+		return nil, fmt.Errorf("no execution branch matched, or all of them were ignored, for kind %s", material.MaterialType.String())
```
migmartri, line 123: is there any better way we could do this? I mean, is there any way to distinguish them? Maybe we can add some debug logging so if you run `eval` with debug you get info about if the execution path was ignored or executed?

jiparis: That would be ideal, yes. It has implications in the engine, though. Let me check them

jiparis:
Check the change in policy_develop_eval.go. I noticed that the `debug` flag in that command is overriding the global `debug` flag value, so by default it was disabling all CLI debug messages.
Now, the message can be clearly seen:
```
DBG evaluating policy sbom-banned-components against auto-detected-material
DBG policy sbom-banned-components explicitly ignored by definition <---------
ERR no execution branch matched, or all of them were ignored, for kind SBOM_CYCLONEDX_JSON
```

app/cli/internal/policydevel/eval.go
jiparis: The message was already there, but a bug prevented it from being shown. Check my comment in the PR

PR #2584, opened 2025-11-26
chore(policy): Bootstrap WASM policy engine

javirln:
## Architecture

The WASM engine is implemented using Extism, a universal plugin system that allows running WebAssembly modules safely across different platforms.

Here's a small video hands on:

https://github.com/user-attachments/assets/7dae8e0e-7c44-4884-b17e-dbfac32c6370

### Policy Type Detection

`pkg/policies/engine/detect.go`

Automatically detects policy type by checking for WASM magic bytes (0x00 0x61 0x73 0x6d). Falls back to Rego for text-based policies. Detection happens transparently at runtime.

### WASM Engine

`pkg/policies/engine/wasm/engine.go`

Implements the PolicyEngine interface using Extism Go SDK.

Features:
- Timeout control with configurable execution timeout (default 60 seconds)
- Logging integration that maps Extism log levels to zerolog
- HTTP restrictions via AllowedHostnames
- WASI support enabled
- Configuration passing via Extism config as JSON

### Policy Execution Flow

Material or attestation is passed to PolicyVerifier, which detects the policy type. For WASM policies, the engine creates an Extism plugin, calls the Execute function, and parses JSON output.

  ```text
  Material/Attestation → PolicyVerifier.executeScript()
                             ↓
                      DetectPolicyType()
                             ↓
               ┌─────────────┴─────────────┐
               ↓                           ↓
          PolicyTypeRego              PolicyTypeWASM
               ↓                           ↓
        Rego.NewEngine()            WASM.NewEngine()
               ↓                           ↓
          OPA Evaluation          Extism Plugin Execution
                             ↓
                      Call "Execute" function
                             ↓
                      Parse JSON output
```
### Common Options

Shared with Rego engine:

- AllowedHostnames: List of domains for HTTP requests (defaults: www.chainloop.dev, www.cisa.gov)
- IncludeRawData: Include raw input/output in results
- EnablePrint: Enable print statements

### WASM-Specific Options

- ExecutionTimeout: Maximum execution time (default 60s)
- Logger: zerolog logger instance

## WASM Policy Contract

WASM policies must export an Execute function.

Input: Raw material bytes as input parameter, with arguments accessible via Extism config under "args" key.

Output: JSON object with violations array, skipped boolean, skip_reason string, and ignore boolean.

## Error Handling

File: `pkg/policies/engine/wasm/errors.go`

Provides sophisticated error parsing for:
- HTTP Forbidden errors with hostname extraction
- Timeout detection
- Out-of-memory conditions
- Runtime errors with helpful messages
- User-friendly hints and suggestions

## Security Features

- Sandboxing through WASM memory isolation
- Network restrictions via allowed hostnames list
- Timeout protection against infinite loops
- No custom host functions exposed
- WASI-only standard interfaces

### Policy Loader

File: `pkg/policies/policies.go`

Automatically selects engine based on detected policy type. For WASM policies, creates engine with logger and common options.

### Material Verification

File: `pkg/attestation/crafter/crafter.go`

WASM policies are evaluated alongside Rego policies with the same violation reporting structure, transparent to users.

## Current Limitations

- MatchesParameters and MatchesEvaluation are stub implementations that always return true
- No automatic boilerplate injection like Rego policies
- Must be pre-compiled WASM modules

### File Structure

- engine.go: Main WASM engine implementation
- engine_test.go: Engine tests
- errors.go: Error parsing and handling
- errors_test.go: Error handling tests

### Key Methods

NewEngine creates a new WASM engine with specified options and applies common configuration.

Verify is the main execution method that creates an Extism plugin, configures it, calls the Execute function, and parses results.

MatchesParameters and MatchesEvaluation are stub implementations that delegate to WASM policy internals.

### DIFF
--- a/pkg/policies/engine/wasm/engine_test.go
+++ b/pkg/policies/engine/wasm/engine_test.go
@@ -28,7 +28,7 @@ func TestNewEngine(t *testing.T) {
 	t.Run("default options", func(t *testing.T) {
 		eng := NewEngine()
 		assert.NotNil(t, eng)
-		assert.Equal(t, 60*time.Second, eng.executionTimeout)
+		assert.Equal(t, 5*time.Second, eng.executionTimeout)
 		assert.False(t, eng.IncludeRawData)
 		assert.False(t, eng.EnablePrint)
 		// Base allowed hostnames should be included by default

### CODE REVIEW

File: app/cli/internal/policydevel/lint.go [hunk 1]
```diff
@@ -173,8 +175,18 @@ func (p *PolicyToLint) processFile(filePath string) error {
 			Path:    filePath,
 			Content: content,
 		})
+	case ".wasm":
+		// Detect if file is actually WASM using magic bytes
+		policyType := engine.DetectPolicyType(content)
+		if policyType != engine.PolicyTypeWASM {
+			return fmt.Errorf("file has .wasm extension but is not a valid WASM file")
+		}
+		p.WASMFiles = append(p.WASMFiles, &File{
```
jiparis: In these cases, `Content` is not needed because the actual binary is not linted. Instead of `os.ReadFile` we could just do a `Stat` to check that it exists.

javirln: I've actually change it a bit. We can perform some "linting" that is basically check:
1. Incoming bytes are from a WASM module
2. Check the WASM module has the required entry point, `Execute` function

And for the last bit we need the content yes, have a look at the new version :)

app/cli/internal/policydevel/lint.go
```diff
@@ -156,12 +159,30 @@ func (p *PolicyToLint) loadReferencedRegoFiles(baseDir string) error {
 }
 
 func (p *PolicyToLint) processFile(filePath string) error {
+	ext := strings.ToLower(filepath.Ext(filePath))
+
+	// WASM files: validate magic bytes
+	if ext == ".wasm" {
+		content, err := os.ReadFile(filePath)
```
jiparis: This could be moved outside `if`. The file is now being read in any case. 
And then, the body of this condition could be moved to the `switch` below.

pkg/policies/engine/rego/rego.go
```diff
@@ -52,55 +47,35 @@ func WithOperatingMode(mode EnvironmentMode) EngineOption {
 	}
 }
 
-func WithAllowedNetworkDomains(domains ...string) EngineOption {
+// ToEngineOption converts common engine options to rego-specific engine options
+func ToEngineOption(commonOpts ...engine.CommonEngineOption) EngineOption {
```
jiparis: Why don't just accept `engine.CommonEngineOption` (I would call it just EngineOption) directly? Implementations could choose whether to accept or not them if they don't apply.

pkg/policies/policies.go
```diff
@@ -347,26 +348,60 @@ func (pv *PolicyVerifier) VerifyStatement(ctx context.Context, statement *intoto
...
+	switch policyType {
+	case engine.PolicyTypeRego:
+		// Rego engine with common options
+		engineOpts := []rego.EngineOption{rego.ToEngineOption(commonOpts...)}
+		policyEngine = rego.NewEngine(engineOpts...)
+
+	case engine.PolicyTypeWASM:
+		// WASM engine with common options plus logger
+		wasmOpts := []wasm.EngineOption{
+			wasm.WithLogger(pv.logger),
```
jiparis: I think options can be common to both to simplify this. Rego might also receive the logger option (even if the current impl doesn't use it).

```go
opts := []engine.Options{ ... } // WithAllowedHostNames, etc...
switch policyType {
	case engine.PolicyTypeRego:
       policyEngine := rego.NewEngine(opts...)
    case engine.PolicyTypeWASM:
       policyEngine := wasm.NewEngine(opts...)
    default:
       ...
}
```

pkg/policies/policies.go
```diff
@@ -637,10 +674,12 @@ func LoadPolicyScriptsFromSpec(policy *v1.Policy, kind v1.CraftingSchema_Materia
 					return nil, fmt.Errorf("failed to load policy script: %w", err)
 				}
 
-				// Inject boilerplate if needed
-				script, err = rego.InjectBoilerplate(script, policy.GetMetadata().GetName())
-				if err != nil {
-					return nil, fmt.Errorf("failed to inject boilerplate: %w", err)
+				// Inject boilerplate only for Rego policies, not WASM
+				if engine.DetectPolicyType(script) == engine.PolicyTypeRego {
```
jiparis, line 684: I wonder if the policy type should be a property of `spec`. I'm ok if we feel comfortable with this magic detection, but this might not stand in the future if we support more engines (`jq` or something else), since here ware just detecting that this is NOT WASM, but it might be anything else.

javirln:
All comments addressed

PR #2592, opened 2025-12-01
feat(policy-devel): Adapt policy develop init to WASM

javirln:
This PR modifies the command `chainloop policy develop init` so it can bootstrap all needed files to start writing policies for Golang and Javascript.


migmartri:
I'd not add this since wasm support is still experimental

PR #2596, opened 2025-12-04
fix(config): allow commands to run without valid connection

Piskoo:
Fixes an issue where `chainloop config reset` and other config management commands would fail when certificates couldn't be loaded, creating situation where users couldn't fix broken configurations.

### DIFF
--- a/app/cli/cmd/config_reset.go
+++ b/app/cli/cmd/config_reset.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023 The Chainloop Authors.
+// Copyright 2023-2025 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -26,6 +26,9 @@ func newConfigResetCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "reset",
 		Short: "Reset the CLI configuration",
+		Annotations: map[string]string{
+			skipActionOptsInit: trueString,
+		},
 		Run: func(cmd *cobra.Command, args []string) {
 			configFile := viper.ConfigFileUsed()
 			err := os.Remove(configFile)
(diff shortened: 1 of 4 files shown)


app/cli/cmd/root.go
```diff
@@ -483,3 +491,22 @@ func isAPITokenPreferred(cmd *cobra.Command) bool {
...
+// isConfigManagementCommand checks if the command is a config management command
+// that doesn't require action options initialization
+func isConfigManagementCommand(cmd *cobra.Command) bool {
+	// Walk up the command tree to find the parent
+	current := cmd
+	for current.Parent() != nil {
+		if current.Parent().Use == "config" {
+			switch current.Use {
+			// These config subcommands don't need action options
+			case "reset", "view", "save":
```
migmartri: this looks very brittle, is there any other way to do it? like an annotation in the command?

Piskoo: Sure, updated it with the annotation

app/cli/cmd/root.go
```diff
@@ -102,6 +104,12 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 
 			logger.Debug().Str("path", viper.ConfigFileUsed()).Msg("using config file")
 
+			// Commands annotated with skipActionOptsInit don't need ActionOpts initialization
+			// These are local-only commands that don't interact with the control plane
+			if cmd.Annotations[skipActionOptsInit] == trueString {
```
migmartri, line 109: I am assuming this will not fail if the annotation is not set correct?

app/cli/cmd/config_reset.go
```diff
@@ -26,6 +26,9 @@ func newConfigResetCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "reset",
 		Short: "Reset the CLI configuration",
+		Annotations: map[string]string{
```
migmartri, line 29: shouldn't this affect all the config commands? Otherwise if it's failing you'll not be able to set it again?

app/cli/cmd/root.go
Piskoo: Correct

app/cli/cmd/config_reset.go
Piskoo: It does affect all of them, but I've excluded `config plugin` because it uses some of the fields that actionopts contain. Subcommands don't inherit annotations from parents so it has to be added to each command.

PR #2527, opened 2025-11-07
feat: add setting to prevent implicit workflow creation

migmartri:
## Summary

Add organization-level setting to control whether workflows and projects can be created implicitly during attestation init. This helps prevent workflow and contract sprawl in automated environments.

When `prevent_implicit_workflow_creation` is enabled, workflows must be created explicitly via `chainloop workflow create`.

```sh
$ go run main.go org update --name testing-123  --prevent-implicit-workflow-creation=true
$ go run main.go att init --workflow sast --project my-project-2222 --release  --replace
ERR creating workflows during the attestation process is disabled for this organization. Please create them in advance or contact your administrator
```

## Changes

- Add `prevent_implicit_workflow_creation` field to organization
- Add validation in workflow creation to check organization setting
- Add CLI flag `--prevent-implicit-workflow-creation` to organization update command

Fixes #2526

PR #2602, opened 2025-12-05
feat: add organization setting to restrict project-scoped contracts

Piskoo:
This PR adds a new organization-level setting that allows administrators to restrict contract creation to be only available for organization admins, preventing their creation by org members and project admins.

When setting is enabled and non-admin user tries to run `att init` without specifying contract an error is returned: 
```
chainloop att init --project myproject --workflow newworkflow --replace
ERR failed to initialize the attestation: can't create the workflow: authorization error: only organization admins can create contracts
```

### DIFF
--- a/app/controlplane/internal/service/attestation.go
+++ b/app/controlplane/internal/service/attestation.go
@@ -778,11 +778,7 @@ func (s *AttestationService) FindOrCreateWorkflow(ctx context.Context, req *cpAP
 		createOpts.Owner = &userID
 
 		// Check if user is an org admin
-		orgUUID, err := uuid.Parse(apiToken.OrgID)
-		if err != nil {
-			return nil, handleUseCaseErr(err, s.log)
-		}
-		createOpts.UserIsOrgAdmin = isUserOrgAdmin(ctx, orgUUID)
+		createOpts.UserIsOrgAdmin = isUserOrgAdmin(ctx)
 	}
 
 	wf, err := s.workflowUseCase.Create(ctx, createOpts)
--- a/app/controlplane/internal/service/service.go
+++ b/app/controlplane/internal/service/service.go
@@ -330,20 +330,9 @@ func (s *service) visibleProjects(ctx context.Context) []uuid.UUID {
 }
 
 // isUserOrgAdmin checks if the current user is an org admin or owner
-func isUserOrgAdmin(ctx context.Context, orgUUID uuid.UUID) bool {
-	membership := entities.CurrentMembership(ctx)
-	if membership == nil {
-		return false
-	}
-
-	// Find the resource membership for this organization
-	for _, resource := range membership.Resources {
-		if resource.ResourceType == authz.ResourceTypeOrganization && resource.ResourceID == orgUUID {
-			return resource.Role.IsAdmin()
-		}
-	}
-
-	return false
+func isUserOrgAdmin(ctx context.Context) bool {
+	userRole := usercontext.CurrentAuthzSubject(ctx)
+	return authz.Role(userRole).IsAdmin()
 }
 
 // initializePaginationOpts initializes the pagination options with the provided request pagination options.
--- a/app/controlplane/internal/service/workflow.go
+++ b/app/controlplane/internal/service/workflow.go
@@ -102,11 +102,7 @@ func (s *WorkflowService) Create(ctx context.Context, req *pb.WorkflowServiceCre
 		createOpts.Owner = &userID
 
 		// Check if user is an org admin
-		orgUUID, err := uuid.Parse(currentOrg.ID)
-		if err != nil {
-			return nil, handleUseCaseErr(err, s.log)
-		}
-		createOpts.UserIsOrgAdmin = isUserOrgAdmin(ctx, orgUUID)
+		createOpts.UserIsOrgAdmin = isUserOrgAdmin(ctx)
 	}
 
 	p, err := s.useCase.Create(ctx, createOpts)
--- a/app/controlplane/internal/service/workflowcontract.go
+++ b/app/controlplane/internal/service/workflowcontract.go
@@ -121,17 +121,7 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
 
 	// If setting is enabled, only org admins can create contracts (org-level or project-level)
 	if org.RestrictContractCreationToOrgAdmins {
-		currentUser, err := requireCurrentUser(ctx)
-		if err != nil {
-			return nil, err
-		}
-
-		membership, err := s.userUC.MembershipInOrg(ctx, currentUser.ID, currentOrg.Name)
-		if err != nil {
-			return nil, handleUseCaseErr(err, s.log)
-		}
-
-		if !membership.Role.IsAdmin() {
+		if !isUserOrgAdmin(ctx) {
 			return nil, errors.Forbidden("forbidden", "contract creation is restricted to organization administrators. Please contact your administrator")
 		}
 	}


app/controlplane/internal/service/workflowcontract.go
```diff
@@ -109,9 +111,20 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
...
+	org, err := s.orgUseCase.FindByID(ctx, currentOrg.ID)
+	if err != nil {
+		return nil, handleUseCaseErr(err, s.log)
+	}
+
 	// if the project is provided we make sure it exists and the user has permission to it
 	var projectID *uuid.UUID
 	if req.ProjectReference.IsSet() {
+		// Check if organization prevents project-scoped contracts
+		if org.PreventProjectScopedContracts {
```
migmartri: is this the only place a contract is created? can you check during attestation init?

migmartri: We need to double check that a contract can only be created on-demand from this use case. I remember it was also created during a workflow creation, or attestation init.

Would you mind posting there an evaluation of those entrypoints and how this feature will work with the other `preventWorkflowCreation` feature?

migmartri:
also, what's the use-case here? Can anybody create project contracts? can admins do it?

app/controlplane/internal/service/workflowcontract.go
migmartri: I don't think this is what we want. What we want is to only allow `administrators` to create contracts if this setting is set

Piskoo:
Updated with what have discussed. 

> is this the only place a contract is created? can you check during attestation init?

We could also restrict creation of the default contract that is created when no contract was specified in workflow creation, but is that desired behavior?

migmartri:
> Updated with what have discussed.
> 
> > is this the only place a contract is created? can you check during attestation init?
> 
> We could also restrict creation of the default contract that is created when no contract was specified in workflow creation, but is that desired behavior?

I think so yes

The idea of this feature is to prevent contract sprawl, and the main source of sprawl is the attestation process.

Piskoo:
> > Updated with what have discussed.
> > > is this the only place a contract is created? can you check during attestation init?
> > 
> > 
> > We could also restrict creation of the default contract that is created when no contract was specified in workflow creation, but is that desired behavior?
> 
> I think so yes
> 
> The idea of this feature is to prevent contract sprawl, and the main source of sprawl is the attestation process.

Alright, added validation to `att init` as well

migmartri:
> > > Updated with what have discussed.
> > > > is this the only place a contract is created? can you check during attestation init?
> > > 
> > > 
> > > We could also restrict creation of the default contract that is created when no contract was specified in workflow creation, but is that desired behavior?
> > 
> > 
> > I think so yes
> > The idea of this feature is to prevent contract sprawl, and the main source of sprawl is the attestation process.
> 
> Alright, added validation to `att init` as well

Thanks

Does it work in `workflow create` too?

Piskoo:
> Does it work in `workflow create` too?

Yes, there shouldn't be any other remaining cases

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -109,6 +113,29 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
...
+			return nil, err
+		}
+
+		membership, err := s.userUC.MembershipInOrg(ctx, currentUser.ID, currentOrg.Name)
+		if err != nil {
+			return nil, handleUseCaseErr(err, s.log)
+		}
+
+		if !membership.Role.IsAdmin() {
+			return nil, errors.Forbidden("forbidden", "only organization admins can create contracts")
```
migmartri: can we make sure the error is cleaner and consistent with what we did in this other check? https://github.com/chainloop-dev/chainloop/pull/2527

```
ERR creating workflows during the attestation process is disabled for this organization. Please create them in advance or contact your administrator
```

Piskoo: Updated direct creation (e.g. wf contract create) to
```
ERR contract creation is restricted to organization administrators. Please contact your administrator
```

And indirect creation (e.g. att with --contract flag or wf create) to
```
ERR contract creation is restricted to organization administrators. Please use existing contracts or contact your administrator
```

app/controlplane/internal/service/attestation.go
```diff
@@ -766,6 +776,14 @@ func (s *AttestationService) FindOrCreateWorkflow(ctx context.Context, req *cpAP
 			return nil, handleUseCaseErr(err, s.log)
 		}
 		createOpts.Owner = &userID
+
+		// Check if user is an org admin
+		membership, err := s.userUseCase.MembershipInOrg(ctx, user.ID, org.Name)
```
migmartri: I think this roles are already in the membership

app/controlplane/internal/service/workflow.go
```diff
@@ -89,6 +100,12 @@ func (s *WorkflowService) Create(ctx context.Context, req *pb.WorkflowServiceCre
 			return nil, handleUseCaseErr(err, s.log)
 		}
 		createOpts.Owner = &userID
+
+		// Check if user is an org admin
+		membership, err := s.userUseCase.MembershipInOrg(ctx, user.ID, org.Name)
+		if err == nil && membership != nil {
+			createOpts.UserIsOrgAdmin = membership.Role.IsAdmin()
```
migmartri: `membership.Role.IsAdmin()` does this include also owner?

Piskoo: Yes

app/controlplane/internal/service/attestation.go
Piskoo: Moved it to a helper that uses context instead of db query

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -109,6 +113,29 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
...
+	}
+
+	// If setting is enabled, only org admins can create contracts (org-level or project-level)
+	if org.RestrictContractCreationToOrgAdmins {
+		currentUser, err := requireCurrentUser(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		membership, err := s.userUC.MembershipInOrg(ctx, currentUser.ID, currentOrg.Name)
```
migmartri: this is already part of the membership doesn't it? cc/ @jiparis 

I don't think you need to retrieve it again, it was done by a middleware afaik, am I mistaken?

app/controlplane/internal/service/service.go
```diff
@@ -329,6 +329,23 @@ func (s *service) visibleProjects(ctx context.Context) []uuid.UUID {
 	return projects
 }
 
+// isUserOrgAdmin checks if the current user is an org admin or owner
+func isUserOrgAdmin(ctx context.Context, orgUUID uuid.UUID) bool {
```
migmartri: could you confirm we don't have this already?

Piskoo: We don't but I found another helper to simplify it even further

app/controlplane/internal/service/workflowcontract.go
Piskoo: Fixed, thanks

migmartri: thanks

Please re-review the implementation to make sure there is no duplication, feel free to mark it as `draft` if needed, thanks

migmartri (approved):

migmartri:
@Piskoo one thing it's useful for the reviewers is that after you apply some of the feedback, you update the description. For example, the description here still talks about the old error message.

app/controlplane/api/controlplane/v1/organization.proto
```diff
@@ -96,6 +96,9 @@ message OrganizationServiceUpdateRequest {
 
   // prevent workflows and projects from being created implicitly during attestation init
   optional bool prevent_implicit_workflow_creation = 5;
+
+  // restrict_contract_creation_to_org_admins restricts contract creation (org-level and project-level) to only organization admins (owner/admin roles)
+  optional bool restrict_contract_creation_to_org_admins = 6;
```
migmartri, line 101: Didn't we add support for the CLI to update this setting?

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -109,6 +113,19 @@ func (s *WorkflowContractService) Create(ctx context.Context, req *pb.WorkflowCo
...
 	}
 
+	// Check organization settings for contract creation restriction
+	org, err := s.orgUseCase.FindByID(ctx, currentOrg.ID)
+	if err != nil {
+		return nil, handleUseCaseErr(err, s.log)
+	}
+
+	// If setting is enabled, only org admins can create contracts (org-level or project-level)
+	if org.RestrictContractCreationToOrgAdmins {
```
migmartri, line 123: have you tried to create a contract using an org level API token?

PR #2606, opened 2025-12-05
chore(signing): verify signature on att push

jiparis:
This change ensures that the attestation bundle is verified on attestation push, failing if the signature is not recognized (only when chainloop is the signer)

### DIFF
--- a/app/controlplane/pkg/biz/workflowrun.go
+++ b/app/controlplane/pkg/biz/workflowrun.go
@@ -456,7 +456,6 @@ func (uc *WorkflowRunUseCase) VerifyRun(ctx context.Context, run *WorkflowRun) (
 }
 
 func (uc *WorkflowRunUseCase) verifyBundle(ctx context.Context, bundle []byte) (*VerificationResult, error) {
-
 	tr, err := uc.signingUseCase.GetTrustedRoot(ctx)
 	if err != nil {
 		if IsErrNotImplemented(err) {


app/controlplane/pkg/biz/workflowrun.go
```diff
@@ -322,8 +322,19 @@ func (uc *WorkflowRunUseCase) SaveAttestation(ctx context.Context, id string, en
 		return nil, fmt.Errorf("extracting predicate: %w", err)
 	}
 
+	// verify attestation (only if chainloop is the signer)
+	result, err := uc.verifyBundle(ctx, rawContent)
```
migmartri, line 326: where is the logic that makes sure the bundle is verified when keyless signing was used?

app/controlplane/pkg/biz/workflowrun.go
```diff
@@ -440,7 +451,11 @@ type VerificationResult struct {
 	FailureReason string
 }
 
-func (uc *WorkflowRunUseCase) Verify(ctx context.Context, run *WorkflowRun) (*VerificationResult, error) {
+func (uc *WorkflowRunUseCase) VerifyRun(ctx context.Context, run *WorkflowRun) (*VerificationResult, error) {
+	return uc.verifyBundle(ctx, run.Attestation.Bundle)
```
migmartri, line 455: is there any case where run.Attestation would be empty?

migmartri: it seems it will not, makes sense.

app/controlplane/pkg/biz/workflowrun.go
```diff
@@ -322,8 +322,19 @@ func (uc *WorkflowRunUseCase) SaveAttestation(ctx context.Context, id string, en
...
 	}
 
+	// verify attestation (only if chainloop is the signer)
+	result, err := uc.verifyBundle(ctx, rawContent)
+	if err != nil {
+		return nil, err
+	}
+
+	// if it's verifiable, make sure it passed
+	if result != nil && !result.Result {
```
migmartri, line 332: oh, so result will be nil if can't be verified? I'd make it more explicit somehow

jiparis: yes in that case it would return `nil, nil`. I agree it should return something more explicit.

PR #2608, opened 2025-12-05
chore: remove old Dockerfiles

jiparis:
These dockerfiles are not used anymore since Dockerfile.goreleaser ones are used for building images. We might want to keep them for development purposes, but there is no other reason to maintain them.


migmartri (approved):
there might be some references to these files for CLAUDE somewhere. 

Thanks

PR #2611, opened 2025-12-07
feat(policy): implement policy eval

Piskoo:
## Summary

Introduces a new `chainloop policy eval` command that evaluates policies using the control plane with organization settings, supporting both generic and material-based policy evaluation.

## Differences

| Feature           | policy eval               | policy develop eval             |
|-------------------|---------------------------|---------------------------------|
| Authentication    | Required                  | Optional                        |
| Allowed Hostnames | From org config           | Manual `--allowed-hostnames`    |
| Debug Mode        | Not supported             | Supported                       |


app/controlplane/api/controlplane/v1/policy_evaluation.proto
```diff
@@ -0,0 +1,21 @@
...
+
+package controlplane.v1;
+
+import "attestation/v1/crafting_state.proto";
+import "buf/validate/validate.proto";
+
+option go_package = "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1;v1";
+
+service PolicyEvaluationService {
+  rpc Evaluate(PolicyEvaluationServiceEvaluateRequest) returns (PolicyEvaluationServiceEvaluateResponse);
```
migmartri: is it possible that this PR contains too much information, including remote evaluations or smth like that?

app/controlplane/configs/config.devel.yaml
```diff
@@ -97,11 +97,11 @@ auth:
 prometheus_integration:
   - org_name: "development"
 
-# Policy providers configuration
-# policy_providers:
-#   - name: chainloop
-#     default: true
-#     url: http://localhost:8002/v1
+#     Policy providers configuration
+policy_providers:
```
migmartri: please keep it commented out

app/controlplane/internal/service/policyevaluation.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (s *PolicyEvaluationService) Evaluate(ctx context.Context, req *pb.PolicyEvaluationServiceEvaluateRequest) (*pb.PolicyEvaluationServiceEvaluateResponse, error) {
+	// Verify current org is set (middleware handles this)
+	_, err := requireCurrentOrg(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Call business layer to evaluate the policy
+	result, err := s.uc.Evaluate(ctx, &biz.PolicyEvaluationEvaluateOpts{
```
migmartri: why isn't the evaluation happening client side? Why did we move it server-side?

Piskoo:
> is it possible that this PR contains too much information, including remote evaluations or smth like that?

Yes, this PR before was only about remote execution, the idea was to use server side execution as part of declarative control gates (which we decided to not implement). But remote execution was moved out of scope for this project so I cleaned up the PR. Now it contains only changes related to the new command. Sorry about that.

PR #2620, opened 2025-12-15
fix(craftingschema): move annotations to spec

Piskoo:
This PRs fixes the placement of attestation-level annotations in the v2 schema moving them back to spec like in v1 schema.
```
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: chainloop-contract
spec:
  annotations:
    - name: environment
    # value is empty so environment will have to be specified during attestation process
```


app/cli/pkg/action/testdata/contract_v2.yaml
```diff
@@ -3,10 +3,12 @@ kind: Contract
...
   name: test-contract-v2
   description: Test contract in V2 format
-  annotations:
-    version: "1.0.0"
-    team: test-team
 spec:
+  annotations:
+    - name: version
+      value: "1.0.0"
+    - name: team
```
migmartri, line 10: does this work by only providing the name not the value, so the value is recovered at runtime?

Piskoo: Yes, the annotation value is required if it wasn't provided in the contract
```
$ cl att push
ERR annotations ["environment"] required
```

PR #2622, opened 2025-12-15
feat(policies): gated policy attachments

jiparis:
This PR allows to configure individual policy evaluations as attestation gates during `att push` operations. A new `gate` property has been added to the policy attachment:

Example contract:
```yaml
apiVersion: chainloop.dev/v1
kind: Contract
metadata:
  name: gate-contract
  description: Test contract with gate
spec:
  policies:
    attestation:
      - ref: file://test/gate-policy.yaml
        gate: true
```

Example policy:
```yaml
kind: Policy
metadata:
  name: policy-with-gate
spec:
  policies:
    - kind: ATTESTATION
      embedded: |
        package main
        import rego.v1
        
        result := {"violations": ["Gated policy violation"]}
```

Attestation process:
```
> chainloop att init --workflow gate-wf --project gate-project --contract test/gate-contract.yaml
This command will run against the organization "my-org"
y
┌───────────────────────────┬──────────────────────────────────────────┐
│ Initialized At            │ 15 Dec 25 19:04 UTC                      │
├───────────────────────────┼──────────────────────────────────────────┤
│ Attestation ID            │ 084d6158-c7c4-4efc-9cb3-35cc904301ce     │
│ Organization              │ my-org                                   │
│ Name                      │ gate-wf                                  │
│ Project                   │ gate-project                             │
│ Version                   │ v1.62.1+next (prerelease)                │
│ Contract                  │ gate-project-gate-wf (revision 1)        │
│ Policy violation strategy │ ADVISORY                                 │
│ Policies                  │ ------                                   │
│                           │ policy-with-gate: Gated policy violation │
└───────────────────────────┴──────────────────────────────────────────┘

> chainloop att push
INF push completed
│ Initialized At            │ 16 Dec 25 09:15 UTC                                                     │
│ Attestation ID            │ f5c9fda7-47f5-46bb-8146-bdad1f390cfa                                    │
│ Digest                    │ sha256:526b12ec0318637c77267f6625ad59d81383d93becf8bfbc1e3dac8baae95837 │
│ Organization              │ my-org                                                                  │
│ Name                      │ gate-wf                                                                 │
│ Project                   │ gate-project                                                            │
│ Version                   │ v1.62.1+next (prerelease)                                               │
│ Contract                  │ gate-project-gate-wf (revision 1)                                       │
│ Policies                  │ ------                                                                  │
│                           │ policy-with-gate (gate): gate policy triggered                          │
ERR the policy "policy-with-gate" is configured as a gate and has violations
exit status 4
```

Note that this setting is independent to the global policy violation strategy configured at the org level.


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -229,6 +229,9 @@ message PolicyAttachment {
     }
   }];
 
+  // If true, the policy will act as a gate, failing the attestation if the policy fails
```
migmartri: does it fail the attestation or it will make the client fail but the attestation will be still be sent and the run marked as success?

app/cli/cmd/attestation_push.go
```diff
@@ -122,8 +122,20 @@ func newAttestationPushCmd() *cobra.Command {
...
-			// and fail the command if needed
+			// Block if any of the policies has been configured as a gate
+			for _, evaluations := range res.Status.PolicyEvaluations {
+				for _, eval := range evaluations {
+					if len(eval.Violations) > 0 && eval.Gate {
+						if bypassPolicyCheck {
+							logger.Warn().Msg(exceptionBypassPolicyCheck)
+							continue
+						}
+						return fmt.Errorf("your contract requires the policy %q to pass before continuing", eval.Name)
```
migmartri: shouldn't we return `ErrBlockedByPolicyViolation` so a custom error that also returns code error 3 or 4?

migmartri:
A couple of questions

- Do we still send the attestation up but mark is as `gated`? similarly to what we do with the policy evaluation enforced?
- If the gated policy is a material one, do we still only fail on `att push`? This seems consistent but double checking

migmartri (commented):
Thanks @jiparis 

I left a couple of comments, I am interested in knowing how this behavior related to our current policy enforcement mechanism, as in if the data is sent to chainloop, marked as enforced, etc.

app/cli/cmd/attestation_push.go
jiparis: I've just implemented a custom error. It now returns a `4`

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
jiparis: I've changed the comment. It will be consistent with the global block strategy, returning an error code, but always sending the attestation.

migmartri:
It would be nice if we could also add next to the policy information if it's a `gate` or not. Meaning that when you initialize, status or push, you get `policy-name` (gate)

PR #2629, opened 2025-12-21
feat(cas-backend): introduce fallback mechanism

Piskoo:
Adds support for configuring a fallback CAS backend that is automatically used when the default backend becomes unavailable. 

## Changes
- Added `fallback` boolean field to CAS backends
- A backend cannot be both `default` and `fallback`
- Creating a new fallback backend automatically unsets the previous one
- Fallback backends are included in periodic validation (every 30 minutes)
- During attestation process cas backend is decided as follows:
  - Try default backend
  - If default invalid, try fallback backend
  - If fallback invalid or missing, return error
- Added `--fallback` flag to all `casbackend add` and `update` commands.

Example
```
$ chainloop casbackend add oci --repo index.docker.io/repo --username name --password password  --name fallbacktest --fallback

┌───────────────┬────────────────────────┬──────────┬─────────────┬───────────────┬─────────┬──────────┬────────┐
│ NAME          │ LOCATION               │ PROVIDER │ DESCRIPTION │ LIMITS        │ DEFAULT │ FALLBACK │ STATUS │
├───────────────┼────────────────────────┼──────────┼─────────────┼───────────────┼─────────┼──────────┼────────┤
│ fallbacktest  │ index.docker.io/repo   │ OCI      │             │ MaxSize: 300M │ false   │ true     │ valid  │
└───────────────┴────────────────────────┴──────────┴─────────────┴───────────────┴─────────┴──────────┴────────┘
```

During `attestation init` cas backend is decided seamlessly, we will provide information on what cas backend was used in the attestation itself.

### DIFF
--- a/app/controlplane/pkg/biz/casbackend.go
+++ b/app/controlplane/pkg/biz/casbackend.go
@@ -111,8 +111,8 @@ type CASBackendRepo interface {
 	List(ctx context.Context, orgID uuid.UUID) ([]*CASBackend, error)
 	UpdateValidationStatus(ctx context.Context, ID uuid.UUID, status CASBackendValidationStatus, validationError *string) error
 	// ListBackends returns CAS backends across all organizations
-	// If onlyDefaults is true, only default backends are returned
-	ListBackends(ctx context.Context, onlyDefaults bool) ([]*CASBackend, error)
+	// If defaultsOrFallbacks is true, only default and fallback backends are returned
+	ListBackends(ctx context.Context, defaultsOrFallbacks bool) ([]*CASBackend, error)
 	Create(context.Context, *CASBackendCreateOpts) (*CASBackend, error)
 	Update(context.Context, *CASBackendUpdateOpts) (*CASBackend, error)
 	Delete(ctx context.Context, ID uuid.UUID) error
@@ -246,6 +246,7 @@ func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, o
 	}
 
 	if defaultBackend == nil {
+		// Should never happen
 		return nil, NewErrNotFound("default CAS Backend")
 	}
 
@@ -260,10 +261,12 @@ func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, o
 
 	fallbackBackend, err := uc.FindFallbackBackend(ctx, orgID)
 	if err != nil {
-		if IsNotFound(err) {
-			return nil, NewErrValidationStr("default CAS backend is unreachable and no fallback backend is configured")
-		}
 		return nil, err
+	}
+
+	if fallbackBackend == nil {
+		// Should never happen
+		return nil, NewErrNotFound("fallback CAS Backend")
 	}
 
 	// Check if fallback backend is valid
@@ -287,7 +290,6 @@ func (uc *CASBackendUseCase) CreateInlineBackend(ctx context.Context, orgID stri
 		CASBackendOpts: &CASBackendOpts{
 			Provider:    CASBackendInline,
 			Default:     ToPtr(true),
-			Fallback:    ToPtr(false),
 			Description: &CASBackendInlineDescription,
 			OrgID:       orgUUID,
 		},
@@ -295,6 +297,7 @@ func (uc *CASBackendUseCase) CreateInlineBackend(ctx context.Context, orgID stri
 }
 
 // promoteNextAvailableBackend promotes the next available backend to default.
+// Promotion priority: try fallback backend first, then inline backend if no fallback exists.
 func (uc *CASBackendUseCase) promoteNextAvailableBackend(ctx context.Context, orgID string) (*CASBackend, error) {
 	orgUUID, err := uuid.Parse(orgID)
 	if err != nil {
@@ -304,15 +307,17 @@ func (uc *CASBackendUseCase) promoteNextAvailableBackend(ctx context.Context, or
 	backend, err := uc.repo.FindFallbackBackend(ctx, orgUUID)
 	if err != nil {
 		return nil, err
-	} else if backend == nil {
+	}
+
+	if backend == nil {
 		// If there is no fallback backend, try to find and use inline backend
-		inlineBackend, err := uc.repo.FindInlineBackend(ctx, orgUUID)
+		backend, err = uc.repo.FindInlineBackend(ctx, orgUUID)
 		if err != nil {
 			return nil, err
-		} else if inlineBackend == nil {
+		}
+		if backend == nil {
 			return nil, nil
 		}
-		return uc.repo.Update(ctx, &CASBackendUpdateOpts{ID: inlineBackend.ID, CASBackendOpts: &CASBackendOpts{Default: ToPtr(true)}})
 	}
 
 	return uc.repo.Update(ctx, &CASBackendUpdateOpts{ID: backend.ID, CASBackendOpts: &CASBackendOpts{Default: ToPtr(true)}})
@@ -321,6 +326,11 @@ func (uc *CASBackendUseCase) promoteNextAvailableBackend(ctx context.Context, or
 func (uc *CASBackendUseCase) Create(ctx context.Context, orgID, name, location, description string, provider CASBackendProvider, creds any, defaultB bool, fallbackB bool, maxBytes *int64) (*CASBackend, error) {
 	if orgID == "" || name == "" {
 		return nil, NewErrValidationStr("organization and name are required")
+	}
+
+	// Backend cannot be both default and fallback
+	if defaultB && fallbackB {
+		return nil, NewErrValidationStr("a backend cannot be both default and fallback")
 	}
 
 	orgUUID, err := uuid.Parse(orgID)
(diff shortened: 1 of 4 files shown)


app/controlplane/pkg/biz/casbackend.go
```diff
@@ -233,26 +236,62 @@ func (uc *CASBackendUseCase) FindFallbackBackend(ctx context.Context, orgID stri
 	return backend, nil
 }
 
-func (uc *CASBackendUseCase) CreateInlineFallbackBackend(ctx context.Context, orgID string) (*CASBackend, error) {
+// FindDefaultOrFallbackBackend finds a valid CAS backend for the organization.
+// Attempts to use the default backend first, if invalid it uses the fallback backend.
+func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, orgID string) (*CASBackend, error) {
+	// Find the default backend
+	defaultBackend, err := uc.FindDefaultBackend(ctx, orgID)
```
migmartri, line 243: can defaultBackend be nil?

app/controlplane/internal/service/attestation.go
```diff
@@ -174,17 +174,15 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
-	}
-
-	// Check the status of the backend
-	if backend.ValidationStatus != biz.CASBackendValidationOK {
-		return nil, cpAPI.ErrorCasBackendErrorReasonInvalid("your CAS backend can't be reached")
+	// Find the default or fallback CAS backend to associate the workflow
+	backend, err := s.casUC.FindDefaultOrFallbackBackend(context.Background(), robotAccount.OrgID)
+	if err != nil {
+		if biz.IsNotFound(err) {
+			return nil, errors.NotFound("not found", "default CAS backend not found")
```
migmartri: we might be able to use something called handle_errors?

migmartri: this error message is not accurate, let's try to use `handle_errors`

app/controlplane/internal/service/attestation.go
```diff
@@ -174,17 +174,15 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
+		} else if biz.IsErrValidation(err) {
```
migmartri, line 182: this should return `ErrorCasBackendErrorReasonInvalid` no?

app/controlplane/pkg/biz/casbackend.go
Piskoo: no, but I'll add a check

app/controlplane/internal/service/attestation.go
```diff
@@ -174,17 +174,15 @@ func (s *AttestationService) Init(ctx context.Context, req *cpAPI.AttestationSer
...
-
-	// Check the status of the backend
-	if backend.ValidationStatus != biz.CASBackendValidationOK {
-		return nil, cpAPI.ErrorCasBackendErrorReasonInvalid("your CAS backend can't be reached")
+	// Find the default or fallback CAS backend to associate the workflow
+	backend, err := s.casUC.FindDefaultOrFallbackBackend(context.Background(), robotAccount.OrgID)
+	if err != nil {
+		if biz.IsNotFound(err) {
+			return nil, cpAPI.ErrorCasBackendErrorReasonRequired(err.Error())
+		} else if biz.IsErrValidation(err) {
```
migmartri, line 182: Do we need double conditional here?

Piskoo: Yes if we want to return more specific error or we can pass it in all cases to  `handleUseCaseErr`

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -412,10 +462,10 @@ func (uc *CASBackendUseCase) Update(ctx context.Context, orgID, id string, descr
 		}
 	}
 
-	// If we just updated the backend from default=true => default=false, we need to set up the fallback as default
+	// If we just updated the backend from default=true => default=false, we need to promote the next available backend
```
jiparis, line 480: Is this comment wrong?, since it's just comparing "before" with "after", but I don't see such transition from `true` to `false`. 
Is this logic promoting the next backend no matter the value of `after.Default`?

Piskoo: The comment is correct. If there is a default backend(before.Default=true) and the user decides to demote it in the update endpoint(after.Default=false) the next one in the order gets promoted. Previously it was inline backend, now we promote fallback first if available.

app/controlplane/pkg/data/casbackend.go
```diff
@@ -145,14 +173,39 @@ func (r *CASBackendRepo) Update(ctx context.Context, opts *biz.CASBackendUpdateO
...
+			updateChain = updateChain.SetDefault(*opts.Default)
+			if *opts.Default {
+				updateChain = updateChain.SetFallback(false)
+			}
+		}
+
+		if opts.Fallback != nil {
+			updateChain = updateChain.SetFallback(*opts.Fallback)
+			if *opts.Fallback {
+				updateChain = updateChain.SetDefault(false)
```
migmartri, line 205: We should not get there since we have a guard cause right?

app/controlplane/pkg/data/casbackend.go
```diff
@@ -253,7 +306,7 @@ func (r *CASBackendRepo) UpdateValidationStatus(ctx context.Context, id uuid.UUI
 }
 
 // ListBackends returns CAS backends across all organizations. Only not inline backends are returned
-// If onlyDefaults is true, only default backends are returned
+// If onlyDefaults is true, only default and fallback backends are returned
 func (r *CASBackendRepo) ListBackends(ctx context.Context, onlyDefaults bool) ([]*biz.CASBackend, error) {
```
migmartri: defaultsOrFallbacks

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -233,26 +236,66 @@ func (uc *CASBackendUseCase) FindFallbackBackend(ctx context.Context, orgID stri
 	return backend, nil
 }
 
-func (uc *CASBackendUseCase) CreateInlineFallbackBackend(ctx context.Context, orgID string) (*CASBackend, error) {
+// FindDefaultOrFallbackBackend finds a valid CAS backend for the organization.
+// Attempts to use the default backend first, if invalid it uses the fallback backend.
+func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, orgID string) (*CASBackend, error) {
+	// Find the default backend
+	defaultBackend, err := uc.FindDefaultBackend(ctx, orgID)
```
migmartri, line 243: FindDefaultbackend and `FindFallbackBackend` dont' seem to be consistent in the way that the former might return `nil` if no backend is present and the latter returns a notFoundError?

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -233,26 +236,66 @@ func (uc *CASBackendUseCase) FindFallbackBackend(ctx context.Context, orgID stri
...
+		return defaultBackend, nil
+	}
+
+	// Default backend is invalid, try fallback
+	uc.logger.Infow("msg", "default CAS backend validation failed, attempting fallback",
+		"backend", defaultBackend.Name, "status", defaultBackend.ValidationStatus, "orgID", orgID)
+
+	fallbackBackend, err := uc.FindFallbackBackend(ctx, orgID)
+	if err != nil {
+		if IsNotFound(err) {
```
migmartri: to my comment before, this one is checking isNotFound instead of `fallbackBackend` == nil.

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -233,26 +236,66 @@ func (uc *CASBackendUseCase) FindFallbackBackend(ctx context.Context, orgID stri
...
 
 	return uc.repo.Create(ctx, &CASBackendCreateOpts{
 		Name:     "default-inline",
-		Fallback: true,
 		MaxBytes: CASBackendInlineDefaultMaxBytes,
 		CASBackendOpts: &CASBackendOpts{
-			Provider: CASBackendInline, Default: ToPtr(true),
+			Provider:    CASBackendInline,
+			Default:     ToPtr(true),
+			Fallback:    ToPtr(false),
```
migmartri: passing false might not be necessary isn't it?

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -262,14 +305,20 @@ func (uc *CASBackendUseCase) defaultFallbackBackend(ctx context.Context, orgID s
 	if err != nil {
 		return nil, err
 	} else if backend == nil {
-		// If there is no fallback backend, we skip the update
-		return nil, nil
+		// If there is no fallback backend, try to find and use inline backend
+		inlineBackend, err := uc.repo.FindInlineBackend(ctx, orgUUID)
```
migmartri: should we promote the inline or the fallback? Or both?

migmartri: nevermind, the closure is what confused me, can you call this variable `backend` and use the update outside?

app/controlplane/pkg/data/casbackend.go
Piskoo: Yes, but I added it as safety net

app/controlplane/pkg/biz/casbackend.go
Piskoo: I see what you meant now, I made it consistent in both cases

migmartri: Did you double check the the data layer is returning a nil object when it cannot be found?

migmartri: Did you double check the the data layer is returning a nil object when it cannot be found?

Piskoo: yes it does, thanks

PR #2632, opened 2025-12-22
feat: store cas backend information in attestation

### LINKED ISSUE #2631: feat: add cas backend tracking to attestations
In the attestation process, we want to track which CAS backend was used and whether it was a fallback backend. Add CAS backend information to the attestation in the same format as the existing auth information e.g.
 ```json
  "cas_backend": {
    "id": "1c879a79-11db-465b-bf3e-da5e338bdaf2"
    "fallback": true
}
```

### PR DESCRIPTION
**Pull request by Piskoo**
Adds information about cas backend used in attestation process to the attestation

Part of attestation statement
```
  "predicate": {
    "casBackend": {
      "casBackendID": "07ddd65b-ed78-43ab-8914-659aab11f5c3",
      "casBackendName": "casb",
      "fallback": false
    },
```

Closes #2631

### DIFF
--- a/pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
+++ b/pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
@@ -179,7 +179,7 @@ message Attestation {
 
   message CASBackend {
     // UUID of the CAS backend
-    string cas_backend_id = 1 [(buf.validate.field).string = {min_len: 1}];
+    string cas_backend_id = 1 [(buf.validate.field).string.uuid = true];
     // Name of the CAS backend
     string cas_backend_name = 2 [(buf.validate.field).string = {min_len: 1}];
     // Whether this is a fallback backend
--- a/app/controlplane/api/gen/jsonschema/attestation.v1.Attestation.CASBackend.jsonschema.json
+++ b/app/controlplane/api/gen/jsonschema/attestation.v1.Attestation.CASBackend.jsonschema.json
@@ -5,7 +5,7 @@
   "patternProperties": {
     "^(cas_backend_id)$": {
       "description": "UUID of the CAS backend",
-      "minLength": 1,
+      "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$",
       "type": "string"
     },
     "^(cas_backend_name)$": {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: app/cli/pkg/action/action.go [hunk 1]
```diff
@@ -100,7 +101,7 @@ func newCrafter(stateOpts *newCrafterStateOpts, conn *grpc.ClientConn, opts ...c
 }
 
 // getCASBackend tries to get CAS upload credentials and set up a CAS client
-func getCASBackend(ctx context.Context, client pb.AttestationServiceClient, workflowRunID, casCAPath, casURI string, casConnectionInsecure bool, logger zerolog.Logger, casBackend *casclient.CASBackend) (func() error, error) {
+func getCASBackend(ctx context.Context, client pb.AttestationServiceClient, workflowRunID, casCAPath, casURI string, casConnectionInsecure bool, logger zerolog.Logger, casBackend *casclient.CASBackend, casBackendInfo **clientAPI.Attestation_CASBackend) (func() error, error) {
```
jiparis: I find this pattern super opaque. Why not returning the `*Attestation_CASBackend` always as part of the results? 
```golang
func getCASBackend(ctx context.Context, client pb.AttestationServiceClient, workflowRunID, casCAPath, casURI string, casConnectionInsecure bool, logger zerolog.Logger, casBackend *casclient.CASBackend) (*clientAPI.Attestation_CASBackend, func() error, error) {
```

Piskoo: You're right, thanks

pkg/attestation/crafter/api/attestation/v1/crafting_state.proto
```diff
@@ -174,6 +177,15 @@ message Attestation {
     }
   }
 
+  message CASBackend {
+    // UUID of the CAS backend
+    string cas_backend_id = 1 [(buf.validate.field).string = {min_len: 1}];
```
migmartri: we should use the uuid validation here instead

pkg/attestation/renderer/chainloop/chainloop.go
```diff
@@ -136,6 +137,12 @@ type Auth struct {
 	Type string `json:"type"`
 }
 
+type CASBackend struct {
+	CASBackendID   string `json:"casBackendId"`
```
migmartri:
```suggestion
	CASBackendID   string `json:"casBackendID"`
```

migmartri (approved):
ptal at the my comments.

Thanks!

PR #2636, opened 2025-12-23
feat: add disable requirements auto matching setting

Piskoo:
Adds organization-level setting to disable automatic matching of policies tocompliance requirements.

### Changes

- Added `disable_requirements_auto_matching` bool field to organizations
- Added `--disable-requirements-auto-matching flag` to `organization update` cli command


migmartri (changes requested):
I think this option `disable_requirements_auto_matching ` should be added in org_settings in platform

PR #2651, opened 2026-01-08
feat(api): add team field to WorkflowRef message

migmartri:
## Summary
Add optional team field to WorkflowRef message to provide team context when referencing workflows in contract responses.

This change extends the WorkflowRef message to include the team field, aligning it with WorkflowItem which already exposes team information. The team field is populated from the workflow entity and included in API responses for workflow contract describe and list operations.

## Changes
- Extended WorkflowRef protobuf message with team field
- Updated business layer, data layer, and service layer to handle team
- Updated CLI action layer for proper JSON serialization
- Updated CLAUDE.md with current year (2026)

The change is fully backward compatible as the team field is optional.

### DIFF
--- a/app/controlplane/api/controlplane/v1/response_messages.pb.go
+++ b/app/controlplane/api/controlplane/v1/response_messages.pb.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024-2025 The Chainloop Authors.
+// Copyright 2024-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


jiparis:
I thought we were going to remove the `team` field. Is it still used?

migmartri:
> I thought we were going to remove the `team` field. Is it still used?

yes it is, although behind `--wide` flags

I am implementing it to maintain feature parity, but let's discuss the deprecation of team

app/cli/pkg/action/workflow_contract_list.go
```diff
@@ -131,5 +132,5 @@ func pbWorkflowContractVersionItemToAction(in *pb.WorkflowContractVersionItem) *
 }
 
 func pbWorkflowRefToAction(in *pb.WorkflowRef) *WorkflowRef {
-	return &WorkflowRef{ID: in.GetId(), Name: in.GetName(), ProjectName: in.GetProjectName()}
+	return &WorkflowRef{ID: in.GetId(), Name: in.GetName(), ProjectName: in.GetProjectName(), Team: in.GetTeam()}
```
javirln, line 135: Is the team needed to identify a Workflow? Isn't it enough with the ID or Name?

migmartri: it is used in the UI though

PR #2652, opened 2026-01-09
feat(user-verification): Introduce user verification on attestation init

javirln:
This PR adds platform-based commit signature verification during attestation initialization. When creating an attestation, the system now queries GitHub or GitLab APIs to verify if the commit was signed and validated by the platform.

Extended the attestation initialization process to include commit signature verification status. The verification captures whether a commit's signature was validated by GitHub or GitLab, including details about the signature algorithm, key ID, and verification status.

Added new `VerifyCommitSignature` method to the `SupportedRunner` interface. Implemented verification for:
- GitHub Actions: Uses GitHub REST API with `GITHUB_TOKEN`
- GitLab CI: Uses GitLab API /signature endpoint with `CI_JOB_TOKEN`
- Dagger Pipeline: Detects underlying CI platform and delegates appropriately

### Protobuf Schema

Extended `Commit` message with optional `platform_verification` field containing:
- Verification status (verified, unverified, unavailable, not applicable)
- Reason for the verification result
- Platform name (github, gitlab)
- Key ID and signature algorithm

Verification failures never block attestation creation. All errors result in recording the unavailable or not applicable status, allowing the attestation process to continue.

### Example
Example from a GitHub Action. This is part of the crafting state:
```json
"runnerType": "GITHUB_ACTION",
    "head": {
      "hash": "6a6f5761863d107f39df09b7cdc34a464c44a0f7",
      "authorEmail": "REDACTED",
      "authorName": "Javier Rodríguez",
      "message": "Update test-commit-verification.yml",
      "date": "2026-01-09T12:25:23Z",
      "remotes": [
        {
          "name": "origin",
          "url": "https://github.com/javirln/goreleaser"
        }
      ],
      "signature": "-----BEGIN PGP SIGNATURE-----SIGNATURE-----END PGP SIGNATURE-----\n\n",
      "platformVerification": {
        "attempted": true,
        "status": "VERIFICATION_STATUS_VERIFIED",
        "reason": "valid",
        "platform": "github",
        "signatureAlgorithm": "PGP"
      }
    },
    "signingOptions": {
      "timestampAuthorityUrl": "http://timestamp.digicert.com/",
      "signingCa": "fileCA"
    },
```
Example of attestation:
```json
"annotations": {
  "author.email": "REDACTED",
  "author.name": "Javier Rodríguez",
  "author.verified": "VERIFICATION_STATUS_VERIFIED",
  "date": "2026-01-09T12:30:26Z",
  "message": "Update test-commit-verification.yml",
  "remotes": [
    {
      "name": "origin",
      "url": "https://github.com/javirln/goreleaser"
    }
  ],
  "signature": "-----BEGIN PGP SIGNATURE-----SIGNATURE-----END PGP SIGNATURE-----\n\n",
  "signature.algorithm": "PGP"
}
```

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -106,6 +106,9 @@ modules:
       except:
         - FIELD_NOT_REQUIRED
         - PACKAGE_NO_IMPORT_CYCLE
+        - ENUM_VALUE_UPPER_SNAKE_CASE
+        - ENUM_VALUE_PREFIX
+        - ENUM_ZERO_VALUE_SUFFIX
       disallow_comment_ignores: true
     breaking:
       use:


jiparis (approved):
Looks good to me, thanks!

pkg/attestation/renderer/chainloop/v02.go
```diff
@@ -144,6 +144,13 @@ func commitAnnotations(c *v1.Commit) (*structpb.Struct, error) {
 		annotationsRaw[subjectGitAnnotationSignature] = c.GetSignature()
 	}
 
+	// add author verification only if exists
+	if c.GetPlatformVerification() != nil {
+		pv := c.GetPlatformVerification()
+		annotationsRaw[subjectGitAnnotationAuthorVerified] = pv.GetStatus().String()
```
migmartri: I am wondering if we shoudl just put `true` or `false` here since the key seems to be a bool not a state

migmartri: then if false we can add another annotaiton wth the reason on why it's not verified? That would make policy authoring easier no?

pkg/attestation/crafter/crafter.go
```diff
@@ -168,6 +168,8 @@ type InitOpts struct {
 	PoliciesAllowedHostnames []string
 	// CAS backend information
 	CASBackend *api.Attestation_CASBackend
+	// Logger for verification logging
+	Logger *zerolog.Logger
```
migmartri, line 173: thanks

pkg/attestation/crafter/runner.go
```diff
@@ -50,8 +51,13 @@ type SupportedRunner interface {
...
 	IsAuthenticated() bool
 
-	// RunnerEnvironment returns the runner environment
+	// Environment returns the runner environment
 	Environment() runners.RunnerEnvironment
+
+	// VerifyCommitSignature checks if a commit's signature is verified by the platform.
+	// Returns nil if verification is not supported or not applicable for this runner.
+	// Non-blocking: errors are logged and returned as unavailable status.
+	VerifyCommitSignature(ctx context.Context, commitHash string) *api.Commit_CommitVerification
```
migmartri: awesome although nitpick. I'd not return `api` but its own struct, that way we don't couple the implementation of the underlying runners to it

migmartri (approved):
Thanks @javirln, see my comment about updating the annotations

pkg/attestation/renderer/chainloop/v02.go
javirln: Good question and I have the feeling that instead of suggesting a bool on the annotation it should be a `status`. The reason is the possible values of the verification already includes the possible results:
```
VERIFICATION_STATUS_UNSPECIFIED = 0;
// Successfully verified by platform
VERIFICATION_STATUS_VERIFIED = 1;
// Platform checked but signature is invalid/unverified
VERIFICATION_STATUS_UNVERIFIED = 2;
// Verification could not be performed (no API access, network error, etc.)
VERIFICATION_STATUS_UNAVAILABLE = 3;
// Platform doesn't support verification or no commit signature present
VERIFICATION_STATUS_NOT_APPLICABLE = 4;
```

So my proposal would be to go for `author.verification_status` instead of `author.verified`

migmartri: ok, I'd also change `VERIFICATION_STATUS_UNVERIFIED` to not to be that PROTOBUFFY since we are bypassing that linting reqiurement in other places

meaning `verified`, `undefified`,  ...

migmartri:
Btw, this feature is awesome, lert's make sure we update our docs and changelog to reflect it

PR #2654, opened 2026-01-12
fix(wasm): Fixes WASM policy examples, add missing documentation and reference to docs

javirln:
Fixes references to example policies of WASM policies in Go and Javascript, add missing READMEs and updates link to official documentation.


javirln:
> Should we remove the WASM binaries?
> Image
> 
> https://securityscorecards.dev/viewer/?uri=github.com/chainloop-dev/chainloop

Ok, let's remove them, they can be easily reproducible.

PR #2657, opened 2026-01-13
feat(api-token): add instance admin token

Piskoo:
## Summary
This PR adds support for instance admin tokens to existing api token logic 

### Changes
- Organization in api tokens is now optional
- Adds an explicit JWT scope claim for instance admin API tokens to enable clearer identification
- Add permissions(policies) planned for instance admin token 
- Adds correct handling for instance admin tokens to api token middleware

### DIFF
--- a/app/controlplane/pkg/biz/apitoken.go
+++ b/app/controlplane/pkg/biz/apitoken.go
@@ -237,16 +237,14 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description 
 	}
 
 	// Dispatch the event to the auditor to notify the creation of the token
-	if orgUUID != nil {
-		uc.auditorUC.Dispatch(ctx, &events.APITokenCreated{
-			APITokenBase: &events.APITokenBase{
-				APITokenID:   &token.ID,
-				APITokenName: name,
-			},
-			APITokenDescription: description,
-			ExpiresAt:           expiresAt,
-		}, orgUUID)
-	}
+	uc.auditorUC.Dispatch(ctx, &events.APITokenCreated{
+		APITokenBase: &events.APITokenBase{
+			APITokenID:   &token.ID,
+			APITokenName: name,
+		},
+		APITokenDescription: description,
+		ExpiresAt:           expiresAt,
+	}, orgUUID)
 
 	return token, nil
 }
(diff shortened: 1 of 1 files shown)


app/controlplane/pkg/data/ent/migrate/migrations/20260112115927.sql
```diff
@@ -0,0 +1,6 @@
+-- Make organization_id nullable to support instance-level API tokens
+ALTER TABLE "api_tokens" ALTER COLUMN "organization_id" DROP NOT NULL;
+
+-- Create index "apitoken_name" to table: "api_tokens"
+CREATE UNIQUE INDEX "apitoken_name" ON "api_tokens" ("name") WHERE ((revoked_at IS NULL) AND (organization_id IS NULL));
```
migmartri, line 5: Does this mean that now you can have an API token with the same name in the same org?

jiparis: If I'm reading correctly, it's unique when organization_id is null. In other conditions, the existing constraints keep the same.

Piskoo: No, you can have an instance admin API token with the same name if the other one with the same name was revoked, the same way as org API tokens in here https://github.com/chainloop-dev/chainloop/blob/4dc6b0ae16ecdeba3ecefd240205def84257c02c/app/controlplane/pkg/data/ent/schema/apitoken.go#L68-L69

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -194,34 +212,41 @@ func (uc *APITokenUseCase) Create(ctx context.Context, name string, description
...
 	// Dispatch the event to the auditor to notify the creation of the token
-	uc.auditorUC.Dispatch(ctx, &events.APITokenCreated{
-		APITokenBase: &events.APITokenBase{
-			APITokenID:   &token.ID,
-			APITokenName: name,
-		},
-		APITokenDescription: description,
-		ExpiresAt:           expiresAt,
-	}, &orgUUID)
+	if orgUUID != nil {
```
jiparis: `Dispatch` already accepts a `nil` UUID, so this condition is not needed.

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -347,12 +389,14 @@ func (uc *APITokenUseCase) Revoke(ctx context.Context, orgID, id string) error {
 	}
 
 	// Dispatch the event to the auditor to notify the revocation of the token
-	uc.auditorUC.Dispatch(ctx, &events.APITokenRevoked{
-		APITokenBase: &events.APITokenBase{
-			APITokenID:   &tokenUUID,
-			APITokenName: token.Name,
-		},
-	}, &orgUUID)
+	if orgUUID != nil {
```
jiparis: Same here, a `nil` `orgUUID` is valid.

app/controlplane/pkg/authz/authz.go
```diff
@@ -346,8 +356,10 @@ var ServerOperationsMap = map[string][]*Policy{
 	"/controlplane.v1.CASRedirectService/DownloadRedirect": {PolicyArtifactDownload},
 	// Or to retrieve a download url
 	"/controlplane.v1.CASRedirectService/GetDownloadURL": {PolicyArtifactDownload},
-	// CAS Backend listing
+	// CAS Backend
+	"/controlplane.v1.CASBackendService/Create":     {PolicyCASBackendCreate},
```
jiparis: All these permission changes are breaking changes and are not needed for this PR. 
I'd leave them for another PR or set of PRs (we can start just with OrganizationCreate) once we figure out the middlewares configuration.

Piskoo: Alright, I'll add them as we need

PR #2665, opened 2026-01-14
feat(source-commit): Update source-commit to enforce signature verification

javirln:
This PR updates some contracts of the workflows present on the repository to enforce author's commit signature verification by using the policy `source-commit`.

This enforcement will check if the commit's author has a signature that has been verified, in this case by GitHub.

More information what commit verification is, can be found on this [link](https://docs.chainloop.dev/concepts/attestations#commit-verification) to the documentation.

### DIFF
--- a/.github/workflows/package_chart.yaml
+++ b/.github/workflows/package_chart.yaml
@@ -90,6 +90,8 @@ jobs:
         env:
           COSIGN_PRIVATE_KEY: ${{secrets.COSIGN_KEY}}
           COSIGN_PASSWORD: ${{secrets.COSIGN_PASSWORD}}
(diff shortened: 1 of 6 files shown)


.github/workflows/package_chart.yaml
```diff
@@ -90,6 +90,8 @@ jobs:
         env:
           COSIGN_PRIVATE_KEY: ${{secrets.COSIGN_KEY}}
           COSIGN_PASSWORD: ${{secrets.COSIGN_PASSWORD}}
+          # Needed for commit signature verification: https://docs.chainloop.dev/concepts/attestations#commit-verification
```
migmartri, line 93: thanks for adding these comments :)

PR #2677, opened 2026-01-21
feat(attestation): Make attestation view URL clickable in CI/CD reports

### LINKED ISSUE #2669: Include attestation link on attestation status once attestation is pushed
Similar to the work done on https://github.com/chainloop-dev/chainloop/issues/2667, it would be ideal for attestations to print the attestation link for easy inspection and access.

migmartri:
## Summary

- changed dashboard URL, closes #2669 
- Updated GitHub Actions and GitLab CI attestation reports to include clickable markdown links to attestation details
- The attestation view URL is now rendered as a markdown link after the status table in CI/CD platform reports
- Modified the Report() method signature across all runner implementations to accept the attestation view URL parameter

image

### DIFF
--- a/app/cli/cmd/attestation_push.go
+++ b/app/cli/cmd/attestation_push.go
@@ -151,12 +151,6 @@ func newAttestationPushCmd() *cobra.Command {
 				}
 			}
 
-			if flagOutputFormat == output.FormatTable {
-				if res.Status.AttestationViewURL != "" {
-					logger.Info().Msg(res.Status.AttestationViewURL)
-				}
-			}
-
 			// Do a final check in case the operator has configured the attestation to be blocked on any policy violation
 			if res.Status.MustBlockOnPolicyViolations {
 				if bypassPolicyCheck {


pkg/attestation/crafter/runners/githubaction.go
```diff
@@ -168,18 +168,19 @@ func (r *GitHubAction) VerifyCommitSignature(ctx context.Context, commitHash str
...
-	content.WriteString("```\n")
-	content.Write(tableOutput)
-	content.WriteString("```\n")
+	// Build the markdown content
+	content := fmt.Sprintf("## Chainloop Attestation Report\n\n```\n%s```\n", tableOutput)
+
+	// Add clickable markdown link after the table if URL is provided
+	if attestationViewURL != "" {
+		content += fmt.Sprintf("\n[View Attestation Details](%s)\n", attestationViewURL)
+	}
```
javirln, line 183: This and the one in GitLab will basically print twice the URL, once inside the table with the attestation status and then after the table was rendered as markdown, is that expected?

migmartri: I think so, I don't want to mess with the ascii table, I looked into changing the content inside but it will mess with its formatting, the solution in reality is to translate the whole table into markdown

Issue #2065, opened 2025-05-22
Error decoding attestations with INLINE binary data


jiparis:
When using binary data, the `value` field is interpreted as a String, so the GRPC marshaler complains about invalid UTF-8 data:

```bash
➜  head -c 10 /dev/urandom > random.bin
➜  chainloop att add --value random.bin
INF material kind detected kind=ARTIFACT
INF material added to attestation
┌──────────┬─────────────────────────────────────────────────────────────────────────┐
│ Name     │ material-1747928434860035000                                            │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Type     │ ARTIFACT                                                                │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Required │ No                                                                      │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Value    │ random.bin                                                              │
├──────────┼─────────────────────────────────────────────────────────────────────────┤
│ Digest   │ sha256:f12b89ec8f19b3f2ed7e86eb88ee0063d6ef1f9aca9f69dc7a270537bca9a45e │
└──────────┴─────────────────────────────────────────────────────────────────────────┘
➜ chainloop att push
INF push completed
┌───────────────────────────┬─────────────────────────────────────────────────────────────────────────┐
│ Initialized At            │ 22 May 25 15:37 UTC                                                     │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────┤
│ Attestation ID            │ 4df0fb42-c8d9-49cd-9e75-fa73c4eaf73c                                    │
│ Digest                    │ sha256:d83bae2cdc3bff4dc2ab7d6caad0e1ac045e41bf5a9cf88a0aafce0226d1c51b │
│ Organization              │ my-org                                                                  │
│ Name                      │ mywf                                                                    │
│ Project                   │ myproject                                                               │
│ Version                   │ v1.6.0 (prerelease)                                                     │
│ Contract                  │ myproject-mywf (revision 101)                                           │
│ Policy violation strategy │ ADVISORY                                                                │
└───────────────────────────┴─────────────────────────────────────────────────────────────────────────┘
┌────────────────────────────────────────────────────────────────────────────────────┐
│ Materials                                                                          │
├──────────┬─────────────────────────────────────────────────────────────────────────┤
│ Name     │ material-1747928434860035000                                            │
│ Type     │ ARTIFACT                                                                │
│ Set      │ Yes                                                                     │
[...]

PR #2683, opened 2026-01-22
fix(policies): prevent binary content parsing for HELM_CHART materials

### LINKED ISSUE #2684: Policy develop evaluation fails for Helm charts
When using `chainloop policy develop eval` with a local `HELM_CHART` tarball, the command fails with:

`ERR failed to parse input: invalid character '\x1f' looking for beginning of value`
 
Character `\x1f` is the first byte of gzip-compressed data.

Piskoo:
When using inline CAS storage, `HELM_CHART` materials containing binary data were incorrectly passed to JSON decoder, causing errors during policy evaluation.

Closes #2684

### DIFF
--- a/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
+++ b/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
@@ -96,9 +96,14 @@ func (m *Attestation_Material) GetEvaluableContent(value string) ([]byte, error)
 	if artifact != nil {
 		if m.InlineCas && m.MaterialType != v1.CraftingSchema_Material_HELM_CHART {
 			rawMaterial = artifact.GetContent()
-		} else if value == "" {
+		}
+
+		if !m.InlineCas && value == "" {
 			return nil, errors.New("artifact path required")
-		} else if m.MaterialType != v1.CraftingSchema_Material_HELM_CHART &&
+		}
+
+		if !m.InlineCas &&
+			m.MaterialType != v1.CraftingSchema_Material_HELM_CHART &&
 			m.MaterialType != v1.CraftingSchema_Material_JUNIT_XML {
 			// read content from local filesystem (except for tgz charts)
 			rawMaterial, err = os.ReadFile(value)


migmartri:
Thank you for the PR. Do you mind creating an issue explaining the problem and when it happens? Because in Helm charts we support both OCI and tarballs. Does this happen with other artifacts as well?

Piskoo:
> Do you mind creating an issue explaining the problem and when it happens? Because in Helm charts we support both OCI and tarballs. Does this happen with other artifacts as well?

`OCI HELM_CHART` is a material of `ContainerImage` type, where tarball is handled as `Artifact`. For `ContainerImage` materials we don't store content anywhere, we only use the metadata so it's not affected. Added the issue.

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -94,11 +94,16 @@ func (m *Attestation_Material) GetEvaluableContent(value string) ([]byte, error)
 	}
 
 	if artifact != nil {
-		if m.InlineCas {
+		if m.InlineCas && m.MaterialType != v1.CraftingSchema_Material_HELM_CHART {
```
migmartri, line 97: so this means taht the helm chart no longer get stored?

Piskoo: It is, I just tested it, which leads to another bug. Helm charts stored inline cannot be properly marshaled, created a task https://github.com/chainloop-dev/chainloop/issues/2685

migmartri:
I don't understand how an issue in the policy engine requires changing if/when to store an artifact?

Piskoo:
> I don't understand how an issue in the policy engine requires changing if/when to store an artifact?

We don't change WHEN it is stored, we change WHAT is being passed for the evaluation. For evaluations we accept only json content, for xml we marshal it into json first for processing. The same thing was being done for helm charts, we took binary content and tried to marshal into json before passing to policy engine, hence the issue. For non inline backends it was already excluded, with this PR it is being also excluded for the inline backend.

migmartri:
> > I don't understand how an issue in the policy engine requires changing if/when to store an artifact?
> 
> We don't change WHEN it is stored, we change WHAT is being passed for the evaluation. For evaluations we accept only json content, for xml we marshal it into json first for processing. The same thing was being done for helm charts, we took binary content and tried to marshal into json before passing to policy engine, hence the issue. For non inline backends it was already excluded, with this PR it is being also excluded for the inline backend.

Ah, I missunderstood, your response, sorry. 

So what happens if you have a `evidence` or `artifact` policy type?

Piskoo:
> So what happens if you have a `evidence` or `artifact` policy type?

It will fail in the same place, where we would try to marshal binary into json content. Kind detection works correctly for helm charts, so that will only happen if the user tries to evaluate helm chart that was explicitly passed as `artifact` or `evidence`

javirln:
> We don't change WHEN it is stored, we change WHAT is being passed for the evaluation. For evaluations we accept only json content, for xml we marshal it into json first for processing. 

Hmm have a look at the WASM engine because that engine allows to pass any format, including raw bytes to the policy.

migmartri:
I am closing it because I think this is a more wirespread problem that needs to be solved https://github.com/chainloop-dev/chainloop/issues/2065

PR #2692, opened 2026-01-28
chore(chart): disable Vault Kubernetes Injector

philiposs:
Disabling vault injector in development as we are not using it and it can cause problems because of mutatingwebhook resource.
In case that injector has problems communicating with vault it can stop all new pods (in whole cluster) from being deployed.
By default its better to disable it and if someone want to use it then he has to explicitly enable it.


migmartri (approved):
thanks for your contribution, please make sure to signoff your commits, thanks

PR #2693, opened 2026-01-29
feat(materials): add gitleaks json material type support  

Piskoo:
## Summary
This PR adds a new `GITLEAKS_JSON` material type. JSON report returned by `gitleaks` is an array of objects, for validation we unmarshal each finding and then check for `fingerprint` field

Closes https://github.com/chainloop-dev/chainloop/issues/2694

### DIFF
--- a/pkg/attestation/crafter/materials/gitleaks_test.go
+++ b/pkg/attestation/crafter/materials/gitleaks_test.go
@@ -88,14 +88,14 @@ func TestGitleaksReportCrafter_Craft(t *testing.T) {
 			name:     "clean report (no secrets)",
 			filePath: "./testdata/gitleaks-report-clean.json",
 			annotations: map[string]string{
-				"chainloop.material.tool.name": "Gitleaks",
+				"chainloop.material.tool.name": "gitleaks",
 			},
 		},
 		{
 			name:     "report with violations",
 			filePath: "./testdata/gitleaks-report-violations.json",
 			annotations: map[string]string{
-				"chainloop.material.tool.name": "Gitleaks",
+				"chainloop.material.tool.name": "gitleaks",
 			},
 		},
 	}


pkg/attestation/crafter/materials/gitleaks.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (i *GitleaksReportCrafter) Craft(ctx context.Context, filePath string) (*api.Attestation_Material, error) {
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("can't open the file: %w", err)
+	}
+
+	var findings []GitleaksFinding
+	if err = json.Unmarshal(data, &findings); err != nil {
```
migmartri, line 52: what kind of validation are we making here to make sure it's a gitleaks report? 

Do we check the tool? Does the json output have any openAPI spec published? I believe we make spec validations in some of those materials.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -147,6 +147,8 @@ message CraftingSchema {
       CHAINLOOP_RUNNER_CONTEXT = 25;
       // Pull Request / Merge Request metadata collected automatically during attestation
       CHAINLOOP_PR_INFO = 26;
+      // Gitleaks report json https://github.com/gitleaks/gitleaks/blob/master/README.md#reporting
+      GITLEAKS_REPORT = 27;
```
migmartri: I don't think we call REPORT to any other material, to be consistent we could do [TOOL]_[FORMAT, i.e GITLEAKS_JSON

Piskoo: We have `GITLAB_SECURITY_REPORT` but also `GHAS_SECRET_SCAN` so I went with `REPORT` as that's how it's also called in gitleaks repo

migmartri: That makes sense, it might make sense to indicate JSON though, since it has different kinds of reports?

pkg/attestation/crafter/materials/gitleaks.go
Piskoo: We can't check the tool, the report is returned as an json array `[{},{}]` there's no tool information attached. There's no offical schema published. I decided to do validation similar to `GHAS_SECRET_SCAN` where we unmarshal and then check for a required field

pkg/attestation/crafter/materials/gitleaks.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, fmt.Errorf("can't open the file: %w", err)
+	}
+
+	if err = json.Unmarshal(data, &findings); err != nil {
+		return nil, fmt.Errorf("invalid gitleaks report file: %w", ErrInvalidMaterialType)
+	}
+
+	// if list is empty. It's ambiguous, but we accept it
+	if len(findings) == 0 {
+		i.logger.Debug().Msg("Accepting an empty report. Make sure it's a valid Gitleaks report")
```
javirln: I would personally remove the part `Make sure it's a valid Gitleaks report`.

pkg/attestation/crafter/materials/gitleaks.go
```diff
@@ -0,0 +1,86 @@
...
+	return m, nil
+}
+
+func (i *GitleaksReportCrafter) injectAnnotations(m *api.Attestation_Material) {
+	// Gitleaks doesn't include version information in the JSON output
+	// Set the tool name annotation
+	if m.Annotations == nil {
+		m.Annotations = make(map[string]string)
+	}
+	m.Annotations[AnnotationToolNameKey] = "Gitleaks"
```
javirln: So it's similar to the tool reported on SARIF, I would write it like: https://github.com/gitleaks/gitleaks/blob/ca20267a84aa1fa2c2a9c1a13cdb50cafb48eeb0/report/constants.go#L4

PR #2703, opened 2026-02-04
feat(rbac): allow creating orgs to tokens with proper permissions

jiparis:
This PR adds the ability for tokens to create organizations, if they have the "organization:create" policy associated.

I've removed the ad hoc permissions check that this endpoint used to have, and replaced with policy checks, both for users and API tokens.

See some execution logs:

With `restrict_org_creation: false`:

regular user (Instance Admin): allowed 🟢 
```
✗ cldev org create --name org16
INF Organization "org16" created!
```

regular user (Org Viewer in another org): allowed 🟢 
```
cldev org create --name org17
INF Organization "org17" created!
```

regular API token in a different org: not allowed 🔴 
```
cldev org create --name org18
ERR not allowed
exit status 1
```

**Instance Admin API token**: allowed 🟢 
```
 cldev org create --name org19
INF Organization "org19" created!
```

With `restrict_org_creation: true`:

regular user (Instance admin): allowed 🟢 
```
cldev org create --name org22
INF Organization "org22" created!
```

regular user (Org viewer in different org): not allowed 🔴 
```
cldev org create --name org23
ERR not allowed
```

regular API token from existing org: not allowed 🔴 
```
cldev org create --name org21
ERR not allowed
exit status 1
```

**Instance admin API token**: allowed 🟢 
```
cldev org create --name org20
INF Organization "org20" created!
```

Closes https://github.com/chainloop-dev/chainloop/issues/2704

### DIFF
--- a/app/controlplane/internal/service/organization.go
+++ b/app/controlplane/internal/service/organization.go
@@ -215,7 +215,7 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
 
 func (s *OrganizationService) canCreateOrganization(ctx context.Context) (bool, error) {
 	// if org creation restriction is disabled, allow creation to all users
-	if !s.authz.RestrictOrgCreation && entities.CurrentMembership(ctx) != nil {
+	if !s.authz.RestrictOrgCreation && entities.CurrentUser(ctx) != nil {
 		return true, nil
 	}
 


app/controlplane/internal/service/organization.go
```diff
@@ -211,25 +214,18 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
 }
 
 func (s *OrganizationService) canCreateOrganization(ctx context.Context) (bool, error) {
-	// Restricted org creation is disabled, allow creation
+	// Restricted org creation is disabled, allow creation only to users
```
migmartri: I don't understand this new logic, can you elaborate?

app/controlplane/internal/service/organization.go
```diff
@@ -211,25 +214,18 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
...
-		}
-		if pass {
+		if entities.CurrentMembership(ctx) != nil {
 			return true, nil
 		}
+		return false, nil
+	}
+
+	// otherwise, check for permissions
+	if err := s.checkPolicy(ctx, authz.PolicyOrganizationCreate); err != nil {
```
migmartri, line 223: why is this only checked if org creation is restricted? Does it mean that if it's not restricted we can't create orgs with tokens?

app/controlplane/internal/service/service.go
```diff
@@ -329,6 +329,41 @@ func (s *service) visibleProjects(ctx context.Context) []uuid.UUID {
...
+// checkPolicy Checks a policy against a user or a token
+func (s *service) checkPolicy(ctx context.Context, policy *authz.Policy) error {
+	_, token, err := requireCurrentUserOrAPIToken(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Token case
+	if token != nil {
+		for _, p := range token.Policies {
```
migmartri: we don't have any logic that does this already?

migmartri (commented):
Added some comments, I'd like to get some understanding on what's the expected behavior of the org create with both token/users and restrictedCreation/non-restricted

migmartri (changes requested):
Please double check my comments

app/controlplane/internal/service/organization.go
jiparis: I'll add some call examples in the description.

But basically, `RestrictOrgCreation` setting is limited to users (this is for backwards compatiblity).

Then, in any other condition (for both users and tokens), the enforcer or token policies are used.

app/controlplane/internal/service/organization.go
jiparis: Yes, do you think tokens should be also enabled when there is no restriction?

app/controlplane/internal/service/service.go
jiparis: I've just seen it's also in the AuthzUseCase, which is available in this context. Let me refactor it.

app/controlplane/internal/service/organization.go
jiparis: I've just changed the behaviour to only bypass the check if it's a user and the restriction is not set.

app/controlplane/internal/service/organization.go
```diff
@@ -211,25 +214,15 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
 }
 
 func (s *OrganizationService) canCreateOrganization(ctx context.Context) (bool, error) {
-	// Restricted org creation is disabled, allow creation
-	if !s.authz.RestrictOrgCreation {
+	// if org creation restriction is disabled, allow creation to all users
+	if !s.authz.RestrictOrgCreation && entities.CurrentMembership(ctx) != nil {
```
migmartri: but wouldn't `entities.CurrentMembership(ctx)` return true also for API tokens? Does it mean that we are allowing any API token to create orgs?

jiparis: No, `currentMembership` is `nil` for API tokens. I agree it's not consistent and we should change it at some point.

jiparis: I can change it to `entities.CurrentUser(ctx)` to be more specific

jiparis: Done

jiparis:
> Added some comments, I'd like to get some understanding on what's the expected behavior of the org create with both token/users and restrictedCreation/non-restricted

I've updated the description with some tests with and without restrict_org_description setting

app/controlplane/internal/service/organization.go
```diff
@@ -211,25 +214,15 @@ func (s *OrganizationService) UpdateMembership(ctx context.Context, req *pb.Orga
 }
 
 func (s *OrganizationService) canCreateOrganization(ctx context.Context) (bool, error) {
-	// Restricted org creation is disabled, allow creation
-	if !s.authz.RestrictOrgCreation {
+	// if org creation restriction is disabled, allow creation to all users
+	if !s.authz.RestrictOrgCreation && entities.CurrentUser(ctx) != nil {
```
migmartri, line 218: yeah, more clear now, thanks

migmartri (approved):
thanks

PR #2721, opened 2026-02-10
feat(cli): support inline CA certificates in gRPC connections and config

javirln:
Store CA certificate content (base64-encoded) in config instead of file paths, enabling portable configurations across environments. The gRPC connection layer now accepts CA content directly via WithCAContent option.

PFM-4423

### DIFF
--- a/app/cli/cmd/root.go
+++ b/app/cli/cmd/root.go
@@ -18,7 +18,6 @@ package cmd
 import (
 	"context"
 	"crypto/sha256"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
@@ -502,23 +501,22 @@ func getConfigDir(appName string) string {
 	return filepath.Join(xdg.ConfigHome, appName)
 }
 
-// processCAFlag reads CA file content and encodes it to base64 if value is a file path
+// processCAFlag reads CA file content and stores it as PEM if value is a file path
 func processCAFlag(opt *confOpt) error {
 	value := viper.GetString(opt.viperKey)
 	if value == "" {
 		return nil
 	}
 
-	// If it's a file path, read and encode
+	// If it's a file path, read and store PEM content directly
 	if _, err := os.Stat(value); err == nil {
 		content, err := os.ReadFile(value)
 		if err != nil {
 			return fmt.Errorf("failed to read CA file %s: %w", value, err)
 		}
 
-		// Store base64-encoded content in viper (will be persisted by config save)
-		encoded := base64.StdEncoding.EncodeToString(content)
-		viper.Set(opt.viperKey, encoded)
+		// Store PEM content directly
+		viper.Set(opt.viperKey, string(content))
 	}
 
 	return nil


javirln:
> @javirln we need to make sure new clients work with old configuration and the other way around.
> 
> In other words, new clients should be compatible with the deprecated path loading

Yes, it is. I've added tests for that. It will attempt to load the path, and if successful, store it as base64. Otherwise, It will try to load the base64 directly.

migmartri:
> > @javirln we need to make sure new clients work with old configuration and the other way around.
> > In other words, new clients should be compatible with the deprecated path loading
> 
> Yes, it is. I've added tests for that. It will attempt to load the path, and if successful, store it as base64. Otherwise, It will try to load the base64 directly.

I don't mean the storing, but the consumption of an already stored path, just to confirm, is that covered?

javirln:
> > > @javirln we need to make sure new clients work with old configuration and the other way around.
> > > In other words, new clients should be compatible with the deprecated path loading
> > 
> > 
> > Yes, it is. I've added tests for that. It will attempt to load the path, and if successful, store it as base64. Otherwise, It will try to load the base64 directly.
> 
> I don't mean the storing, but the consumption of an already stored path, just to confirm, is that covered?

Yes, it's backwards compatible with CAs as paths stored in the configuration from all the tests I could ran.

app/cli/cmd/root.go
```diff
@@ -495,3 +501,25 @@ func isAPITokenPreferred(cmd *cobra.Command) bool {
...
+func processCAFlag(opt *confOpt) {
+	value := viper.GetString(opt.viperKey)
+	if value == "" {
+		return
+	}
+
+	// If it's a file path, read and encode
+	if grpcconn.IsFilePath(value) {
+		content, err := os.ReadFile(value)
+		if err != nil {
```
migmartri, line 514: please don't swallow errors, if the path is not found this should fail

pkg/grpcconn/grpcconn.go
```diff
@@ -140,3 +177,30 @@ func (t tokenAuth) GetRequestMetadata(_ context.Context, _ ...string) (map[strin
 func (t tokenAuth) RequireTransportSecurity() bool {
 	return !t.insecure
 }
+
+// IsFilePath checks if a value looks like a file path (vs base64/PEM content).
+// It returns true if the value appears to be a file path, false if it appears to be content.
+func IsFilePath(value string) bool {
```
migmartri: I feel we are reinventing something here? Don't we have access to ioutils and other packages that offer this?

app/cli/cmd/root.go
```diff
@@ -495,3 +501,25 @@ func isAPITokenPreferred(cmd *cobra.Command) bool {
...
+
+	// If it's a file path, read and encode
+	if _, err := os.Stat(value); err == nil {
+		content, err := os.ReadFile(value)
+		if err != nil {
+			return fmt.Errorf("failed to read CA file %s: %w", value, err)
+		}
+
+		// Store base64-encoded content in viper (will be persisted by config save)
+		encoded := base64.StdEncoding.EncodeToString(content)
```
jiparis: Note that the certificate is already PEM-encoded (base64), there is no need to encode in base64 again (which will make it grow). Chainloop only accepts PEM encoded certificates. https://github.com/chainloop-dev/chainloop/blob/61cac886433584e0c1e465c3b4629bd10a366e12/pkg/grpcconn/grpcconn.go#L112

app/cli/cmd/root.go
```diff
@@ -123,8 +124,13 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 				grpcconn.WithInsecure(apiInsecure()),
 			}
 
-			if caFilePath := viper.GetString(confOptions.controlplaneCA.viperKey); caFilePath != "" {
-				opts = append(opts, grpcconn.WithCAFile(caFilePath))
+			if caValue := viper.GetString(confOptions.controlplaneCA.viperKey); caValue != "" {
+				// Check if the value is a file path, if it is we read the content and encode it to base64, if not we assume it's the content already
```
jiparis, line 127: Should it use different keys instead of trying to do a `os.Stat`? Not sure about the security consequences of this.

javirln: If we use different keys, which would take precedence then?

javirln: As far as I remember os.Stat is safe to use because it does not open the file descriptor. It calls directly the underlying OS to read just the file attributes

app/cli/cmd/config_save.go
```diff
@@ -29,6 +29,13 @@ func newConfigSaveCmd() *cobra.Command {
 			skipActionOptsInit: trueString,
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Process CA flags - read file contents and encode to base64 if needed
+			if err := processCAFlag(confOptions.controlplaneCA); err != nil {
```
migmartri, line 33: btw, how's is this change gonna affect the platform CLI?

javirln: As we do in other commands being inherited from OSS, we will extend the `config save` on EE so it handles Platforms' CAs. The changes are in draft already there :)

PR #2727, opened 2026-02-10
chore(ci): Replace contract sync workflow with direct CLI usage

migmartri:
Replace the reusable workflow reference with inline steps that install the Chainloop CLI and apply contracts directly. Standardize contract file extensions to .yaml for consistency.

### DIFF
--- a/.github/workflows/sync_contracts.yml
+++ b/.github/workflows/sync_contracts.yml
@@ -12,7 +12,7 @@ permissions: read-all
 
 jobs:
   chainloop_contract_sync:
-    name: "[chainloop] Chainloop Contract Sync"
+    name: "Chainloop Contract Sync"
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_contracts.yml
```diff
@@ -16,7 +12,18 @@ permissions: read-all
 
 jobs:
   chainloop_contract_sync:
-    name: Chainloop Contract Sync
-    uses: chainloop-dev/labs/.github/workflows/chainloop_contract_sync.yml@5344787a72876bd22e199b94c09a86532521b4b3
-    secrets:
-      api_token: ${{ secrets.CHAINLOOP_TOKEN }}
+    name: "[chainloop] Chainloop Contract Sync"
```
Piskoo: Was it renamed on purpose?

migmartri: Good catch, renamed, thanks

PR #2734, opened 2026-02-11
feat: add auto-revoke mechanism for stale API tokens

### LINKED ISSUE #2742: Auto-revoke stale API tokens based on organization inactivity threshold
Organizations should be able to configure an inactivity threshold (in days) for API tokens. Tokens that have not been used within the configured period should be automatically revoked by a background process.

migmartri:
Adds an organization-level configurable inactivity threshold for automatic API token revocation. A background job periodically sweeps for inactive tokens and revokes them based on each organization's configured threshold.

Closes #2742

### DIFF
--- a/app/cli/documentation/cli-reference.mdx
+++ b/app/cli/documentation/cli-reference.mdx
@@ -2775,6 +2775,7 @@ chainloop organization update [flags]
 Options
 
 ```
+--api-token-max-days-inactive string   maximum days of inactivity before API tokens are auto-revoked (e.g. '90', '0' to disable)
 --block                                set the default policy violation blocking strategy
 -h, --help                                 help for update
 --name string                          organization name
--- a/app/controlplane/api/controlplane/v1/organization.pb.go
+++ b/app/controlplane/api/controlplane/v1/organization.pb.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024-2025 The Chainloop Authors.
+// Copyright 2024-2026 The Chainloop Authors.
 //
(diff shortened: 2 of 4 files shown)


app/controlplane/pkg/data/ent/schema/organization.go
```diff
@@ -55,6 +55,9 @@ func (Organization) Fields() []ent.Field {
 		field.Bool("prevent_implicit_workflow_creation").Default(false),
 		// restrict_contract_creation_to_org_admins restricts contract creation (org-level and project-level) to only organization admins
 		field.Bool("restrict_contract_creation_to_org_admins").Default(false),
+		// api_token_inactivity_threshold_days is the number of days after which inactive API tokens are auto-revoked.
+		// nil = disabled, any positive int = threshold in days
+		field.Int("api_token_inactivity_threshold_days").Optional().Nillable().Default(30),
```
jiparis: If it has a default, it's forcing the setting. Is that the desired effect?

migmartri: agreed, changed to not to have it by default and 90 days (not 30)

app/cli/pkg/action/membership_list.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2024 The Chainloop Authors.
+// Copyright 2024-2025 The Chainloop Authors.
```
migmartri: 2026

jiparis (approved):

PR #2748, opened 2026-02-17
fix(contracts): allow org tokens to create contracts

jiparis:
In https://github.com/chainloop-dev/chainloop/pull/2602, a new setting was added to restrict contract creation to org admins, but tokens were not considered. This PR fixes that.


migmartri:
But I think that's what we wanted to prevent no? We don't want contracts for being created during attesation process, does this mean that now we can?

migmartri (approved):

PR #2762, opened 2026-02-19
fix(cli): skip policy evaluation during attestation init status display

### LINKED ISSUE #2761: feat: skip policy evaluation when status is called from attestation init
## Problem

When running `attestation init`, the CLI automatically calls `attestation status` afterward to display the initial state. The status action evaluates attestation-level policies against an empty attestation statement. This stops being an optimization now that we are doing API calls using built-in functions, in which case we were doing them at `init` and at push.

## Proposed solution

Pass `WithSkipPolicyEvaluation()` to the `statusAction.Run()` call inside the init command. The mechanism already exists in the status action, it's just not being used from init.

migmartri:
Both `attestation init` and `attestation push` call `AttestationStatus.Run()` internally for display purposes. Previously, each of these calls triggered a full policy re-evaluation which was either unnecessary (init evaluating against an empty attestation) or redundant (push evaluating twice).

This change passes `WithSkipPolicyEvaluation()` to these internal status calls so they skip re-evaluation. Existing policy evaluations stored in the crafting state are always included in the result regardless of this flag. In the push flow, the status result is explicitly populated with policy evaluations after the push performs its own evaluation.

Fixes #2761

### DIFF
--- a/app/cli/pkg/action/attestation_status.go
+++ b/app/cli/pkg/action/attestation_status.go
@@ -93,6 +93,9 @@ func NewAttestationStatus(cfg *AttestationStatusOpts) (*AttestationStatus, error
 	}, nil
 }
 
+// WithSkipPolicyEvaluation skips re-evaluating policies against the attestation statement.
+// Existing policy evaluations already stored in the crafting state (e.g. from material additions)
+// will still be included in the result.
 func WithSkipPolicyEvaluation() func(*AttestationStatus) {
 	return func(opts *AttestationStatus) {
 		opts.skipPolicyEvaluation = true


app/cli/pkg/action/attestation_push.go
```diff
@@ -100,7 +100,7 @@ func (action *AttestationPush) Run(ctx context.Context, attestationID string, ru
 	if err != nil {
 		return nil, fmt.Errorf("creating status action: %w", err)
 	}
-	attestationStatus, err := statusAction.Run(ctx, attestationID)
+	attestationStatus, err := statusAction.Run(ctx, attestationID, WithSkipPolicyEvaluation())
```
jiparis, line 103: Looking at the code, it might not make sense to call StatusAction at this point, but doing it after it has been rendered and all evaluations have been done.

app/cli/cmd/attestation_init.go
```diff
@@ -127,7 +127,7 @@ func newAttestationInitCmd() *cobra.Command {
 				return newGracefulError(err)
 			}
 
-			res, err := statusAction.Run(cmd.Context(), attestationID)
+			res, err := statusAction.Run(cmd.Context(), attestationID, action.WithSkipPolicyEvaluation())
```
jiparis, line 130: Here we may be losing relevant evaluations (git commit signature, crafter checks, gather info, PR metadata, etc)

migmartri: ok, I am going to stop this PR and instead work on a lifecycle management per-policy.

Thanks

migmartri:
replaced by https://github.com/chainloop-dev/chainloop/pull/2765

PR #2765, opened 2026-02-20
feat(policies): add attestation phase lifecycle control

### LINKED ISSUE #2764: Add attestation phase lifecycle control to policy evaluation
## Summary

Attestation-level policies currently evaluate at both `status` and `push` phases unconditionally. This adds a mechanism for policy authors to control at which attestation phase(s) their policy runs via a new `attestation_phases` field on `PolicySpecV2` (the per-kind policy definition).

Default behavior is preserved (all phases) for backwards compatibility — when no phases are specified, the policy runs at all phases as it does today.

## Motivation

Some policies only make sense at specific points in the attestation lifecycle:
- A final approval check should only run at `push` time, not during `status`
- A quick lint/feedback policy might only need to run at `status` time for early feedback

Currently there is no way to express this — all attestation policies always run at both phases.

## Proposed Changes

- Add an `AttestationPhase` enum (`INIT`, `STATUS`, `PUSH`) to the crafting schema proto
- Add `repeated AttestationPhase attestation_phases` field to `PolicySpecV2` (alongside `kind` and source)
- Filter policies during evaluation based on the current phase
- Pass the current phase from the `status` and `push` CLI actions through to the policy verifiers

## Example

```yaml
kind: Policy
metadata:
  name: final-approval-check
spec:
  policies:
    - kind: ATTESTATION
      path: approval-check.rego
      attestation_phases:
        - PUSH
```

This policy would only be evaluated during `attestation push`, not during `attestation status`.

migmartri:
supersedes https://github.com/chainloop-dev/chainloop/pull/2762

## Summary

- Add `AttestationPhase` enum (`INIT`, `PUSH`) to the crafting schema proto to control when attestation policies are evaluated
- Add `attestation_phases` field on `PolicySpecV2` so policy authors can restrict evaluation to specific lifecycle phases
- When no phases are specified, the policy runs at all phases (backwards compatible)
- Policy evaluation is decoupled from the `attestation status` command — status is now a read-only observer that displays existing evaluations from crafting state
- `attestation init` evaluates policies explicitly at `INIT` phase, `attestation push` evaluates at `PUSH` phase
- Cross-phase results are preserved: PUSH-phase evaluation overwrites INIT results for the same policy, while INIT-only policies are kept intact

Closes #2764

## Example

```yaml
kind: Policy
metadata:
  name: final-approval-check
spec:
  policies:
    - kind: ATTESTATION
      path: approval-check.rego
      attestation_phases:
        - PUSH
```


app/cli/pkg/action/attestation_status.go
```diff
@@ -141,27 +126,8 @@ func (action *AttestationStatus) Run(ctx context.Context, attestationID string,
 		TimestampAuthority:          att.GetSigningOptions().GetTimestampAuthorityUrl(),
 	}
 
-	if !action.skipPolicyEvaluation {
```
jiparis, line 144: So in Status we don't evaluate policies anymore? makes sense.

migmartri: Yeah, I don't think we need it tbh.

Issue #2769, opened 2026-02-23
feat(policies): allow disabling gating restrictions for specific policies


migmartri:
## Problem

Currently, policy gating (control gates) can be configured at two levels:

1. **Organization-wide**: Enable enforcement globally via `chainloop org update --name <org> --block`
2. **Per-policy opt-in**: Enable gating on individual policies in a contract with `gate: true`

```yaml
policies:
  attestation:
    - ref: my-policy
      gate: true  # opt-in to gating
```

However, there is **no way to disable gating for a specific policy** when organization-wide enforcement is enabled. If you turn on `--block` at the org level, all policy violations become blocking — you cannot exempt individual policies.

## Proposed Solution

Allow `gate: false` in contract policy definitions to explicitly disable gating for specific policies, even when organization-wide enforcement is active.

```yaml
policies:
  attestation:
    - ref: critical-security-policy
      # inherits org-wide gating (blocking)
    - ref: informational-lint-policy
      gate: false  # explicitly non-blocking, overrides org default
```

This would enable a "default-on, opt-out per policy" model — the inverse of the current "default-off, opt-in per policy" approach.

## Use Case

Organizations that want strict enforcement by default but need to:
- Roll out new experimental policies in a non-blocking/advisory mode
- Keep informational or low-severity policies as warnings without blocking pipelines
- Gradually migrate policies from advisory to enforced without toggling the org-wide setting

## References

- [Control Gates documentation](https://docs.chainloop.dev/concepts/control-gates#configuring-enforcement)

PR #2777, opened 2026-02-24
feat(policies): allow gate=false to override org-wide blocking

matiasinsaurralde:
PR for #2769

This change implements per-policy gate override behavior for policy enforcement.
- Makes contract gate presence-aware so unset and explicit false are distinct.
- Adds effective gate resolution in policy evaluation:
  - unset gate inherits organization default strategy
  - gate: true enforces blocking for that policy
  - gate: false forces non-blocking for that policy
- Updates attestation push enforcement to rely on effective gated violations, with a safe fallback when policy evaluations are unavailable.
- Adds test coverage for gate inheritance and explicit override behavior.

### DIFF
--- a/app/cli/cmd/attestation_push.go
+++ b/app/cli/cmd/attestation_push.go
@@ -163,8 +163,7 @@ func newAttestationPushCmd() *cobra.Command {
 const exceptionFlagName = "exception-bypass-policy-check"
 
 var (
-	ErrBlockedByPolicyViolation = fmt.Errorf("the operator requires all policies to pass before continuing, please fix them and try again or temporarily bypass the policy check using --%s", exceptionFlagName)
-	exceptionBypassPolicyCheck  = fmt.Sprintf("Attention: You have opted to bypass the policy enforcement check and an operator has been notified of this exception.\nPlease make sure you are back on track with the policy evaluations and remove the --%s as soon as possible.", exceptionFlagName)
+	exceptionBypassPolicyCheck = fmt.Sprintf("Attention: You have opted to bypass the policy enforcement check and an operator has been notified of this exception.\nPlease make sure you are back on track with the policy evaluations and remove the --%s as soon as possible.", exceptionFlagName)
 )
 
 type GateError struct {
@@ -202,12 +201,6 @@ func validatePolicyEnforcement(status *action.AttestationStatusResult, bypassPol
 			return nil
 		}
 
-		// Effective gate semantics are already resolved in policy evaluations.
-		// For backwards compatibility, fall back to aggregate status only if
-		// no evaluations are available.
-		if len(status.PolicyEvaluations) == 0 && status.HasPolicyViolations {
-			return ErrBlockedByPolicyViolation
-		}
 	}
 
 	return nil
--- a/app/cli/main.go
+++ b/app/cli/main.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023-2025 The Chainloop Authors.
+// Copyright 2023-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -100,9 +100,6 @@ func errorInfo(err error, logger zerolog.Logger) (string, int) {
 			logger.Debug().Msg("GracefulErrorExit enabled (exitCode 0). If you want to disable it set --graceful-exit=false")
 			exitCode = 0
 		}
-	case errors.Is(err, cmd.ErrBlockedByPolicyViolation):
-		// default exit code for policy violations
-		exitCode = 3
 	case errors.As(err, &gateErr):
 		// exit code for gate errors
 		exitCode = 4


matiasinsaurralde:
`buf:breaking:ignore` per-field comment annotations do not exist in buf (only `buf:lint:ignore` do). To suppress the `FIELD_SAME_CARDINALITY` violation introduced by changing `bool gate` to `optional bool gate`, the exception was added to `buf.yaml` for the `app/controlplane/api` module.

Additionally, `--exclude-imports` was added to the `buf breaking` CI step to prevent `crafting_schema.proto` from being re-evaluated as an import in other modules (e.g. `pkg/attestation/crafter/api`), which would trigger the same violation a second time outside the module where the exception is defined.

Both the `buf.yaml` exception and the `--exclude-imports` flag can be removed once this change lands in `main`, since the baseline will no longer include the old field cardinality.

In order to try the `buf breaking` locally you can do (where b46ef29bb9c471b401a1dd083fcaad95fe72f021 is latest `main`):

```
buf breaking --against "https://github.com/chainloop-dev/chainloop.git#format=git,commit=b46ef29bb9c471b401a1dd083fcaad95fe72f021" --exclude-imports
```

buf.yaml
```diff
@@ -49,6 +49,7 @@ modules:
       except:
         - EXTENSION_NO_DELETE
         - FIELD_SAME_DEFAULT
+        - FIELD_SAME_CARDINALITY
```
migmartri, line 52: I am not sure we want to disable this globally but instead you coudl add an annotation in your specific case?

app/cli/cmd/attestation_push.go
```diff
@@ -202,7 +202,10 @@ func validatePolicyEnforcement(status *action.AttestationStatusResult, bypassPol
 			return nil
 		}
 
-		if status.HasPolicyViolations {
+		// Effective gate semantics are already resolved in policy evaluations.
+		// For backwards compatibility, fall back to aggregate status only if
+		// no evaluations are available.
+		if len(status.PolicyEvaluations) == 0 && status.HasPolicyViolations {
```
migmartri: I don't understand this change, mind elaborating a bit?

pkg/policies/policies.go
```diff
@@ -336,10 +345,22 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
 		SkipReasons:  reasons,
 		Requirements: attachment.Requirements,
 		RawResults:   engineRawResultsToAPIRawResults(rawResults),
-		Gate:         attachment.GetGate(),
+		Gate:         policyAttachmentGate(attachment, pv.defaultGate),
```
migmartri, line 348: ok, so before the default gate was outside the context of policy verifier and now you are bringing both in?

migmartri (commented):
great, ptal at my comments, thanks

buf.yaml
matiasinsaurralde: Note that breaking change detection doesn't support inline comment ignores.
 - [buf:lint](https://github.com/search?q=repo%3Abufbuild%2Fbuf+%22buf%3Alint%22&type=code) references in Buf code.
 - [buf:breaking](https://github.com/search?q=repo%3Abufbuild%2Fbuf+%22buf%3Abreaking%22&type=code) references in Buf code (no occurrences).

Also exceptions are only per file/directory level.

Given that we control the gRPC clients/servers that use these messages and that the `bool` to `optional bool` doesn't imply a wire format change in PB, it should be safe to regenerate the bindings and update any other parts of the code that use this message.

Also `buf breaking` always compares to latest `main`. If this PR lands (with or without the exceptions, which we can also leave out), a subsequent PR won't raise a breaking change error - cc @jiparis

app/cli/cmd/attestation_push.go
```diff
@@ -194,17 +194,11 @@ func validatePolicyEnforcement(status *action.AttestationStatusResult, bypassPol
...
-
-		if status.HasPolicyViolations {
-			return ErrBlockedByPolicyViolation
-		}
+	// Block on any policy violation only when configured (bypass handled above).
+	// When we have policy evaluations, gate semantics are already enforced in the loop above.
+	if status.MustBlockOnPolicyViolations && bypassPolicyCheck {
+		logger.Warn().Msg(exceptionBypassPolicyCheck)
+		return nil
 	}
```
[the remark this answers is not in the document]
matiasinsaurralde: Both values (PolicyEvaluations and HasPolicyViolations) come from the same loop. In normal execution, `len(PolicyEvaluations)==0 && HasPolicyViolations==true` cannot happen.

pkg/policies/policies.go
matiasinsaurralde: Right, `PolicyEvaluation.Gate` was populated with the raw attachment value (`attachment.GetGate()`), while the org-level default (`block_on_policy_violation`) was applied later in CLI enforcement.

The **efective gate** is now resolved inside `PolicyVerifier` (`policyAttachmentGate(attachment, pv.defaultGate)`), following the 3 state behavior (to achieve default on / opt-out per policy):

- Explicit `gate: true`: blocking.
- Explicit `gate: false`: non-blocking (opt-out)
- Unset `gate`: inherits org default.

buf.yaml
jiparis: Yeah, unfortunately `buf breaking` doesn't support inline ignores as `buf lint` does. To be honest, I'm wondering how this is the first time we are facing this issue. Did we update buf recently? 
The only solution is adding this top level `except` option. The change to `optional` is safe, as it only affects to generated code. Wired bytes are exactly the same (although with different semantics from now on).

app/cli/cmd/attestation_push.go
matiasinsaurralde: This was dead code from backwards compatibility code (that we don't need for this case), I've added comments about it on the cubic thread in this same PR.

app/cli/cmd/attestation_push.go
```diff
@@ -194,17 +194,11 @@ func validatePolicyEnforcement(status *action.AttestationStatusResult, bypassPol
...
-		if bypassPolicyCheck {
-			logger.Warn().Msg(exceptionBypassPolicyCheck)
-			return nil
-		}
-
-		if status.HasPolicyViolations {
-			return ErrBlockedByPolicyViolation
-		}
+	// Block on any policy violation only when configured (bypass handled above).
+	// When we have policy evaluations, gate semantics are already enforced in the loop above.
```
jiparis: I see. Now the gating logic resides in the policy itself, where it handles the default global value accordingly.
But, what about `ErrBlockedByPolicyViolation`. Is it used anymore? If not, please remove it from main.go

pkg/attestation/crafter/crafter.go
```diff
@@ -739,7 +746,13 @@ func (c *Crafter) addMaterial(ctx context.Context, m *schemaapi.CraftingSchema_M
 	policies.LogPolicyEvaluations(policyGroupResults, c.Logger)
 
 	// Validate policies
-	pv := policies.NewPolicyVerifier(c.CraftingState.GetPolicies(), c.attClient, c.Logger, policies.WithAllowedHostnames(c.CraftingState.Attestation.PoliciesAllowedHostnames...))
+	pv := policies.NewPolicyVerifier(
+		c.CraftingState.GetPolicies(),
+		c.attClient,
+		c.Logger,
+		policies.WithAllowedHostnames(c.CraftingState.Attestation.PoliciesAllowedHostnames...),
+		policies.WithDefaultGate(c.CraftingState.Attestation.GetBlockOnPolicyViolation()),
```
jiparis, line 754: Very elegant, thanks.

app/cli/cmd/attestation_push.go
matiasinsaurralde: Removed this one and its reference in `errorInfo`.

PR #2801, opened 2026-03-02
feat(chart): show real endpoints in Helm post-install notes

### LINKED ISSUE #97: chart: improve installation output (notes.md)
Currently, once you install or upgrade the Chainloop Helm Chart you get this information

```
###########################################################################
  CONFIGURE CLI
###########################################################################

Configure the CLI to point to this instance, for example

  chainloop --insecure config save \
    --control-plane my-controlplane.acme.com:80 \
    --artifact-cas cas.acme.com:80
```

We could improve this output by actually showing you the command with the real hostname:port (my-controlplane.acme.com is a placeholder)

We could do this by checking if the services are set as load balancer or the ingresses has been configured. We can take a look at how some other Bitnami Charts do it [[1]](https://github.com/bitnami/charts/blob/main/bitnami/grafana/templates/NOTES.txt#L9) since the ingress logic has been mimicked from Bitnami Charts

ghophp:
Closes #97

## Summary

The Helm chart NOTES.txt previously showed hardcoded placeholder endpoints (`my-controlplane.acme.com:80`). It now detects the actual service configuration (Ingress, LoadBalancer, NodePort, or ClusterIP) for both the Control Plane and Artifact CAS gRPC services and shows actionable connection instructions.

- When both endpoints use Ingress, a single ready-to-use `chainloop config save` command is shown with real hostnames
- For LoadBalancer/NodePort, `kubectl` commands to discover the external IP/port are provided
- For ClusterIP (default), `kubectl port-forward` instructions are shown
- The `--insecure` flag is only included when at least one endpoint lacks TLS
- Follows the same pattern used by [Bitnami charts](https://github.com/bitnami/charts/blob/main/bitnami/grafana/templates/NOTES.txt)

## Example output

Default (ClusterIP):

    ###########################################################################
      CONFIGURE CLI
    ###########################################################################

    To configure the CLI, you need the Control Plane and Artifact CAS gRPC endpoints.

    1. The Control Plane gRPC API is not externally accessible (ClusterIP). Use port-forwarding:

         kubectl port-forward --namespace default svc/test-chainloop-controlplane-api 9000:80 &

    2. The Artifact CAS gRPC API is not externally accessible (ClusterIP). Use port-forwarding:

         kubectl port-forward --namespace default svc/test-chainloop-cas-api 9001:80 &

    3. Configure the CLI:

         chainloop --insecure config save \
           --control-plane localhost:9000 \
           --artifact-cas localhost:9001

Ingress with TLS:

    ###########################################################################
      CONFIGURE CLI
    ###########################################################################

    Configure the CLI to point to this instance:

      chainloop config save \
        --control-plane api.cp.example.com:443 \
        --artifact-cas api.cas.example.com:443

## Helm unit tests

Added snapshot tests using [helm-unittest](https://github.com/helm-unittest/helm-unittest) with `matchSnapshotRaw` covering all 6 service type combinations (ClusterIP, Ingress, Ingress+TLS, LoadBalancer, NodePort, mixed). The tests are **opt-in for now** and not part of the CI pipeline. To run them locally:

    helm plugin install https://github.com/helm-unittest/helm-unittest.git
    helm unittest deployment/chainloop

Team: worth discussing whether we want to add `helm unittest` to the CI pipeline and `helm plugin install` to `make init` so these run automatically going forward.

### DIFF
--- a/deployment/chainloop/templates/NOTES.txt
+++ b/deployment/chainloop/templates/NOTES.txt
@@ -12,11 +12,11 @@ APP VERSION: {{ .Chart.AppVersion  }}
 {{- $casGrpcURL := include "chainloop.cas.external_grpc_url" . }}
 
 {{- $cpSvcType := .Values.controlplane.serviceAPI.type }}
-{{- $cpSvcPort := coalesce .Values.controlplane.serviceAPI.port .Values.controlplane.serviceAPI.ports.http }}
+{{- $cpSvcPort := .Values.controlplane.serviceAPI.ports.http }}
 {{- $cpSvcName := printf "%s-api" (include "chainloop.controlplane.fullname" .) }}
 
 {{- $casSvcType := .Values.cas.serviceAPI.type }}
-{{- $casSvcPort := coalesce .Values.cas.serviceAPI.port .Values.cas.serviceAPI.ports.http }}
+{{- $casSvcPort := .Values.cas.serviceAPI.ports.http }}
 {{- $casSvcName := printf "%s-api" (include "chainloop.cas.fullname" .) }}
 
 {{- $namespace := include "common.names.namespace" . }}


deployment/chainloop/templates/NOTES.txt
```diff
@@ -8,11 +8,120 @@ APP VERSION: {{ .Chart.AppVersion  }}
   CONFIGURE CLI
 ###########################################################################
 
-Configure the CLI to point to this instance, for example
+{{- $cpIngressEnabled := .Values.controlplane.ingressAPI.enabled }}
```
migmartri: we might alerady have most of these templates already, if not would you mind adding them to `template` instead?

philiposs:
Hi, thanks for the contribution, i would create a new PR with implementation of unit tests as they are not covered right now (and are a big thing) and keep only `NOTES.txt` changes in this one.

deployment/chainloop/templates/NOTES.txt
philiposs: this chart has only has external https endpoints in helpers, would be good idea to add `chainloop.controlplane.external_grpc_url` and `chainloop.cas.external_grpc_url` that returns ingressAPI urls for both services

ghophp:
Hey @migmartri @Filip7656, thanks for the feedback! I've addressed both points:

- Extracted the gRPC endpoint resolution into reusable helper templates (chainloop.controlplane.external_grpc_url and chainloop.cas.external_grpc_url) in _helpers.tpl, following the same pattern as the existing HTTP helpers
- Refactored NOTES.txt to use those helpers instead of inlining the logic
- Removed the helm-unittest introduction from this PR to keep the scope focused

deployment/chainloop/templates/NOTES.txt
```diff
@@ -8,11 +8,115 @@ APP VERSION: {{ .Chart.AppVersion  }}
...
 
-Configure the CLI to point to this instance, for example
+{{- $cpGrpcURL := include "chainloop.controlplane.external_grpc_url" . }}
+{{- $casGrpcURL := include "chainloop.cas.external_grpc_url" . }}
 
-  chainloop --insecure config save \
-    --control-plane my-controlplane.acme.com:80 \
-    --artifact-cas cas.acme.com:80
+{{- $cpSvcType := .Values.controlplane.serviceAPI.type }}
+{{- $cpSvcPort := coalesce .Values.controlplane.serviceAPI.port .Values.controlplane.serviceAPI.ports.http }}
```
philiposs: .Values.controlplane.serviceAPI.port this key does not exist, it should be .ports

deployment/chainloop/templates/NOTES.txt
```diff
@@ -8,11 +8,115 @@ APP VERSION: {{ .Chart.AppVersion  }}
...
+{{- $cpSvcName := printf "%s-api" (include "chainloop.controlplane.fullname" .) }}
+
+{{- $casSvcType := .Values.cas.serviceAPI.type }}
+{{- $casSvcPort := coalesce .Values.cas.serviceAPI.port .Values.cas.serviceAPI.ports.http }}
```
philiposs: .Values.cas.serviceAPI.port same here

ghophp: Good catch @Filip7656 🙏🏼 the coalesce pattern was copied from existing templates (service-grpc.yaml, configmap.yaml) that use the same incorrect .serviceAPI.port reference. It works at runtime because coalesce silently skips nil, but the key doesn't exist in the schema. Fixed to use .serviceAPI.ports.http directly.

ghophp: https://github.com/chainloop-dev/chainloop/pull/2819 this address the problem in the Helm templates 🙋🏼‍♂️

philiposs:
LGTM.
Final last touch i would move CONFIGURE CLI section at the bottom, after an installation there is a lot of output and you can miss it
@migmartri  what do you think?

migmartri:
@philiposs makes sense to me

migmartri (approved):

PR #2802, opened 2026-03-02
feat(contracts): move attestation phases policy references

Piskoo:
The `attestation_phases` field was incorrectly placed on `PolicySpecV2` (the policy definition) instead of `PolicyAttachment` (the contract attachment). This moves it to `PolicyAttachment` so users can control at which lifecycle phase (INIT/PUSH) a policy is evaluated directly in the contract. `attestation_phases` in `PolicySpecV2` was marked as depracted and acts as a fallback logic now.

### DIFF
--- a/pkg/policies/policies_test.go
+++ b/pkg/policies/policies_test.go
@@ -1298,6 +1298,27 @@ func (s *testSuite) TestAttestationPhaseFiltering() {
 			phase:            EvalPhasePush,
 			npolicies:        0,
 		},
+		// Fallback: phases from deprecated PolicySpecV2 field
+		{
+			name:      "spec-level push-only fallback, runs at push",
+			policyRef: "file://testdata/workflow_push_only.yaml",
+			phase:     EvalPhasePush,
+			npolicies: 1,
+		},
+		{
+			name:      "spec-level push-only fallback, skipped at init",
+			policyRef: "file://testdata/workflow_push_only.yaml",
+			phase:     EvalPhaseInit,
+			npolicies: 0,
+		},
+		// Attachment phases take precedence over spec-level phases
+		{
+			name:             "attachment overrides spec-level phases",
+			policyRef:        "file://testdata/workflow_push_only.yaml",
+			attachmentPhases: []v12.AttestationPhase{v12.AttestationPhase_INIT},
+			phase:            EvalPhaseInit,
+			npolicies:        1,
+		},
 	}
 
 	for _, tc := range cases {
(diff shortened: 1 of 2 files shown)


app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -350,11 +355,6 @@ message PolicySpecV2 {
   CraftingSchema.Material.MaterialType kind = 3 [(buf.validate.field).enum = {
     not_in: [3]
   }];
-
-  // Controls at which attestation phases this policy is evaluated.
-  // Empty means evaluate at all phases (INIT and PUSH) for backwards compatibility.
-  // Only applicable when kind is ATTESTATION.
-  repeated AttestationPhase attestation_phases = 5;
```
jiparis, line 357: I don't believe this is being used anywhere. But better mark it as "deprecated" and we can remove it later on.

jiparis: Or maybe we can have both behaviours (apply the one contract, and if not present, use the defaults in the policy if they exist)

Piskoo: Added it back as fallback for now, but it's marked as depracted

migmartri:
> Wait. That this feature is meant to be in the policy itself not in the contract. I am wondering if we want to support both 

https://docs.chainloop.dev/concepts/policies#configuring-phases-in-a-policy-spec

jiparis:
> > Wait. That this feature is meant to be in the policy itself not in the contract. I am wondering if we want to support both
> 
> https://docs.chainloop.dev/concepts/policies#configuring-phases-in-a-policy-spec

Having it in the policy spec makes it less reusable. I wonder then why this change was applied https://github.com/chainloop-dev/chainloop/commit/13386c84a484d91f286e3f578325d78c24396a66

jiparis:
> Wait. That this feature is meant to be in the policy itself not in the contract. I am wondering if we want to support both

We only have one use case for the release gate, and it should be placed in the contract, since it's using a policy that's also useful in different contexts.

jiparis:
@Piskoo We're going to close this issue. The former issue was a policy being executed with wrong data. Instead of skipping the policy, we are fixing the workflow to ensure the data is consistent. 
Check https://github.com/chainloop-dev/chainloop/pull/2807 for the actual fix.

We can come back to this PR if/when we need an actual override at the contract level for policy phases.

Issue #2803, opened 2026-03-02
Add chainloop apply command


Piskoo:
Add `chainloop apply` command to detect and apply workflow contract resources from YAML files. The declarative apply workflow allows contracts to be version-controlled and applied in bulk alongside other compliance resources.

PR #2804, opened 2026-03-02
feat(cli): add apply command

Piskoo:
Adds a `chainloop apply -f <file|dir>` command that applies contracts from YAML files or directories.

All succeeded
```
$ chainloop apply -f ./contracts
INF 2 contracts applied
```

1 failed
```
$ chainloop apply -f ./contracts
ERR parsing file /home/ubuntu/Desktop/contracts/contract2.yml: missing 'metadata.name' field in YAML document of kind "Contract"
exit status 1
```

Closes https://github.com/chainloop-dev/chainloop/issues/2803

### DIFF
--- a/app/cli/cmd/apply.go
+++ b/app/cli/cmd/apply.go
@@ -16,8 +16,6 @@
 package cmd
 
 import (
-	"fmt"
-
 	"github.com/chainloop-dev/chainloop/app/cli/pkg/action"
 	"github.com/spf13/cobra"
 )
@@ -41,26 +39,7 @@ Supports multi-document YAML files. Each document must have a 'kind' field.`,
 				return err
 			}
 
-			var contracts int
-			var errors []string
-			for _, r := range results {
-				if r.Error != nil {
-					errors = append(errors, fmt.Sprintf("  %s/%s: %s", r.Kind, r.Name, r.Error))
-					continue
-				}
-				if r.Kind == action.KindContract {
-					contracts++
-				}
-			}
-
-			if len(errors) > 0 {
-				for _, e := range errors {
-					logger.Error().Msg(e)
-				}
-				return fmt.Errorf("%d of %d resources failed to apply", len(errors), len(results))
-			}
-
-			logger.Info().Msgf("%d contracts applied", contracts)
+			logger.Info().Msgf("%d contracts applied", len(results))
 
 			return nil
 		},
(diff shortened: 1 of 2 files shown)


app/cli/pkg/action/apply.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		results = append(results, result)
+	}
+
+	return results, nil
+}
+
+// ParseYAMLPath collects all YAML files from a path (file or directory),
+// reads them, and splits multi-document files into individual YAMLDoc entries.
+func ParseYAMLPath(path string) ([]*YAMLDoc, error) {
```
matiasinsaurralde, line 87: Do we want to export this and `CollectYAMLFiles`, etc.?

Piskoo: yes, it will be reused in CLI EE

matiasinsaurralde: Great!

app/controlplane/internal/service/workflowcontract.go
```diff
@@ -233,6 +233,77 @@ func (s *WorkflowContractService) Update(ctx context.Context, req *pb.WorkflowCo
...
+		return nil, handleUseCaseErr(err, s.log)
+	}
+
+	if org.RestrictContractCreationToOrgAdmins {
+		if !canCreateContractsInRestrictedMode(ctx) {
+			return nil, errors.Forbidden("forbidden", "contract creation is restricted to organization administrators and service accounts. Please contact your administrator")
+		}
+	}
+
+	schema, err := s.contractUseCase.Create(ctx, &biz.WorkflowContractCreateOpts{
```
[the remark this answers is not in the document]
Piskoo: false positive

app/cli/cmd/apply.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/chainloop-dev/chainloop/app/cli/pkg/action"
+	"github.com/spf13/cobra"
+)
+
+func newApplyCmd() *cobra.Command {
+	var filePath string
+
+	cmd := &cobra.Command{
+		Use:   "apply",
```
jiparis, line 27: Since OSS only supports declarative workflow contracts, I think this command should be placed in `chainloop wf contract apply`.

Piskoo: We already have `wf contract apply`, this one is there to expose generic apply for all resources in chainloop, it will be extended in CLI EE. We can refactor `wf contract apply` in a separate task so it uses the new endpoint, although if we do that, we will lose the capability scoping contracts per project. We don't support projects in the schema

PR #2819, opened 2026-03-05
fix(chart): use correct serviceAPI.ports.http key in templates

ghophp:
- Fix incorrect `.Values.controlplane.serviceAPI.port` and `.Values.cas.serviceAPI.port` references in Helm templates
- The `serviceAPI.port` key does not exist in the values schema — the correct path is `serviceAPI.ports.http`
- The `coalesce` fallback silently masked the issue since nil was always skipped at runtime

Spotted during review of #2801, these occurrences pre-date that PR and exist in `service-grpc.yaml` (controlplane and CAS) and `configmap.yaml`.

### DIFF
--- a/deployment/chainloop/Chart.yaml
+++ b/deployment/chainloop/Chart.yaml
@@ -7,7 +7,7 @@ description: Chainloop is an open source software supply chain control plane, a 
 
 type: application
 # Bump the patch (not minor, not major) version on each change in the Chart Source code
-version: 1.338.0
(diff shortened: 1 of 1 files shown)


philiposs:
Thanks for the PR, please bump the chart version and feel free to merge

ghophp:
@Filip7656 `At least 1 approving review is required by reviewers with write access.` cc @migmartri

migmartri (approved):

PR #2825, opened 2026-03-08
fix(cli): handle GitHub federated OIDC tokens in attestation auth parsing

matiasinsaurralde:
## Summary
This change fixes a CLI federated-auth parsing issue that affected GitHub OIDC flows.  
In the failing scenario, `att init` and `att add` succeeded, but `att push` failed because crafting state contained invalid `attestation.auth` values (`type=0`, empty `id`).

## What changed
- Added explicit GitHub federated token recognition via issuer `https://token.actions.githubusercontent.com`.
- Kept federated classification for GitLab tokens.
- Added a guard to avoid returning partially filled parsed tokens that would result in invalid crafting-state auth metadata.
- Extended token parsing tests to cover:
  - GitHub federated token parsing
  - Federated token missing issuer
  - GitHub issuer detection helper behavior

## Why
Issue analysis showed the CLI parser handled federated tokens with a GitLab-specific claim heuristic, so valid GitHub OIDC tokens could bypass proper classification and still produce a non-nil parsed token with zero-value auth fields. Those invalid auth fields were then rejected later by strict push-time crafting-state validation.

## Context
This is related to **GitHub support** for keyless/federated CLI workflows, ensuring GitHub OIDC users can complete the expected attestation flow without deferred validation errors.

### DIFF
--- a/app/cli/internal/token/token.go
+++ b/app/cli/internal/token/token.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024-2025 The Chainloop Authors.
+// Copyright 2024-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -36,6 +36,7 @@ const (
 	//nolint:gosec
 	apiTokenAudience       = "api-token-auth.chainloop"
 	federatedTokenAudience = "chainloop"
+	githubFederatedIssuer  = "https://token.actions.githubusercontent.com"
 )
 
 // Parse the token and return the type of token. At the moment in Chainloop we have 3 types of tokens:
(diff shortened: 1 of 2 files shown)


app/cli/internal/token/token.go
```diff
@@ -172,3 +186,14 @@ func isGitLabFederatedToken(claims jwt.MapClaims) bool {
 
 	return found >= requiredClaims
 }
+
+// TODO: Evaluate adding claim-structure validation for GitHub tokens, similar to
+// the GitLab heuristic above, once we stabilize the expected claim set.
+func isGitHubFederatedToken(claims jwt.MapClaims) bool {
```
jiparis, line 192: I would set this behaviour the default one for all federated issuers. Not sure what are we trying to aim for in `isGitLabFederatedToken`, I'd remove that function and substitute it by a generic `validIssuer` that gets the valid issuers from the configuration. So I would just do:
```
case federatedTokenAudience:
  pToken.TokenType = v1.Attestation_Auth_AUTH_TYPE_FEDERATED
  if issuer, ok := claims["iss"].(string); ok {
      if !validIssuer(issuer) {
        return nil, nil
      }
      pToken.ID = issuer
  } else {
      return nil, nil
  }
```

matiasinsaurralde:
This was simplified and solve in #2826 - 029e62a1200f83294ec8c9cf39eaf42a87ac88d8

We can close it.

PR #2832, opened 2026-03-10
fix(auth): improve JWT and GitHub OIDC token handling

matiasinsaurralde:
## Summary

- Expose the raw GitHub OIDC token on decoded claims, enabling downstream consumers to reuse the exact token string when needed.
- Treat empty JWT strings as invalid in the HTTP JWT middleware and surface a consistent `ErrTokenInvalid` instead of letting parsing proceed.

## Rationale

These changes tighten token handling across runners and HTTP middleware, making behavior more predictable in edge cases (empty tokens) and more flexible when integrating with external systems that require access to the original OIDC token.

### DIFF
--- a/pkg/middlewares/http/jwt.go
+++ b/pkg/middlewares/http/jwt.go
@@ -84,6 +84,10 @@ func AuthFromAuthorizationHeader(keyFunc jwt.Keyfunc, claimsFunc ClaimsFunc, sig
 func verifyAndMarshalJWT(token string, keyFunc jwt.Keyfunc, claimsFunc ClaimsFunc, signingMethod jwt.SigningMethod) (*jwt.Claims, error) {
 	var tokenInfo *jwt.Token
 
+	if token == "" {
+		return nil, jwtMiddleware.ErrTokenInvalid
+	}
+
 	tokenInfo, err := jwt.ParseWithClaims(token, claimsFunc(), keyFunc)
 	if err != nil {
 		var ve *jwt.ValidationError


app/cli/cmd/root.go
```diff
@@ -121,8 +121,11 @@ func NewRootCmd(l zerolog.Logger) *cobra.Command {
 				return err
 			}
 
-			// If the auth token is not set and the command supports federated auth, we try to discover the runner and use the federated token for the runner if available
-			if authToken == "" && cmdSupportsFederatedAuth(cmd) {
+			// If the auth token is not set (or it's an expired stored user token) and the command supports federated
+			// auth, we try to discover the runner and use the federated token if available.
+			// The isUserToken guard ensures we only fall through for implicitly loaded config tokens, not for
+			// explicit API tokens provided via flag or CHAINLOOP_TOKEN env var.
+			if (authToken == "" || (isUserToken && token.IsExpired(authToken))) && cmdSupportsFederatedAuth(cmd) {
```
migmartri: I don't understand this change, isn't it empty? Why check for expired?

pkg/attestation/crafter/runners/oidc/github.go
```diff
@@ -262,13 +262,14 @@ func (c *GitHubOIDCClient) verifyToken(ctx context.Context, audience []string, a
...
 
-func (c *GitHubOIDCClient) decodeToken(token *oidc.IDToken) (*Token, error) {
+func (c *GitHubOIDCClient) decodeToken(token *oidc.IDToken, rawToken string) (*Token, error) {
 	c.logger.Debug().Msgf("decoding token: %s", token)
 	var t Token
 	if err := token.Claims(&t); err != nil {
 		return nil, fmt.Errorf("%w: getting claims: %w", errToken, err)
 	}
 
+	t.RawToken = rawToken
```
migmartri: Ok so you are saying that we didn't setup raw token which was then sent, that makes sense

But I thought the raw token part was already here from before, maybe it never worked?

pkg/middlewares/http/jwt.go
```diff
@@ -84,6 +84,10 @@ func AuthFromAuthorizationHeader(keyFunc jwt.Keyfunc, claimsFunc ClaimsFunc, sig
 func verifyAndMarshalJWT(token string, keyFunc jwt.Keyfunc, claimsFunc ClaimsFunc, signingMethod jwt.SigningMethod) (*jwt.Claims, error) {
 	var tokenInfo *jwt.Token
 
+	if token == "" {
```
migmartri, line 87: btw I don't think this solves our problem. Our verification is actually handled somewhere else (platform middleware)

matiasinsaurralde: Yes, I think this was just a safeguard we've discussed with @jiparis to improve debugging, etc.

migmartri: But I think the error you saw yesterday was in platform middle ware not here

jiparis: When token == "", the `nbf` and `exp` claims are initialized to `time.Nil`, and the error reported is `ErrTokenExpired`.
You can test it quickly, remove your user auth info, and any `CHAINLOOP_TOKEN`, you get an expired error:
```
➜  cl org ls
ERR your authentication token has expired, please run chainloop auth login again
```

migmartri: Yes, but this is for chainloop but not for federated

migmartri: Just pointing it out that we'd need to do a similar fix in platform to prevent misleading error in federated tokens

jiparis: Right, there is a similar logic.

matiasinsaurralde: Agree, will do it!

PR #2840, opened 2026-03-13
chore(ci): migrate CodeQL workflow to GitHub keyless auth

matiasinsaurralde:
## Summary

- Migrate CodeQL GH workflow to use [GitHub keyless attestation](https://docs.chainloop.dev/guides/github-keyless).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -11,6 +11,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write # required for SLSA attestation
 
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -11,6 +11,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write # required for SLSA attestation
```
migmartri, line 14: do we need the permission here at the top. see #2841

migmartri (changes requested):

PR #2854, opened 2026-03-14
chore(ci): migrate github_release workflow to GitHub keyless auth

matiasinsaurralde:
Migrate `github_release` workflow to use [GitHub keyless attestation](https://docs.chainloop.dev/guides/github-keyless).


javirln:
This workflow is not used anymore and probably can be removed at all.

PR #2870, opened 2026-03-17
feat: wrap AI agent config in evidence envelope

jiparis:
## Summary

- Wraps `CHAINLOOP_AI_AGENT_CONFIG` material type in the standard custom schema envelope (`chainloop.material.evidence.id`, `schema`, `data`), aligning it with the pattern already used by `CHAINLOOP_PR_INFO`
- Renames the internal `Evidence` struct to `Data` and introduces a new `Evidence` wrapper with `EvidenceID` and `EvidenceSchemaURL` constants
- Updates the crafter to unwrap the envelope for schema validation while preserving raw JSON for strict field checking

Closes PFM-4976

### DIFF
--- a/internal/schemavalidators/testdata/ai_agent_config_empty_config_files.json
+++ b/internal/schemavalidators/testdata/ai_agent_config_empty_config_files.json
@@ -1,5 +1,4 @@
 {
-  "schema_version": "0.1",
   "agent": {
     "name": "claude"
   },
--- a/internal/schemavalidators/testdata/ai_agent_config_extra_fields.json
+++ b/internal/schemavalidators/testdata/ai_agent_config_extra_fields.json
@@ -1,5 +1,4 @@
 {
-  "schema_version": "0.1",
   "agent": {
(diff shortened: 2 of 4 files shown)


internal/aiagentconfig/aiagentconfig.go
```diff
@@ -53,8 +58,8 @@ type ConfigFile struct {
 	Content string         `json:"content"`
 }
 
-// Evidence is the AI agent configuration payload
-type Evidence struct {
+// Data is the AI agent configuration payload
+type Data struct {
 	SchemaVersion string       `json:"schema_version"`
```
javirln: We won't probably need this now right? We have the schema version outside?

jiparis: right. I'll remove it

javirln: Thanks Jose!

jiparis: done

PR #2871, opened 2026-03-17
fix: return early in DiscoverRunner on first match

### LINKED ISSUE #2834: Do not try all runners in runner discovery
In the runner discover function (`DiscoverRunner`), the logic loops through all registered runners, and even if many are found (which is likely impossible), it only takes the first one.

When executing with the `--debug` option, you can see how some of those runners are trying to open system files (to discover k8s configuration for Tekton, for example), or doing some API calls (Github for example). This is a waste of resources and makes the `attestation init` step slower as we add more runner implementations.

Instead, the for loop should stop after the first one is found, and return it directly.

waveywaves:
## Summary

- Stop iterating through all runner factories in `DiscoverRunner` after the first match — return it immediately instead of collecting all matches into a slice
- Avoids unnecessary `CheckEnv()` calls from remaining runners, which becomes more impactful as more runner implementations are added

Fixes #2834

### DIFF
--- a/pkg/attestation/crafter/runner.go
+++ b/pkg/attestation/crafter/runner.go
@@ -114,35 +114,18 @@ func NewRunner(t schemaapi.CraftingSchema_Runner_RunnerType, authToken string, l
 
 // DiscoverRunner the runner environment
 // This method does a simple check to see which runner is available in the environment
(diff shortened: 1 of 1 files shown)


waveywaves:
cc @jiparis

migmartri:
### Code review

PR #2874, opened 2026-03-17
feat(attestation): preserve chainloop.* annotations in NormalizedMaterial

### LINKED ISSUE #2872: Preserve chainloop.* annotations in NormalizedMaterial during normalization
## Problem

`normalizeMaterial()` in `pkg/attestation/renderer/chainloop/v02.go` (lines 436-440) strips all annotations with the `chainloop.` prefix from materials:

```go
for k, v := range mAnnotationsMap {
    if strings.HasPrefix(k, v1.AnnotationPrefix) {
        continue
    }
    m.Annotations[k] = v.GetStringValue()
}
```

System annotations like `chainloop.material.size`, `chainloop.material.cas`, `chainloop.material.type`, etc. are used to populate dedicated struct fields (`UploadedToCAS`, `Type`, `Name`, etc.) but are then discarded from the `Annotations` map.

This causes issues downstream when consumers need access to these annotations. For example, `chainloop.material.size` is needed by the frontend to perform a pre-fetch size check — without it, the content viewer downloads the full file before discovering it exceeds the size limit.

## Proposed fix

Preserve `chainloop.*` annotations in `NormalizedMaterial.Annotations` alongside the dedicated struct fields. The annotations should be kept as-is in the map rather than being stripped during normalization.

Alternatively, we could expose them through another method in the interface

migmartri:
## Summary

- `normalizeMaterial()` was stripping all `chainloop.*` prefixed annotations from `NormalizedMaterial.Annotations`, making internal system annotations (e.g. `chainloop.material.size`, `chainloop.material.cas`, `chainloop.material.tool.name`) unavailable to downstream consumers
- Removed the `continue` that skipped `chainloop.*` keys so all annotations flow through to the map
- Added a `structValueToString` helper to correctly serialize all structpb value types — necessary because some annotations like `chainloop.material.cas` are stored as `BoolValue` (for which `GetStringValue()` returns `""`)

Closes #2872

### DIFF
--- a/app/controlplane/plugins/sdk/v1/testdata/attestations/full.md
+++ b/app/controlplane/plugins/sdk/v1/testdata/attestations/full.md
@@ -18,16 +18,26 @@
 | Type | CONTAINER_IMAGE |
 | Value | index.docker.io/bitnami/nginx |
 | Digest | sha256:264f55a6ff9cec2f4742a9faacc033b29f65c04dd4480e71e23579d484288d61 |
+| Annotations | ------ |
+|  | chainloop.material.name: image |
+|  | chainloop.material.type: CONTAINER_IMAGE |
 | Name | skynet-sbom |
 | Type | SBOM_CYCLONEDX_JSON |
 | Value | sbom.cyclonedx.json |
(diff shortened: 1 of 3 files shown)


migmartri:
### Code review

PR #2895, opened 2026-03-20
feat(attestation): expand PR reviewers with requested status and review state

javirln:
Add Requested and ReviewStatus fields to track which reviewers were explicitly requested versus who actually submitted reviews. For GitHub PRs, merge requested reviewer lists from both event payload and /requested_reviewers API endpoint, then overlay review states from /reviews API. For GitLab, mark API reviewers as requested. Introduces PR info schema v1.2.

### DIFF
--- a/pkg/attestation/crafter/prmetadata.go
+++ b/pkg/attestation/crafter/prmetadata.go
@@ -121,8 +121,8 @@ func extractGitHubPRMetadata(ctx context.Context, envVars map[string]string) (bo
 
 	// GITHUB_TOKEN is read via os.Getenv to avoid persisting it in the attestation envVars map.
 	// GITHUB_API_BASE_URL can be overridden (e.g. in tests); defaults to api.github.com.
-	parts := splitOwnerRepo(envVars["GITHUB_REPOSITORY"])
-	owner, repo := parts[0], parts[1]
+	owner := envVars["GITHUB_REPOSITORY_OWNER"]
+	_, repo, _ := strings.Cut(envVars["GITHUB_REPOSITORY"], "/")
 	prNumber := fmt.Sprintf("%d", event.PullRequest.Number)
 	githubAPIBase := os.Getenv("GITHUB_API_BASE_URL")
 	if githubAPIBase == "" {
@@ -152,6 +152,9 @@ func extractGitHubPRMetadata(ctx context.Context, envVars map[string]string) (bo
 		})
 	}
 
+	// Also fetch from the API: the event payload is a snapshot at dispatch time, so reviewers
+	// added after the event fires won't appear in event.PullRequest.RequestedReviewers.
+	// The API reflects current state and may include those late additions.
 	for _, r := range fetchGitHubRequestedReviewers(ctx, githubAPIBase, owner, repo, prNumber, token) {
 		if _, exists := reviewerMap[r.Login]; !exists {
 			reviewerMap[r.Login] = len(reviewers)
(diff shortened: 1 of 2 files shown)


pkg/attestation/crafter/prmetadata.go
```diff
@@ -118,18 +119,56 @@ func extractGitHubPRMetadata(envVars map[string]string) (bool, *PRMetadata, erro
 		return false, nil, fmt.Errorf("failed to parse event JSON: %w", err)
 	}
 
+	// GITHUB_TOKEN is read via os.Getenv to avoid persisting it in the attestation envVars map.
+	// GITHUB_API_BASE_URL can be overridden (e.g. in tests); defaults to api.github.com.
+	parts := splitOwnerRepo(envVars["GITHUB_REPOSITORY"])
+	owner, repo := parts[0], parts[1]
```
jiparis: Isn't owner available as GITHUB_REPOSITORY_OWNER? That's already available in the runner's `envVars`.

pkg/attestation/crafter/prmetadata.go
```diff
@@ -118,18 +119,56 @@ func extractGitHubPRMetadata(envVars map[string]string) (bool, *PRMetadata, erro
...
 		reviewers = append(reviewers, prinfo.Reviewer{
-			Login: r.Login,
-			Type:  reviewerType,
+			Login:     r.Login,
+			Type:      reviewerType,
+			Requested: true,
 		})
 	}
 
+	for _, r := range fetchGitHubRequestedReviewers(ctx, githubAPIBase, owner, repo, prNumber, token) {
```
jiparis, line 158: Aren't these the ones the ones gathered from `event.PullRequest` a few lines above? Is there a possibility that they don't match?

pkg/attestation/crafter/prmetadata.go
```diff
@@ -146,6 +185,166 @@ func extractGitHubPRMetadata(envVars map[string]string) (bool, *PRMetadata, erro
...
+		return nil
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil
+	}
+
+	var result struct {
+		Users []struct {
```
jiparis: Why not just use the same `prinfo.Reviewer` struct? I think it has the same fields.

pkg/attestation/crafter/prmetadata.go
```diff
@@ -79,7 +80,7 @@ func DetectPRContext(ctx context.Context, runner SupportedRunner) (bool, *PRMeta
 }
 
 // extractGitHubPRMetadata reads GITHUB_EVENT_PATH JSON and extracts PR metadata
-func extractGitHubPRMetadata(envVars map[string]string) (bool, *PRMetadata, error) {
+func extractGitHubPRMetadata(ctx context.Context, envVars map[string]string) (bool, *PRMetadata, error) {
```
jiparis, line 83: I think we could move these `extractXXXPRMetadata` to the Runner interface and implement it there, don't you think?

jiparis:
Thanks @javirln. I've left some comments.

javirln:
> Thanks @javirln. I've left some comments.

Checking them Jose, thanks!

pkg/attestation/crafter/prmetadata.go
javirln: I found yesterday a case where the event didn't have all reviewers, I added an User and the API had it and since we are asking for the actual reviewers I don't it harms to do another API request

migmartri (approved):

PR #2897, opened 2026-03-20
feat: adding metric configuration and services for cas and controlplane

lemonprogis:
Hey y'all 👋  

In our deployments of Chainloop OSS, we have created a couple of services in a config chart to expose the metrics endpoints to be scraped by prometheus via a [servicemonitor](https://prometheus-operator.dev/docs/getting-started/design/#servicemonitor).

I did a couple of simplifications compared to the http and grpc services. Keeping this one ClusterIP only etc. 

It's configurable similarly to the others using the same conventions.  This allows prometheus to lookup the endpoints of the pods and scrape every pod as the replicasets scale.

### DIFF
--- a/deployment/chainloop/templates/cas/service-metrics.yaml
+++ b/deployment/chainloop/templates/cas/service-metrics.yaml
@@ -3,22 +3,22 @@ Copyright Chainloop, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- if .Values.metrics.enabled }}
+{{- if .Values.cas.metrics.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "chainloop.cas.fullname" . }}-metrics
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{ include "chainloop.cas.labels" . | nindent 4 }}
-  {{- if or .Values.metrics.service.annotations .Values.commonAnnotations }}
-  {{- $annotations := include "common.tplvalues.merge" (dict "values" (list .Values.metrics.service.annotations .Values.commonAnnotations) "context" .) }}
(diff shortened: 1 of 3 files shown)


migmartri:
### Code review

A few concerns about consistency with the rest of the chart:

**1. Shared top-level `metrics:` breaks the per-component convention**

Every other component setting lives under `controlplane.*` or `cas.*` (e.g., `controlplane.service`, `cas.serviceAPI`, `controlplane.containerPorts.metrics`). Even the dex sub-chart uses `dex.metrics.enabled`. A single global `metrics.enabled` means you can't enable metrics for one component independently. The convention would be `controlplane.metrics.*` / `cas.metrics.*`.

https://github.com/chainloop-dev/chainloop/blob/05478c1ba74333893922d35b8573d1cc478cfa79/deployment/chainloop/values.yaml#L54-L65

**2. Duplicate port default**

`metrics.service.port: 5000` introduces a new independent value, but `controlplane.containerPorts.metrics: 5000` and `cas.containerPorts.metrics: 5000` already exist. If someone changes `containerPorts.metrics`, the service port won't follow. The templates should reference the existing per-component values instead of introducing a duplicate.

**3. Dead `$podLabels` variable in controlplane template**

`$podLabels` is computed but never used — the selector uses `chainloop.controlplane.selectorLabels` instead. The CAS counterpart correctly omits this line.

https://github.com/chainloop-dev/chainloop/blob/05478c1ba74333893922d35b8573d1cc478cfa79/deployment/chainloop/templates/controlplane/service-metrics.yaml#L23-L25

-- Miguel

migmartri:
also make sure to run sign off the commits, thanks!

lemonprogis:
Thank you @migmartri for the review! Address the comment above, also adding sign-off as well. What's funny is I had it per one before but thought it made sense as a top level since it's shared :D

lemonprogis:
Issue ref: https://github.com/chainloop-dev/chainloop/issues/2900

Issue #2918, opened 2026-03-22
feat: clean up orphaned credentials on CAS backend rotation and allow custom secret name


migmartri:
## Problem

When rotating credentials on a CAS backend (via the `Update` endpoint), the current implementation always creates a **new secret** at a new UUID-based path in the secret manager (Vault, AWS Secrets Manager, etc):

```go
secretName := strings.Join([]string{m.secretPrefix, orgID, uuid.Generate().String()}, "/")
```

The old secret is never deleted — it is simply orphaned. Over time, this leads to credential sprawl in the secret manager with no way to identify or clean up stale entries.

### 1. Delete old secret on credential rotation

When `Update` is called with new credentials, after successfully saving the new secret and updating the DB reference, the old `secret_name` should be deleted from the secret manager via `DeleteCredentials`.

### 2. Allow callers to provide their own secret name / UUID

Rather than always generating a random UUID, `SaveCredentials` (or the update flow) could accept an optional caller-supplied name or UUID. This would allow:
- Updating a secret **in place** at the same path (if the secret manager supports updates, e.g. Vault's `Put` already overwrites)
- Predictable, human-readable secret paths for operational visibility
- Simpler auditing — the secret path stays stable across rotations

## Current Behavior

1. `Update()` calls `SaveCredentials()` → new UUID path created in secret manager
2. DB `secret_name` is updated to new path
3. Old secret at previous path is left behind indefinitely

## Expected Behavior

Option A: Always clean up the old secret after a successful rotation.
Option B: Allow reusing the same secret name/path across rotations (update in place).
Option C: Both — reuse path by default, with cleanup as a fallback for backends that don't support updates.

## Affected Files

- `pkg/credentials/vault/keyval.go` — `SaveCredentials` always generates new UUID
- `pkg/credentials/aws/secretmanager.go` — same pattern
- `pkg/credentials/azurekv/keyvault.go` — same pattern
- `pkg/credentials/gcp/secretmanager.go` — same pattern
- `app/controlplane/pkg/biz/casbackend.go` — `Update()` does not delete old secret after rotation

PR #2919, opened 2026-03-22
feat(credentials): add in-place credential rotation via upsert

migmartri:
## Summary

- Introduces `WithSecretName` functional option to `SaveCredentials` across all credential backends (Vault, AWS Secrets Manager, GCP Secret Manager, Azure Key Vault), enabling upsert at a known path instead of always generating a new UUID-based path
- Updates `CASBackendUseCase.Update()` to pass the existing `SecretName` when rotating credentials, so rotations overwrite in-place and stop creating orphaned secret entries
- Adds `PutSecretValue` to the AWS `SecretsManagerIface` with fallback to `CreateSecret` on not-found, and `GetSecret`-based existence check for GCP before adding a new secret version
- Updates `Writer`/`ReaderWriter` mocks to handle the variadic `opts ...SaveOption` parameter; adds table-driven upsert tests for all four backends and a `casbackend` rotation test verifying `WithSecretName` is forwarded

closes https://github.com/chainloop-dev/chainloop/issues/2918

### DIFF
--- a/pkg/credentials/aws/secretmanager.go
+++ b/pkg/credentials/aws/secretmanager.go
@@ -81,7 +81,7 @@ func NewManager(opts *NewManagerOpts) (*Manager, error) {
 	}, nil
 }
 
-// SaveCredentials saves credentials. If opts includes WithSecretName, upserts at the given path.
+// SaveCredentials saves credentials. If opts includes WithExistingSecret, upserts at the given path.
 func (m *Manager) SaveCredentials(ctx context.Context, orgID string, creds any, opts ...credentials.SaveOption) (string, error) {
 	o := credentials.ApplySaveOptions(opts...)
 	secretName := o.SecretName
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -34,7 +34,7 @@ type ReaderWriter interface {
 	Writer
 }
 
-// SaveOpts holds optional parameters for SaveCredentials. Exported for use by sub-packages.
+// SaveOpts holds optional parameters for SaveCredentials.
 type SaveOpts struct {
 	// SecretName when non-empty causes SaveCredentials to upsert at the given path
 	// instead of generating a new UUID-based path.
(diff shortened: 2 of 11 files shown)


pkg/credentials/credentials.go
```diff
@@ -34,8 +34,33 @@ type ReaderWriter interface {
 	Writer
 }
 
+// SaveOpts holds optional parameters for SaveCredentials. Exported for use by sub-packages.
+type SaveOpts struct {
+	// SecretName when non-empty causes SaveCredentials to upsert at the given path
+	// instead of generating a new UUID-based path.
+	SecretName string
```
migmartri, line 41: check if it needs to be exposed

pkg/credentials/credentials.go
```diff
@@ -34,8 +34,33 @@ type ReaderWriter interface {
...
+// An empty name is a no-op (auto-generate behavior is preserved).
+func WithSecretName(name string) SaveOption {
+	return func(o *SaveOpts) { o.SecretName = name }
+}
+
+// ApplySaveOptions applies the given options and returns the resulting SaveOpts.
+func ApplySaveOptions(opts []SaveOption) SaveOpts {
+	o := SaveOpts{}
+	for _, opt := range opts {
+		opt(&o)
```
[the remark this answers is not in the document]
migmartri: makes sense

pkg/credentials/aws/secretmanager.go
```diff
@@ -79,21 +81,43 @@ func NewManager(opts *NewManagerOpts) (*Manager, error) {
...
 }
 
-// Save Credentials, this is a generic function that can be used to save any type of credentials
-// as long as they can be passed to json.Marshal
-func (m *Manager) SaveCredentials(ctx context.Context, orgID string, creds any) (string, error) {
-	secretName := strings.Join([]string{m.secretPrefix, orgID, uuid.New().String()}, "/")
+// SaveCredentials saves credentials. If opts includes WithSecretName, upserts at the given path.
+func (m *Manager) SaveCredentials(ctx context.Context, orgID string, creds any, opts ...credentials.SaveOption) (string, error) {
+	o := credentials.ApplySaveOptions(opts...)
+	secretName := o.SecretName
```
migmartri, line 87: let's call this, explicitSecretName

migmartri: or existingSecretName

pkg/credentials/credentials.go
```diff
@@ -34,8 +34,33 @@ type ReaderWriter interface {
...
+	// instead of generating a new UUID-based path.
+	SecretName string
+}
+
+// SaveOption is a functional option for SaveCredentials.
+type SaveOption func(*SaveOpts)
+
+// WithSecretName instructs SaveCredentials to upsert at the provided path.
+// An empty name is a no-op (auto-generate behavior is preserved).
+func WithSecretName(name string) SaveOption {
```
migmartri: let's call this WithExistingSecret

PR #2933, opened 2026-03-24
fix(sast): fix codeql SAST warning

jiparis:
Add cookie settings for OAuth to fix CodeQL errors: https://github.com/chainloop-dev/chainloop/actions/runs/23467110462

### DIFF
--- a/app/controlplane/internal/service/auth.go
+++ b/app/controlplane/internal/service/auth.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024-2026 The Chainloop Authors.
+// Copyright 2024-2025 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


app/controlplane/internal/service/auth.go
```diff
@@ -434,7 +434,15 @@ func generateUserJWT(userID, passphrase string, expiration time.Duration) (strin
 }
 
 func setOauthCookie(w http.ResponseWriter, name, value string) {
-	http.SetCookie(w, &http.Cookie{Name: name, Value: value, Path: "/", Expires: time.Now().Add(10 * time.Minute)})
+	http.SetCookie(w, &http.Cookie{
+		Name:     name,
+		Value:    value,
+		Path:     "/",
+		Expires:  time.Now().Add(10 * time.Minute),
+		HttpOnly: true,
```
migmartri, line 442: this might break local development, have you tried?

jiparis: Yes, both CLI and UI auth work perfectly.

jiparis: image

jiparis: The server response:
image

jiparis: And they are sent back after login:
image

jiparis: But it's strange, as I agree it should fail locally. I'll dig deeper into this.

jiparis: Ok, it seems that only Safari would break. Others consider "localhost" as secure: https://github.com/httpwg/http-extensions/issues/2605

app/controlplane/internal/service/auth.go
```diff
@@ -434,7 +434,15 @@ func generateUserJWT(userID, passphrase string, expiration time.Duration) (strin
...
+		Secure:   true,
```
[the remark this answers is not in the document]
jiparis: Reconsider this. Modern browsers seem to trust localhost for secure cookies.

PR #2943, opened 2026-03-27
fix(crafter): use deterministic filename for AI config collector

### LINKED ISSUE #2907: AI config collector creates duplicate uploads on retry
The AI config collector generates a non-deterministic digest for the config file. When a retry happens due to a remote state conflict, the file gets uploaded again with a different name, resulting in duplicate artifacts in the CAS.

The digest should be deterministic by incorporating the attestation `created_at` timestamp (or similar stable identifier) so that retries produce the same file and skip re-uploading.

### Example

On retry, the same config gets uploaded multiple times with different filenames:

```
2026-03-21T12:50:30.713+0100    INFO    {"component": "service", "msg": "artifact does not exist, uploading", "digest": "5b3bfaf2f3e8862b2a9aef668589546fa13c4f0465bd115547e0ae8da0bbd5c3", "name": "ai-agent-config-claude-169015807.json"}
2026-03-21T12:50:30.714+0100    INFO    {"component": "service", "msg": "artifact received, uploading now to backend", "name": "ai-agent-config-claude-169015807.json", "digest": "5b3bfaf2f3e8862b2a9aef668589546fa13c4f0465bd115547e0ae8da0bbd5c3", "size": 50566}
2026-03-21T12:50:31.230+0100    INFO    {"component": "service", "msg": "upload finished", "name": "ai-agent-config-claude-169015807.json", "digest": "5b3bfaf2f3e8862b2a9aef668589546fa13c4f0465bd115547e0ae8da0bbd5c3", "size": 50566}
```

Since the digest changes on each attempt, the CAS treats each retry as a new artifact.

waveywaves:
## Summary
- Replace `os.CreateTemp` (random suffix) with deterministic filenames in both `collector_aiagentconfig.go` and `collector_prmetadata.go`, eliminating duplicate CAS uploads on attestation retries
- Write temp files directly to `os.TempDir()` with deterministic names and `defer os.Remove`
- Remove `createPRMetadataTempFile` helper — inlined at the call site with explanatory comment

> **Note:** PR #2917 already fixed the primary root cause (non-deterministic content digest from `captured_at` timestamp). This PR provides **additional hardening** by making temp filenames deterministic, so that `Artifact.Name` (derived from `os.Stat(filepath).Name()`) is also stable across retries.

## Root Cause Analysis

The reviewer questioned whether the filename is actually the cause of duplicates since `Materials[m.Name]` already uses a deterministic key. The answer is yes — the map key is deterministic, but the **Artifact struct inside the material** is not.

Here is the chain of causation:

1. `uploadAgentConfig()` calls `os.CreateTemp("", "ai-agent-config-claude-*.json")`, which produces a random filename like `/tmp/ai-agent-config-claude-3847291.json`
2. `AddMaterialContractFree()` calls `addMaterial()`, which calls `uploadAndCraft()`
3. `uploadAndCraft()` calls `fileStats(artifactPath)`, which does `os.Stat(filepath).Name()` to get the **basename** of the temp file
4. That basename becomes `Artifact.Name` in the material proto (line 136 of `materials.go`)
5. `fileStats()` also computes `sha256(file_contents)` which becomes `Artifact.Digest`

Because the JSON payload is identical across retries, `Artifact.Digest` stays the same. But `Artifact.Name` changes every time (random temp suffix), so the CAS sees a different resource name on each retry. The map key `Materials["ai-agent-config-claude"]` is overwritten (good), but each retry still triggers a new CAS upload because the artifact metadata differs.

The fix makes the filename deterministic by deriving it from a content hash, so `Artifact.Name` is stable across retries.

## Changes

- **`pkg/attestation/crafter/collector_aiagentconfig.go`**:
  - Replace `os.CreateTemp` + manual write/close with `os.WriteFile` to a deterministic path in `os.TempDir()`
  - Filename uses full `ConfigHash` (no truncation)
  - Inline the path construction at the call site with root-cause comment
  - Remove `aiConfigTempFilePath()` helper and `MkdirTemp` wrapper

- **`pkg/attestation/crafter/collector_prmetadata.go`**:
  - Replace `createPRMetadataTempFile()` helper (which used `os.CreateTemp` with random suffix) with inline deterministic path using `sha256(content)`
  - Same pattern: `os.WriteFile` + `defer os.Remove` in `os.TempDir()`

- **Removed test files**:
  - `collector_aiagentconfig_test.go` — tested the removed `aiConfigTempFilePath` helper
  - `collector_prmetadata_test.go` — tested the removed `createPRMetadataTempFile` helper

## Test Plan
- [x] `go vet ./pkg/attestation/crafter/...` passes
- [x] `go test ./pkg/attestation/crafter/...` — all sub-packages pass

Fixes #2907


waveywaves:
@jiparis @migmartri gentle ping — this is ready for review. CI is green except for the known flaky `TestReferrerIntegration` deadlock (unrelated to this PR). Fixes both `collector_aiagentconfig.go` and `collector_prmetadata.go` deterministic filenames. Note: PR #2917 already fixed the primary root cause; this provides additional hardening.

pkg/attestation/crafter/collector_prmetadata.go
```diff
@@ -69,40 +72,22 @@ func (c *PRMetadataCollector) Collect(ctx context.Context, cr *Crafter, attestat
 		return fmt.Errorf("marshaling PR/MR metadata: %w", err)
 	}
 
-	materialName, tmpFile, err := createPRMetadataTempFile(metadata.Number, jsonData)
```
jiparis, line 72: PR metadata uses the PR number intentionally. It's already deduplicated across different PRs, but they share the name within the same PR since basically the latest one is the one that matters.

pkg/attestation/crafter/collector_aiagentconfig.go
```diff
@@ -99,20 +100,18 @@ func (c *AIAgentConfigCollector) uploadAgentConfig(
...
 
-	tmpFile, err := os.CreateTemp("", fmt.Sprintf("ai-agent-config-%s-*.json", agentName))
-	if err != nil {
-		return fmt.Errorf("creating temp file: %w", err)
-	}
-	defer os.Remove(tmpFile.Name())
-
-	if _, err := tmpFile.Write(jsonData); err != nil {
-		tmpFile.Close()
+	// Use a deterministic filename derived from the config hash so that retries
```
jiparis: This sentence is not accurate. Deduplication is done through the digest, which is already addressed https://github.com/chainloop-dev/chainloop/pull/2917

pkg/attestation/crafter/collector_aiagentconfig.go
```diff
@@ -99,20 +100,18 @@ func (c *AIAgentConfigCollector) uploadAgentConfig(
...
+	// produce the same Artifact.Name (via fileStats -> os.Stat().Name()) and
+	// avoid duplicate CAS uploads. PR #2917 fixed the primary root cause
+	// (non-deterministic content from captured_at); this is additional hardening.
+	tmpPath := filepath.Join(os.TempDir(), fmt.Sprintf("ai-agent-config-%s-%s.json", agentName, data.ConfigHash))
```
jiparis: Note that `ConfigHash` has a length of 64 characters, too long for a filename. I'd rather take the first 6 characters, or better, just use a constant name like `ai-agent-config.json`. Name conflicts are ok here, assuming that the project/workflow combo is always constant in the attestation (they refer to the same repository).

jiparis (changes requested):
Thanks for the contribution @waveywaves. 
Please remove those PR-metadata changes since they are not related to the main issue.
Regarding the AI config deduplication in the CAS, note that it was already solved at https://github.com/chainloop-dev/chainloop/pull/2917 (filename doesn't affect to CAS uploads), however, I agree that using the digest is better than a random rumber. Just check my comment about the filename length.

pkg/attestation/crafter/collector_prmetadata.go
waveywaves: Done. Reverted all PR metadata changes — the PR number naming is intentional. Restored the original file and test.

pkg/attestation/crafter/collector_aiagentconfig.go
waveywaves: Fixed. Removed the misleading dedup claim — updated the comment to just note the constant filename for consistent Artifact.Name.

pkg/attestation/crafter/collector_aiagentconfig.go
waveywaves: Good call. Switched to a constant filename: `ai-agent-config-{agentName}.json` in os.TempDir(). No hash in the filename at all — name conflicts within the same attestation are expected.

jiparis (approved):

jiparis (changes requested):
@waveywaves none of the commits are signed. This PR cannot be accepted for that reason.

waveywaves:
All commits are now GPG-signed. Also addressed all the review feedback in the second commit — reverted PR metadata changes and simplified the AI config filename to a constant name. Ready for re-review!

migmartri:
@jiparis ptal, thanks

waveywaves:
@jiparis Gentle nudge — all review feedback has been addressed and commits are GPG-signed since Apr 8. Would you be able to clear the CHANGES_REQUESTED so this can land? Thanks!

jiparis (approved):

Issue #2951, opened 2026-03-28
Add unified cache infrastructure with NATS KV and in-memory LRU backends


migmartri:
## Problem

The control plane has three ad-hoc caching mechanisms with inconsistent patterns:

1. **Membership cache** — package-level `expirable.LRU` with 1s TTL
2. **JWT claims cache** — local `expirable.LRU` inside middleware with 10s TTL
3. **Remote policy/group caches** — package-level `map` + `sync.Mutex` with no TTL (unbounded growth, latent memory leak)

These caches are not configurable, use global state, and are limited to single-process in-memory storage with no option to share cache state across replicas.

## Proposal

Introduce a unified `Cache[T any]` interface in `pkg/cache/` with two backends:

- **NATS JetStream KV** — used when NATS is configured, enables shared caching across replicas
- **In-memory LRU** — automatic fallback when NATS is not configured

A single constructor auto-selects the backend based on whether a NATS connection is provided. All existing ad-hoc caches are migrated to use this abstraction.

Key properties:
- Generic interface: `Get`, `Set`, `Delete`, `Purge`
- TTL-based expiration on both backends
- Structured debug/warn logging
- No external framework dependencies in the interface — reusable across projects
- Eliminates package-level global cache state

Spec: `docs/superpowers/specs/2026-03-28-cache-infrastructure-design.md`

PR #2901, opened 2026-03-20
feat(policies): parallelize policy evaluations with bounded concurrency

waveywaves:
## Summary

- Parallelize policy attachment evaluations within `VerifyStatement()` and `VerifyMaterial()` using `errgroup` with bounded concurrency (`max(NumCPU*2, 10)`)
- Add `WithMaxConcurrency(n)` option to `PolicyVerifier` for configurable parallelism
- Fix pre-existing concurrency safety issues:
  - Move `remotePolicyCache`/`remoteGroupCache` mutexes from instance-level to package-level (the caches are package-level maps but were protected by per-instance mutexes — a latent data race)
  - Use `singleflight.Group` to coalesce concurrent gRPC fetches for the same policy/group ref — ensures exactly one fetch per key, no duplicate RPCs or head-of-line blocking
  - Use `context.WithoutCancel` inside singleflight callbacks to prevent the winning goroutine's cancellation from propagating to waiters
  - Move `extism.SetLogLevel()` to `sync.Once` (it modifies global state; first engine's log level wins)
- Marshal statement JSON once per evaluation batch instead of per-policy
- Add 5 race detector tests:
  - Concurrent `VerifyStatement` from 10 goroutines
  - Concurrent `VerifyMaterial` from 10 goroutines (with real SPDX fixture)
  - `WithMaxConcurrency` option validation
  - Sequential (`maxConcurrency=1`) vs parallel result equivalence
  - `errgroup` cancellation on policy load failure

All existing + new tests pass with `-race` enabled. All CI checks green.

Fixes #2899

PR #2952, opened 2026-03-28
feat(cache): unified cache abstraction with in-memory and NATS KV backends

migmartri:
## Summary

- Add new `pkg/cache/` shared library with a generic `Cache[T]` interface, backed by in-memory LRU (default) or NATS JetStream KV (when configured)
- Migrate all four existing ad-hoc caching mechanisms (membership cache, JWT claims cache, remote policy cache, remote policy group cache) to use the unified interface
- Eliminate package-level global cache state by injecting cache instances through constructors and Wire
- Fix unbounded growth in remote policy/group caches by adding 5-minute TTL expiration

Closes https://github.com/chainloop-dev/chainloop/issues/2951

### DIFF
--- a/app/controlplane/cmd/wire.go
+++ b/app/controlplane/cmd/wire.go
@@ -140,18 +140,42 @@ var cacheProviderSet = wire.NewSet(
 	newClaimsCache,
 )
 
-func newClaimsCache(conn *nats.Conn) (cache.Cache[*jwt.MapClaims], error) {
-	opts := []cache.Option{cache.WithTTL(10 * time.Second)}
+func newClaimsCache(conn *nats.Conn, logger log.Logger) (cache.Cache[*jwt.MapClaims], error) {
+	l := log.NewHelper(logger)
+	backend := "memory"
+	opts := []cache.Option{cache.WithTTL(10 * time.Second), cache.WithLogger(&kratosLogAdapter{h: l})}
 	if conn != nil {
-		opts = append(opts, cache.WithNATS(conn, "jwt-claims"))
+		backend = "nats"
+		opts = append(opts, cache.WithNATS(conn, "chainloop-jwt-claims"))
 	}
+	l.Infow("msg", "cache initialized", "bucket", "chainloop-jwt-claims", "backend", backend, "ttl", "10s")
 	return cache.New[*jwt.MapClaims](opts...)
 }
 
-func newMembershipsCache(conn *nats.Conn) (cache.Cache[*entities.Membership], error) {
-	opts := []cache.Option{cache.WithTTL(time.Second)}
+func newMembershipsCache(conn *nats.Conn, logger log.Logger) (cache.Cache[*entities.Membership], error) {
+	l := log.NewHelper(logger)
+	backend := "memory"
+	opts := []cache.Option{cache.WithTTL(time.Second), cache.WithLogger(&kratosLogAdapter{h: l})}
 	if conn != nil {
-		opts = append(opts, cache.WithNATS(conn, "memberships"))
+		backend = "nats"
+		opts = append(opts, cache.WithNATS(conn, "chainloop-memberships"))
 	}
+	l.Infow("msg", "cache initialized", "bucket", "chainloop-memberships", "backend", backend, "ttl", "1s")
 	return cache.New[*entities.Membership](opts...)
+}
+
+// kratosLogAdapter adapts kratos log.Helper (Debugw(...interface{})) to cache.Logger (Debugw(string, ...any)).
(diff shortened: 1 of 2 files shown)


migmartri:
### Code review

Found 4 issues:

1. **Nil pointer dereference in `ChainloopGroupLoader.Load` when `GroupCache` is nil.** `attestation_init.go` calls `LoadPolicyGroup` without setting `GroupCache`, so `ChainloopGroupLoader` is created with a nil cache. `c.cache.Get(ctx, ref)` on the nil interface will panic at runtime for `chainloop://` scheme policy group references. The server-side path correctly initializes `groupCache` when nil (in `NewPolicyGroupVerifier`), but the CLI call site was not updated.

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/pkg/policies/group_loader.go#L116-L118

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/app/cli/pkg/action/attestation_init.go#L353-L358

2. **Regression: singleflight deduplication and bounded timeouts removed from policy/group loaders.** PR #2901 (commit `c7f0b9b7`, merged to `main` after this branch diverged) added `singleflight.Group` with 30-second bounded timeouts to `ChainloopLoader.Load` and `ChainloopGroupLoader.Load` to prevent thundering-herd on concurrent gRPC fetches. The `Cache[T]` abstraction provides caching but not singleflight coalescing -- multiple goroutines can race past a cache miss and each issue separate gRPC calls. Branch needs rebasing on `main`.

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/pkg/policies/loader.go#L161-L215

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/pkg/policies/group_loader.go#L109-L157

3. **Regression: parallel policy evaluation (`errgroup`) reverted to sequential loops.** PR #2901 added `errgroup` with bounded concurrency (`max(runtime.NumCPU(), 5)`) to parallelize policy evaluations in `VerifyMaterial` and `VerifyStatement`. This branch predates that merge, so evaluation is sequential. Additionally, `VerifyStatement` re-marshals the statement on every loop iteration instead of once before the loop.

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/pkg/policies/policies.go#L200-L215

https://github.com/chainloop-dev/chainloop/blob/7a2e219e92d584f3e78e918d1955f41089abb908/pkg/policies/policies.go#L455-L475

4. **Regression: `concurrency_test.go` will be dropped on merge.** PR #2901 added `pkg/policies/concurrency_test.go` to exercise race conditions under concurrent policy evaluation. This file exists on `main` but not on this branch. Merging without rebasing removes that test coverage.

Note: Issues 2-4 share a root cause -- the branch diverged from `main` before PR #2901 was merged. Rebasing on `main` and integrating the cache abstraction with the singleflight, errgroup, and bounded-timeout patterns from #2901 should resolve all three.

app/controlplane/internal/usercontext/attjwtmiddleware/attmiddleware.go
```diff
@@ -251,8 +257,10 @@ func WithJWTMulti(l log.Logger, opts ...JWTOption) middleware.Middleware {
 		logger.Infof("federated authentication enabled, using URL: %s", o.federatedAuthURL)
 	}
 
-	// claims cache with 10s TTL and unlimited keys
-	claimsCache := expirable.NewLRU[string, *jwt.MapClaims](0, nil, time.Second*10)
+	if o.claimsCache == nil {
```
migmartri: double chekc if ewe need to fallback here or just require a cache by default. Or is there any path there the cache is not present?

pkg/policies/policies.go
```diff
@@ -133,12 +140,30 @@ func WithEvalPhase(phase EvalPhase) PolicyVerifierOption {
...
+const defaultPolicyCacheTTL = 5 * time.Minute
+
 func NewPolicyVerifier(policies *v1.Policies, client v13.AttestationServiceClient, logger *zerolog.Logger, opts ...PolicyVerifierOption) *PolicyVerifier {
 	options := &PolicyVerifierOptions{}
 	for _, opt := range opts {
 		opt(options)
 	}
 
+	if options.PolicyCache == nil {
+		options.PolicyCache, _ = cache.New[*policyWithReference](cache.WithTTL(defaultPolicyCacheTTL))
```
migmartri, line 184: not sure we need to fallback herer, aren't we falling back in the constructor itself?

PR #2967, opened 2026-03-30
feat(policies): lenient structured finding validation at runtime

migmartri:
## Summary

- Add `WithLenientFindingValidation()` option to `PolicyVerifier` that makes structured finding validation errors non-fatal at runtime
- When validation fails in lenient mode, violations are preserved as plain strings with `FindingDegraded=true` and a warning instead of failing the entire evaluation
- Runtime crafter uses lenient mode; CLI `policy develop eval` keeps strict validation for fast feedback during policy authoring
- Add `finding_degraded` boolean field to the `Violation` proto message so downstream consumers can distinguish intentionally unstructured violations from degraded ones


migmartri:
Will close it for now

PR #2976, opened 2026-04-01
feat: add policyEvaluationsCount and policyViolationsCount to attestation predicate

### LINKED ISSUE #2975: feat: add policyEvaluationsCount and policyViolationsCount to attestation predicate
## Summary

Add `policyEvaluationsCount` and `policyViolationsCount` fields to the attestation predicate, alongside the existing `policyHasViolations` boolean.

## Motivation

Policy evaluations are moving from inline `policyEvaluations` to `policyEvaluationsRef` (stored in CAS). External consumers that call `ExtractPredicate` only need summary counts — not the full evaluation data. Currently, the only summary field is `policyHasViolations` (bool), which doesn't tell consumers how many evaluations or violations exist.

## Changes

- Add `policyEvaluationsCount` (total number of policy evaluations) and `policyViolationsCount` (total number of violations across all evaluations) to the predicate
- Expose via `PolicyEvaluationStatus` in Go and the proto API response
- Compute counts during attestation rendering in `groupEvaluations()`

migmartri:
## Summary

- Add `policyEvaluationsCount` and `policyViolationsCount` fields to the attestation predicate, alongside the existing `policyHasViolations` boolean
- Expose the new counts through `PolicyEvaluationStatus` in the gRPC API response
- Consolidate gated-violation detection into `groupEvaluations`, eliminating a redundant loop in `predicate()`

Closes #2975

### DIFF
--- a/pkg/attestation/renderer/chainloop/v02.go
+++ b/pkg/attestation/renderer/chainloop/v02.go
@@ -52,6 +52,10 @@ type ProvenancePredicateV02 struct {
 
 	// Whether the attestation has policy violations
 	PolicyHasViolations bool `json:"policyHasViolations"`
+	// Total number of policy evaluations
+	PolicyEvaluationsCount int `json:"policyEvaluationsCount"`
+	// Total number of policy violations across all evaluations
+	PolicyViolationsCount int `json:"policyViolationsCount"`
 	// Whether the attestation has policy violations in gated policies
 	PolicyHasGatedViolations bool `json:"policyHasGatedViolations,omitempty"`
 	// Whether we want to block the attestation on policy violations
@@ -234,7 +238,7 @@ func (r *RendererV02) predicate() (*structpb.Struct, error) {
 		return nil, fmt.Errorf("error normalizing materials: %w", err)
 	}
 
-	policies, hasViolations, err := mappedPolicyEvaluations(r.att)
+	evalResult, err := mappedPolicyEvaluations(r.att)
 	if err != nil {
 		return nil, fmt.Errorf("error rendering policy evaluations: %w", err)
 	}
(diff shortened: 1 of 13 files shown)


pkg/attestation/renderer/chainloop/v02.go
```diff
@@ -244,27 +248,21 @@ func (r *RendererV02) predicate() (*structpb.Struct, error) {
...
 	// In all cases, if the bypass flag is set, the attestation is not blocked
-	gated := false
-	for _, eval := range r.att.PolicyEvaluations {
-		if len(eval.Violations) > 0 && eval.Gate {
-			gated = true
-			break
-		}
-	}
-	blocked := !r.att.GetBypassPolicyCheck() && (gated || hasViolations && r.att.GetBlockOnPolicyViolation())
+	blocked := !r.att.GetBypassPolicyCheck() && (evalResult.hasGatedViolations || (evalResult.hasViolations && r.att.GetBlockOnPolicyViolation()))
```
migmartri, line 255: this logic might be incorrect, we can gate specific evaluations

migmartri: The logic is equivalent to the original. In `groupEvaluations`, the gated check is per-evaluation:

```go
if len(ev.Violations) > 0 {
    // ...
    if ev.Gate {
        res.hasGatedViolations = true
    }
}
```

`ev.Gate` comes from the individual evaluation's `Gate` field (mapped at `renderEvaluation` line 382: `Gate: ev.Gate`). So `hasGatedViolations` is only set when a specific gated evaluation has violations — same as the original loop that checked `len(eval.Violations) > 0 && eval.Gate` per evaluation.

PR #2981, opened 2026-04-03
feat(validators): lazy-initialize schema validators with sync.Once

### LINKED ISSUE #919: Change initialization of schema validators
Some of the Chainloop crafters use a JSON schema approach for validation, meaning that incoming evidence is validated against a specified schema. Currently, the schemas are loaded in an `init` function, which loads all schemas registered by the CLI into memory.

At the moment, this approach loads the schemas for every single command, regardless of whether it's for attestation or not. As the CLI grows, this could become a memory issue.

The goal of this task is to refactor the initialization of the JSON schemas so that they are only loaded when needed.

As reference, the package can be found here: https://github.com/chainloop-dev/chainloop/blob/main/internal/schemavalidators/schemavalidators.go

waveywaves:
## Summary

- Replace eager `init()` that compiles all JSON schema validators at package load time with lazy initialization using `sync.Once`, one per schema family (CycloneDX, CSAF, RunnerContext, PRInfo, AIAgentConfig, AICodingSession)
- Each family now has its own private initializer that creates a dedicated `jsonschema.Compiler` with only the resources needed for that family, avoiding unnecessary work when only a subset of validators is used
- Remove the package-level `schemaURLMapping` map that was only used by `init()`

Fixes #919

## Test plan

- [x] `go vet ./internal/schemavalidators/...` passes with no issues
- [x] `go test ./internal/schemavalidators/...` -- all existing tests pass (CycloneDX 1.5/1.6, CSAF 2.0/2.1, RunnerContext, AIAgentConfig)
- [ ] Verify no other packages depend on `schemaURLMapping` or the `init()` side effects
- [ ] Confirm lazy initialization does not introduce data races (`go test -race`)

### DIFF
--- a/internal/schemavalidators/schemavalidators.go
+++ b/internal/schemavalidators/schemavalidators.go
@@ -18,8 +18,10 @@ package schemavalidators
 import (
 	_ "embed"
 	"errors"
+	"fmt"
 	"slices"
 	"strings"
+	"sync"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/santhosh-tekuri/jsonschema/v5"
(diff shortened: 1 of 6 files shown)


jiparis:
@waveywaves Thanks for the contribution. Please sign your commits so that it can be merged.

PR #2983, opened 2026-04-03
fix(crafter): surface auto-discovery validation errors

### LINKED ISSUE #2394: Material type auto-discovery hides material name not being dns-1123 compliant
When using auto-discovery for material type, when the material type is not dns-1123 compliant, so contains characters others than  lowercase letters, numbers, and hyphens and error is thrown hiding the real issue. The error that is reported:

```
ERR adding material: failed to auto-discover material kind: %!w(<nil>)
```

waveywaves:
## Summary

- Stop auto-detecting further material kinds when `AddMaterialContractFree` returns a `protovalidate.ValidationError`.
- Surface invalid material-name validation errors directly instead of masking them behind a generic auto-discovery failure.

Fixes #2394

Assisted-by: pi

### DIFF
--- a/pkg/attestation/crafter/crafter.go
+++ b/pkg/attestation/crafter/crafter.go
@@ -653,31 +653,33 @@ func (c *Crafter) IsMaterialInContract(key string) bool {
 // AddMaterialContactFreeWithAutoDetectedKind adds a material to the crafting state checking the incoming material matches any of the
 // supported types in validation order. If the material is not found it will return an error.
 func (c *Crafter) AddMaterialContactFreeWithAutoDetectedKind(ctx context.Context, attestationID, name, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string, opts ...AddOpt) (*api.Attestation_Material, error) {
-	var err error
 	for _, kind := range schemaapi.CraftingMaterialInValidationOrder {
 		m, err := c.AddMaterialContractFree(ctx, attestationID, kind.String(), name, value, casBackend, runtimeAnnotations, opts...)
 		if err == nil {
-			// Successfully added material, return the kind
 			return m, nil
 		}
 
 		c.Logger.Debug().Err(err).Str("kind", kind.String()).Msg("failed to add material")
 
-		// Handle base error for upload and craft errors except the opening file error
-		// TODO: have an error to detect validation error instead
 		var policyError *policies.PolicyError
 		if errors.Is(err, materials.ErrBaseUploadAndCraft) || errors.As(err, &policyError) {
 			return nil, err
 		}
 
-		// This is a final error, we detected the kind
 		if v1.IsAttestationStateErrorConflict(err) {
+			return nil, err
+		}
+
+		// Proto-validation errors (e.g. invalid material name) are schema-level
+		// failures, not kind mismatches. Stop probing immediately so the real
+		// error is surfaced to the user instead of being masked by the loop.
+		var valErr *protovalidate.ValidationError
+		if errors.As(err, &valErr) {
 			return nil, err
 		}
 	}
 
-	// Return an error if no material could be added
-	return nil, fmt.Errorf("failed to auto-discover material kind: %w", err)
+	return nil, errors.New("failed to auto-discover material kind")
 }
 
 // addMaterials adds the incoming material m to the crafting state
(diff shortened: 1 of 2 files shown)


javirln:
@waveywaves is this ready to be reviewed?

waveywaves:
@jiparis Ready for review — all commits GPG-signed and rebased on main. This fixes a variable shadowing bug that silently masks DNS-1123 validation errors in AddMaterialContractFree (issue #2394).

waveywaves:
@javirln Yes, this is ready for review! All commits are GPG-signed and rebased on latest main. Sorry for the delayed reply.

waveywaves:
@migmartri Would you be able to take a look at this one? It fixes a variable shadowing bug where DNS-1123 validation errors get silently masked in AddMaterialContractFree (issue #2394). All commits GPG-signed, CI green, cubic-dev-ai's finding already addressed. Happy to answer any questions!

pkg/attestation/crafter/crafter.go
```diff
@@ -628,9 +639,12 @@ func (c *Crafter) IsMaterialInContract(key string) bool {
...
 // supported types in validation order. If the material is not found it will return an error.
 func (c *Crafter) AddMaterialContactFreeWithAutoDetectedKind(ctx context.Context, attestationID, name, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string) (*api.Attestation_Material, error) {
-	var err error
+	var (
+		err error
+		m   *api.Attestation_Material
+	)
 	for _, kind := range schemaapi.CraftingMaterialInValidationOrder {
-		m, err := c.AddMaterialContractFree(ctx, attestationID, kind.String(), name, value, casBackend, runtimeAnnotations)
+		m, err = c.AddMaterialContractFree(ctx, attestationID, kind.String(), name, value, casBackend, runtimeAnnotations)
```
jiparis: This is the actual fix that solves https://github.com/chainloop-dev/chainloop/issues/2394. `err` was being override in the in the loop scope, making it always `nil` outside it.

jiparis:
Thanks @waveywaves . Check my comment for the actual fix. https://github.com/chainloop-dev/chainloop/issues/2394 is not about DNS names, but just surfacing the error as you did in that line. Please check.

waveywaves:
Thanks @jiparis — you're right, the DNS-1123 regex was out of scope. Stripped the PR back to just the two relevant changes:

1. Variable shadowing fix: `m, err :=` → `m, err =` so the error propagates out of the loop
2. `protovalidate.ValidationError` early break so schema-level failures aren't masked by the kind-probing loop

All the unrelated changes (gracefulGitRepoHead refactor, policyEvalMatches removal, RunCollectors optimization) are gone. One file, 13 insertions, 7 deletions.

pkg/attestation/crafter/crafter.go
```diff
@@ -653,30 +653,36 @@ func (c *Crafter) IsMaterialInContract(key string) bool {
 // AddMaterialContactFreeWithAutoDetectedKind adds a material to the crafting state checking the incoming material matches any of the
 // supported types in validation order. If the material is not found it will return an error.
 func (c *Crafter) AddMaterialContactFreeWithAutoDetectedKind(ctx context.Context, attestationID, name, value string, casBackend *casclient.CASBackend, runtimeAnnotations map[string]string, opts ...AddOpt) (*api.Attestation_Material, error) {
-	var err error
+	var (
+		err error
```
jiparis: I think this is not needed, otherwise we would be propagating the error from one iteration to the next in the for loop.  It's immediately replaced by the `c.AddMaterialContractFree` call, but still semantically wrong. The actual fix is the `ValidationError` check that you added below, returning early the first time we detect such error.

waveywaves:
Addressed the latest feedback: dropped the outer `m/err` change, kept the early `protovalidate.ValidationError` return as the actual fix, added a regression test for invalid auto-detected material names, and refreshed the PR/commit AI disclosure to use `Assisted-by: pi` instead of the stale Claude Code wording.

Validated with:

```bash
go test ./pkg/attestation/crafter/...
```

jiparis (approved):

jiparis (approved):

jiparis:
Thanks @waveywaves !

Issue #2977, opened 2026-04-01
When provided path to install script does not exist doesn't provide error output


jorgemoralespou:
When installing chainloop via the script I was providing a non existing target dir, and could not figure out what was the problem of the CLI not being installed:

```
[~] $ curl -sfL https://dl.chainloop.dev/cli/install.sh | bash -s -- --path ~/.local/bin2
Step 1: Downloading: chainloop-linux-arm64, Version: 1.89.6
######################################################################## 100.0%
Done...

Step 1.2: Verifying checksum
chainloop-linux-arm64: OK
Checksum OK


Signature verification skipped, cosign is not installed

Step 2: Installing: chainloop to /home/eduk8s/.local/bin2
sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
```

When used an existing dir, everything worked fine:
```
[~] $ mkdir -p ~/.local/bin2
[~] $ curl -sfL https://dl.chainloop.dev/cli/install.sh | bash -s -- --path ~/.local/bin2
Step 1: Downloading: chainloop-linux-arm64, Version: 1.89.6
######################################################################## 100.0%
Done...

Step 1.2: Verifying checksum
chainloop-linux-arm64: OK
Checksum OK


Signature verification skipped, cosign is not installed

Step 2: Installing: chainloop to /home/eduk8s/.local/bin2
Step 3: Cleanup
Done...

Client Version: 1.89.6
Server Version: 1.89.6
Check here for the next steps: https://docs.chainloop.dev

Run 'chainloop auth login' to get started
```

migmartri:
Thanks @jorgemoralespou, this is now fixed in the upstream script

```
curl -sfL https://dl.chainloop.dev/cli/install.sh | bash -s -- --path non-existing
Installation path does not exist: non-existing
Please create it first or provide a valid path with --path.
```

PR #2984, opened 2026-04-03
feat(materials): extract main component info from SPDX files

### LINKED ISSUE #2580: Include main component information from SPDX files
Like we do with CycloneDX SBOM files, we shall include the main component information as well for the SPDX files so the information is available at attestation level as annotations in the material.

waveywaves:
## Summary

- Extract main component information (name, version, kind) from SPDX JSON SBOM files, following the existing CycloneDX pattern
- Use `spdxlib.GetDescribedPackageIDs` to identify the described package, then populate `SBOMArtifact.MainComponent` with the package's name, version, and lowercased `PrimaryPackagePurpose`
- Standardize container image names via `go-containerregistry` (same as CycloneDX) and gracefully skip when no described package is found

Fixes #2580

## Test plan
- [x] Existing SPDX tests continue to pass (no described package case handled gracefully)
- [x] New test case: SPDX with described APPLICATION package extracts name, version, kind
- [x] New test case: SPDX with described CONTAINER package standardizes image name
- [x] `go vet` passes cleanly
- [x] Full materials test suite passes

### DIFF
--- a/pkg/attestation/crafter/materials/spdxjson.go
+++ b/pkg/attestation/crafter/materials/spdxjson.go
@@ -138,11 +138,7 @@ func (i *SPDXJSONCrafter) extractMainComponent(m *api.Attestation_Material, doc 
 		}
 	}
 
-	m.M.(*api.Attestation_Material_SbomArtifact).SbomArtifact.MainComponent = &api.Attestation_Material_SBOMArtifact_MainComponent{
-		Name:    name,
-		Kind:    kind,
-		Version: version,
-	}
+	setMainComponent(m, name, kind, version)
 
 	return nil
 }
--- a/pkg/attestation/crafter/materials/cyclonedxjson.go
+++ b/pkg/attestation/crafter/materials/cyclonedxjson.go
@@ -251,8 +251,6 @@ func (i *CyclonedxJSONCrafter) extractMetadata(m *api.Attestation_Material, meta
 
 // extractMainComponent inspects the SBOM and extracts the main component if any and available
 func (i *CyclonedxJSONCrafter) extractMainComponent(m *api.Attestation_Material, component *cyclonedxComponent) error {
-	var mainComponent *SBOMMainComponentInfo
-
 	// If the version is empty, try to extract it from the properties
 	if component.Version == "" {
 		for _, prop := range component.Properties {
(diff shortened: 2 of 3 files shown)


waveywaves:
@jiparis Ready for review — GPG-signed and rebased. Extracts main component info from SPDX SBOMs (matching the existing CycloneDX behavior), issue #2977. Also fixed the SA1019 staticcheck lint for the deprecated Artifact.Id field.

pkg/attestation/crafter/materials/spdxjson.go
```diff
@@ -65,13 +66,92 @@ func (i *SPDXJSONCrafter) Craft(ctx context.Context, filePath string) (*api.Atte
...
+	}
+
+	name := describedPkg.PackageName
+	version := describedPkg.PackageVersion
+
+	// PrimaryPackagePurpose is optional in SPDX 2.3. If absent, skip main component extraction
+	// since we cannot determine the component kind.
+	kind := strings.ToLower(describedPkg.PrimaryPackagePurpose)
+	if kind == "" {
+		return fmt.Errorf("described package %q has no PrimaryPackagePurpose set", describedPkg.PackageName)
```
jiparis: In this case, as a best effort, it could just return name and version. Not sure if Kind is expected in other parts of the code.

pkg/attestation/crafter/materials/spdxjson.go
```diff
@@ -65,13 +66,92 @@ func (i *SPDXJSONCrafter) Craft(ctx context.Context, filePath string) (*api.Atte
...
+	}
+
+	// For container packages, standardize the name via go-containerregistry
+	// to get the full repository name and strip any tag (matching CycloneDX behavior)
+	if kind == containerComponentKind {
+		ref, err := remotename.ParseReference(name)
```
jiparis, line 133: Here it's likely that chainloop doesn't have the credentials for this query in most cases. Instead of returning, I'd just continue and return the existing values.

pkg/attestation/crafter/materials/spdxjson.go
waveywaves: Good point — now returns name and version as best effort when PrimaryPackagePurpose is absent. Kind will be empty string.

pkg/attestation/crafter/materials/spdxjson.go
waveywaves: Makes sense — chainloop likely won't have registry credentials in most cases. Now logs at debug level and continues with the original name instead of returning an error.

pkg/attestation/crafter/materials/spdxjson.go
```diff
@@ -65,13 +66,91 @@ func (i *SPDXJSONCrafter) Craft(ctx context.Context, filePath string) (*api.Atte
...
+	}
+
+	m.M.(*api.Attestation_Material_SbomArtifact).SbomArtifact.MainComponent = &api.Attestation_Material_SBOMArtifact_MainComponent{
+		Name:    name,
+		Kind:    kind,
+		Version: version,
+	}
+
+	return nil
 }
```
migmartri, line 144: nit: This type-assertion + assignment block is identical to the one in `cyclonedxjson.go:290-294`. Consider extracting a shared helper in `materials.go`, e.g.:

```go
func setMainComponent(m *api.Attestation_Material, name, kind, version string) {
    m.M.(*api.Attestation_Material_SbomArtifact).SbomArtifact.MainComponent =
        &api.Attestation_Material_SBOMArtifact_MainComponent{
            Name: name, Kind: kind, Version: version,
        }
}
```

Also — there's an existing `SBOMMainComponentInfo` struct in `materials.go:88` that the CycloneDX crafter uses as an intermediate but this PR bypasses. If we add the helper above, `SBOMMainComponentInfo` becomes unnecessary and can be dropped from both crafters.

waveywaves: Great suggestion — extracted a shared `setMainComponent()` helper in `materials.go` and updated both CycloneDX and SPDX crafters to use it. Dropped the `SBOMMainComponentInfo` struct entirely. Net -22 lines.

migmartri:
Thanks!

PR #2987, opened 2026-04-04
fix: add reloadable NATS connection with automatic reconnect handling

### LINKED ISSUE #2986: fix: control plane loses NATS connection when NATS pods restart
## Problem

When NATS pods restart (e.g., during rolling updates, scaling events, or crashes), the control plane loses its connection to NATS and does not recover. The only workaround is to restart the control plane pods as well.

This affects audit log publishing (via JetStream) and all three NATS-backed KV caches (JWT claims, org memberships, policy eval bundles).

## Root Cause

The NATS connection in `app/controlplane/cmd/main.go:newNatsConnection` is established with minimal options — no explicit reconnect configuration and no error/disconnect/reconnect callbacks are registered.

While the NATS Go client has reconnection enabled by default, the higher-level consumers (audit log publisher, KV caches) don't handle the underlying connection being re-established:

- **Audit log publisher** (`app/controlplane/pkg/auditor/nats.go`): Creates a JetStream context once at startup. After a reconnect, the JetStream context may become stale.
- **KV caches** (`pkg/cache/natskv.go`): A `WithReconnect()` option and `watchReconnect()` handler already exist in the cache layer, but none of the three cache constructors in `app/controlplane/cmd/wire.go` pass a reconnect channel.

## Suggested Fix

1. **Register reconnect/disconnect callbacks** on the NATS connection in `newNatsConnection` to log events and signal dependent components.
2. **Wire the reconnect channel** into the cache constructors — the infrastructure (`WithReconnect`, `watchReconnect`) already exists but is unused.
3. **Re-initialize JetStream context** in the audit log publisher after a NATS reconnection event.
4. **Add connection health monitoring** — expose NATS connection status in health checks or metrics so pod liveness probes can detect a permanently broken connection.

migmartri:
## Summary

When NATS pods restart, the control plane loses its connection and does not recover until restarted. This adds a `ReloadableConnection` wrapper in `pkg/natsconn` that broadcasts reconnection events to all consumers (caches and audit publisher), allowing them to reinitialize their JetStream handles automatically.

- New `pkg/natsconn` package with `ReloadableConnection` type providing Subscribe/Broadcast fan-out for reconnect events, decoupled from controlplane proto config for cross-repo importability
- Audit log publisher now reinitializes JetStream stream on NATS reconnection
- All three NATS KV caches now use the existing `WithReconnect` plumbing that was previously unwired
- NATS connection is properly drained on shutdown via Wire cleanup

Fixes #2986

### DIFF
--- a/app/controlplane/cmd/wire.go
+++ b/app/controlplane/cmd/wire.go
@@ -45,7 +45,7 @@ import (
 	"github.com/google/wire"
 )
 
-func wireApp(*conf.Bootstrap, credentials.ReaderWriter, log.Logger, sdk.AvailablePlugins) (*app, func(), error) {
+func wireApp(context.Context, *conf.Bootstrap, credentials.ReaderWriter, log.Logger, sdk.AvailablePlugins) (*app, func(), error) {
 	panic(
 		wire.Build(
 			wire.Bind(new(credentials.Reader), new(credentials.ReaderWriter)),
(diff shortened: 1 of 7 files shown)


app/controlplane/cmd/wire.go
```diff
@@ -141,37 +143,40 @@ var cacheProviderSet = wire.NewSet(
 	newPolicyEvalBundleCache,
 )
 
-func newClaimsCache(conn *nats.Conn, logger log.Logger) (cache.Cache[*jwt.MapClaims], error) {
+func newClaimsCache(rc *natsconn.ReloadableConnection, logger log.Logger) (cache.Cache[*jwt.MapClaims], error) {
```
migmartri: Would it make sense to provide context in all these initializer?

Issue #2979, opened 2026-04-02
Policy evaluation fails on older CLIs when policies use newer custom builtins


migmartri:
## Problem

When a Rego policy uses a custom builtin function (e.g. `chainloop.vulnerability`) that was introduced in a newer CLI version, older CLI clients fail with a hard error during attestation:

```
ERR adding material: error applying policy groups to material: failed to execute policy: failed to evaluate policy: 1 error occurred: vulnerabilities:115: rego_type_error: undefined function chainloop.vulnerability
```

OPA performs static type checking at compile time. If a builtin isn't registered in the client's engine, the entire policy is rejected before evaluation — there's no way to catch or guard against this in Rego.

This means any policy updated to use a new builtin becomes incompatible with all older CLI versions.

## Proposed fix

**Graceful degradation**: In `executeScript` (`pkg/policies/policies.go`), catch `rego_type_error: undefined function` errors and degrade the policy to `skipped` with a reason instead of hard-failing the attestation. This uses the existing `skipped`/`skip_reasons` proto fields. This will not fix existing CLIs but will help moving fowarwd.

PR #2988, opened 2026-04-06
feat(policies): add description field and split findings for backward compat

migmartri:
## Summary
- Add optional `description` field to `PolicyVulnerabilityFinding` proto for richer vulnerability context
- Fix forward compatibility: use `DiscardUnknown` in protojson unmarshal so older CLIs ignore unrecognized fields instead of failing
- Split structured violation data into a new `findings` field in policy result output, keeping `violations` as strings-only for backward compatibility with older CLIs that don't support structured violations

refs https://github.com/chainloop-dev/chainloop/issues/2979

### DIFF
--- a/app/cli/internal/policydevel/eval.go
+++ b/app/cli/internal/policydevel/eval.go
@@ -49,7 +49,8 @@ type EvalOptions struct {
 }
 
 type EvalResult struct {
-	Violations  []json.RawMessage `json:"violations"`
+	Violations  []string          `json:"violations"`
+	Findings    []json.RawMessage `json:"findings,omitempty"`
 	SkipReasons []string          `json:"skip_reasons"`
 	Skipped     bool              `json:"skipped"`
 }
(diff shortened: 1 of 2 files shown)


javirln (approved):

pkg/policies/findings/registry.go
```diff
@@ -60,7 +60,9 @@ func ValidateFinding(findingType string, raw map[string]any) (proto.Message, err
 	}
 
 	msg := factory()
-	if err := protojson.Unmarshal(data, msg); err != nil {
+	// DiscardUnknown allows policies to include fields added in newer proto
+	// versions without breaking older CLIs that haven't been updated yet.
+	if err := (protojson.UnmarshalOptions{DiscardUnknown: true}).Unmarshal(data, msg); err != nil {
```
jiparis, line 65: I think this should be the default.

migmartri: do you mean to set it as default?

PR #2992, opened 2026-04-07
fix: surface proper parsing errors for malformed v2 contracts

Piskoo:
When a user provided a v2 contract YAML with valid metadata but a structural error in the spec (e.g., `materials: ref: vulnerabilities` instead of `materials: - ref: vulnerabilities`), the server returned a misleading "name is required" error. The actual parsing error was silently swallowed in `extractMetadata()`.

The fix uses a lenient proto parse to detect v2 fields (`apiVersion` and `kind`) even when the contract has structural errors. When a v2 contract is detected but has parse errors, the actual error is now surfaced to the user.

#### Before
```
$ chainloop apply -f testcontract.yml 
ERR parsing file testcontract.yml: missing 'metadata.name' field in YAML document of kind "Contract"
exit status 1
```

#### After
```
$ chainlooop apply -f testcontract.yml 
ERR invalid contract: error unmarshalling raw message: :8:13 expected sequence, got mapping
   8 |             ref: vulnerabilities
   8 | ............^

exit status 1
```

### DIFF
--- a/app/controlplane/internal/service/workflowcontract.go
+++ b/app/controlplane/internal/service/workflowcontract.go
@@ -17,6 +17,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 
 	pb "github.com/chainloop-dev/chainloop/app/controlplane/api/controlplane/v1"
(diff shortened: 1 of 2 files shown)


migmartri:
can you provide an example of the before and after?

PR #2999, opened 2026-04-08
fix(cli): resolve attestation init digest mismatch with old servers

javirln:
## Summary

- Fix inverted condition in `RunCollectors` that skipped state reload when the client-side digest was stale (old servers not returning digests from Save RPC)
- Always reload crafting state after collectors run, ensuring the digest is in sync with the server regardless of server version

Closes PFM-5378


migmartri:
Thanks.

PR #3021, opened 2026-04-10
feat: add operation authorization forward middleware

Piskoo:
## Summary
Added a middleware that forwards operations to an external authorization endpoint before allowing them to proceed. Disabled by default.

### DIFF
--- a/app/controlplane/cmd/wire.go
+++ b/app/controlplane/cmd/wire.go
@@ -59,7 +59,7 @@ func wireApp(context.Context, *conf.Bootstrap, credentials.ReaderWriter, log.Log
 			wire.Bind(new(biz.CASClient), new(*biz.CASClientUseCase)),
 			serviceOpts,
 			wire.Value([]biz.CASClientOpts{}),
-			wire.FieldsOf(new(*conf.Bootstrap), "Server", "Auth", "Data", "CasServer", "ReferrerSharedIndex", "Onboarding", "PrometheusIntegration", "PolicyProviders", "NatsServer", "FederatedAuthentication"),
+			wire.FieldsOf(new(*conf.Bootstrap), "Server", "Auth", "Data", "CasServer", "ReferrerSharedIndex", "Onboarding", "PrometheusIntegration", "PolicyProviders", "NatsServer", "FederatedAuthentication", "OperationAuthorizationProvider"),
 			wire.FieldsOf(new(*conf.Data), "Database"),
 			dispatcher.New,
 			authz.NewCasbinEnforcer,
--- a/app/controlplane/internal/usercontext/operation_authorization_middleware.go
+++ b/app/controlplane/internal/usercontext/operation_authorization_middleware.go
@@ -33,9 +33,7 @@ import (
 )
 
 type operationAuthRequest struct {
-	Operation      string `json:"operation"`
-	UserID         string `json:"user_id"`
-	OrganizationID string `json:"organization_id,omitempty"`
+	Operation string `json:"operation"`
 }
 
 type operationAuthResponse struct {
@@ -67,20 +65,8 @@ func WithOperationAuthorizationMiddleware(conf *conf.OperationAuthorizationProvi
 				return handler(ctx, req)
 			}
 
-			user := entities.CurrentUser(ctx)
-			if user == nil {
-				return handler(ctx, req)
-			}
-
-			var orgID string
-			if org := entities.CurrentOrg(ctx); org != nil {
-				orgID = org.ID
-			}
-
 			result, err := callAuthorizationEndpoint(ctx, client, url, &operationAuthRequest{
-				Operation:      operation,
-				UserID:         user.ID,
-				OrganizationID: orgID,
+				Operation: operation,
 			})
 			if err != nil {
 				logger.Errorw("msg", "operation authorization call failed, denying request (fail-closed)", "error", err, "operation", operation)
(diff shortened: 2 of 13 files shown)


app/controlplane/internal/usercontext/operation_authorization_middleware.go
```diff
@@ -0,0 +1,160 @@
...
+	errorsAPI "github.com/go-kratos/kratos/v2/errors"
+	"github.com/go-kratos/kratos/v2/log"
+	"github.com/go-kratos/kratos/v2/middleware"
+	"github.com/go-kratos/kratos/v2/transport"
+
+	"github.com/chainloop-dev/chainloop/pkg/cache"
+)
+
+// Operations that require authorization checks
+var operationAuthTargets = map[string]bool{
```
migmartri: woudl it be possible to add this property to authz.go? As in an extension of the current system?

app/controlplane/internal/usercontext/operation_authorization_middleware.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+func callAuthorizationEndpoint(ctx context.Context, client *http.Client, url string, authReq *operationAuthRequest) (*operationAuthResponse, error) {
+	body, err := json.Marshal(authReq)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling request: %w", err)
+	}
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
```
migmartri, line 91: we might want to forward the API token as a header

migmartri (changes requested):

app/controlplane/internal/usercontext/operation_authorization_middleware.go
jiparis: I agree, that would be great

app/controlplane/internal/usercontext/operation_authorization_middleware.go
```diff
@@ -0,0 +1,160 @@
...
+			if user == nil {
+				return handler(ctx, req)
+			}
+
+			var orgID string
+			if org := entities.CurrentOrg(ctx); org != nil {
+				orgID = org.ID
+			}
+
+			cacheKey := fmt.Sprintf("%s:%s:%s", user.ID, orgID, operation)
```
jiparis: Not sure the kind of operations we want to authorize this way, but I think we should add the parameters here. Most RBAC operations are done against specific resources.

app/controlplane/internal/usercontext/operation_authorization_middleware.go
jiparis: That should be the way. @Piskoo check how we do with policy providers. We use plain `Bearer` header so that providers can authenticate it automatically using standard middlewares.

app/controlplane/internal/usercontext/operation_authorization_middleware.go
Piskoo: Updated, thanks

app/controlplane/internal/usercontext/operation_authorization_middleware.go
Piskoo: Updated, thanks

app/controlplane/internal/usercontext/operation_authorization_middleware.go
Piskoo: Removed the cache from v1, we can add it later

app/controlplane/cmd/wire.go
```diff
@@ -159,3 +162,26 @@ func newMembershipsCache(logger log.Logger) (cache.Cache[*entities.Membership],
...
+// newOperationAuthzClient creates a gRPC client for the external operation authorization service.
+// Returns nil (disabled) if the config is missing or disabled.
+// The connection uses no static token since the middleware forwards per-request Bearer tokens via gRPC metadata.
+func newOperationAuthzClient(bootstrap *conf.Bootstrap, logger log.Logger) (v1.OperationAuthorizationServiceClient, func(), error) {
+	cfg := bootstrap.GetOperationAuthorizationProvider()
+	if cfg == nil || !cfg.GetEnabled() || cfg.GetGrpc() == nil {
+		return nil, func() {}, nil
+	}
+
+	conn, err := grpcconn.New(cfg.GetGrpc().GetAddr(), "", grpcconn.WithInsecure(cfg.GetInsecure()))
```
migmartri: this might require also take into account custom certs? btw, why did we chose to use grpc here?

migmartri (commented):
My main question is why did we decide to use gRPC for this forwarder, when the policyProvider one uses http

Piskoo:
> My main question is why did we decide to use gRPC for this forwarder, when the policyProvider one uses http

I've misread the comment in the other PR, reverted it back to HTTP

app/controlplane/pkg/authz/authz.go
```diff
@@ -351,67 +358,67 @@ var RolesMap = map[Role][]*Policy{
...
 }
 
-// ServerOperationsMap is a map of server operations to the ResourceAction tuples that are
-// required to perform the operation
-// If it contains more than one policy, all of them need to be true
-var ServerOperationsMap = map[string][]*Policy{
+// ServerOperationsMap is a map of server operations to their authorization requirements.
+// Each entry specifies the Casbin policies required (all must pass) and whether the
+// operation also requires external authorization.
+var ServerOperationsMap = map[string]*OperationPolicy{
```
jiparis, line 364: Note that this is a breaking change. Please create an issue to update consumers (platform) to this new API.

jiparis (approved):

migmartri (approved):

app/controlplane/pkg/authz/authz.go
Piskoo: @jiparis Seems like platform has it's own type defined, so no changes needed there

PR #3024, opened 2026-04-13
feat(cache): add configurable JetStream KV replica count

jiparis:
## Summary

- Add `WithReplicas` option to `pkg/cache` so NATS JetStream KV buckets can be created with a configurable replica count. Max 3 replicas
- Thread replica count from Helm values → proto config → `natsconn.Config` → `ReloadableConnection` → attestationbundle/policyevalbundle caches
- Fixes inconsistent reads in clustered NATS deployments where KV buckets defaulted to 1 replica

Follow-up of PFM-5440.

### DIFF
--- a/app/controlplane/cmd/main.go
+++ b/app/controlplane/cmd/main.go
@@ -221,9 +221,17 @@ func newNatsConfig(c *conf.Bootstrap_NatsServer) *natsconn.Config {
 		return nil
 	}
 
+	replicas := int(c.GetReplicas())
+	if replicas < 1 {
+		replicas = 1
+	} else if replicas > 3 {
+		replicas = 3
+	}
+
 	cfg := &natsconn.Config{
-		URI:  uri,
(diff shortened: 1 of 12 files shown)


app/controlplane/cmd/main.go
```diff
@@ -221,9 +221,15 @@ func newNatsConfig(c *conf.Bootstrap_NatsServer) *natsconn.Config {
 		return nil
 	}
 
+	replicas := int(c.GetReplicas())
+	if replicas < 0 {
+		replicas = 0
```
javirln: Is 0 replica a valid replica o should it be 1?

javirln: Ok min can be 0 but, it's actually 1: https://github.com/nats-io/nats-server/blob/71d4264b90e0e064bc3389b835c058563270a07c/server/stream.go#L1644-L1652

migmartri (approved):

PR #3030, opened 2026-04-14
feat(materials): add binary-safe inline material support

### LINKED ISSUE #2065: Error decoding attestations with INLINE binary data

migmartri:
## Problem

When binary data is stored inline (no CAS backend configured), `wf run describe` crashes with:

```
grpc: error while marshaling: string field contains invalid UTF-8
```

The root cause: binary file content flows through `NormalizedMaterial.Value` (Go `string`) and into `AttestationItem.Material.value` (proto `string`). Protobuf 3 requires `string` fields to be valid UTF-8, so gRPC rejects the response. The write path is correct — `Artifact.content` in the crafting state proto is `bytes`, and the in-toto `ResourceDescriptor.content` is also `bytes`. The corruption only happens on the read/response path when `normalizeMaterial()` casts `[]byte` to `string`.

## Approach

Add a new `bytes raw_value` field to the API response proto alongside the existing (now deprecated) `string value`. Internally, change `NormalizedMaterial.Value` from `string` to `[]byte` so binary content flows cleanly through the pipeline. Dual-populate at the API boundary for backward compatibility.

## Changes

- Add `bytes raw_value` field to `AttestationItem.Material` proto, mark `string value` as deprecated
- Change internal `NormalizedMaterial.Value` from `string` to `[]byte`
- Dual-populate at API boundary: `raw_value` always set, deprecated `value` set only for valid UTF-8
- CLI prefers `raw_value` with fallback to `value` for backward compat with older servers
- Plugin fanout proto keeps `string value` with UTF-8 guard — binary content available via `Content` bytes field

## Testing

- Binary round-trip test (`TestBinaryContentStructRoundTrip`) verifies data survives the `structpb.Struct` serialization path (json.Marshal -> protojson.Unmarshal -> protojson.Marshal -> json.Unmarshal) with null bytes, ELF headers, all 256 byte values, and 64KB payloads
- Unit test for `extractMaterials` confirms binary content populates `raw_value` while `value` stays empty
- Manual verification with real attestation: a 293KB Helm chart `.tgz` (gzip binary) round-trips intact via `raw_value`, while a 348KB CycloneDX JSON SBOM populates both fields correctly

## Compatibility

| Scenario | Behavior |
|----------|----------|
| Old CLI + New Server | Reads `value` (deprecated) — works for UTF-8 content, empty for binary (no crash) |
| New CLI + New Server | Reads `raw_value` — works for all content |
| New CLI + Old Server | Falls back to `value` — same as today |
| Existing attestations | Read path fixed, no migration needed |

Closes #2065

### DIFF
--- a/app/cli/pkg/action/workflow_run_describe.go
+++ b/app/cli/pkg/action/workflow_run_describe.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024-2025 The Chainloop Authors.
+// Copyright 2024-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -312,9 +312,19 @@ func policyEvaluationPBToAction(in *pb.PolicyEvaluation) *PolicyEvaluation {
 }
 
 func materialPBToAction(in *pb.AttestationItem_Material) *Material {
+	// Prefer raw_value (binary-safe) when available.
+	// Fall back to deprecated string value field for compatibility with
+	// older control plane versions that don't populate raw_value.
+	var value string
+	if len(in.GetRawValue()) > 0 {
+		value = string(in.GetRawValue())
+	} else {
+		value = in.GetValue() //nolint:staticcheck // fallback for older servers
+	}
+
 	m := &Material{
 		Name:           in.Name,
-		Value:          in.Value,
+		Value:          value,
 		Type:           in.Type,
 		Hash:           in.Hash,
 		Tag:            in.Tag,
--- a/app/cli/pkg/action/workflow_run_describe.go
+++ b/app/cli/pkg/action/workflow_run_describe.go
@@ -318,18 +318,16 @@ func materialPBToAction(in *pb.AttestationItem_Material) *Material {
 	// Fall back to deprecated string value field for compatibility with
 	// older control plane versions that don't populate raw_value.
 	var value string
-	if len(in.GetRawValue()) > 0 {
-		value = string(in.GetRawValue())
-		if !utf8.Valid(in.GetRawValue()) {
-			value = ""
+	if raw := in.GetRawValue(); len(raw) > 0 {
+		if utf8.Valid(raw) {
+			value = string(raw)
 		}
 	} else {
 		value = in.GetValue() //nolint:staticcheck // fallback for older servers
 	}
 
 	m := &Material{
-		Name: in.Name,
-		// kept for compatibility in case we have users that are still using the string value field
+		Name:           in.Name,
 		Value:          value,
 		RawValue:       in.GetRawValue(),
 		Type:           in.Type,
--- a/app/cli/pkg/action/workflow_run_describe.go
+++ b/app/cli/pkg/action/workflow_run_describe.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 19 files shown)


app/cli/pkg/action/workflow_run_describe.go
```diff
@@ -312,9 +312,19 @@ func policyEvaluationPBToAction(in *pb.PolicyEvaluation) *PolicyEvaluation {
 }
 
 func materialPBToAction(in *pb.AttestationItem_Material) *Material {
+	// Prefer raw_value (binary-safe) when available.
+	// Fall back to deprecated string value field for compatibility with
+	// older control plane versions that don't populate raw_value.
+	var value string
+	if len(in.GetRawValue()) > 0 {
+		value = string(in.GetRawValue())
```
jiparis: What if RawValue doesn't contain UTF8 content? This line will panic. I think you'll need to validate here as well with `utf8.Valid`

jiparis (approved):

PR #3032, opened 2026-04-14
fix(crafter): handle invalid merge branch config during repo open

### LINKED ISSUE #3031: error getting git commit hash: opening repository: branch config: invalid merge
```
chainloop att init --workflow test-crash --project b --replace
WRN User-attended mode detected. This is intended for local testing only. For CI/CD or automated workflows, please use an API token.
This command will run against the organization "test-crash"
y
ERR initializing crafting state: getting git commit hash: opening repository: branch config: invalid merge
```

I believe this started since we upgraded Go-Git.

migmartri:
## Summary

- Upgrade go-git from v5.17.1 to v5.17.2
- Gracefully handle `branch config: invalid merge` errors from go-git when opening a repository, not just when listing remotes. This is a protective measure to prevent the crafter from crashing on repos with invalid merge references in their git config, similar to the existing protection we already had for the remote-listing path
- Extract `isBranchInvalidMergeError` helper to deduplicate the string-based error check

Closes #3031

### DIFF
--- a/pkg/attestation/crafter/crafter.go
+++ b/pkg/attestation/crafter/crafter.go
@@ -310,7 +310,12 @@ var errBranchInvalidMerge = errors.New("branch config: invalid merge")
 
 // Returns the current directory git commit hash if possible
 // If we are not in a git repo it will return an empty string
-func gracefulGitRepoHead(path string) (*HeadCommit, error) {
+func gracefulGitRepoHead(path string, logger *zerolog.Logger) (*HeadCommit, error) {
+	if logger == nil {
+		l := zerolog.Nop()
+		logger = &l
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/attestation/crafter/crafter.go
```diff
@@ -387,6 +399,10 @@ func gracefulGitRepoHead(path string) (*HeadCommit, error) {
 	return c, nil
 }
 
+func isBranchInvalidMergeError(err error) bool {
```
jiparis, line 405: Please add a comment to state that this is to mitigate https://github.com/go-git/go-git/issues/331.
The current 6 alpha version doesn't return this error anymore, so this is likely to be invalid when we upgrade.

jiparis (approved):

PR #3035, opened 2026-04-14
feat: add --latest-version flag to attestation init

### LINKED ISSUE #3034: AI_SESSION attestations should always target latest project version
**Problem**

AI_SESSION (trace) attestations use `projectVersion` from `.chainloop.yml`, which can reference a closed or removed version. Attestations should always go to the latest version, letting release management control which version that is.

**Proposed solution**

The client sends `"latest"` as a magic constant in the `projectVersion` field during `attestation init`. The control plane resolves it to the actual latest version (the one with `latest=true` in the DB). If no version exists, the server returns an error.

jiparis:
## Summary

- Add `--latest-version` boolean flag to `attestation init`, mutually exclusive with `--version`
- Add `use_latest_version` field to the `AttestationServiceInitRequest` proto
- When set, the control plane resolves to the project version with `latest=true` in the database
- Returns a validation error if no project version exists for the project
- Skip loading version from `.chainloop.yml` when `--latest-version` is used

Closes #3034

### DIFF
--- a/app/controlplane/pkg/biz/biz.go
+++ b/app/controlplane/pkg/biz/biz.go
@@ -65,10 +65,6 @@ var ProviderSet = wire.NewSet(
 	wire.Struct(new(NewUserUseCaseParams), "*"),
 )
 
-// LatestVersionMagicConstant is a reserved version identifier that resolves to
-// the project version with latest=true in the database.
-const LatestVersionMagicConstant = "latest"
-
 var (
 	// versionRegexp allows alphanumeric, dots, hyphens, underscores, plus signs, and build metadata
 	versionRegexp = regexp.MustCompile(`^[a-zA-Z0-9.\-_+]+(?:\+[a-zA-Z0-9.\-_]+)?$`)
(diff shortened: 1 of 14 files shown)


app/controlplane/pkg/biz/biz.go
```diff
@@ -65,6 +65,10 @@ var ProviderSet = wire.NewSet(
 	wire.Struct(new(NewUserUseCaseParams), "*"),
 )
 
+// LatestVersionMagicConstant is a reserved version identifier that resolves to
+// the project version with latest=true in the database.
+const LatestVersionMagicConstant = "latest"
```
migmartri: we need to be careful if `latest` exists already, what would happen?

migmartri (approved):

app/controlplane/pkg/biz/biz.go
jiparis: I was thinking that, maybe we should add a new flag like `--latest-version` instead of reusing the `--version` flag.

jiparis: I'll actually do this. I think it's safer and cleaner.

PR #3040, opened 2026-04-16
feat: add billing account resource type and roles to membership model

Piskoo:
This PR extends the membership model to support billing accounts as a resource type. Adds Owner and Admin roles for billing account access control.


app/controlplane/pkg/authz/authz.go
```diff
@@ -106,6 +106,10 @@ const (
 	RoleProductViewer Role = "role:product:viewer"
 	RoleProductAdmin  Role = "role:product:admin"
 
+	// BillingAccount roles
+	RoleBillingAccountOwner Role = "role:billing_account:owner"
```
migmartri, line 110: is this needed here? in this codebase? cc/ @jiparis

Piskoo: Same as `RoleProductAdmin`/`RoleProductViewer` - defined here because the membership schema lives here

Piskoo:
Closing  as discussed

PR #3048, opened 2026-04-16
feat(grpc): add configurable maximum size limit for received gRPC messages

corang:
Adds a configurable `max_recv_msg_size` option to the controlplane gRPC server, allowing operators to increase the default 4MB message size limit. This resolves `ResourceExhausted` errors when pushing attestations larger than 4MB (e.g., attestations with large SBOMs or SARIF reports).

The setting is exposed end-to-end:
- Proto config field on `Server.GRPC`
- Applied as a gRPC server option in the controlplane
- Configurable via Helm with `controlplane.grpcMaxRecvMsgSize`

When unset, the standard gRPC default of 4MB applies.

Example Helm usage:
```yaml
controlplane:
  grpcMaxRecvMsgSize: 16777216  # 16MB
```

### DIFF
--- a/app/controlplane/internal/conf/controlplane/config/v1/conf.pb.go
+++ b/app/controlplane/internal/conf/controlplane/config/v1/conf.pb.go
@@ -1264,13 +1264,16 @@ func (x *Server_TLS) GetPrivateKey() string {
 }
 
 type Server_GRPC struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	Network       string                 `protobuf:"bytes,1,opt,name=network,proto3" json:"network,omitempty"`
-	Addr          string                 `protobuf:"bytes,2,opt,name=addr,proto3" json:"addr,omitempty"`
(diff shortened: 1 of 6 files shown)


migmartri (approved):
LGTM, please fix the conflict and let me know.

BTW, while testing your PR we noticed some mayor improvements in the attestation https://github.com/chainloop-dev/chainloop/pull/3056 push payload that will reduce the size of that payload ~60%. What API calls were causing you the error?

corang:
> LGTM, please fix the conflict and let me know.
> 
> BTW, while testing your PR we noticed some mayor improvements in the attestation #3056 push payload that will reduce the size of that payload ~60%. What API calls were causing you the error?

It was just using the `chainloop attestation push` command but there are a lot of pieces of evidence and an entire other attestation attached to it so I think it's just gonna be large regardless. Fixed that conflict too!

migmartri (approved):

migmartri:
@philiposs could you please backport this fix to chainloop-ee chart? thanks!

PR #3049, opened 2026-04-16
fix(cache): retry bucket init on transient NATS disconnect

migmartri:
## Summary

- When NATS dropped mid-init, `initBucket` surfaced the error immediately and the control plane panicked on startup. The NATS client already retries connections in the background (`MaxReconnects=-1`, `ReconnectWait=2s`), so the cache init now retries alongside it for up to 30s before giving up.
- Only connection-level errors are retried (`context.DeadlineExceeded`, `nats.ErrConnectionClosed`, `ErrNoServers`, `ErrTimeout`, `ErrDisconnected`, `ErrConnectionDraining`). Configuration errors — e.g. unsupported replica count — still fail fast.
- `UpdateStream` on the backing KV stream is now skipped when `Discard` is already `DiscardOld` (idempotent on redeploys), and has an explicit 5s context timeout.
- `panic(err)` in `app/controlplane/cmd/main.go` is replaced with a logged exit so a genuinely unavailable NATS fails startup with a readable error message instead of a stack trace. Deferred cleanups (sentry flush, plugin cleanup, context cancel) are invoked explicitly before exit.

### DIFF
--- a/app/controlplane/cmd/main.go
+++ b/app/controlplane/cmd/main.go
@@ -146,7 +146,12 @@ func main() {
 
 	app, cleanup, err := wireApp(ctx, &bc, credsWriter, logger, availablePlugins)
 	if err != nil {
-		panic(err)
+		_ = logger.Log(log.LevelError, "msg", "failed to initialize control plane", "error", err.Error())
+		// Invoke critical deferred cleanups explicitly since os.Exit skips defers.
+		cancel()
+		availablePlugins.Cleanup()
+		flush()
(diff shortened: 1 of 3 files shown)


migmartri:
Valid catch — `info.Created` is creation-time metadata and never changes on `UpdateStream`, so the assertion was a no-op. Replaced the test in baaf3969: now exercises `ensureDiscardOld` directly on both branches and asserts the skip path via `nc.Stats().OutMsgs` delta (confirmed failing with delta=2 when skip is disabled, passing with delta=1 when enabled).

jiparis (approved):

PR #3056, opened 2026-04-17
refactor(api): remove deprecated attestation and bundle fields from Store

### LINKED ISSUE #3055: CLI attestation push sends the payload three times in a single Store request
## Summary

During attestation push, the CLI sends three representations of the same attestation data in a single `AttestationService.Store` gRPC request:

- `Attestation` — DSSE envelope (JSON)
- `Bundle` — full Sigstore bundle (protojson, contains the DSSE envelope inside)
- `AttestationBundle` — the same Sigstore bundle with the base64 signature fix applied

See `app/cli/pkg/action/attestation_push.go` (`pushToControlPlane`).

## How the server uses each field

Only one of them is actually consumed. The rest are fallbacks for older peers:

- **Service layer** (`app/controlplane/internal/service/attestation.go`): prefers `AttestationBundle`, falls back to `Bundle` only if it is nil.
- **Biz layer** (`app/controlplane/pkg/biz/workflowrun.go`): uses the bundle; only falls back to the envelope when no bundle is provided.
- **Data layer** (`app/controlplane/pkg/data/workflowrun.go`) has an explicit comment: *"the envelope will come empty in normal attestations, since bundles are stored separately. But old CLIs might still send the envelope instead of the bundle... DEPRECATED behaviour that will be removed eventually."*

In other words, when a modern CLI talks to a modern server, the envelope and the legacy `Bundle` field are entirely ignored.

## Deprecation timeline

The two redundant fields are already marked `[deprecated = true]` in `app/controlplane/api/controlplane/v1/workflow_run.proto`, and the replacement has been available for a long time:

| Date | Commit | Event |
|---|---|---|
| 2025-02-03 | `38e140d7` | `attestation_bundle` field added |
| 2025-03-05 | `b83223a8` | `bundle` annotated as carrying the #1832 base64 bug, `attestation_bundle` established as the fix |
| 2025-06-18 | `14c9c944` | `[deprecated = true]` markers added to both `attestation` and `bundle` (PR #2116) |

`attestation_bundle` has existed for ~14 months and the deprecation markers have been in place for ~10 months, so any active control plane today should accept `attestation_bundle`.

## Impact

The request size is roughly tripled for no benefit, because the same attestation data is serialized three times (with protojson base64-encoding the `bytes` fields in the bundle, adding ~33% overhead on top). Observed concretely: a ~6 MB logical attestation produces a ~25 MB gRPC request, which now exceeds the server's 16 MB `max_recv_msg_size` limit introduced in 561ae62c.

## How this was found

Surfaced while debugging a `ResourceExhausted: received message larger than max (25509399 vs. 16777216)` error from the CLI. Investigation traced the size inflation back to the triple-send in `pushToControlPlane`, rather than to a genuinely large payload.

## Capability discovery

Nothing in `AttestationServiceInitResponse` currently signals which of the three fields the server actually consumes; the only version information exposed to the CLI is `StatusService.Infoz.version` (the control plane semver), which the CLI already fetches in `auth_login`, `version`, and `action.go`.

## Suggested direction

- **Short term:** stop sending the legacy `Attestation` and `Bundle` fields from the CLI. Given the 14-month / 10-month rollout of `attestation_bundle` and its deprecation markers, this is low-risk.
- **Optional, cleaner contract:** expose a capability on `AttestationServiceInitResponse` (e.g. `bool attestation_bundle_is_authoritative` or a `repeated string supported_store_fields`) so the CLI doesn't have to bake server-version knowledge in.
- **Follow-up:** remove the deprecated fields from the proto entirely once older CLIs are no longer supported (the `TODO` in `pkg/attestation/attestations.go` near `DSSEEnvelopeFromRaw` already flags this).

migmartri:
## Summary

- Remove the deprecated `attestation` and `bundle` fields from `AttestationServiceStoreRequest`, keeping only `attestation_bundle`.
- Drop the envelope/legacy-bundle fallback paths in the CLI, controlplane service/biz/data layers, and the `pkg/attestation` helper.
- The CLI now sends a single serialized bundle per `Store` RPC instead of three, roughly tripling the maximum attestation size that fits under the server's 16 MB limit.

Closes #3055

### DIFF
--- a/app/controlplane/pkg/biz/workflowrun.go
+++ b/app/controlplane/pkg/biz/workflowrun.go
@@ -93,7 +93,7 @@ type WorkflowRunRepo interface {
 	FindByAttestationDigest(ctx context.Context, digest string) (*WorkflowRun, error)
 	FindByIDInOrg(ctx context.Context, orgID, ID uuid.UUID) (*WorkflowRun, error)
 	MarkAsFinished(ctx context.Context, ID uuid.UUID, status WorkflowRunStatus, reason string) error
-	SaveAttestation(ctx context.Context, ID uuid.UUID, att *dsse.Envelope, digest string) error
+	SaveAttestation(ctx context.Context, ID uuid.UUID, digest string) error
 	SaveBundle(ctx context.Context, ID uuid.UUID, bundle []byte) error
 	GetBundle(ctx context.Context, wrID uuid.UUID) ([]byte, error)
 	UpdatePolicyViolationsStatus(ctx context.Context, ID uuid.UUID, hasPolicyViolations bool) error
@@ -326,25 +326,20 @@ func (uc *WorkflowRunUseCase) MarkAsFinished(ctx context.Context, id string, sta
 	return uc.wfRunRepo.MarkAsFinished(ctx, runID, status, reason)
 }
 
-func (uc *WorkflowRunUseCase) SaveAttestation(ctx context.Context, id string, envelope, bundle []byte) (*v1.Hash, error) {
+func (uc *WorkflowRunUseCase) SaveAttestation(ctx context.Context, id string, bundle []byte) (*v1.Hash, error) {
 	runID, err := uuid.Parse(id)
 	if err != nil {
 		return nil, NewErrInvalidUUID(err)
 	}
 
-	rawContent := bundle
-	if rawContent == nil {
-		rawContent = envelope
-	}
-
 	// calculate the content digest
 	// Todo: this should be calculated in the use case
-	digest, _, err := v1.SHA256(bytes.NewReader(rawContent))
+	digest, _, err := v1.SHA256(bytes.NewReader(bundle))
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate SHA256 of attestation: %w", err)
 	}
 
-	dsseEnv, err := attestation.DSSEEnvelopeFromRaw(bundle, envelope)
+	dsseEnv, err := attestation.DSSEEnvelopeFromBundleBytes(bundle)
 	if err != nil {
 		return nil, fmt.Errorf("extracting DSSE envelope: %w", err)
 	}
@@ -356,7 +351,7 @@ func (uc *WorkflowRunUseCase) SaveAttestation(ctx context.Context, id string, en
 	}
 
 	// verify attestation (only if chainloop is the signer)
-	validation, err := uc.verifyBundle(ctx, rawContent)
+	validation, err := uc.verifyBundle(ctx, bundle)
 	if err != nil {
 		if !errors.Is(err, verifier.ErrInvalidBundle) {
 			return nil, err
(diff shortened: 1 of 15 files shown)


matiasinsaurralde (approved):

app/controlplane/pkg/biz/workflowrun.go
```diff
@@ -326,25 +326,20 @@ func (uc *WorkflowRunUseCase) MarkAsFinished(ctx context.Context, id string, sta
...
+	digest, _, err := v1.SHA256(bytes.NewReader(bundle))
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate SHA256 of attestation: %w", err)
 	}
 
-	dsseEnv, err := attestation.DSSEEnvelopeFromRaw(bundle, envelope)
+	dsseEnv, err := attestation.DSSEEnvelopeFromBundleBytes(bundle)
 	if err != nil {
 		return nil, fmt.Errorf("extracting DSSE envelope: %w", err)
 	}
```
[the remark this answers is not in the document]
migmartri: Good catch — it's pre-existing (the old `DSSEEnvelopeFromRaw` bundle path hit the same `.Signatures[0]` assumption), but this PR narrowed the Store boundary to a single bytes-parsing entry point, so it's the right place to harden. Pushed a guard in `DSSEEnvelopeFromBundleBytes` in a87928bb that returns a typed error when the bundle has no DSSE signatures.

app/controlplane/pkg/biz/workflowrun_integration_test.go
```diff
@@ -489,6 +468,18 @@ func testBundle(t *testing.T, path string) (*v1.Bundle, []byte) {
 	return &bundle, bundleJSON
 }
 
+// testBundleBytesFromEnvelope wraps a DSSE envelope file into a Sigstore bundle and returns its protojson bytes.
+func testBundleBytesFromEnvelope(t *testing.T, path string) []byte {
+	_, envBytes := testEnvelope(t, path)
+	var env dsse.Envelope
+	require.NoError(t, json.Unmarshal(envBytes, &env))
+	b, err := attestation.BundleFromDSSEEnvelope(&env)
```
jiparis, line 476: This method won't be needed anymore, right?

pkg/attestation/attestations.go
```diff
@@ -91,25 +91,20 @@ func BundleFromDSSEEnvelope(dsseEnvelope *dsse.Envelope) (*protobundle.Bundle, e
...
+		return nil, fmt.Errorf("unmarshalling bundle: %w", err)
 	}
-	return &dsseEnv, nil
+	if len(attBundle.GetDsseEnvelope().GetSignatures()) == 0 {
+		return nil, fmt.Errorf("invalid attestation bundle: missing DSSE signature")
+	}
+	return DSSEEnvelopeFromBundle(&attBundle), nil
 }
 
-// TODO: remove this fix once `AttestationServiceStoreRequest.Bundle` is fully removed,
```
jiparis, line 110: What about this? I think we can remove this logic as well already. Or at least leave the comment as it is

PR #3065, opened 2026-04-21
feat(cli): load project from .chainloop.yaml

### LINKED ISSUE #3063: feat: allow to force project and organization in .chainloop.yaml
Currently, .chainloop.yaml is used to enforce the project version attestations should push to. We could extend it to also force the organization and project.

waveywaves:
## Summary

* Allow `chainloop attestation init` to load the project name from `.chainloop.yaml` when `--project` is omitted.
* Preserve precedence: CLI flag over repository configuration.

Fixes chainloop-dev/chainloop#3063

Assisted-by: pi

### DIFF
--- a/app/cli/cmd/attestation_init.go
+++ b/app/cli/cmd/attestation_init.go
@@ -57,19 +57,23 @@ func newAttestationInitCmd() *cobra.Command {
 				return errors.New("workflow name is required, set it via --workflow flag")
 			}
 
-			// load version from the file if not set and not using --latest-version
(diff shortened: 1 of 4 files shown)


waveywaves:
@migmartri @jiparis Ready for review. Extends `.chainloop.yaml` to support a `project` field — same precedence pattern as `projectVersion` (CLI flag > config > default). Organization field left for follow-up due to gRPC connection lifecycle constraints. All commits GPG-signed.

PR #3066, opened 2026-04-21
fix(crafter): use actual PR head commit instead of merge commit in GitHub Actions

### LINKED ISSUE #3064: bug: attestation git.head commit SHA is reporting the merge commit SHA instead of the actual commit
This issue only happens in attestations done from PRs in Github actions. Since Github creates a temporary merge commit for the PR, the attestation reports it as the commit at HEAD, instead of the latest commit in the PR (the actual branch head).

This is a known feature / issue in Github and probably others, well explained here https://www.codegenes.net/blog/get-commit-sha-in-github-actions/.

The result is that attestations end up being referred to ghost commits in the referral graph. This is specially relevant when users also create attestations from agentic workflows: local attestations don't match any PR because their commit SHAs are different (and they shouldn't)

waveywaves:
## Summary

Fixes #3064.

GitHub Actions creates a temporary merge commit for `pull_request` events, so `.git/HEAD` points to that synthetic commit. This causes `attestation init` to record a ghost commit SHA in `git.head` that doesn't exist on any branch — breaking the referral graph when cross-referencing with local or agentic attestations.

### Root cause

`gracefulGitRepoHead()` reads `repo.Head()` directly via go-git, which in a GH Actions PR checkout returns the merge commit. The CI runner already captures `GITHUB_EVENT_NAME` and `GITHUB_EVENT_PATH` but these weren't used to correct the git HEAD.

### Fix

After `gracefulGitRepoHead()` resolves HEAD, check if we're in a GitHub Actions `pull_request` or `pull_request_target` event. If so, read the actual PR head SHA from the event payload (`pull_request.head.sha`) and look up that commit in the local repo instead.

- **New file**: `pkg/attestation/crafter/cioverride.go`
  - `resolveGitHubPRHeadSHA()` — reads event JSON, extracts SHA
  - `resolveCommitByHash()` — opens repo, looks up commit by hash
- **Modified**: `pkg/attestation/crafter/crafter.go` — `initialCraftingState` now calls the override after `gracefulGitRepoHead`
- **Falls back gracefully**: if the override SHA can't be resolved (shallow clone, missing object), keeps the merge commit and logs a warning

### Design choices

- **Separate file** (`cioverride.go`) keeps CI-specific logic out of the general-purpose `crafter.go`
- **Reads env vars directly** rather than going through the runner interface — `GITHUB_EVENT_NAME` and `GITHUB_EVENT_PATH` only exist in GitHub Actions, so a runner-type check would be redundant
- **Re-opens the repo** instead of threading the repo handle from `gracefulGitRepoHead` — the perf cost is negligible for a one-time `att init` operation, and it keeps the function signatures simple

### Future extensibility

The same pattern can be extended for GitLab MR pipelines (`CI_MERGE_REQUEST_SOURCE_BRANCH_SHA`) by adding another resolver in `cioverride.go`.

## Test plan

- [x] `go build ./pkg/attestation/crafter/...` passes
- [x] 6 table-driven subtests for `resolveGitHubPRHeadSHA`: PR event, PR target event, push event, no event, malformed JSON, missing head SHA
- [x] `gofmt -l` clean
- [x] Commits GPG-signed and DCO signed-off

### DIFF
--- a/pkg/attestation/crafter/cioverride.go
+++ b/pkg/attestation/crafter/cioverride.go
@@ -1,0 +1,117 @@
+//
+// Copyright 2026 The Chainloop Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crafter
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/go-git/go-git/v6"
+	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/rs/zerolog"
+)
+
+// resolveGitHubPRHeadSHA returns the actual PR branch head SHA when running
+// in a GitHub Actions pull_request event.
+//
+// GitHub Actions creates a temporary merge commit for PR workflows, so
+// .git/HEAD (and GITHUB_SHA) points to the merge commit instead of the
+// actual PR head. The real SHA is available in the event payload at
+// pull_request.head.sha.
+//
+// Note: pull_request_target is intentionally excluded because it checks out
+// the base branch, not the PR branch — the PR head commit may not be
+// available in the local checkout at all.
+//
+// Returns "" when not in a GitHub Actions PR context, or if the event
+// payload is missing/unreadable.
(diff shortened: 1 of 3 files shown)


waveywaves:
@migmartri @jiparis Ready for review. Fixes the merge commit SHA bug in GitHub Actions PR workflows (#3064). The hash is always overridden from the event payload — works with default `actions/checkout` shallow clones (depth=1). Full commit metadata is resolved best-effort when the object is available locally. All commits GPG-signed.

waveywaves:
@migmartri @jiparis ptal

migmartri (approved):
Thanks!

migmartri:
@waveywaves please fix the commit signatures, thanks

image

migmartri:
> @waveywaves please fix the commit signatures, thanks

Friendly reminder about the above ^^ 

Thanks

PR #3087, opened 2026-05-04
feat(attestation): opt-in CAS-only attestation storage

### LINKED ISSUE #2126: feat: do not store attestation in DB
Currently, attestations are stored both in DB and in the CAS. This causes the DB usage to grow, especially now that we also store policy evaluations, etc in them.

One solution would be to

a) stop storing the attestation in the DB<br>b) update the biz layer so nothing changes from the API POV, just an inevitable response slowdown for which we could<br>c) Add a cache layer, i.e something simple in memory [like](<https://github.com/hashicorp/golang-lru>), [Nats KV with TTL](<https://docs.nats.io/nats-concepts/jetstream/key-value-store>)

In the future we might want to not to couple the `wf run describe` API call to retrieving attestation too, and have another API endpoint/flag to do so, that way the API call can be fast regardless of cache MISS

cc/ @jiparis @javirln

migmartri:
## Summary

Adds a controlplane config flag `attestations.skip_db_storage` that opts the platform out of writing attestation bundles to the database when an external CAS backend is configured.

When the flag is **on** and the workflow run's CAS backend is **non-inline**:
- The bundle is uploaded to CAS synchronously (60s timeout) and CAS becomes the only durable copy.
- Only the attestation digest is recorded on the workflow run; the `attestation` table row is skipped.
- Reads transparently fall back to CAS via the existing layered resolver (cache → DB → CAS).

When the flag is **off** (default) or the CAS backend is **inline**, behaviour is unchanged: the bundle is persisted to the DB and, for non-inline backends, uploaded to CAS asynchronously.

Closes #2126

---

_AI disclosure: this contribution was assisted by Claude Code._

### DIFF
--- a/app/controlplane/internal/service/attestation.go
+++ b/app/controlplane/internal/service/attestation.go
@@ -16,6 +16,7 @@
 package service
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -280,6 +281,25 @@ func (s *AttestationService) Store(ctx context.Context, req *cpAPI.AttestationSe
 	}, nil
 }
 
+// casAttestationUploadMaxElapsedTime caps the total time spent retrying the
+// CAS upload across both the sync (skipDbStorage) and async paths.
+const casAttestationUploadMaxElapsedTime = 1 * time.Minute
+
+// uploadAttestationToCASWithRetry uploads the attestation bundle to CAS with
+// exponential backoff bounded by casAttestationUploadMaxElapsedTime. It is
+// shared by the synchronous (skipDbStorage) and asynchronous paths.
+func (s *AttestationService) uploadAttestationToCASWithRetry(ctx context.Context, bundle []byte, casBackend *biz.CASBackend, workflowRunID string, digest v1.Hash) error {
+	b := backoff.NewExponentialBackOff()
+	b.MaxElapsedTime = casAttestationUploadMaxElapsedTime
+	if err := backoff.Retry(func() error {
+		return s.attestationUseCase.UploadAttestationToCAS(ctx, bundle, casBackend, workflowRunID, digest)
+	}, b); err != nil {
+		return err
+	}
+	s.log.Infow("msg", "attestation uploaded to CAS", "digest", digest, "runID", workflowRunID)
+	return nil
+}
+
 // storeAttestation stores and processes a Sigstore attestation bundle.
 func (s *AttestationService) storeAttestation(ctx context.Context, bundle []byte, robotAccount *usercontext.RobotAccount, wf *biz.Workflow, wfRun *biz.WorkflowRun, markAsReleased *bool) (*v1.Hash, error) {
 	workflowRunID := wfRun.ID.String()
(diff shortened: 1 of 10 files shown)


app/controlplane/pkg/biz/workflowrun_integration_test.go
```diff
@@ -178,6 +178,29 @@ func (s *workflowRunIntegrationTestSuite) TestSaveAttestation() {
 		assert.NoError(err)
 		assert.True(exists)
 	})
+
+	s.T().Run("WithSkipBundlePersistence stores digest only", func(_ *testing.T) {
```
migmartri: s.Run()

app/controlplane/internal/service/attestation.go
```diff
@@ -291,40 +297,66 @@ func (s *AttestationService) storeAttestation(ctx context.Context, bundle []byte
...
+	// Inline backends have no external CAS to fall back to, so the bundle
+	// always stays in the DB regardless of the flag.
+	preferCAS := s.bootstrapConfig.GetAttestations().GetPreferCasStorage() && !casBackend.Inline
 
-	// If we have an external CAS backend, we will push there the attestation
-	if !casBackend.Inline {
-		// Check the validation status of the backend. The backend might be different from the one configured as default
+	var digest *v1.Hash
+	if preferCAS {
 		if casBackend.ValidationStatus != biz.CASBackendValidationOK {
```
migmartri: let's have the validation code only once

app/controlplane/pkg/biz/workflowrun_integration_test.go
migmartri: Done — switched to s.Run in d02551e0

app/controlplane/internal/service/attestation.go
migmartri: Hoisted the CAS validation check above the if/else in d02551e0.

app/controlplane/internal/service/attestation.go
```diff
@@ -291,40 +297,61 @@ func (s *AttestationService) storeAttestation(ctx context.Context, bundle []byte
...
+	var digest *v1.Hash
+	if skipDB {
+		digestHash, _, hashErr := v1.SHA256(bytes.NewReader(bundle))
+		if hashErr != nil {
+			return nil, handleUseCaseErr(hashErr, s.log)
+		}
+
+		uploadCtx, cancel := context.WithTimeout(ctx, casAttestationUploadTimeout)
+		defer cancel()
+		if err = s.attestationUseCase.UploadAttestationToCAS(uploadCtx, bundle, casBackend, workflowRunID, digestHash); err != nil {
```
migmartri: this should also be opart of backoff retry mechanism, maybe the function can be reused

migmartri: Extracted `uploadAttestationToCASWithRetry` and called it from both the sync (skipDbStorage) and async paths. Both now share the same exponential backoff with a 1-minute MaxElapsedTime. Pushed in 1f42cb53.

matiasinsaurralde (approved):

PR #3088, opened 2026-05-04
feat(cli): send CLI version on every request and forward it to policy providers

migmartri:
## Summary

- The CLI attaches a `Chainloop-Cli-Version` header (formatted as `<version>-<edition>`, e.g. `v1.94.2-oss`) on every gRPC request to the Control Plane and CAS via unary/stream client interceptors in `pkg/grpcconn`.
- The Control Plane reads the header and forwards it on outbound HTTP calls to policy providers (`Resolve`, `ResolveGroup`, `ValidateAttachment`).
- Biz-layer methods that reach policy providers now accept a `context.Context`; `ValidateAttachment` adopts `ProviderAuthOpts` for parity with `Resolve`/`ResolveGroup`.


matiasinsaurralde (approved):
LGTM

PR #3092, opened 2026-05-05
feat(cas): add managed flag

Piskoo:
Added a `managed` flag on the CAS backend to mark backends as Chainloop-managed, and blocked users from updating or deleting them.


app/controlplane/internal/service/casbackend.go
```diff
@@ -203,6 +203,7 @@ func bizCASBackendToPb(in *biz.CASBackend) *pb.CASBackendItem {
 		Default:     in.Default,
 		Fallback:    in.Fallback,
 		IsInline:    in.Inline,
+		IsManaged:   in.Managed,
```
jiparis, line 206: I'd say that only users with the INSTANCE_ADMIN role can set this flag to true. what do you think?

migmartri: nodody should do it no? Isn't setting this value out of the scope of the PR?

Piskoo: yes, nobody should have access to that, it should be set automatically by the backend during auto provisioning

migmartri (approved):

javirln (approved):

PR #3093, opened 2026-05-05
feat(policies): add suppressed_findings to policy evaluation result

### LINKED ISSUE #3091: policy engine: add suppressed_findings field on policy result
## Summary

Add a new top-level key `suppressed_findings` to the policy evaluation result, alongside the existing `findings` and `violations`.

```json
{
  "skipped": false,
  "skip_reason": "",
  "violations": ["..."],
  "findings": [
    { "external_id": "CVE-2023-45288", "package_purl": "pkg:golang/...", "severity": "HIGH", ... }
  ],
  "suppressed_findings": [
    {
      "external_id": "CVE-2023-45283",
      "package_purl": "pkg:golang/...",
      "severity": "HIGH",
      "chainloop_finding_id": "fnd_01J...",
      "chainloop_assessment_ids": ["asm_01J...", "asm_01K..."]
    }
  ]
}
```

## Semantics

- Each item in `suppressed_findings` uses the **same Rego object** as items in `findings` (e.g. a `PolicyVulnerabilityFinding`), plus two new optional fields:
  - `chainloop_finding_id` (string) — the Chainloop finding ID this Rego finding maps to.
  - `chainloop_assessment_ids` (array of strings) — the assessment IDs that drove the suppression.
- `findings` keeps containing every emitted finding (suppressed and non-suppressed alike) so the ingestion pipeline keeps tracking them. `suppressed_findings` is the subset that did **not** contribute to violations because the platform-side assessment matched.
- **Invariant:** for every `f ∈ suppressed_findings`, `f ∈ findings`. Ingestion can ignore `suppressed_findings` entirely if it wants.
- Empty array (or absent) when no suppression happened — same convention as `violations`.

## Scope

- Extend `engine.EvaluationResult` with a new `SuppressedFindings` slice.
- Parse `suppressed_findings` from the Rego `result` rule output and from the WASM policy output JSON.
- Add `chainloop_finding_id` and `chainloop_assessment_ids` to the three finding proto types (`PolicyVulnerabilityFinding`, `PolicySASTFinding`, `PolicyLicenseViolationFinding`) as optional fields.
- Update the Rego boilerplate so policies can declare `suppressed_findings` and have it default to `[]` and be wired into the `result` object automatically.
- Tests for parsing in both engines.

Ingestion-side use of the new field (filtering violations, surfacing suppressed in UI) is out of scope for this issue and can be done in follow-ups.

migmartri:
## Summary

- Adds a new top-level `suppressed_findings` key on the policy evaluation result, alongside the existing `findings` and `violations`.
- Items in `suppressed_findings` use the same Rego object schema as items in `findings` (e.g. `PolicyVulnerabilityFinding`), and the invariant is that every entry in `suppressed_findings` also appears in `findings`.
- Adds two new optional correlation fields on the three finding proto types (`PolicyVulnerabilityFinding`, `PolicySASTFinding`, `PolicyLicenseViolationFinding`): `chainloop_finding_id` and `chainloop_assessment_ids`, used by policies to reference the platform-side finding and the assessments that drove the suppression.
- Surfaces the new field through both the Rego and WASM engine result parsers as `EvaluationResult.SuppressedFindings`.

Closes #3091

This change does not yet wire `suppressed_findings` into the ingestion pipeline; downstream consumers can ignore it until follow-up work lands.

Assisted-by: Claude Code

### DIFF
--- a/pkg/policies/engine/rego/rego.go
+++ b/pkg/policies/engine/rego/rego.go
@@ -224,6 +224,22 @@ func parseResultRule(res rego.ResultSet, policy *engine.Policy, rawData *engine.
 					}
 					result.Violations = append(result.Violations, pv)
 				}
+
+				// Parse the optional "suppressed_findings" array. Each entry must be a
+				// structured finding object using the same schema as `findings`.
+				if suppressedRaw, ok := ruleResult["suppressed_findings"].([]any); ok {
+					for _, f := range suppressedRaw {
+						obj, ok := f.(map[string]any)
+						if !ok {
+							return nil, fmt.Errorf("suppressed finding must be an object, got %T", f)
+						}
+						pv, err := engine.NewStructuredViolation(policy.Name, obj)
+						if err != nil {
+							return nil, fmt.Errorf("suppressed finding in policy %q: %w", policy.Name, err)
+						}
+						result.SuppressedFindings = append(result.SuppressedFindings, pv)
+					}
+				}
 			} else if violations, ok := ruleResult["violations"].([]any); ok {
 				// Fallback: violations (strings or deprecated structured objects).
 				// TODO: remove structured object support once policies are fully migrated to findings.
--- a/pkg/policies/engine/rego/rego.go
+++ b/pkg/policies/engine/rego/rego.go
@@ -224,22 +224,6 @@ func parseResultRule(res rego.ResultSet, policy *engine.Policy, rawData *engine.
 					}
 					result.Violations = append(result.Violations, pv)
 				}
-
-				// Parse the optional "suppressed_findings" array. Each entry must be a
-				// structured finding object using the same schema as `findings`.
-				if suppressedRaw, ok := ruleResult["suppressed_findings"].([]any); ok {
-					for _, f := range suppressedRaw {
-						obj, ok := f.(map[string]any)
-						if !ok {
-							return nil, fmt.Errorf("suppressed finding must be an object, got %T", f)
-						}
-						pv, err := engine.NewStructuredViolation(policy.Name, obj)
-						if err != nil {
-							return nil, fmt.Errorf("suppressed finding in policy %q: %w", policy.Name, err)
-						}
-						result.SuppressedFindings = append(result.SuppressedFindings, pv)
-					}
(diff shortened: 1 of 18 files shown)


jiparis (dismissed):

pkg/policies/engine/rego/testfiles/suppressed_findings.rego
```diff
@@ -0,0 +1,58 @@
...
+# the corresponding entry in `findings`, plus the chainloop_* correlation fields.
+# The correlation fields are read with object.get so that omitted fields fall
+# back to safe zero values instead of making the whole entry undefined.
+suppressed_findings contains v if {
+	some vuln in input.vulnerabilities
+	vuln.suppressed == true
+	v := {
+		"message": sprintf("Found vulnerability %s", [vuln.id]),
+		"external_id": vuln.id,
+		"package_purl": vuln.purl,
```
migmartri, line 56: Fixed in bb5af38d — switched to object.get with safe defaults ("" and []) so suppressed findings surface even when correlation fields are absent. Added a test case covering the missing-fields path.

pkg/policies/engine/rego/rego.go
```diff
@@ -224,6 +224,22 @@ func parseResultRule(res rego.ResultSet, policy *engine.Policy, rawData *engine.
 					}
 					result.Violations = append(result.Violations, pv)
 				}
+
+				// Parse the optional "suppressed_findings" array. Each entry must be a
+				// structured finding object using the same schema as `findings`.
+				if suppressedRaw, ok := ruleResult["suppressed_findings"].([]any); ok {
```
migmartri: superssed findinfgs might be added even if ew don't have findings

migmartri: Right — fixed in 0bbe6de4. Reworked the model so suppressed_findings are disjoint from findings (a vulnerability is either a finding or a suppressed finding, never both) and parsing now happens independently of the findings branch, so a policy can emit suppressed_findings without any findings. Also added a new `suppressed_findings` repeated field on the `PolicyEvaluation` proto so they survive into the CAS bundle.

PR #3094, opened 2026-05-05
feat(policies): inject per-evaluation project context for chainloop.* built-ins

### LINKED ISSUE #3090: Engine-injected project context for chainloop.* built-ins
## Context

`chainloop.findings` (and its sibling `chainloop.project_compliance`) require `project_name` and `project_version_name` as request fields. This is incompatible with material-kind policies (vulnerability policies, SAST, etc.), whose `input` is the parsed material with no project metadata anywhere — there is no way for a rego author to populate those fields, so policies fail to type-check at evaluation time:

```
rego_type_error: chainloop.findings: invalid argument(s)
  have: (object<finding_type: array<string>, status: array<string>>, ???)
  want: (request: object<
      finding_type, status, severity: any<string, array[string]>,
      project_name: string,
      project_version_name: string,
   >, response: array[any])
```

As a result, vulnerability policies that should be calling `chainloop.findings` (to suppress assessed findings) cannot, and any future built-in that needs project scope hits the same wall.

> ## Goals

1. Make `chainloop.findings` callable from material-kind policies without the rego author writing project context.
2. Establish a single, reusable convention so future `chainloop.*` built-ins inherit the same behaviour.
3. Keep `chainloop.project_compliance` working unchanged for `check-compliance-requirement` (no migration churn for attestation-kind policies).

## Non-Goals

* Changing how attestation-kind policies receive project context (`input.predicate.metadata.project` keeps working).
* Exposing project context to user policies as readable input fields. Engine-side injection only.
* Designing how server-side rego evaluation gets the same context (tracked separately).

## Proposed Approach

The CLI's policy engine maintains a per-evaluation context with the project name and version it is currently attesting against. When a `chainloop.*` built-in is invoked, the engine fills in `project_name` and `project_version_name` on the request *before* OPA type-checking runs.

Two separate problems to solve:

1. **Type-check time** — OPA validates built-in argument types at compile time. Required fields make the policy fail to compile if absent. Mark `project_name` and `project_version_name` *optional* in the built-in's declared schema.
2. **Runtime** — When the built-in is invoked without project context, the engine fills it in from the per-evaluation context. Author-provided values still win.

### Engine context plumbing

Add per-evaluation project name/version fields to the rego engine options. The attestation crafter populates them at evaluation start (it already knows project + version from the run config). For `chainloop policy develop eval`, populate from new optional `--project` / `--project-version` flags, or leave empty.

Surface the values to built-ins via `bctx.Context` so any built-in (including EE-only) can pull them from the per-evaluation context.

### Built-in schema relaxation + runtime fill-in

Mark `project_name` / `project_version_name` optional in the built-in schemas. In the Go implementation:

```go
if req.ProjectName == "" {
    req.ProjectName = engineCtx.ProjectName
}
if req.ProjectVersionName == "" {
    req.ProjectVersionName = engineCtx.ProjectVersionName
}
if req.ProjectName == "" || req.ProjectVersionName == "" {
    return emptyList, nil // graceful degradation for local dev
}
```

Same change to `chainloop.project_compliance`.

### Local dev experience

`chainloop policy develop eval` gains optional `--project` / `--project-version` flags. When omitted, built-ins return empty results so the rego author can still iterate locally; suppression simply never fires.

### Graceful failure on entitlement / network errors

`connect.CodeFailedPrecondition` / `Unauthenticated` from the platform should not kill the whole policy. Built-ins log and return an empty list rather than erroring.

## Risks

* Schema relaxation could break `check-compliance-requirement` (still passes the fields explicitly) — mitigated because author-provided values win.
* Local dev eval silently produces no suppression and dev assumes feature is broken — mitigate with a debug log line and clear flag documentation.
* Entitlement-error fall-through hiding a real outage — match on specific connect codes only.

## Related

* Spec 051 (Allow-list Vulnerabilities via Risk Assessments) is blocked on this.

migmartri:
## Summary

Adds engine-level project name + version that propagates into the per-evaluation `context.Context` so `chainloop.*` built-ins (e.g. `chainloop.findings`, `chainloop.project_compliance`) can scope their queries automatically without the rego author writing project metadata in the policy.

- New `engine.WithProjectContext(name, version)` option, propagated to built-ins through `bctx.Context` via `builtins.WithProjectContext` / `ProjectContextFromContext` helpers.
- Crafter passes project context to every `NewPolicyVerifier` / `NewPolicyGroupVerifier` call (material verification and attestation-level evaluation) by reading workflow metadata.
- `chainloop policy develop eval` gains `--project` / `--project-version` flags so authors can supply the same context locally.
- Schema relaxation and runtime fill-in for the affected built-ins themselves live in EE and will consume this OSS plumbing in a separate change.

Closes #3090

### DIFF
--- a/app/cli/cmd/policy_develop_eval.go
+++ b/app/cli/cmd/policy_develop_eval.go
@@ -27,12 +27,14 @@ import (
 
 func newPolicyDevelopEvalCmd() *cobra.Command {
 	var (
-		materialPath     string
-		kind             string
-		annotations      []string
-		policyPath       string
-		inputs           []string
-		allowedHostnames []string
+		materialPath       string
+		kind               string
+		annotations        []string
+		policyPath         string
+		inputs             []string
+		allowedHostnames   []string
+		projectName        string
+		projectVersionName string
 	)
 
 	cmd := &cobra.Command{
@@ -41,18 +43,20 @@ func newPolicyDevelopEvalCmd() *cobra.Command {
 		Long: `Perform a full evaluation of the policy against the provided material type.
 The command checks if there is a path in the policy for the specified kind and
 evaluates the policy against the provided material or attestation.`,
-		Example: `  
+		Example: `
   # Evaluate policy against a material file
   chainloop policy develop eval --policy policy.yaml --material sbom.json --kind SBOM_CYCLONEDX_JSON --annotation key1=value1,key2=value2 --input key3=value3`,
 		RunE: func(_ *cobra.Command, _ []string) error {
 			opts := &action.PolicyEvalOpts{
-				MaterialPath:     materialPath,
-				Kind:             kind,
-				Annotations:      parseKeyValue(annotations),
-				PolicyPath:       policyPath,
-				Inputs:           parseKeyValue(inputs),
-				AllowedHostnames: allowedHostnames,
-				Debug:            flagDebug,
+				MaterialPath:       materialPath,
+				Kind:               kind,
+				Annotations:        parseKeyValue(annotations),
+				PolicyPath:         policyPath,
+				Inputs:             parseKeyValue(inputs),
+				AllowedHostnames:   allowedHostnames,
+				Debug:              flagDebug,
+				ProjectName:        projectName,
+				ProjectVersionName: projectVersionName,
 			}
 
 			policyEval, err := action.NewPolicyEval(opts, ActionOpts)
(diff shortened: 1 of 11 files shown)


jiparis (approved):

jiparis:
awesome! I managed something similar just for ATTESTATION materials (since they bring the project already). This is smarter.

migmartri:
> awesome! I managed something similar just for ATTESTATION materials (since they bring the project already). This is smarter.

Yep I know, I thought it was possible but then I realized is only available for attestation-type policies :)

PR #3107, opened 2026-05-12
feat(audit): emit OrgUserRemoved when admins remove members

Piskoo:
Added a new `OrgUserRemoved` audit event and dispatched from `MembershipUseCase.DeleteOther`.


jiparis (approved):

jiparis:
thanks!

PR #3116, opened 2026-05-14
feat(controlplane): API token workflow scope, attestation enforcement, system flag

### LINKED ISSUE #3115: API token: workflow scope, attestation enforcement, internal system tokens
## Summary

Extend the API token model with three related, internal capabilities. None of these change the public proto / service API.

### 1. Workflow scope on API tokens

API tokens can already be scoped to an organization and, optionally, a project. Today the `apitoken` row persists `project_id` and the minted JWT carries `project_id`/`project_name` claims.

Extend this one level deeper: a token may optionally be scoped to a single workflow within a project. The workflow scope is persisted on the row and embedded in the JWT as `workflow_id`/`workflow_name` claims, mirroring the project pattern. Expressed in biz through a new `APITokenWithWorkflow(*Workflow)` functional option on `APITokenUseCase.Create`, with validation that the workflow belongs to the requested project.

### 2. Attestation-process enforcement for workflow-scoped tokens

When an attestation call (`AttestationService.Init`, `Store`, `Cancel`, `GetUploadCreds`, `GetContract`, attestation state) is authenticated with a workflow-scoped API token, the workflow resolved from the request must equal the workflow encoded in the token. Mismatches are rejected.

Two layers of enforcement:

- Auth middleware verifies the JWT `workflow_id` claim matches the row's `workflow_id` (mirrors the existing project mismatch check).
- A check in `findWorkflowFromTokenOrNameOrRunID` — the single chokepoint every attestation endpoint goes through — rejects when the resolved workflow differs from the token's workflow.

`FindOrCreateWorkflow` additionally forbids workflow-scoped tokens from creating or looking up other workflows.

### 3. Internal "system" tokens hidden from the API

Add an immutable `is_system` flag on the `apitoken` row. System tokens are minted by internal code paths (a new `APITokenAsSystem()` biz option) and continue to mint and validate JWTs normally, but are hidden from the public API:

- `APITokenService/List` excludes them by default; opt-in via a new biz-level `WithIncludeSystemTokens()` filter.
- Service-layer handlers that look up tokens by id/name (e.g. `Revoke`) return `NotFound` when the row is a system token.

The mint path is not exposed in the proto; only internal callers can create system tokens.

## Scope

All changes are internal: builders, persistence layer, auth middleware, and the attestation service chokepoint. No proto, no service-level API change.

migmartri:
Closes #3115

## Summary

Extends the API token model with three related internal capabilities. No public proto / service API surface changes.

- **Workflow scope.** API tokens can already carry an org and optional project scope (`project_id` column + claim). This adds an analogous, optional workflow scope persisted on the `apitoken` row and embedded in the JWT as `workflow_id` / `workflow_name` claims. In biz it surfaces as a new `APITokenWithWorkflow(*Workflow)` functional option on `APITokenUseCase.Create`, validated against the requested project.

- **Attestation enforcement.** The auth middleware verifies the `workflow_id` JWT claim matches the DB row (mirroring the existing project mismatch check). The single chokepoint `findWorkflowFromTokenOrNameOrRunID` rejects when a workflow-scoped token operates on a different workflow; every attestation endpoint (`Init`, `Store`, `Cancel`, `GetUploadCreds`, `GetContract`, attestation state) inherits this. `FindOrCreateWorkflow` additionally forbids workflow-scoped tokens from creating or looking up other workflows.

- **System tokens.** A new immutable `is_system` flag on the row. System tokens mint and validate JWTs normally but are hidden from `APITokenService/List` by default, and `Revoke` against a system token returns `NotFound`. List opt-in via biz `WithIncludeSystemTokens()`; mint via `APITokenAsSystem()`. Public proto does not expose `isSystem` — internal code paths only.


migmartri:
Addressed CI + cubic review:

- `go mod tidy` failure (root cause for all 4 failing CI test jobs): cleaned `go.sum` of unused entries leftover from the rebase resolution.
- P1 panic risk in `jwt/apitoken.GenerateJWT`: added an explicit error when `WorkflowID` is set without `WorkflowName`.
- P1 `RegenerateJWT` claim drop: project and workflow scopes from the persisted row are now re-emitted on the regenerated JWT.
- P2 middleware tests: workflow-mismatch cases now assert the error message contains `workflow mismatch`, so a regression that silently drops the check would surface as a wrong-error test failure rather than passing for the wrong reason.
- P2 system-token list test: now asserts the visible token by ID and that `IsSystem` is false, and that the opt-in path returns both tokens by ID — not just by count.

jiparis (approved):

PR #3121, opened 2026-05-15
feat(blobmanager): add managed CAS backend via S3 Access Points

### LINKED ISSUE #3114: feature: support S3 Access points as CAS storage backend
This is a special case for on-prem installations, where a single bucket is shared across different organizations. By using S3 access points, there is some degree of isolation by using AP ARNs (https://aws.amazon.com/s3/features/access-points/). The solution must include the proper checks to avoid cross-organization access and collisions.

jiparis:
## Summary

- Introduces a new `AWS-S3-ACCESS-POINT` CAS backend that targets a single shared bucket via per-tenant S3 Access Points. Each request mints a scoped session via `sts:AssumeRole` with a session policy and `RoleSessionName` derived from the authenticated requesting org carried in `ctx`.
- All AWS coordinates (`AccessPointARN`, `Region`, `BaseRoleARN`) live in the per-tenant secret blob; there is no deployment-level config block. The provider is registered unconditionally, so on-prem deployments without managed CAS are unaffected — they simply never have managed rows.
- Per-tenant key prefix and STS session name both come from the authenticated requesting org (not the secret blob), so a secrets-store rewrite alone cannot reroute uploads to another tenant's AP. The AP resource policy enforces this server-side via `aws:userid`.
- Carries the requesting org through the CAS robotaccount JWT (`org-id` claim). An auth middleware in `artifact-cas` reads the claim and enriches the request `ctx` via `blobmanager.WithRequestingOrg` before any service handler runs, so callers cannot accidentally skip the binding. Non-managed providers ignore the key.
- Dev-mode bypass (skip `sts:AssumeRole`, use the SDK's default credential chain) is opt-in via the `CHAINLOOP_S3_ACCESS_POINT_DEV_MODE` env var.
- For `Managed=true` rows, redacts AWS implementation details from any wire output: the AP ARN (`Location`) becomes `"managed by Chainloop"` and the provider ID (`Provider`) becomes `"Chainloop"` in both API responses and audit-event payloads. The DB and biz layer keep the real values.
- The user-facing `CASBackend.Create` RPC rejects the managed provider; managed rows are provisioned by the platform-side reconciler.

## AI Assistance

This change was developed with Claude Code; per-commit `Assisted-by:` trailers record the specific commits.

Closes https://github.com/chainloop-dev/chainloop/issues/3114


app/artifact-cas/internal/conf/conf.proto
```diff
@@ -79,6 +83,23 @@ message Server {
   HTTP http_metrics = 3;
 }
 
+// BlobBackends mirrors the controlplane's `BlobBackends` block. Defined
+// independently here so the artifact-cas binary doesn't depend on the
+// controlplane's protobuf package. Keep field numbering in sync across
+// both definitions.
+message BlobBackends {
```
migmartri: I'd call it ManagedCASBackends

app/artifact-cas/internal/service/service.go
```diff
@@ -58,6 +63,28 @@ func (s *commonService) loadBackend(ctx context.Context, providerType, secretID
...
+// The OrgID claim is empty for legacy tokens minted before this PR; in
+// that case WithRequestingOrg sets an empty value and only managed
+// providers (which fail closed) will reject the request — every existing
+// provider ignores the key entirely, so non-managed flows stay unchanged.
+//
+// Returns the loaded backend together with the enriched context so the
+// caller can reuse it for the subsequent Upload/Download calls. Callers
+// MUST use the returned context, not the original one.
+func (s *commonService) loadBackendForClaims(ctx context.Context, claims *casJWT.Claims) (context.Context, backend.UploaderDownloader, error) {
+	ctx = s3accesspoint.WithRequestingOrg(ctx, claims.OrgID)
```
migmartri: this should probably be a middleware instead, it doesn't seem explicitly related to loading the backend no?

migmartri:
@jiparis how does this work? Does it create a backend DB entry automatically if the managed setup is configured in the instance in a similar way we do it with inline?

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
```diff
@@ -126,6 +126,56 @@ message Bootstrap {
...
+  // so on-prem deployments without managed CAS are unaffected.
+  ManagedCASBackends managed_cas_backends = 22;
+}
+
+// ManagedCASBackends groups the additive, deployment-level config
+// blocks for the storage providers that back Chainloop-managed CAS
+// backends. New managed providers append a nested message rather
+// than adding top-level fields to Bootstrap, so the surface stays
+// organised as more backends are added.
+message ManagedCASBackends {
```
migmartri: you might want to add bool enabled = 2;

app/artifact-cas/internal/conf/conf.proto
```diff
@@ -79,6 +83,23 @@ message Server {
   HTTP http_metrics = 3;
 }
 
+// ManagedCASBackends mirrors the controlplane's `ManagedCASBackends`
+// block. Defined independently here so the artifact-cas binary doesn't
+// depend on the controlplane's protobuf package. Keep field numbering
+// in sync across both definitions.
+message ManagedCASBackends {
```
migmartri: why do we need to configure this in CAS?

jiparis: I've removed the configuration. role and AP ARNs are placed now in the secret.

app/controlplane/internal/conf/controlplane/config/v1/conf.proto
jiparis: There is no config anymore.

jiparis:
> @jiparis how does this work? Does it create a backend DB entry automatically if the managed setup is configured in the instance in a similar way we do it with inline?

Row creation in the database is out of scope for this PR. Here you can find just the provider implementation for S3 accesspoints. In practice, it will expect an AP and an AP resource policy generated for every organization. The creation of those resources are not for this PR since here we are not yet provisioning backends.

pkg/blobmanager/loader/loader.go
```diff
@@ -20,18 +20,31 @@ import (
...
+// registered unconditionally — the s3accesspoint provider has no
+// deployment-level config of its own (everything per-tenant lives in the
+// secret blob), so on-prem deployments without managed CAS simply never
+// have managed rows and the provider is dormant.
 func LoadProviders(creader credentials.Reader) backends.Providers {
-	// Initialize CAS backend providers
 	ociProvider := oci.NewBackendProvider(creader)
 	azureBlobProvider := azureblob.NewBackendProvider(creader)
 	s3Provider := s3.NewBackendProvider(creader)
+	apProvider, err := s3accesspoint.NewBackendProvider(creader)
```
migmartri: why is this one different? Is it mandatory? Why a panic?

pkg/blobmanager/s3accesspoint/backend.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+	return nil
+}
+
+func (b *Backend) Describe(ctx context.Context, digest string) (*pb.CASResource, error) {
+	key, err := b.keyFor(ctx, digest)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := b.s3Client.HeadObject(ctx, &s3.HeadObjectInput{
```
migmartri, line 165: would it be possible to reuse the s3 one?

internal/robotaccount/cas/robotaccount.go
```diff
@@ -103,7 +110,12 @@ func NewBuilder(opts ...NewOpt) (*Builder, error) {
 	return b, nil
 }
 
-func (ra *Builder) GenerateJWT(backendType, secretID, audience string, role Role, maxBytes int64) (string, error) {
+// GenerateJWT mints a CAS token. orgID is required for tokens that will
+// touch managed providers (e.g. AWS-S3-ACCESS-POINT) and otherwise
+// optional — pass "" if the targeted backend doesn't need per-tenant
+// attribution. The token always carries the CAS audience and a short
+// expiry window.
+func (ra *Builder) GenerateJWT(backendType, secretID, audience string, role Role, maxBytes int64, orgID string) (string, error) {
```
migmartri, line 119: why is orgID optional?

migmartri (dismissed):

internal/robotaccount/cas/robotaccount.go
jiparis: I guess it can be mandatory now.

pkg/blobmanager/s3accesspoint/backend.go
jiparis: No, because here we are pointing to the Access point ARN, using a specific session and credentials.

migmartri (approved):

PR #3142, opened 2026-05-21
perf(crafter): avoid redundant SHA256 pass on every CAS upload

matiasinsaurralde:
## Summary

When uploading a material to a CAS backend, the file's SHA256 was being computed twice and the file opened three times:

1. `fileStats` opens the file, computes SHA256, rewinds — producing an `*os.File` that the caller holds.
2. `uploadAndCraft` then called `Uploader.UploadFile`, which re-opened the file, recomputed SHA256, rewound, then streamed it.

The reader from step 1 was already in the right state to be streamed. This change calls `Uploader.Upload` with that reader and the digest already produced by `fileStats`, eliminating the redundant open + hash pass while leaving the CAS protocol and on-the-wire bytes unchanged.

Per material, local work drops from `3 reads + 2 SHA256 passes` to `2 reads + 1 SHA256 pass`. Savings scale linearly with material size.

Also closes a file-handle leak in `fileStats` when SHA256 or Seek returns an error.

`Uploader.UploadFile` is retained — the CLI's `artifact upload` command still uses it.

### DIFF
--- a/pkg/attestation/crafter/crafter_test.go
+++ b/pkg/attestation/crafter/crafter_test.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023-2025 The Chainloop Authors.
+// Copyright 2023-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


migmartri (approved):
Of the curiosity how did you find this?

jiparis (approved):

matiasinsaurralde:
> Of the curiosity how did you find this?

Mostly digging into `casclient` code and uploader logic plus some Claude insights.

`Upload` and `UploadFile` have been around for a long time. `UploadFile` just computes the hash and ends up calling `Upload`, so in this change we can call it directly from materials.

PR #3168, opened 2026-06-05
fix(controlplane): bound CAS mapping resolution on artifact download

### LINKED ISSUE #3167: N+1 query in CASMapping FindByDigest causes context deadlines on artifact download
## Summary

Downloading an artifact can hit `context deadline exceeded` when resolving the CAS mapping for a digest. The root cause is an N+1 query pattern in `CASMappingRepo.FindByDigest`.

## Details

`FindByDigest` (`app/controlplane/pkg/data/casmapping.go`) runs:

1. One indexed query to fetch all CAS mappings for the digest.
2. For every returned mapping, `IsPublic` runs two additional sequential queries (workflow run lookup, then workflow lookup).

This results in `1 + 2N` sequential round-trips, where `N` is the number of mappings sharing the digest. The digest lookup itself is global (not scoped by org/project at the SQL level; org/project filtering happens in memory afterwards), so a digest reused across many workflow runs accumulates mappings unbounded over time. Once `N` is large enough, the sequential follow-up queries exceed the request deadline.

This explains the intermittent nature ("in some cases"): only digests with a large number of accumulated mappings are affected.

## Proposed fix

Batch the public-visibility resolution into a constant number of queries (collect the referenced workflow run IDs and resolve their public flag in bulk) instead of two queries per mapping, removing the N+1.

## Follow-up

The control plane currently has no distributed tracing (OpenTelemetry/Jaeger) instrumented, which made this slow path harder to observe in production. Adding DB-level tracing is tracked separately.

Linear: PFM-6309

migmartri:
Resolving the CAS mapping for a digest could exhaust the request deadline and surface as a context deadline error on artifact download. Two problems compounded, both fixed here.

1. `FindByDigest` resolved each mapping's public visibility with two queries per mapping, producing `1 + 2N` sequential round-trips. Visibility for all referenced workflow runs is now resolved in a single edge-loaded query.

2. The download path fetched every mapping for a digest globally and filtered by org in memory. Because the same artifact can be pushed across thousands of workflow runs, a single digest accumulates thousands of mappings, yet only one is needed to locate the CAS backend. The org-scoped path now selects a single mapping directly in the database (organization + project RBAC predicates, preferring the default backend, `LIMIT 1`), so its cost no longer grows with the number of mappings. The public mapping lookup remains a fallback, used only when the requester has no org-level access to the digest.

Closes #3167

Assisted-by: Claude Code


app/controlplane/pkg/biz/casmapping.go
```diff
@@ -131,28 +133,25 @@ func (uc *CASMappingUseCase) FindCASMappingForDownloadByOrg(ctx context.Context,
...
+		return nil, fmt.Errorf("failed to find cas mapping in orgs: %w", err)
+	} else if mapping != nil {
+		return mapping, nil
 	}
 
-	// 2 - CAS mappings associated with the given list of orgs and project IDs
-	orgMappings, err := filterByOrgs(mappings, orgs, projectIDs)
+	// 2 - Otherwise, fall back to a public mapping. This only runs when the requester has no
+	// org-level access to the digest.
+	mappings, err := uc.repo.FindByDigest(ctx, digest)
```
migmartri: should'we then in this case be more targetted and search just for public ones?

migmartri: Good call — done. The public fallback no longer goes through the global `FindByDigest` + in-memory filter. It now uses a targeted `FindPublicByDigest` that selects a single public mapping directly in the DB (matched on the workflow's visibility via a subquery on `workflow_run_id`, ordered to prefer the default backend, `LIMIT 1`).

Since that was the last caller of `FindByDigest`, I removed the now-dead global-scan stack (`FindByDigest`, its public-visibility batch helper, and the unused use-case wrapper). The download path is now two bounded queries: org-scoped first, public fallback only when the requester has no org access.

jiparis (approved):

PR #3180, opened 2026-06-10
fix(controlplane): propagate FailedPrecondition errors instead of masking them

migmartri:
When an error that was already converted by `handleUseCaseErr` is processed again (e.g. `AttestationService.Store` wrapping `storeAttestation`), the resulting gRPC `FailedPrecondition` status error was not recognized by the conversion switch. It fell into the default branch, where it was reported to Sentry and masked as a generic internal server error, hiding the actionable message (such as pushing an attestation to a released, immutable project version) from the client.

This change passes through gRPC `FailedPrecondition` status errors in the default branch, making the conversion idempotent:

- The error is no longer reported to Sentry, since it is a client error, not a server fault.
- The original descriptive message reaches the client instead of a masked "internal error".

This PR was created with AI assistance (Claude Code).

### DIFF
--- a/app/controlplane/internal/service/service.go
+++ b/app/controlplane/internal/service/service.go
@@ -422,19 +422,30 @@ func handleUseCaseErr(err error, l *log.Helper) error {
 	case biz.IsErrReleasedVersionImmutable(err):
 		return status.Error(codes.FailedPrecondition, err.Error())
 	default:
-		// Business precondition failures (e.g. adding attestations to a released project version)
-		// are client errors: propagate them instead of masking and reporting them to Sentry.
-		// They can reach this point as gRPC status errors when an error already converted by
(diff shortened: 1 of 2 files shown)


migmartri:
Addressed the review finding (identified by cubic): the double-processing masking was not limited to FailedPrecondition — kratos errors also implement `GRPCStatus()`, so an already-converted `NotFound` (404) or `AlreadyExists` was masked as a 500 and reported to Sentry on a second pass as well.

`handleUseCaseErr` is now idempotent for every client-error code it produces (`Canceled`, `InvalidArgument`, `NotFound`, `PermissionDenied`, `Unimplemented`, `AlreadyExists`, `FailedPrecondition`), with test cases covering re-processing of converted errors and confirming server-side codes (e.g. `Unavailable`) are still masked.

jiparis (approved):

PR #3181, opened 2026-06-10
feat: add Yelp detect-secrets baseline as a material type

javirln:
## Description

Adds support for [Yelp detect-secrets](https://github.com/Yelp/detect-secrets) as a first-class material type via the new `YELP_DETECT_SECRETS_BASELINE` kind.

Users can now attest detect-secrets baseline files as a dedicated evidence type. The crafter validates that the input is a genuine detect-secrets baseline (presence of the `version`, `plugins_used` and `results` fields) and records the tool name and version as material annotations, which the policy engine can read at evaluation time.

## AI assistance disclosure

This contribution was produced with AI assistance (Claude Code), as disclosed in the commit `Assisted-by:` trailer.


migmartri (approved):
did we add it to auto-detection? should we? I don't think so

javirln:
> did we add it to auto-detection? should we? I don't think so

No I didn't add it. I left the auto-detection as it is. I would say that we need to revisit it to include recently added types?

PR #3182, opened 2026-06-10
feat(dagger): expose mark-latest as a tri-state enum on attestation init

javirln:
## Summary

Adds support for controlling "latest" promotion from the Dagger module's `Init`, exposed as a `MarkLatest` enum with three values:

- `ON_CREATE` (default): newly created versions become latest; existing/pre-release versions are left unchanged.
- `TRUE`: force-promote a pre-release version to latest.
- `FALSE`: skip latest promotion entirely, even for newly created versions.

The flag is modelled as an enum rather than a `*bool` because the Dagger v0.19.11 SDK collapses `*bool` parameters to plain `bool` and drops the false value before it reaches the wire, making the skip-promotion state unreachable through consuming modules. A non-empty string enum makes all three states reachable.

The enum maps onto the existing CLI `--mark-latest` flag: `TRUE` → `--mark-latest=true`, `FALSE` → `--mark-latest=false`, `ON_CREATE` → flag omitted (server default).

## AI disclosure

Assisted by Claude Code.

### DIFF
--- a/extras/dagger/README.md
+++ b/extras/dagger/README.md
@@ -80,9 +80,11 @@ dagger call -m github.com/chainloop-dev/chainloop \
 
 The optional `--mark-latest` flag controls whether the project version is promoted to `latest`. It is an enum with three values:
 
-- `ON_CREATE` (default): newly created versions become `latest`; existing/pre-release versions are left unchanged.
+- `ON_CREATE` (default): newly created versions become `latest`; existing versions are left unchanged.
 - `TRUE`: force-promote a pre-release version to `latest`.
-- `FALSE`: skip `latest` promotion entirely, even for newly created versions.
+- `FALSE`: do not promote the version to `latest`.
+
+`TRUE` and `FALSE` only affect promotion: they never demote a version that is already `latest`. In particular, `FALSE` prevents a newly created version from becoming `latest`, but it has no effect on an existing version (its `latest` status is left untouched).
 
 ```sh
 dagger call -m github.com/chainloop-dev/chainloop \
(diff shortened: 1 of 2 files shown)


extras/dagger/README.md
```diff
@@ -76,6 +76,24 @@ dagger call -m github.com/chainloop-dev/chainloop \
...
+- `FALSE`: skip `latest` promotion entirely, even for newly created versions.
+
+```sh
+dagger call -m github.com/chainloop-dev/chainloop \
+  init \
+  --token env:CHAINLOOP_TOKEN \
+  --workflow-name the-name-of-the-workflow \
+  --project-name the-name-of-the-project \
+  --version 1.0.0 \
+  --mark-latest FALSE # create the version without promoting it to latest
```
migmartri: does this update the existing version to false?

migmartri (dismissed):

extras/dagger/README.md
```diff
@@ -76,6 +76,24 @@ dagger call -m github.com/chainloop-dev/chainloop \
   --contract-name my-existing-contract \ # optional flag to specify an existing contract that will be used during the creation of a workflow
 ```
 
+##### Controlling "latest" promotion
+
+The optional `--mark-latest` flag controls whether the project version is promoted to `latest`. It is an enum with three values:
+
+- `ON_CREATE` (default): newly created versions become `latest`; existing/pre-release versions are left unchanged.
+- `TRUE`: force-promote a pre-release version to `latest`.
+- `FALSE`: skip `latest` promotion entirely, even for newly created versions.
```
jiparis: Should it accept derivatives like "true", "yes", "y", "1", etc., all case-insensitive?

extras/dagger/README.md
javirln: `mark-latest=false` creates the version without promoting it to latest, if the version exists and is latest, it's a no-op, it won't unset latest

extras/dagger/README.md
javirln: Not really, a Dagger enum compiles to a case-sensitive GraphQL enum with a fixed symbol set (and 1/0 aren't even legal enum names), so accepting case-insensitive derivatives would mean dropping the enum for a free-form string we parse by hand, losing validation/autocomplete; I'd rather keep the strict ON_CREATE/TRUE/FALSE to match the existing OutputFormat enum :/

jiparis (approved):

PR #3184, opened 2026-06-10
feat: add Sysinternals sigcheck output as a material type

javirln:
Adds a `SYSINTERNALS_SIGCHECK` material type that ingests [Sysinternals sigcheck](https://learn.microsoft.com/en-us/sysinternals/downloads/sigcheck) output.

sigcheck emits CSV, but the Rego policy engine evaluates JSON only. The material is therefore transformed into a JSON array of records at evaluation time (delimiter auto-detected between comma and tab, BOM-aware), following the existing JUnit and Jacoco handling in `GetEvaluableContent`. Policies can then read the records under `input.elements[_]`.

The crafter validates the report via a structural fingerprint (presence of the `Path` and `Verified` columns), stores the original CSV in the CAS, and annotates the material with the tool name.

This change was developed with AI assistance (Claude Code), as disclosed via the `Assisted-by` commit trailer.


migmartri (approved):
We probably want to update the readme and the documentation page

PR #3189, opened 2026-06-11
feat(cas): emit audit events on artifact uploads and downloads

### LINKED ISSUE #3187: Audit events for CAS artifact uploads, downloads and deduplication savings
## Problem

There is currently no record of Artifact CAS data-plane activity per organization. It is not possible to answer questions like:

- How many bytes has an organization uploaded to the CAS?
- How many bytes has it downloaded?
- How much storage/transfer was saved through deduplication (uploads skipped because the artifact already existed)?

## Proposal

Have the Artifact CAS emit audit events for artifact **uploads** (including a flag marking deduplicated/skipped uploads) and **downloads**, reusing the existing audit events infrastructure (NATS JetStream `chainloop-audit` stream) already used by the control plane.

High-level:

- Events carry the organization, artifact digest, size in bytes, filename, and backend type, so a stream consumer can aggregate per-organization upload volume, download volume, and deduplication savings.
- NATS remains optional for the CAS: when not configured, nothing changes. Event emission is best-effort and never affects the upload/download path.
- Internal control-plane traffic (e.g. fetching materials for policy evaluation) is excluded so it doesn't pollute organization usage numbers.

migmartri:
Closes #3187

The Artifact CAS now publishes audit events for artifact uploads, downloads and deduplicated (skipped) uploads to the existing `chainloop-audit` NATS JetStream stream, so a consumer of the stream can compute per-organization upload volume, download volume and deduplication savings.

## Implementation overview

- Two new event types in the existing controlplane events package, following the same `auditor.LogEntry` format as the rest of the audit events: `CASArtifactUploaded` and `CASArtifactDownloaded`. Events carry the organization, digest, size in bytes, filename and backend type, and are published to `audit.casartifact.*` subjects.
- The events are emitted by the CAS data plane itself (gRPC bytestream write/read and the HTTP download endpoint), since the CAS is the only component that observes the deduplication decision and the actual byte transfers. Emission happens only after the operation's success point: uploads stored, downloads checksum-verified.
- A new thin `AuditDispatcher` in the CAS resolves the organization from the request JWT claims and publishes through the shared `auditor` machinery.
- New optional `nats_server` section in the CAS configuration, exposed in the Helm chart as `cas.nats.*` values.

## Major design decisions

- **Deduplicated uploads are not a separate event type**: they are `CASArtifactUploaded` events with a `skipped: true` flag (no bytes were transferred or stored). Since the stored size is not known at the dedup point, it is resolved with a best-effort `Describe()` call, falling back to `0`; consumers can join on the digest with a prior non-skipped upload.
- **Internal control plane traffic emits no events**: a new `SourceInternal` JWT claim is set only when the control plane mints tokens for its own CAS client (attestation storage, policy material reads). The boolean zero value means client traffic, so the client-facing token issuance paths needed no changes, and the suppression happens at emission so the stream needs no consumer-side filtering.
- **The control plane remains the single owner of the stream configuration**: the CAS publisher runs in a new publish-only mode (`NewPublishOnlyAuditLogPublisher`), so it can never create or reconfigure the `chainloop-audit` stream (e.g. accidentally downgrade its replica count).
- **Events are best-effort and NATS stays optional**: when `nats_server` is unset the behavior is unchanged; publish failures are logged and never fail or slow down an upload/download.
- **SYSTEM actor, no target ID**: CAS JWTs carry only the organization, so events require no actor; artifacts are identified by their digest in the event payload. Actor attribution (user/API token identity in CAS tokens) is left as a possible follow-up.

## Consumer math

Per organization, over a time window on the stream: uploaded bytes = sum of `size_bytes` where `skipped == false`; downloaded bytes = sum of `CASArtifactDownloaded` sizes; dedup savings = sum of `size_bytes` where `skipped == true`.

## AI disclosure

This contribution was developed with AI assistance (Claude Code).


javirln (approved):

app/artifact-cas/internal/service/auditor.go
```diff
@@ -0,0 +1,87 @@
...
+// AuditDispatcher publishes CAS audit events. Unlike the control plane's
+// biz.AuditorUseCase, the actor is always SYSTEM (CAS JWTs carry no user
+// identity) and the org comes from the JWT claims instead of the request context.
+type AuditDispatcher struct {
+	// nil when NATS is not configured, making the dispatcher a no-op
+	publisher eventPublisher
+	log       *log.Helper
+}
+
+func NewAuditDispatcher(publisher *auditor.AuditLogPublisher, logger log.Logger) *AuditDispatcher {
```
jiparis, line 43: cannot it reuse the dispatcher from controlplane? perhaps moving it out of internal?

migmartri: Will run a follow up, thanks

migmartri: Good call @jiparis. I extracted the shared generate → publish → error-reporting flow into a new `auditor.Dispatcher` in `pkg/auditor` (importable, not internal), and both `biz.AuditorUseCase` and this CAS `AuditDispatcher` now delegate to it — each keeping only its own actor/org resolution. Reusing `biz.AuditorUseCase` directly wasn't viable since it resolves the actor from the control-plane request context and importing `biz` would pull the whole control-plane dependency graph into the CAS, so the shared seam lives at the lower-level dispatcher instead.

Follow-up: #3191

PR #3217, opened 2026-06-16
fix(verifier): require signature verification, reject timestamp-only bundles

jiparis:
## Summary

`VerifyBundle` only verified the DSSE signature when a bundle carried an X.509 certificate. A bundle with no certificate but a valid RFC3161 timestamp skipped signature verification and was reported as verified, even though the signature was never checked against any key. Since the timestamp signs only the DSSE signature bytes and not the payload, a forged in-toto statement with arbitrary signature bytes plus a timestamp from a trusted TSA could pass verification.

This change makes signature verification mandatory:

- A certificate-backed DSSE signature must verify against the trusted root before a bundle is considered verified. The RFC3161 timestamp is now only used to validate the signing window and can never be the sole verification material.
- Public-key verification material is explicitly rejected as unsupported instead of falling through to the timestamp-only path.
- Cert-less bundles with no verifiable key are reported as missing verification material rather than verified.

## AI assistance disclosure

This change was produced with the assistance of Claude Code, as disclosed via the `Assisted-by:` commit trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/verifier/testdata/bundle_with_publickey.json
+++ b/pkg/attestation/verifier/testdata/bundle_with_publickey.json
@@ -1,0 +1,17 @@
+{
+   "mediaType": "application/vnd.dev.sigstore.bundle+json;version=0.3",
+   "verificationMaterial": {
+      "publicKey": {
+         "hint": "some-key-hint"
+      }
+   },
+   "dsseEnvelope": {
+      "payload": "eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1lbnQvdjEifQ==",
+      "payloadType": "application/vnd.in-toto+json",
+      "signatures": [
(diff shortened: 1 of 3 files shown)


migmartri (approved):
Let's make sure this doesn't cause any issue with old CLIs that were not bundling the same format

matiasinsaurralde (approved):
LGTM

PR #3223, opened 2026-06-18
fix(cli): surface policy-group materials during attestation

### LINKED ISSUE #3222: Materials defined in a policy group are not shown during attestation
## Summary

When running an attestation, the CLI/UI should surface **all** materials that are expected for the workflow — both required and optional. Materials that are defined as part of a **policy group** attached to the contract are not being shown during attestation, so users have no visibility into them.

## Expected behavior

During attestation (e.g. `chainloop attestation status` / `add`), the list of expected materials should include materials contributed by any policy group bound to the contract, alongside the materials declared directly in the contract.

## Actual behavior

Materials that originate from a policy group are omitted from the materials shown during attestation. Only materials declared directly on the contract appear; the policy-group materials are silently missing.

## Related: confusing warning on duplicate materials

If the same material is declared in **both** the contract's `materials` list and inside an attached policy group, the user gets a confusing warning during attestation. The two sources of materials aren't reconciled cleanly: instead of treating the contract + policy-group definitions as a single merged set, the overlap is flagged in a way that's unclear to the user (it reads like an error/misconfiguration rather than the expected "same material defined in two places" case). The warning text and the condition that triggers it should be clarified — or the duplicate should be merged silently when the definitions are compatible.

## Impact

- Users can't tell which materials a policy group expects, so they may not provide them.
- Reduced visibility makes it hard to satisfy policy-group requirements and to debug failing/incomplete attestations.
- The duplicate-material warning adds confusion when a material legitimately appears in both the contract and a policy group.

## Notes

The materials are still defined correctly inside the policy group — the bug is specifically in how the expected-materials list is assembled/displayed during attestation, where policy-group materials are not merged in cleanly with the contract materials.

migmartri:
## Summary

Fixes #3222.

Materials declared by a policy group attached to a contract were not surfaced during attestation for V2 contracts. The CLI enriched the contract's expected-materials list only on the V1 schema, but the crafter stores the V2 schema when it is present, so the enriched V1 was discarded and the policy-group materials never appeared in `attestation status`/`add` (nor were they validated as expected).

This change enriches the schema that the crafter actually stores — the V2 schema when present, V1 otherwise — so policy-group materials are merged into the expected-materials list alongside the contract's own materials.

It also clarifies the duplicate-material handling: when the same material is declared in both the contract and an attached policy group, the definitions are now merged silently when compatible, and a clearer warning is emitted only when the material types genuinely conflict.

Tested locally, below is an example of a material called sbom that comes from the policy group

```
hainloop attestation init --project policygrouptest --workflow test --replace
WRN User-attended mode detected. This is intended for local testing only. For CI/CD or automated workflows, please use an API token.
This command will run against the organization "miguel-test"
y
┌───────────────────────────┬────────────────────────────────────────┐
│ Initialized At            │ 18 Jun 26 14:43 UTC                    │
├───────────────────────────┼────────────────────────────────────────┤
│ Attestation ID            │ 918bf69a-6365-480a-a798-f06fd9586e37   │
│ Organization              │ miguel-test                            │
│ Name                      │ test                                   │
│ Project                   │ policygrouptest                        │
│ Version                   │ v1.100.11+next (prerelease)            │
│ Contract                  │ contract-fix-policy-group (revision 1) │
│ Timestamp Authority       │ http://timestamp.digicert.com          │
│ Policy violation strategy │ ADVISORY                               │
│ Policies                  │ ------                                 │
│                           │ sbom-present: missing SBOM material    │
└───────────────────────────┴────────────────────────────────────────┘
┌────────────────────────────────┐
│ Materials                      │
├──────────┬─────────────────────┤
│ Name     │ sbom                │
│ Type     │ SBOM_CYCLONEDX_JSON │
│ Set      │ No                  │
│ Required │ Yes                 │
└──────────┴─────────────────────┘
```

in previous versions of the CLI the material was not present nor enforced

## AI disclosure

This change was produced with the assistance of Claude Code. Affected commits carry an `Assisted-by: Claude Code` trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/cli/pkg/action/attestation_init.go
+++ b/app/cli/pkg/action/attestation_init.go
@@ -181,12 +181,6 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 
 	action.Logger.Debug().Msg("workflow contract and metadata retrieved from the control plane")
 
-	// 3. enrich contract with group materials and policies
-	err = enrichContractMaterials(ctx, contractVersion.GetV1(), client, &action.Logger)
-	if err != nil {
-		return "", fmt.Errorf("failed to apply materials from policy groups: %w", err)
-	}
-
 	// Auto discover the runner context and enforce against the one in the contract if needed
 	// nolint:staticcheck
 	discoveredRunner, err := crafter.DiscoverAndEnforceRunner(contractVersion.GetV1().GetRunner().GetType(), action.dryRun, action.AuthTokenRaw, action.Logger)
@@ -280,6 +274,20 @@ func (action *AttestationInit) Run(ctx context.Context, opts *AttestationInitRun
 		schemaV2 = parseContractV2(contractVersion.GetRawContract())
 	}
 
+	// Enrich the contract with the materials declared by its attached policy
+	// groups, so they show up during attestation. See issue #3222.
+	// Only the schema that the crafter will actually store needs enriching: it
+	// prefers the V2 schema when present and falls back to V1 otherwise.
+	if schemaV2 != nil {
+		err = enrichContractMaterialsV2(ctx, schemaV2, client, &action.Logger)
+	} else {
+		//nolint:staticcheck // TODO: Migrate to new contract version API
+		err = enrichContractMaterials(ctx, contractVersion.GetV1(), client, &action.Logger)
+	}
+	if err != nil {
+		return "", fmt.Errorf("failed to apply materials from policy groups: %w", err)
+	}
+
 	// Initialize the local attestation crafter
 	// NOTE: important to run this initialization here since workflowMeta is populated
 	// with the workflowRunId that comes from the control plane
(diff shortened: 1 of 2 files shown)


migmartri:
Valid catch (identified by cubic). The previous ordering ran the V1 enrichment before the control-plane `Init`; consolidating the V1/V2 enrichment had moved it after `Init`, so a policy-group load failure could occur after a workflow run was created.

Fixed in a76ce6b's follow-up: enrichment now runs before `client.Init` again (both `GetV1()` and `GetRawContract()` are available right after the contract is fetched, so there's no dependency on the init result), restoring fail-fast before any side effect.

jiparis (approved):

PR #3233, opened 2026-06-23
fix(cli): preserve verbatim contract bytes when splitting apply documents

javirln:
## Description

After a clean apply, re-running `chainloop apply -f <contracts> --dry-run` reported a contract as "would be updated" even though nothing changed, while `chainloop wf contract apply` correctly reported it "unchanged".

The cause was entirely client-side: the batch `apply` path re-serialized each YAML document through a yaml.v3 Node (`yaml.Marshal(&node)`), which re-indented and reflowed the content. That produced different bytes than `wf contract apply`, which sends the file verbatim. Since the server detects changes by comparing the stored raw bytes against the incoming ones, the reformatted bytes always looked like a change.

`SplitYAMLDocuments` now splits the stream on document markers and keeps each document's original bytes verbatim, so every apply path sends exactly what the user authored. Re-applying the same file is idempotent regardless of path, and genuine edits — including comment or formatting-only changes — still produce a new revision. No server-side change is needed; raw-byte change detection is left as-is.

Closes PFM-6465

## AI disclosure

Claude Code assisted in producing this change.

### DIFF
--- a/app/cli/pkg/action/apply.go
+++ b/app/cli/pkg/action/apply.go
@@ -18,9 +18,7 @@ package action
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


migmartri (changes requested):
I don't know, this looks strange, shouldn't we take a look at why different clients produce different outputs? at the end, we have the raw content. 

Also, does this mean that if I add a comment now in the yaml it will not be saved because the marshalled output is the same?

javirln:
> _Posted by an AI agent (Claude Code) on behalf of @javirln._

Thanks for the pushback @migmartri — you were right on both counts, so I've reworked the PR to fix it client-side instead of changing the server comparison.

**On "why do different clients produce different outputs? at the end we have the raw content":** it wasn't really two different clients — it was the same file serialized two ways by the CLI. `wf contract apply` sends the file verbatim (`os.ReadFile`), while the batch `apply` path ran each document through `SplitYAMLDocuments`, which decoded into a `yaml.Node` and re-marshalled it (`yaml.Marshal(&node)`), re-indenting and reflowing the content. That reformatted output is what diverged from the stored bytes.

**On "if I add a comment now, will it not be saved because the marshalled output is the same?":** exactly the reason I dropped the semantic-comparison approach. That would have ignored comment/formatting-only edits and skipped the revision. We don't want that.

**New approach:** `SplitYAMLDocuments` now splits the stream on document markers and keeps each document's original bytes verbatim, so every apply path sends exactly what the user authored. The server keeps its existing raw-byte change detection unchanged — no server-side change at all. Result:

- re-applying the same file is idempotent regardless of path (`wf contract apply` / `apply` / `apply --dry-run`);
- genuine edits, **including comment- or whitespace-only changes**, still produce a new revision.

Added CLI unit tests covering verbatim single/multi-document preservation, comment-only-document skipping, and the missing `kind`/`metadata.name` errors.

PTAL when you get a chance.

jiparis (approved):

jiparis:
looks good!

migmartri (approved):

PR #3240, opened 2026-06-25
fix(server): prevent /debug/* exposure via DefaultServeMux fallthrough

matiasinsaurralde:
## Summary

Kratos HTTP servers route unmatched requests to `http.DefaultServeMux` through their NotFoundHandler and `MethodNotAllowedHandler`. Standard-library packages such as `net/http/pprof`, `expvar` and `golang.org/x/net/trace` auto-register handlers on that global mux from their `init()` functions, so any unmatched route on a public server could leak `/debug/pprof/*`, `/debug/vars` and `/debug/requests`.

This adds a shared `DenyDefaultMuxFallthrough` helper that returns plain 404/405 handlers, severing the fallthrough on every network path, and applies it to the Control Plane and Artifact CAS HTTP and metrics servers. Registered routes continue to be matched by the router and are unaffected.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/artifact-cas/internal/server/http.go
+++ b/app/artifact-cas/internal/server/http.go
@@ -40,6 +40,9 @@ func NewHTTPServer(c *conf.Server, authConf *conf.Auth, downloadSvc *service.Dow
 			logging.Server(logger),
 		),
 	}
+	// Stop unmatched routes from falling through to http.DefaultServeMux,
+	// which would expose /debug/vars and /debug/requests (CVE-2026-6993).
+	opts = append(opts, middlewares_http.DenyDefaultMuxFallthrough()...)
 	if c.Http.Network != "" {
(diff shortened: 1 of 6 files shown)


jiparis:
Superseded by https://github.com/chainloop-dev/chainloop/pull/3241 which also removes the default import and configures a specific profiler at 6060

PR #3244, opened 2026-06-26
feat(cli): inject policy inputs from CSV/JSON files at attestation add

### LINKED ISSUE #3250: Support data-driven policy inputs (e.g. ignored_paths) sourced from CSV/JSON files
## Problem

Chainloop ships four Rego policies that evaluate a Sysinternals `sigcheck` manifest (`SYSINTERNALS_SIGCHECK`) to assert that binaries are properly signed: `trusted-binaries-signed`, `trusted-binaries-authenticode-signed`, `trusted-binaries-timestamp`, and `trusted-binaries-vendor-keys`. All four scope the check with path globs, including an `ignored_paths` exclusion list.

Today the only way to express exclusions is a hand-maintained, comma-separated string passed as a policy argument (in the contract `with:` or as a binding). This does not scale for real deployments:

- Exclusion lists routinely run to **hundreds of full file paths** (OS files, third-party redistributables, known-unsigned legacy components).
- The list lives in the contract/argument, **divorced from the system that actually knows** which files are exempt.
- Updating exclusions means **editing policy arguments** by hand rather than updating a data file owned by the relevant team.
- The exemption set itself is **not attested** — there's no tamper-evident record of which exclusions were applied to a given attestation.

## Desired outcome

Allow policy inputs such as `ignored_paths` to be sourced **dynamically from a data file** (CSV or JSON) at attestation time, instead of (or in addition to) an inline argument string:

- Extract a named column/field from the file and inject it as a policy input during material evaluation.
- The mechanism should be **generic** over both the input name and the source column, not hardcoded to `ignored_paths` or to sigcheck.
- Record the source file as attested **evidence**, linked to the evaluated material, and mark in the policy evaluation which inputs were supplied at runtime, so exclusions remain auditable.

## Notes

This is intended to be additive plumbing in the CLI and crafter — no changes to the Rego policies or to the engine's evaluation logic.

Addressed by #3244.

migmartri:
Closes chainloop-dev/chainloop#3250

## Summary

Adds a repeatable `--policy-input-from-file <input>=<file>[:<column>]` flag to `chainloop attestation add` that sources a policy input from a column of a CSV or JSON file and injects it during material policy evaluation. The primary use case is feeding the `ignored_paths` input of the Sysinternals sigcheck binary-signing policies from a large, data-driven exemption list, but the flag is generic over both the input name and the source column.

```sh
chainloop attestation add --name sigcheck --value sigcheckResult.csv --kind SYSINTERNALS_SIGCHECK \
  --policy-input-from-file ignored_paths=exception.csv:Path
```

## Details

* The `:<column>` suffix selects the column and defaults to the input name; a column is always a single, top-level field/header name. The column is the segment after the last `:`, so a Windows drive letter (`C:\...`) or URL scheme (`https://`) in the file path is not mistaken for a column.
* Runtime inputs are merged additively onto contract-declared arguments through a dedicated channel, kept separate from policy-group bindings so group evaluations are not mis-flagged as runtime overrides.
* The source file is recorded as an `EVIDENCE` material named `<material>-<input>` (the input name sanitized to a valid material name, e.g. `ignored_paths` → `ignored-paths`; a `-<n>` suffix disambiguates repeats). The exact input name is preserved in a `chainloop.material.policy_input` annotation.
* The evidence material and the evaluated material are cross-linked **both ways** via the generic `chainloop.material.references` annotation (by material name, modeled on the OCI referrers API). The same primitive can connect other material kinds in the future.
* `PolicyEvaluation` gains a `runtime_input_overrides` field listing which input names were supplied at runtime; the effective merged values remain in `with`.
* CSV/TSV parsing is handled by a new generic `pkg/tabular` package (extracted from the previous sigcheck-only parser): BOM decoding, comma/tab auto-detection, and a case-insensitive `Column` extractor. The sigcheck material crafter now consumes it too. JSON accepts a bare string array, an array of string-valued objects, or an object mapping the column to an array.

## AI assistance

This change was developed with the assistance of Claude Code. Affected commits carry an `Assisted-by: Claude Code` trailer.

🤖 *Posted by Maximus bot (Claude Code) on behalf of* @migmartri

### DIFF
--- a/app/cli/documentation/cli-reference.mdx
+++ b/app/cli/documentation/cli-reference.mdx
@@ -244,21 +244,27 @@ chainloop attestation add --value <material-value>
 
 Add a material by also providing a URL pointing to the material. It will be downloaded to a temporary folder first
 chainloop attestation add --value https://example.com/sbom.json
+
+Feed a policy input from a column of a CSV/JSON file (e.g. the ignored_paths exclusion list for the sigcheck binary-signing policies).
+The :column suffix selects the column; it defaults to the input name when omitted. The file is also recorded as EVIDENCE.
+chainloop attestation add --name sigcheck --value sigcheckResult.csv --kind SYSINTERNALS_SIGCHECK \
+--policy-input-from-file ignored_paths=exception.csv:Path
 ```
 
 Options
 
 ```
---annotation strings         additional annotation in the format of key=value
---attestation-id string      Unique identifier of the in-progress attestation
--h, --help                       help for add
---kind string                kind of the material to be recorded: ["ARTIFACT" "ASYNCAPI_SPEC" "ATTESTATION" "BLACKDUCK_SCA_JSON" "CERTCC_DRANZER" "CHAINLOOP_AI_AGENT_CONFIG" "CHAINLOOP_AI_CODING_SESSION" "CHAINLOOP_PR_INFO" "CHAINLOOP_RUNNER_CONTEXT" "CONTAINER_IMAGE" "CSAF_INFORMATIONAL_ADVISORY" "CSAF_SECURITY_ADVISORY" "CSAF_SECURITY_INCIDENT_RESPONSE" "CSAF_VEX" "EVIDENCE" "GHAS_CODE_SCAN" "GHAS_DEPENDENCY_SCAN" "GHAS_SECRET_SCAN" "GITLAB_SECURITY_REPORT" "GITLEAKS_JSON" "GRAPHQL_SPEC" "HELM_CHART" "JACOCO_XML" "JUNIT_XML" "OPENAPI_SPEC" "OPENVEX" "OSSF_SCORECARD_JSON" "RADAMSA_CRASHES" "RADAMSA_REPORT" "SARIF" "SBOM_CYCLONEDX_JSON" "SBOM_SPDX_JSON" "SLSA_PROVENANCE" "STRING" "SYSINTERNALS_ACCESSCHK" "SYSINTERNALS_SIGCHECK" "TWISTCLI_SCAN_JSON" "YELP_DETECT_SECRETS_BASELINE" "ZAP_DAST_ZIP"]
---name string                name of the material as shown in the contract
---no-strict-validation       skip strict schema validation for structured materials (SBOM_CYCLONEDX_JSON, OPENAPI_SPEC, ASYNCAPI_SPEC, OSSF_SCORECARD_JSON)
---registry-password string   registry password, ($CHAINLOOP_REGISTRY_PASSWORD)
---registry-server string     OCI repository server, ($CHAINLOOP_REGISTRY_SERVER)
---registry-username string   registry username, ($CHAINLOOP_REGISTRY_USERNAME)
---value string               value to be recorded
+--annotation strings                   additional annotation in the format of key=value
+--attestation-id string                Unique identifier of the in-progress attestation
+-h, --help                                 help for add
+--kind string                          kind of the material to be recorded: ["ARTIFACT" "ASYNCAPI_SPEC" "ATTESTATION" "BLACKDUCK_SCA_JSON" "CERTCC_DRANZER" "CHAINLOOP_AI_AGENT_CONFIG" "CHAINLOOP_AI_CODING_SESSION" "CHAINLOOP_PR_INFO" "CHAINLOOP_RUNNER_CONTEXT" "CONTAINER_IMAGE" "CSAF_INFORMATIONAL_ADVISORY" "CSAF_SECURITY_ADVISORY" "CSAF_SECURITY_INCIDENT_RESPONSE" "CSAF_VEX" "EVIDENCE" "GHAS_CODE_SCAN" "GHAS_DEPENDENCY_SCAN" "GHAS_SECRET_SCAN" "GITLAB_SECURITY_REPORT" "GITLEAKS_JSON" "GRAPHQL_SPEC" "HELM_CHART" "JACOCO_XML" "JUNIT_XML" "OPENAPI_SPEC" "OPENVEX" "OSSF_SCORECARD_JSON" "RADAMSA_CRASHES" "RADAMSA_REPORT" "SARIF" "SBOM_CYCLONEDX_JSON" "SBOM_SPDX_JSON" "SLSA_PROVENANCE" "STRING" "SYSINTERNALS_ACCESSCHK" "SYSINTERNALS_SIGCHECK" "TWISTCLI_SCAN_JSON" "YELP_DETECT_SECRETS_BASELINE" "ZAP_DAST_ZIP"]
+--name string                          name of the material as shown in the contract
+--no-strict-validation                 skip strict schema validation for structured materials (SBOM_CYCLONEDX_JSON, OPENAPI_SPEC, ASYNCAPI_SPEC, OSSF_SCORECARD_JSON)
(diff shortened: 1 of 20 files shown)


app/cli/documentation/cli-reference.mdx
```diff
@@ -249,16 +249,17 @@ chainloop attestation add --value https://example.com/sbom.json
...
+--name string                          name of the material as shown in the contract
+--no-strict-validation                 skip strict schema validation for structured materials (SBOM_CYCLONEDX_JSON, OPENAPI_SPEC, ASYNCAPI_SPEC, OSSF_SCORECARD_JSON)
+--policy-input-from-file stringArray   feed a policy input from a column of a CSV or JSON file, in the format <input>[:<column>]=<file> (e.g. ignored_paths:Path=exception.csv); repeatable. The file is also recorded as EVIDENCE.
```
jiparis: I like `ignored_paths=exceptions.csv:path` better.

app/cli/pkg/action/policy_input_file.go
```diff
@@ -0,0 +1,228 @@
...
+		// Content sniff: a leading "[" or "{" means JSON, otherwise treat as CSV.
+		if t := bytes.TrimSpace(content); len(t) > 0 && (t[0] == '[' || t[0] == '{') {
+			return extractJSONColumn(content, column)
+		}
+		return extractCSVColumn(content, column)
+	}
+}
+
+func extractCSVColumn(content []byte, column string) ([]string, error) {
+	report, err := sigcheck.Parse(content)
```
migmartri: shouldn't we move the CSV parser to a generic package or use that one generically? now it looks like we can only support sigcheck.

app/cli/pkg/action/policy_input_file.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+	return "", false
+}
+
+// extractJSONColumn extracts column values from one of three accepted shapes:
+// a bare array of strings, an array of objects (the column field of each), or
+// an object mapping the column to an array of strings. The column is matched
+// only against top-level keys; nested paths are not interpreted.
+func extractJSONColumn(content []byte, column string) ([]string, error) {
```
migmartri, line 129: isn't there a better way of doing this maybe by marshalling to the expected object?

app/cli/pkg/action/policy_input_file.go
```diff
@@ -0,0 +1,212 @@
+//
```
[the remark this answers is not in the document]
migmartri: Good catch — fixed. `ExtractColumnValues` now strips a leading UTF-8 BOM before format detection and JSON parsing, so Windows-authored JSON is detected and parsed correctly (the CSV path already handled it via the parser). Added a regression test.

app/cli/pkg/action/attestation_add.go
```diff
@@ -160,6 +178,60 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
...
+
+	return runtimeInputs, nil
+}
+
+// addPolicyInputEvidence adds each policy input file as an EVIDENCE material,
+// linked back to the evaluated material via the chainloop.material.references
+// annotation. The evidence material name is derived as "<material>-<input>".
+func (action *AttestationAdd) addPolicyInputEvidence(ctx context.Context, c *crafter.Crafter, attestationID, materialName string, policyInputFiles []*PolicyInputFromFile, casBackend *casclient.CASBackend) error {
+	for _, pif := range policyInputFiles {
+		evidenceName := fmt.Sprintf("%s-%s", materialName, pif.Input)
```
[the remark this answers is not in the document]
migmartri: Fixed. When the same input is fed by multiple `--policy-input-from-file` flags, evidence material names are now disambiguated with a `-<n>` suffix so no record is silently overwritten; single-file inputs keep the plain `<material>-<input>` name. Extracted into a pure `policyInputEvidenceNames` helper with a table test covering the collision.

app/cli/pkg/action/attestation_add.go
```diff
@@ -160,6 +178,60 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
...
+		return nil, nil
+	}
+
+	runtimeInputs := make(map[string]string, len(policyInputFiles))
+	for _, pif := range policyInputFiles {
+		values, err := ExtractColumnValues(pif.File, pif.Column)
+		if err != nil {
+			return nil, fmt.Errorf("extracting %q from %q: %w", pif.Column, pif.File, err)
+		}
+		joined := strings.Join(values, "\n")
```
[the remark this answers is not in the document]
migmartri: Reworded. The newline join matches the engine's existing multi-value encoding (it splits inputs back on newlines and commas), so individual values must not embed those delimiters — the same constraint as contract `with:` arguments. Path globs, the intended use, never do. This is inherent to the input mechanism rather than specific to this code path.

app/cli/documentation/cli-reference.mdx
migmartri: Agreed — that is the implemented format: `--policy-input-from-file ignored_paths=exception.csv:Path` (the `:<column>` is optional and defaults to the input name). The column is the segment after the last `:`, and since a column name never contains a path separator, a Windows drive letter (`C:\...`) or URL scheme (`https://`) is not mistaken for a column.

app/cli/pkg/action/policy_input_file.go
migmartri: Done — moved the parser out of the sigcheck subpackage into a generic `pkg/tabular` (`Parse` returning a `Table`, with a `Column(name)` case-insensitive extractor). The sigcheck material crafter, the JSON ingestion in `crafting_state.go`, and this CLI extraction all consume it now, and the old `materials/sigcheck` subpackage is removed.

app/cli/pkg/action/policy_input_file.go
migmartri: Reworked: the array-of-objects shape now decodes straight into `[]map[string]string` (dropped the `json.RawMessage` + manual string decode), and the object→array shape keeps `json.RawMessage` only for the key lookup before decoding the selected value into a typed `[]string`, so sibling fields of other types do not break parsing. `matchKey` is now a single generic helper. Trade-off: array-of-objects now requires string-valued fields (documented in the code and spec).

migmartri: The reason there isn't a single "expected object" to marshal into is that the **column name is chosen at runtime** (`--policy-input-from-file ignored_paths=exception.csv:Path`), so there's no fixed field name to put in a Go struct tag — the target field is `Path` here, but could be anything for another input/file. A typed struct would have to know that field name at compile time, which we don't.

On top of that, the *shape* of the file also varies between the tools that produce these lists, so we accept three:

1. bare array of strings — `["a","b"]`
2. array of string-valued objects — `[{"Path":"a"},{"Path":"b"}]`
3. object mapping the column to an array — `{"Path":["a","b"]}`

Given those two degrees of freedom (dynamic key + variable shape), some generic decoding is unavoidable. What I did do to keep it as typed as possible:

- shape 2 decodes straight into `[]map[string]string` (no `json.RawMessage`, no manual per-field decode),
- shape 3 keeps `json.RawMessage` *only* for the keyed lookup, then decodes the selected value into a typed `[]string` — so sibling fields of other types don't break the parse,
- the key lookup is a single generic helper (`matchKey[T any]`).

This also stays within the repo guideline of not unmarshalling into `any` for data access — we never decode into `interface{}`; the most generic type used is `map[string]string` / `map[string]json.RawMessage` with a typed `[]string` for the actual values.

Trade-off: shape 2 now requires string-valued fields, which is documented in the code and the spec.

javirln (approved):

PR #3245, opened 2026-06-26
fix(authz): close DeleteMembership authorization bypass

matiasinsaurralde:
Closes PFM-6498

## Summary

- Add explicit `ServerOperationsMap` entries for `OrganizationService/DeleteMembership` and `UpdateMembership`, requiring org admin/owner membership-management policies.
- Introduce `PolicyOrganizationMembershipsDelete` and `PolicyOrganizationMembershipsUpdate` and grant them to `RoleAdmin` and `RoleOwner`.
- Prevents any org member (including viewers) from removing or updating other members via the unanchored regex match on the empty `OrganizationService/Delete` policy.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/controlplane/pkg/authz/authz.go
+++ b/app/controlplane/pkg/authz/authz.go
@@ -160,8 +160,10 @@ var (
 	PolicyOrganizationCreate = &Policy{Organization, ActionCreate}
 	PolicyOrganizationDelete = &Policy{Organization, ActionDelete}
 	// User Membership
-	PolicyOrganizationRead            = &Policy{Organization, ActionRead}
-	PolicyOrganizationListMemberships = &Policy{OrganizationMemberships, ActionList}
+	PolicyOrganizationRead              = &Policy{Organization, ActionRead}
+	PolicyOrganizationListMemberships   = &Policy{OrganizationMemberships, ActionList}
+	PolicyOrganizationMembershipsDelete = &Policy{OrganizationMemberships, ActionDelete}
+	PolicyOrganizationMembershipsUpdate = &Policy{OrganizationMemberships, ActionUpdate}
 
 	// Group Memberships
 	PolicyGroupListPendingInvitations = &Policy{ResourceGroup, ActionList}
@@ -203,6 +205,8 @@ var RolesMap = map[Role][]*Policy{
 	RoleOwner: {
 		PolicyOrganizationDelete,
 		PolicyOrganizationManageOwners,
+		PolicyOrganizationMembershipsDelete,
+		PolicyOrganizationMembershipsUpdate,
 	},
 	// RoleAdmin is an org-scoped role that provides super admin privileges (it's the higher role)
 	RoleAdmin: {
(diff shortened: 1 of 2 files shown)


migmartri (commented):
Are you sure that there is no check at the service level for this?

matiasinsaurralde:
> Are you sure that there is no check at the service level for this?

The service handler calls straight into the biz layer with no caller-authz check, and the only biz check (`enforceManageOwners`) fires solely when the target is an owner. For non-owner targets the authz middleware was the only gate, which is exactly where the bypass was. `UpdateMembership` was ok though:

- `DeleteMembership` — the key `OrganizationService/Delete` exists as an empty `{}` entry (authz.go, main), and "…/Delete" is a prefix of "…/DeleteMembership", so the regex matched it → returned an empty policy list → the enforcement loop ran zero checks → allowed. That's the bypass.
- `UpdateMembership` — there is no `OrganizationService/Update` key to collide with (only Create/Delete/ListMemberships exist), so nothing matched → `policiesLookup` returned `errors.New("operation not allowed")` → `checkPolicies` returned Forbidden → denied.

There's also a demo on the ticket.

migmartri (approved):

app/controlplane/pkg/authz/authz.go
```diff
@@ -417,7 +423,9 @@ var ServerOperationsMap = map[string]*OperationPolicy{
 	"/controlplane.v1.OrganizationService/Delete": {},
 
 	// List global memberships
-	"/controlplane.v1.OrganizationService/ListMemberships": {Policies: []*Policy{PolicyOrganizationListMemberships}},
+	"/controlplane.v1.OrganizationService/ListMemberships":  {Policies: []*Policy{PolicyOrganizationListMemberships}},
+	"/controlplane.v1.OrganizationService/DeleteMembership": {Policies: []*Policy{PolicyOrganizationMembershipsDelete}},
```
jiparis, line 427: If it didn't exist previously, only Admins/Owners should be able to call this endpoint.
If you noticed that a Viewer can call this endpoint, then the fix is somewhere else.

jiparis (changes requested):

migmartri (approved):
Thanks

app/controlplane/pkg/authz/authz.go
migmartri: That's correct, if they don't exist it means only admin and owners

jiparis: Ok I see, it's falling under `OrganizationService/Delete: {}` which is permissive. 
Then, I'd prefer to fix the regular expression to look for exact match. The actual error is that `/controlplane.v1.OrganizationService/Delete` is matching `/controlplane.v1.OrganizationService/DeleteMembership` and it shouldn't.

matiasinsaurralde: Yes, this is a hotfix that makes the matching for this route explicit (it was the only vulnerable route found), a follow up wider fix would involve removing the unanchored regex.
I haven't found issues in other paths but in the way we're using regex for paths, it may occur if we accidentally introduce some type of routes. So a follow up PR should make that part better to avoid future issues.

jiparis (approved):

PR #3253, opened 2026-06-30
feat(cli): explode archive materials in chainloop att add

javirln:
## Summary

`chainloop attestation add --kind <KIND> --value <archive>` now unpacks `.zip`, `.tar`, `.tar.gz`, and `.tgz` archives and records each contained regular file as its own material of the given kind, instead of requiring one `att add` invocation per file.

## Behavior

- When `--kind` is provided and the value resolves to a supported archive, each contained regular file is added as an individual material of that kind.
- Archive-native kinds (e.g. `ZAP_DAST_ZIP`), whose value is meant to be the archive itself, are still recorded whole.
- Per-entry material names are derived from the entry filename (sanitized to DNS-1123, with collision suffixes); `--name`, when provided, is used as a name prefix.
- Annotations passed via `--annotation` are applied to every extracted material.
- The operation is atomic: either all qualifying entries are added or none.
- Extraction is guarded against zip bombs and path traversal: `--max-extract-entries` (default 10000) and `--max-extract-size` (default 1GiB) are enforced while streaming, and directory, symlink, absolute, and path-traversal entries are skipped or rejected.

With no `--kind`, and for non-archive values, behavior is unchanged. There are no server-side, contract, or gRPC changes.

## AI disclosure

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


javirln:
Superseded by the same-repo PR (branch pushed directly to chainloop-dev/chainloop so CI runs with full access).

PR #3254, opened 2026-06-30
feat(cli): explode archive materials in chainloop att add

javirln:
## Summary

`chainloop attestation add --kind <KIND> --value <archive>` now unpacks a `.zip`, `.tar`, `.tar.gz`, or `.tgz` for the kinds that have a meaningful "bundle of the same kind" archive form and records each contained regular file as its own material of the given kind, instead of requiring one `att add` invocation per file.

## Behavior

- Explosion applies only to an allowlist of kinds: `SBOM_CYCLONEDX_JSON`, `SBOM_SPDX_JSON`, and `SARIF`. For one of these, when the value resolves to a supported archive, each contained regular file is added as an individual material of that kind.
- Every other kind (e.g. `ARTIFACT`, `EVIDENCE`, `ZAP_DAST_ZIP`) records the archive whole, so a customer can still provide a regular zip as a single material.
- Per-entry material names are sequential and 0-indexed: `material-0`, `material-1`, … with no `--name`, or `<name>-0`, `<name>-1`, … when `--name` is provided (used as the prefix, sanitized to DNS-1123). Names already present in the attestation are skipped. The recorded artifact filename preserves the original entry basename.
- Annotations passed via `--annotation` are applied to every extracted material.
- The operation is atomic: either all qualifying entries are added or none.
- Extraction is guarded against zip bombs and path traversal: `--max-extract-entries` (default 10000) and `--max-extract-size` (default 1GiB) are enforced while streaming, and directory, symlink, absolute, and path-traversal entries are skipped or rejected.

For non-explodable kinds, non-archive values, and no `--kind`, behavior is unchanged. There are no server-side, contract, or gRPC changes.

## Assumptions and scope

- Explosion is limited to the SBOM/SARIF allowlist above; it is intentionally conservative so a regular zip provided for any other kind is never expanded. New kinds can be added to the allowlist as they gain a bundle form.
- Every extracted file is recorded under the same `--kind`; archives mixing material types are not split by type.
- Nested archives are not recursed — an archive contained inside the archive is treated as a regular file.
- `--policy-input-from-file` is ignored on the explode path (a warning is logged); policy inputs and evidence cross-links are not recorded for exploded materials.
- Zero-length entries are skipped and produce no material.
- Only stdlib-supported formats are handled (`.zip`, `.tar`, `.tar.gz`, `.tgz`); bzip2/xz are not supported in this iteration.

## AI disclosure

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


javirln:
## End-to-end examples

Verified by running the locally built CLI against a control plane (dry-run / local state). Server side is unchanged, so this needs only a CLI build.

### Explode a zip into one material per file

```
$ chainloop attestation add --kind SBOM_CYCLONEDX_JSON --value sboms.zip
INF material(s) added to attestation materials=3
```

`sboms.zip` containing `service-a.cdx.json`, `service-b.cdx.json`, `service-c.cdx.json` produces three independent materials, each crafted as `SBOM_CYCLONEDX_JSON` with its own digest:

```
Name    service-a-cdx-json   Type SBOM_CYCLONEDX_JSON   Digest sha256:304ad206…
Name    service-b-cdx-json   Type SBOM_CYCLONEDX_JSON   Digest sha256:51537faa…
Name    service-c-cdx-json   Type SBOM_CYCLONEDX_JSON   Digest sha256:6440a90c…
```

### Behaviors verified

| Command | Result |
|---|---|
| `att add --kind SBOM_CYCLONEDX_JSON --value sboms.zip` | 3 materials, each `SBOM_CYCLONEDX_JSON` |
| `att add --kind ARTIFACT --value collision.zip` (`scan.json`, `nested/scan.json`, `results.xml`) | names `scan-json`, `scan-json-1`, `results-xml` (basenames sanitized to DNS-1123, collisions suffixed) |
| `att add --kind ARTIFACT --name release --value bundle.tar.gz` | names `release-app-bin`, `release-config-yaml`, `release-readme-txt` (`--name` used as prefix) |
| `att add --kind ZAP_DAST_ZIP --value zap_scan.zip` | 1 material — archive-native kind recorded whole (allowlist short-circuit) |
| `att add --kind ARTIFACT --value app.bin` (not an archive) | 1 material — unchanged behavior |
| `att add --kind STRING --value "a literal string"` | 1 material — non-file values are not treated as archives |
| `att add --kind ARTIFACT --value many.zip --max-extract-entries 3` (8 entries) | `ERR adding materials from archive: … archive exceeds the maximum number of entries` (nothing added — atomic) |
| `att add --kind ARTIFACT --value evil.tar.gz` (entry `../escape.txt`) | `ERR adding materials from archive: … unsafe entry path in archive: "../escape.txt"` (nothing added — atomic) |

javirln:
## Additional examples — across material types

Each archive was built from the repo's own `materials/testdata` fixtures, so every entry is crafted and validated as the given `--kind`. Mix of `.zip` and `.tar.gz`.

| Command | Result | Derived materials |
|---|---|---|
| `att add --kind SBOM_CYCLONEDX_JSON --value sboms.zip` | 3 | `service-a-cdx-json`, `service-b-cdx-json`, `service-c-cdx-json` |
| `att add --kind SBOM_SPDX_JSON --value spdx.zip` | 3 | `sbom-spdx-json`, `container-json`, `described-json` |
| `att add --kind SARIF --value sarif.tar.gz` | 2 | `codeql-sarif`, `semgrep-sarif` |
| `att add --kind OPENAPI_SPEC --value openapi.zip` | 3 | `v2-0-yaml`, `v3-0-json`, `v3-1-yaml` (mixed OpenAPI versions, each validated) |
| `att add --kind GITLEAKS_JSON --value gitleaks.zip` | 2 | `clean-json`, `violations-json` |
| `att add --kind OPENVEX --value openvex.zip` | 2 | `vex-a-json`, `vex-b-json` |
| `att add --kind ATTESTATION --value attestation.tar.gz` | 2 | `dsse-json`, `bundle-json` |
| `att add --kind EVIDENCE --value evidence.zip` | 2 | `approval-txt`, `review-json` |

Each entry is crafted with the same `--kind` and validated independently. A single invalid entry fails the whole add atomically:

```
$ chainloop attestation add --kind SBOM_CYCLONEDX_JSON --value mixed.zip
# mixed.zip = one valid CycloneDX SBOM + one non-SBOM file
ERR adding materials from archive: expanding archive "mixed.zip": processing entry "bad.json": \
    staging entry "bad.json" as material "bad-json": crafting material: invalid cyclonedx sbom file: unexpected material type
# attestation status afterwards: 0 materials (atomic rollback)
```

migmartri:
Thanks, let's make the names based on the original zipped material name

javirln:
## Manual end-to-end verification

Run against the labs control plane + CAS (`devel/compose.labs.yml`) with a locally-built CLI, driving the real `chainloop attestation add` surface.

Fixture `evidence.zip`: `scan.json` (7B), `nested/report.sarif` (11B), `summary.txt` (2B), `empty.txt` (0B), `dir/` (directory entry).

```
chainloop attestation init --workflow wf --project proj
chainloop attestation add --kind ARTIFACT --value evidence.zip
```

Observed results:

- Explode (no `--name`) → 3 materials `material-0`, `material-1`, `material-2`; `empty.txt` and `dir/` skipped (5 entries → 3 materials). The material `Value` shows the original filenames (`scan.json`, `report.sarif`, `summary.txt`), confirming the recorded artifact name is preserved.
- `--name foo` → `foo-0`, `foo-1`, `foo-2`.
- `--annotation team=sec` → applied to every extracted material.
- `-o json` → a single valid JSON array of the materials (not concatenated top-level objects).
- Archive-native kind `--kind ZAP_DAST_ZIP` → single-material path (errors validating the whole zip as one ZAP report), i.e. not exploded.
- Guard `--max-extract-entries 1` on the 3-file zip → `archive exceeds the maximum number of entries`, and `attestation status` reports 0 materials, confirming atomic rollback.
- Collision-skip: adding into an attestation that already contained `material-1/2/3` produced `material-0`, `material-4`, `material-5`, skipping the in-use names.

migmartri (dismissed):

app/cli/cmd/attestation_add.go
```diff
@@ -74,6 +78,16 @@ func newAttestationAddCmd() *cobra.Command {
   chainloop attestation add --name sigcheck --value sigcheckResult.csv --kind SYSINTERNALS_SIGCHECK \
     --policy-input-from-file ignored_paths=exception.csv:Path`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			maxExtractSizeBytes, err := bytefmt.ToBytes(maxExtractSize)
```
migmartri, line 86: how do you know it's something to extract?

app/cli/cmd/attestation_add.go
```diff
@@ -74,6 +78,16 @@ func newAttestationAddCmd() *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("invalid --max-extract-size %q: %w", maxExtractSize, err)
+			}
+			// Guard against the uint64->int64 cast wrapping negative, which would
+			// later surface as a misleading "archive too large" error.
+			if maxExtractSizeBytes > math.MaxInt64 {
```
migmartri, line 92: this shoudl not apply to all materials just when we know taht we want to untar smth. Customers can provide a regular zip file.

app/cli/pkg/action/attestation_add.go
```diff
@@ -175,7 +220,21 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
...
 	}
 
-	return materialResult, nil
+	return []*AttestationStatusMaterial{materialResult}, nil
+}
+
+// shouldExplode decides whether an att-add should explode the value into many
+// materials: only when --kind is set, the value is a supported archive, and the
+// kind is not archive-native (e.g. ZAP_DAST_ZIP, which is recorded whole).
+func shouldExplode(materialType, value string) (materials.ArchiveFormat, bool, error) {
```
migmartri: I'd double check this one. This means that if I provide --kind EVIDENCE with a zip file will be marked as unzip it right?

pkg/attestation/crafter/materials/archive.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+}
+
+// archiveNativeKinds lists material kinds whose value is the archive itself.
+// For these, --kind short-circuits the explode path and the archive is
+// recorded whole. Extend this set as new "the archive is the material" kinds
+// are added.
+var archiveNativeKinds = map[string]struct{}{
+	schemaapi.CraftingSchema_Material_ZAP_DAST_ZIP.String(): {},
```
migmartri: This one is not covering evidence or artifact, which we don't want to unzip.

migmartri: It might be safer to do it the other way around, and support only  zip of SBOM, and SARIF for now

pkg/attestation/crafter/materials/archive.go
```diff
@@ -0,0 +1,356 @@
...
+		used[e] = struct{}{}
+	}
+	return &NameAllocator{used: used}
+}
+
+// AllocateSequential returns the next unused "<prefix>-<n>" material name, where
+// n is a zero-based counter that advances across calls and skips names already
+// in use. prefix is sanitized to DNS-1123; an empty or symbol-only prefix yields
+// the base "material" (so entries are named material-0, material-1, …).
+func (a *NameAllocator) AllocateSequential(prefix string) string {
```
migmartri, line 343: this might exist already

pkg/attestation/crafter/materials/archive.go
```diff
@@ -0,0 +1,356 @@
...
+		err = visit(f.Name, rc)
+		rc.Close()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func walkTar(p string, gzipped bool, visit func(name string, r io.Reader) error) error {
```
migmartri, line 232: we don't have tar support?

pkg/attestation/crafter/materials/archive.go
```diff
@@ -0,0 +1,356 @@
...
+// prefix such as "C:" or "c:/", which denotes an absolute path on Windows.
+func hasWindowsDriveLetter(name string) bool {
+	if len(name) < 2 || name[1] != ':' {
+		return false
+	}
+	c := name[0]
+	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')
+}
+
+func walkZip(p string, visit func(name string, r io.Reader) error) error {
```
migmartri, line 200: no zip support?

pkg/attestation/crafter/materials/archive.go
javirln: Good call — inverted in e358fb8f. Replaced the archive-native denylist with an explodable-kind allowlist: only SBOM_CYCLONEDX_JSON, SBOM_SPDX_JSON and SARIF explode; ARTIFACT, EVIDENCE, ZAP_DAST_ZIP and everything else record the archive whole, so a regular zip provided for a non-allowlisted kind is never expanded.

app/cli/pkg/action/attestation_add.go
javirln: Correct — that was the previous behavior. Fixed in e358fb8f: EVIDENCE (and any non-allowlisted kind) no longer explodes; only SBOM/SARIF do. Added a routing test asserting EVIDENCE + zip is recorded whole.

app/cli/cmd/attestation_add.go
javirln: Agreed. As of e358fb8f explosion only applies to the SBOM/SARIF allowlist; a regular zip for any other kind is stored as a single material.

app/cli/cmd/attestation_add.go
javirln: It no longer explodes on detection alone — only when the kind is in the SBOM/SARIF allowlist AND the value sniffs as an archive (e358fb8f).

pkg/attestation/crafter/materials/archive.go
javirln: Right — consolidated in e358fb8f. materials.SanitizeMaterialName is now the single sanitizer (returns the bare component, empty when nothing usable), and the action layer's sanitizeMaterialNamePart delegates to it with its own "input" fallback instead of duplicating the logic.

pkg/attestation/crafter/materials/archive.go
javirln: We do — walkZip handles .zip and walkTar handles .tar/.tar.gz/.tgz. There was no shared/exported archive walker to reuse (zap.go, helmchart.go and radamsa.go each inline their own private zip/tar reading), so WalkArchiveEntries is the new shared helper in the materials package. Happy to refactor those existing readers onto it as a follow-up if you'd like.

pkg/attestation/crafter/materials/archive.go
javirln: Tar is supported — walkTar covers .tar, .tar.gz and .tgz (gzip-wrapped). See the note on the zip thread re: reusing existing private readers.

pkg/attestation/crafter/materials/archive.go
migmartri: I mean if we don't have it already somewhere else

javirln: > 🤖 Automated reply posted by an AI assistant (Claude) running on @javirln's account — not hand-written.

Confirmed there's no shared/reusable archive walker in the codebase today. Every existing archive reader is a private, single-purpose function inlined in its own crafter, each tailored to locate one specific file rather than walk all entries:

- `zap.go:62` — `zip.OpenReader`, extracts a specific `report_json.json`
- `radamsa.go:171,191` — inline `zip.OpenReader` and `gzip`+`tar`
- `helmchart.go:94,100` — inline `gzip`+`tar`, reads `Chart.yaml`
- `junit/junit.go:89,123` — inline `tar` and `zip`, filters `*.xml`

None expose a reusable API. `WalkArchiveEntries` is the first shared, exported helper (zip + tar/tar.gz/tgz, with the zip-bomb size/entry caps and the path-traversal guard). Migrating those four onto it is a reasonable follow-up, but each would still need its bespoke "pick the right file" logic layered on top, so I kept that refactor out of this PR.

migmartri (approved):

PR #3255, opened 2026-06-30
fix(controlplane): fallback CAS downloads from invalid backends

### LINKED ISSUE #3190: Fall back to the org's default CAS backend on artifact download failures
## Problem

CAS mappings are recorded per backend. When a backend's credentials become invalid (e.g. AWS keys rotated by creating a *new* backend instead of updating the existing one), every artifact mapped to the old backend becomes undownloadable — even when the new, healthy backend points at the **same bucket** and the bytes are still there.

Real-world scenario:

1. Org has backend A (S3, bucket `my-bucket`). Artifacts uploaded → mappings point to A.
2. Credentials are rotated by creating backend B (same bucket `my-bucket`, new keys) and making it the default. Backend A is left with revoked keys (`InvalidAccessKeyId`).
3. Downloads of pre-rotation artifacts resolve their mapping to A and fail, even though B could serve the exact same content.

## Proposal

Extend the download path with a fallback mechanism, similar in spirit to the existing write-time `FindDefaultOrFallbackBackend` logic:

- When a download through the mapped backend fails (invalid credentials / validation status not OK), retry the download through the org's current **default** backend (and/or the fallback backend).
- Since storage is content-addressable, the digest verification we already perform guarantees correctness: if the default backend happens to contain the same content (same bucket or replicated storage), the download succeeds and is verified; otherwise it fails as before.
- A cheap heuristic could short-circuit the retry: prefer backends with the same `location` as the broken mapped one, or simply attempt the default backend when the mapped backend's `validationStatus` is invalid.

## Benefits

- Credential rotation done via a new backend (a common operational pattern) no longer strands historical artifacts.
- No data migration or mapping rewrite needed — content addressing makes the retry safe by construction.

waveywaves:
## Summary

- Fall back from an invalid mapped CAS backend to the organization's current default or fallback backend when minting download credentials.
- Apply the same backend selection to browser download redirects.
- Preserve the mapped backend when it remains valid.

Fixes #3190

Assisted-by: pi

### DIFF
--- a/app/controlplane/pkg/biz/casbackend.go
+++ b/app/controlplane/pkg/biz/casbackend.go
@@ -350,6 +350,30 @@ func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, o
 	return fallbackBackend, nil
 }
 
+// FindDownloadBackend returns the backend that should serve a mapped artifact download.
+//
+// CAS mappings are historical: digest -> backend ID. If credentials are rotated by
+// creating a new backend for the same storage location and making it the org default,
+// old mappings still point at the backend with revoked credentials.
+//
+// For that case, try the org's current default backend first, then the configured
+// fallback. This is safe because CAS downloads are verified by digest: if the chosen
+// backend does not contain the exact content, the download fails.
+//
+// The stored mapping is not changed; this only chooses credentials for this read.
+func (uc *CASBackendUseCase) FindDownloadBackend(ctx context.Context, mapped *CASBackend) (*CASBackend, error) {
+	if mapped == nil {
+		return nil, NewErrNotFound("CAS Backend")
+	}
+
+	if mapped.ValidationStatus == CASBackendValidationOK {
+		return mapped, nil
+	}
+
+	uc.logger.Infow("msg", "mapped CAS backend validation failed, attempting default/fallback", "backend", mapped.Name, "status", mapped.ValidationStatus, "orgID", mapped.OrganizationID)
+	return uc.FindDefaultOrFallbackBackend(ctx, mapped.OrganizationID.String())
+}
+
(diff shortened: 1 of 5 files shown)


waveywaves:
Updated tests for Cubic's edge-case feedback: nil mapped backend, missing default, missing fallback, and invalid fallback.

Validation:
```sh
SKIP_INTEGRATION=true go test ./app/controlplane/pkg/biz -run TestCASBackend -count=1
```

app/controlplane/pkg/biz/casbackend.go
```diff
@@ -350,6 +350,20 @@ func (uc *CASBackendUseCase) FindDefaultOrFallbackBackend(ctx context.Context, o
 	return fallbackBackend, nil
 }
 
+// FindDownloadBackend returns the mapped backend, or the org default/fallback when the mapped backend is invalid.
+func (uc *CASBackendUseCase) FindDownloadBackend(ctx context.Context, mapped *CASBackend) (*CASBackend, error) {
```
migmartri, line 364: I am not sure this meethod is correct since it checks default, not just fallback, do we want to also check default?

waveywaves: Yes, default-first is intentional here.

CAS mappings are historical (`digest -> backend ID`). The issue scenario is credential rotation by creating a new backend pointing at the same storage location and making it the org default, while old mappings still point at the backend with revoked credentials.

So for downloads, when the mapped backend is invalid, we try the org's current default backend first, then the configured fallback. This is safe because the CAS download path verifies the content digest; if the default backend does not contain the exact artifact, the download fails.

I added a code comment to make this intent explicit.

waveywaves: The replacement backend is typically made the organization default, so we need to check it before the fallback.

PR #3270, opened 2026-07-07
fix(dispatcher): improve fanout delivery timeouts and retry budget

### LINKED ISSUE #3269: Improve fanout webhook delivery timeouts and retry budget
## Problem

The fanout dispatcher's retry budget was 10 seconds total (`MaxElapsedTime`), with no per-attempt timeout — a single hung HTTP request could consume the entire budget in one attempt. The webhook plugin's `http.Client` also had no timeout set, and `context.TODO()` was passed through to `plugin.Execute`, so there was no hard cutoff other than TCP/OS defaults.

Additionally, the retry loop did not respect parent context cancellation — even if the parent context was cancelled, `backoff.RetryNotify` would keep retrying until `MaxElapsedTime` expired because the backoff was not wrapped with `backoff.WithContext`.

This is a partial fix for the reliability gaps described in chainloop-dev/chainloop#39 — specifically the timeout/retry improvements from Option 1 (improve the current in-process system).

## Solution

* **Bump** `MaxElapsedTime` **from 10s to 5m**: with the existing exponential backoff (500ms initial, ×1.5 multiplier, 60s max interval), this yields ~12-15 attempts — enough to ride out a transient blip, short enough not to leak goroutines in the fire-and-forget dispatch model.
* **Add per-attempt timeout (10s)**: each `Execute` call now runs under `context.WithTimeout(ctx, sdk.PerAttemptTimeout)` so a single hung request cannot eat the whole retry budget.
* **Set** `http.Client.Timeout` **on the webhook plugin**: the webhook `http.Client` now has `Timeout: sdk.PerAttemptTimeout` (10s), matching the dispatcher's per-attempt deadline. The cutoff is enforced at whichever layer sees the stall first.
* **Wrap backoff with parent context**: `backoff.WithContext(b, ctx)` ensures the retry loop stops promptly when the parent context is cancelled, rather than waiting for the full 5m budget.
* **Export** `sdk.PerAttemptTimeout`: shared constant in the plugin SDK so the dispatcher and plugins can't silently drift on the timeout value.
* **Bump webhook plugin version 1.1 → 1.2**: reflects the new per-attempt timeout behavior.
* **Replace deprecated** `io/ioutil` **with** `io` in the webhook plugin.

## Limitations

This is the in-process improvement (Option 1). The dispatch is still fire-and-forget in goroutines — not durable across restarts, no DLQ, delivery failures still only logged. A durable queue solution (NATS/River) is tracked separately in chainloop-dev/chainloop#39.

Refs chainloop-dev/chainloop#39

🤖 *Posted by Maximus bot (OpenCode) on behalf of @migmartri*

migmartri:
## Changes

- **Bump `MaxElapsedTime` from 10s to 5m**: ~12-15 retry attempts with the existing exponential backoff, enough to ride out transient blips without leaking goroutines in the fire-and-forget dispatch model.
- **Add per-attempt timeout (10s)**: each `Execute` call now runs under `context.WithTimeout` so a single hung HTTP request cannot consume the entire retry budget.
- **Set `http.Client.Timeout` on the webhook plugin**: matches the dispatcher's per-attempt deadline via a shared `sdk.PerAttemptTimeout` constant.
- **Wrap backoff with parent context**: `backoff.WithContext` ensures the retry loop stops promptly on parent context cancellation.
- **Bump webhook plugin version 1.1 → 1.2**: reflects the new per-attempt timeout behavior.
- **Replace deprecated `io/ioutil` with `io`** in the webhook plugin.

Closes #3269
Refs #39

Assisted-by: OpenCode

🤖 _Posted by Maximus bot (OpenCode) on behalf of @migmartri_

### DIFF
--- a/app/controlplane/internal/dispatcher/dispatcher.go
+++ b/app/controlplane/internal/dispatcher/dispatcher.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023 The Chainloop Authors.
+// Copyright 2023-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -47,6 +47,17 @@ type FanOutDispatcher struct {
 	l                   log.Logger
 	loaded              sdk.AvailablePlugins
 }
+
+// maxDispatchElapsedTime bounds the total time the dispatcher will keep
+// retrying a single integration delivery. With the default exponential
+// backoff (500ms initial, ×1.5 multiplier, 60s max interval) this yields
+// roughly 12-15 attempts — enough to ride out a transient blip or a brief
+// endpoint restart, while staying short enough not to leak goroutines in
+// the current in-process, fire-and-forget dispatch model.
+//
+// It is a var (not a const) so tests can shrink it to avoid waiting for
+// the real 5m budget. Production code must not mutate it.
+var maxDispatchElapsedTime = 5 * time.Minute
 
 func New(integrationUC *biz.IntegrationUseCase, wfUC *biz.WorkflowUseCase, wfRunUC *biz.WorkflowRunUseCase, creds credentials.ReaderWriter, c biz.CASClient, registered sdk.AvailablePlugins, l log.Logger) *FanOutDispatcher {
 	return &FanOutDispatcher{integrationUC, wfUC, wfRunUC, creds, c, servicelogger.ScopedHelper(l, "fanout-dispatcher"), l, registered}
@@ -280,7 +291,7 @@ func (d *FanOutDispatcher) loadInputs(ctx context.Context, queue dispatchQueue, 
 
 func dispatch(ctx context.Context, plugin sdk.FanOut, opts *sdk.ExecutionRequest, logger *log.Helper) error {
 	b := backoff.NewExponentialBackOff()
-	b.MaxElapsedTime = 10 * time.Second
+	b.MaxElapsedTime = maxDispatchElapsedTime
 
 	var inputType string
 	switch {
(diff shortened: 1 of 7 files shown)


app/controlplane/plugins/sdk/v1/fanout.go
```diff
@@ -33,6 +33,13 @@ import (
 
 const IntegrationKindFanOut = "fan-out"
 
+// PerAttemptTimeout is the hard per-attempt deadline the fanout dispatcher
+// enforces on each Execute call. FanOut plugins that make HTTP calls should
+// set their http.Client.Timeout to this value so a hung endpoint cannot
+// consume the entire retry budget in a single attempt — the cutoff is
+// enforced at whichever layer sees the stall first.
+const PerAttemptTimeout = 10 * time.Second
```
waveywaves: Can this stay out of the SDK? This makes control-plane retry policy part of the plugin SDK API. A local dispatcher/webhook timeout constant is boring duplication, but avoids coupling external plugin authors to this value.

app/controlplane/internal/dispatcher/dispatcher.go
```diff
@@ -299,16 +310,21 @@ func dispatch(ctx context.Context, plugin sdk.FanOut, opts *sdk.ExecutionRequest
...
 		func() error {
+			// Each attempt gets its own deadline so a single hung request
+			// cannot consume the entire retry budget.
+			attemptCtx, cancel := context.WithTimeout(ctx, sdk.PerAttemptTimeout)
+			defer cancel()
+
 			logger.Infow("msg", "executing integration", "integration", plugin.String(), "input", inputType)
 
-			err := plugin.Execute(ctx, opts)
+			err := plugin.Execute(attemptCtx, opts)
```
waveywaves: This only bounds plugins that honor the context. `webhook` does, but `dependency-track` currently ignores ctx and uses `http.DefaultClient` without a timeout, so this isn't a universal fanout attempt timeout yet. Should we update that plugin too or narrow the claim?

app/controlplane/internal/dispatcher/dispatcher.go
```diff
@@ -299,16 +310,21 @@ func dispatch(ctx context.Context, plugin sdk.FanOut, opts *sdk.ExecutionRequest
...
 			if err == nil {
 				logger.Infow("msg", "execution OK!", "integration", plugin.String(), "input", inputType)
 			}
 
 			return err
 		},
-		b,
+		backoff.WithContext(b, ctx),
```
waveywaves: This is good, but the production caller currently uses `dispatcher.Run(context.TODO(), ...)`, so cancellation won't fire there. Should the caller pass a real parent/app context, or should we avoid claiming production cancellation is fixed?

waveywaves (commented):
A few inline questions focused on keeping the fix tight and matching existing timeout patterns.

app/controlplane/plugins/sdk/v1/fanout.go
migmartri: Makes sense, thanks

jiparis (dismissed):

jiparis (approved):

PR #3283, opened 2026-07-11
refactor: remove the public attribute from workflows

matiasinsaurralde:
## Summary

Removes the workflow `public` feature from Chainloop. Public workflows previously made a workflow's runs, attestations, materials and CAS artifacts readable by any authenticated user across organizations. With this change, access to that evidence is organization-scoped only and honors project RBAC.

## What changes

- The `public` attribute is removed from workflows, including the `workflows.public` database column.
- Workflow-run and CAS-mapping lookups are now organization-strict; `WorkflowRunService.View` always applies project RBAC.
- The cross-organization public download fallback for CAS artifacts is removed.
- The public shared referrer index is retired, including the `DiscoverPublicShared` API and its configuration.
- The CLI `--public` flags (`workflow create`, `workflow update`, `referrer discover`) and the `Public` column in `workflow list` are removed.
- The corresponding API fields are removed and their tag numbers reserved; `WorkflowItem.public` is kept as a deprecated, always-false field for wire compatibility.
- A database migration drops the `public` column.

## Notes

This is an API and behavior change: existing public workflows become organization-only, the `DiscoverPublicShared` endpoint is removed, and the CLI `--public` flags no longer exist.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jiparis (dismissed):
thanks!!!

app/controlplane/pkg/data/ent/migrate/migrations/20260710231605.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "workflows" table
+ALTER TABLE "workflows" DROP COLUMN "public";
```
jiparis: Please do not drop the column in this very same PR, since it can break running pods that are still pointing to the old image. Instead, just deprecate it and make sure it's not used, we can run the migration in a follow up after a reasonable time period.

migmartri (approved):

PR #3292, opened 2026-07-16
feat(materials): record checkmarx engine types via scan.types annotation

javirln:
## Description

The Checkmarx One native JSON report (`CHECKMARX_JSON`) is multi-engine: a single report can bundle SAST, SCA, KICS (IaC) and container findings. At the attestation level, the crafter previously injected only `chainloop.material.tool.name = checkmarx`, so there was no way to tell which engines actually produced findings.

This change has the Checkmarx crafter record the distinct engine types present in the report as a new `chainloop.material.scan.types` annotation (lower-cased, sorted, comma-joined; e.g. `kics,sast,sca`). Annotation-only consumers can then distinguish engines without reading material content. A clean or null report advertises no types, so the annotation is omitted rather than set empty.

## AI disclosure

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3292?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/materials/checkmarx.go
+++ b/pkg/attestation/crafter/materials/checkmarx.go
@@ -19,7 +19,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"os"
+	"slices"
+	"strings"
 
 	schemaapi "github.com/chainloop-dev/chainloop/app/controlplane/api/workflowcontract/v1"
 	api "github.com/chainloop-dev/chainloop/pkg/attestation/crafter/api/attestation/v1"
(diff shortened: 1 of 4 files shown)


jiparis:
`kics,sast,sca`. Looks weird t me. It's mixing specific vendors with generic scan types. Do you have a real example? Should we extend this to all materials?

javirln:
> `kics,sast,sca`. Looks weird t me. It's mixing specific vendors with generic scan types. Do you have a real example? Should we extend this to all materials?

Yes it's weird but at the same time is how the tool outputs it. The weird thing is the tool mixes on the same report different angles, `kics,sast,sca`.

This is a real example: https://github.com/chainloop-dev/chainloop/blob/main/pkg/attestation/crafter/materials/testdata/checkmarx.json

I'll try to come up with something different

migmartri (approved):

PR #3293, opened 2026-07-16
fix(controlplane): authorize GroupService/ListProjects at the service layer

matiasinsaurralde:
Closes PFM-6695

## Summary

`GroupService/ListProjects` was the only group sub-resource handler that resolved results without a service-layer permission check, relying solely on the `visibleProjects` row filter. That filter returns `nil` ("RBAC not applied") for legacy `RoleViewer` users and for org-scoped API tokens, and the data layer treats `nil` as "no filter" — so those callers received the full group→project attachment map that the design restricts to org admins/maintainers.

This change adds the same `PolicyGroupListMemberships` guard the sibling handlers already use before resolving projects, matching the earlier `ListMembers` fix.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/controlplane/internal/service/group.go
+++ b/app/controlplane/internal/service/group.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2025 The Chainloop Authors.
+// Copyright 2025-2026 The Chainloop Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jiparis:
Current behaviour is fine: All users can list groups, and they can also see which of their projects are attached to those groups. User membership is hidden though.

PR #3314, opened 2026-07-29
feat: explode archive materials into named contract slots with a prefix policy selector

javirln:
## Summary

Makes `chainloop attestation add` usable with fixed, named contracts when a single logical input is delivered as an archive of many same-kind files (for example a bundle of SARIF reports). Today an archive can be exploded into many materials, but the resulting materials get generic auto-generated names and therefore cannot satisfy a contract's required, named material slot, nor be targeted by a name-based policy selector. This change closes that gap.

Three additive, backward-compatible pieces.

### 1. Stable, reproducible naming for exploded materials

When `--value` is an explodable archive (`.zip` / `.tar` / `.tar.gz`) for an explodable kind, the CLI now:

- sorts the archive entries by their normalized path, deterministically and independent of the order the archive was written in;
- names the first entry with the exact `--name`, and the rest `<name>-1`, `<name>-2`, …

The first entry therefore fills a required, named contract slot, and the file-to-name mapping is fully reproducible for a given archive (no timestamps or randomness).

```
chainloop attestation add --kind SARIF --name scan-report --value scans.zip
# -> scan-report, scan-report-1, scan-report-2, ...
```

### 2. The source archive is recorded as evidence

The original archive is recorded once as an `EVIDENCE` material and cross-linked with every exploded material in both directions via the `chainloop.material.references` annotation, so the source bundle itself is attested and traceable to the materials derived from it.

### 3. Prefix match mode for material selectors

Adds an additive `match_mode` to the contract material selector — `EXACT` (default) and `PREFIX` — on both standalone policy attachments (`PolicyAttachment.MaterialSelector`) and policy groups (`PolicyGroup.Material`). A single shared predicate drives both paths.

`PREFIX` matches the exact name or a `<name>-<suffix>` derivative, so a selector can target an entire exploded set:

```yaml
policies:
  materials:
    - ref: my-policy
      selector:
        name: scan-report
        match_mode: PREFIX   # matches scan-report, scan-report-1, ...
```

`PREFIX` is a literal prefix — it matches any material whose name begins with the value — so choose a discriminating prefix (typically the full slot name, e.g. `scan-report`, which matches `scan-report`, `scan-report-1`, …).

## How policies apply to the exploded set

A policy attachment applies to a material when the policy's declared kind matches the material's type **and** the selector, if present, matches the name. Evaluation is per material: N matching materials times M attachments produce N×M independent evaluations, each recorded against its own material. Exploding N reports into a prefix-selected slot therefore runs every attachment against every report. This also supports multi-slot contracts where each role has its own prefix selector (for example a build-time scan slot and a stripped-release scan slot): the two prefixes never cross-match, so each role's policies apply only to that role's materials.

## What this covers

- Explode into a required, named contract slot, and multi-slot contracts via per-role prefix selectors.
- `.zip`, `.tar`, and `.tar.gz`; deterministic ordering; `--name` collisions; user `--annotation`s propagated to every exploded material; archive-as-evidence cross-linking in both directions.
- Backward compatible: existing selectors with no `match_mode` keep exact-match behavior, and adding a single (non-archive) material is unchanged.

## What this does not cover

- Nested archives are not recursed: a `.zip` inside the archive is treated as a single regular file.
- Archive-native kinds (for example `ZAP_DAST_ZIP`) are still recorded whole, not exploded.
- If `--name` collides with a material already present in the attestation, the allocator never overwrites it, so the first exploded entry becomes `<name>-1` and will not fill an exact required slot named `<name>` in that case.
- No merging of file contents: each exploded entry stays its own material with its own digest.

## Compatibility

The proto change is purely additive (a new optional enum field with `UNSPECIFIED = 0` decoding to exact match). Contracts are persisted as raw body plus serialized proto, so no data migration is required and existing contracts deserialize unchanged.

AI assistance: Claude Code.


app/cli/pkg/action/attestation_add.go
```diff
@@ -160,13 +160,26 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
 	}
 	if format != materials.ArchiveNone {
 		if len(policyInputFiles) > 0 {
-			action.Logger.Warn().Msg("--policy-input-from-file is ignored when expanding an archive; evidence cross-links are not recorded for exploded materials")
+			action.Logger.Warn().Msg("--policy-input-from-file is ignored when expanding an archive")
```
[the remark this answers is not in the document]
javirln: Addressed in dc1ad8a5. Correct: the runtime inputs still flow into every exploded material's policy evaluation via addOpts, so "ignored" was wrong — only the per-input EVIDENCE materials are not recorded on the explode path. The warning now reads: "--policy-input-from-file values apply to policy evaluation but are not recorded as evidence materials when expanding an archive".

app/cli/pkg/action/attestation_add.go
```diff
@@ -160,13 +160,26 @@ func (action *AttestationAdd) Run(ctx context.Context, attestationID, materialNa
...
+
+		// Record the original archive once as an EVIDENCE material, cross-linked
+		// with every exploded material so the source bundle itself is attested.
+		// Note: AddMaterialsFromArchive has already committed the exploded
+		// materials, so this runs as a second commit. If it fails, the exploded
+		// materials remain persisted (they are complete and valid); re-running
+		// the same command is safe because explosion is deterministic — the same
+		// archive yields the same material names, so the retry overwrites
+		// identically and re-records the evidence.
+		if err := action.addSourceArchiveEvidence(ctx, crafter, attestationID, materialName, materialValue, mts, casBackend); err != nil {
```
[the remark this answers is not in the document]
javirln: Addressed in dc1ad8a5. Confirmed valid — the allocator seeds from existing names, so a retry after a failed evidence commit would have created a second exploded set rather than overwriting. Fixed by folding the source-archive EVIDENCE material and its bidirectional cross-links into AddMaterialsFromArchive's single atomic commit (behind a new WithSourceArchiveEvidence option the CLI opts into): if the evidence step fails, rollback() runs and nothing is persisted, so a retry re-explodes cleanly. The misleading "safe to re-run" comment was removed.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
```diff
@@ -295,6 +295,18 @@ message PolicyAttachment {
...
     string name = 1;
+
+    // How `name` is matched against a material's name. Defaults to exact match
+    // (UNSPECIFIED behaves as EXACT), so existing selectors are unchanged. Use
+    // PREFIX to target a set of materials sharing a name prefix (e.g. an
+    // archive exploded into `<name>`, `<name>-1`, `<name>-2`).
+    MatchMode match_mode = 2;
+
+    enum MatchMode {
+      MATCH_MODE_UNSPECIFIED = 0;
```
migmartri: I know that we do this because of linting, but shouldn't we look into not using prefixes like this for the sake of usability?

pkg/attestation/crafter/crafter.go
```diff
@@ -916,37 +941,83 @@ func (c *Crafter) AddMaterialsFromArchive(
...
+		}
+		stagedNames = append(stagedNames, archiveName)
+
+		// Reverse edge: point each exploded material back at the archive. They are
+		// freshly staged, so there are no prior references to merge.
+		for _, mt := range result {
+			if mt.Annotations == nil {
+				mt.Annotations = map[string]string{}
+			}
+			mt.Annotations[materials.AnnotationMaterialReferences] = archiveName
```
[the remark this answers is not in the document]
javirln: Addressed in 9bf834e5. Valid — the assignment clobbered any chainloop.material.references set via runtimeAnnotations. Extracted a shared materials.AppendReferences (preserves existing entries, dedupes) and routed both this reverse edge and the action-layer addReference through it, so the append semantics are now identical. Covered by a new TestAppendReferences.

app/controlplane/api/workflowcontract/v1/crafting_schema.proto
javirln: Good call — dropped the prefix in 063dacac. Values are now UNSPECIFIED / EXACT / PREFIX, so contracts read `match_mode: PREFIX`. ENUM_VALUE_PREFIX is already ignored file-wide (RunnerType/MaterialType are unprefixed too), and the bare UNSPECIFIED zero value needs a single inline `// buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX` on the enum (this module allows comment ignores; same pattern as api_token.proto). Regenerated pb.go/TS/jsonschema and updated the Go refs; wire compatibility is unchanged since values are matched by number.

By Claude

pkg/policies/policies.go
```diff
@@ -970,6 +970,21 @@ func (pv *PolicyVerifier) shouldApplyPolicy(ctx context.Context, policyAtt *v1.P
...
+// name filter under the given match mode. The default (unspecified) mode is an
+// exact match, so existing contracts are unaffected. PREFIX matches the exact
+// name or any "<name>-<suffix>" derivative — the shape produced by exploding an
+// archive into "<name>", "<name>-1", … — and deliberately NOT arbitrary
+// substrings, so a selector "scan" matches "scan" and "scan-1" but not "scanner".
+// It is the single predicate shared by standalone policy attachments
+// (shouldApplyPolicy) and policy groups (VerifyMaterial).
+func nameMatches(name string, mode v1.PolicyAttachment_MaterialSelector_MatchMode, materialID string) bool {
+	if mode == v1.PolicyAttachment_MaterialSelector_PREFIX {
+		return materialID == name || strings.HasPrefix(materialID, name+"-")
```
jiparis: I don't think we should include the `-`. User provides a prefix, which is not what's actually used, so it would lead to inconsistencies.

jiparis:
Looks great, @javirln . Just a nit comment.

pkg/policies/policies.go
javirln: Good call — changed to a literal prefix (strings.HasPrefix) in 327682f2. Agreed: the "-" boundary made the effective prefix differ from what the author writes, which is inconsistent. Authors now pick a discriminating prefix (typically the full slot name, e.g. scan-report). Multi-role separation still holds since distinct full-name prefixes do not cross-match; updated the tests and the PR description to match.

By Claude

jiparis (approved):

PR #3325, opened 2026-08-03
fix: build release images with matching target architecture

### LINKED ISSUE #3324: arm64 release images contain amd64 service binaries
## Problem

The `linux/arm64` release images for `control-plane` and `artifact-cas` contain `x86-64` binaries.

This causes the containers to fail on arm64 nodes with:

```text
exec /control-plane: exec format error
```

## Root Cause

The GoReleaser config builds these binaries only for `linux_amd64`:

- `control-plane`
- `artifact-cas`
- `chainloop-plugin-discord-webhook`
- `chainloop-plugin-smtp`
- `chainloop-plugin-dependency-track`

The same config also builds `linux/arm64` Docker images with buildx. Because the Docker entries do not select `goarch: arm64`, the arm64 image context receives the amd64 binaries.

## Impact

The published arm64 images are labeled as `linux/arm64`, but their entrypoint binaries are `x86-64`. These images cannot run on arm64 nodes.

## Expected Behavior

The arm64 release images must contain arm64 binaries.

## Verification

I verified the fix with:

```sh
goreleaser release --clean --snapshot --skip sign,sbom,publish
```

Then I inspected the rebuilt local arm64 images. The image metadata is `linux/arm64`, and the extracted binaries are `ARM aarch64`.

JeffResc:
## Summary

Fix the release image build for arm64.

The release config now builds the service binaries and copied plugins for `linux_arm64`. It also sets `goarch` on each Docker image entry. This makes GoReleaser use the correct binary set for each image architecture.

Closes #3324

## Root Cause

The arm64 Docker images used `--platform=linux/arm64`, but the Docker build context still received amd64 binaries. The service binaries and copied plugins were only built for `linux_amd64`. The Docker entries also did not select an architecture-specific artifact set.

## Changes

- Build `control-plane` for `linux_arm64`.
- Build `artifact-cas` for `linux_arm64`.
- Build the copied control-plane plugins for `linux_arm64`.
- Set `goarch: amd64` or `goarch: arm64` on Docker image entries.

## Verification

Ran:

```sh
goreleaser release --clean --snapshot --skip sign,sbom,publish
```

Verified that the rebuilt local arm64 images are labeled `linux/arm64`.

Extracted these binaries from the rebuilt arm64 images and verified each one is `ARM aarch64`:

- `/control-plane`
- `/artifact-cas`
- `/plugins/chainloop-plugin-dependency-track`
- `/plugins/chainloop-plugin-discord-webhook`
- `/plugins/chainloop-plugin-smtp`

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3325?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,6 +13,7 @@ builds:
       - -X main.Version={{ .Version }}
     targets:
       - linux_amd64
+      - linux_arm64
   - binary: artifact-cas
     id: artifact-cas
     main: ./app/artifact-cas/cmd
@@ -22,6 +23,7 @@ builds:
       - -X main.Version={{ .Version }}
     targets:
       - linux_amd64
+      - linux_arm64
   - binary: chainloop
     id: cli
     main: ./app/cli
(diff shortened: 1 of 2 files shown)


jiparis (approved):
Good catch, thanks for the fix.

PR #894, opened 2024-06-05
remove robot-account from CLI

migmartri:
- remove robot-account command, references and support from the CLI
- improvements on message, for example

```
ERR Chainloop Token required, please provide it via --token flag or CHAINLOOP_TOKEN environment variable

```

Refs #891

PR #890, opened 2024-06-05
chore: do not create robot-account automatically

migmartri:
```
chainloop wf create --name do-not-create --project bar                    
┌──────────────────────────────────────┬───────────────┬─────────┬─────────────────────┬────────┬─────────────────┐
│ ID                                   │ NAME          │ PROJECT │ CREATED AT          │ RUNNER │ LAST RUN STATUS │
├──────────────────────────────────────┼───────────────┼─────────┼─────────────────────┼────────┼─────────────────┤
│ cc1fa9b0-9db4-4d0a-b1fd-b4744b67bf43 │ do-not-create │ bar     │ 05 Jun 24 12:24 UTC │        │                 │
└──────────────────────────────────────┴───────────────┴─────────┴─────────────────────┴────────┴─────────────────┘
INF To Attest this workflow you'll need to provide an API token. See "chainloop organization api-token" command for more information.
```

closes #889 
refs https://github.com/chainloop-dev/chainloop/issues/891

PR #3326, opened 2026-08-04
fix(controlplane): withhold org-admin-only policies from project-scoped API tokens and remove RobotAccountService

migmartri:
Project-scoped API tokens were minted with two capabilities that no human role holds.

`DefaultAuthzPolicies` granted `robot_account:create` and `registered_integration:create` to every token regardless of scope. Neither policy appears in any role in `RolesMap`, so for users these operations are reachable only through the blanket `IsAdmin()` bypass in the authz middleware — effectively org owner/admin only.

Token authorization is an ACL matched on resource and action; it never reads the token's project. And the two endpoints behind those policies resolve their target by organization, not project: `RobotAccountService/Create` validated the requested workflow with `GetOrgScoped` and discarded the resolved `ProjectID`, while `IntegrationsService/Register` creates an org-level object. Neither called `authorizeResource`.

`chainloop organization api-token create --project my-project` was the whole setup. A project admin holds `api_token:create` for their own project, so the token they got back carried two policies they do not hold themselves — and nothing compares a token's granted policies against what its creator possesses. They could then create a **non-expiring** robot-account credential for *another* project's workflow and push forged attestations into it.

## Change

The two capabilities are handled differently, because their situations differ.

**`RobotAccountService` is removed entirely** — the whole proto (Create/List/Revoke) with its generated stubs, the service handler, `RobotAccountUseCase.Create`/`Revoke`, the matching repo methods, both authz policies (`robot_account:create` and `robot_account:list`) and their `ServerOperationsMap`/`RolesMap` entries. Robot accounts are superseded by workflow-scoped API tokens, which give the same per-workflow granularity plus a finite expiration, revocation coverage and an explicit policy set. Creation left the CLI in 2024 (#890, #894) and has had no caller in any Chainloop component since — no reference in the CLI, the frontend or the platform backend, only generated stubs. Removing the service eliminates the escalation at the root instead of restricting who may reach it.

This also makes that half **retroactive**: tokens minted before the fix still carry `robot_account:create` in their stored policy list, but no operation maps to it any more, so the grant is inert without rotating anything.

**Existing robot-account JWTs keep working for attestation.** The JWT provider, the attestation middleware and a slimmed `FindByID`-only use case/repo stay, so in-flight CI using a legacy robot account is unaffected. One deliberate trade-off: with the management API gone, revoking a compromised legacy account now requires a direct DB update (`revoked_at`) — the retirement path for these credentials is switching to a workflow-scoped API token.

**The `integration_registered` policies move into `orgLevelTokenPolicies`**, granted to org-level tokens only: `create`, `list` and `read`. Registering an integration is a legitimate org-level operation with a live caller (`chainloop integration registered add`), and the two read endpoints behind `list`/`read` — `ListRegistrations` and `DescribeRegistration` — enumerate and read any registered integration in the organization, including its configuration, with no project filter. Only org-scoped roles (viewer, contributor) hold them; no project role does. All three stay available to org-level tokens and are withheld from project-scoped ones.

Supporting changes: `DefaultAuthzPolicies` is extracted to a package-level `defaultAuthzPolicies` so the invariant is assertable without constructing the use case; the org-level grant uses `slices.Concat` so the shared backing array is never written to; the orphaned `authConf` field and `RobotAccount.JWT` are gone (wire regenerated); and the generated jsonschema files for the removed messages are deleted.

## Scope of the escalation — corrected

An earlier draft of this description claimed registering an org-level integration causes it to receive every attestation in the organization. **That is wrong.** Fan-out dispatch delivers only to integrations *attached* to a given workflow (`dispatcher.go:166`), and `IntegrationsService/Attach` does enforce project scope (`integration.go:135`). A project-scoped token cannot route another project's attestations to itself.

So the integration half is a vertical escalation — creating and configuring an org-level object a project admin has no business creating — and the precondition for the separately-tracked SSRF-via-registration finding. It is not a cross-project data-exfiltration path. The cross-project impact, and the substance of this issue, is the robot-account half: forging supply-chain evidence for another project's workflow.

## Tests

- `TestDefaultTokenPoliciesAreHeldByAHumanRole` — no default token policy may be absent from every role in `RolesMap`. This is a static stand-in for the missing invariant that granted policies be a subset of the granter's own.
- `TestOrgLevelPoliciesStayOutOfDefaults` — the two sets stay disjoint, backstopping the above if a project-level role is ever granted one of these policies.

Both were confirmed to fail when a policy is put back into the default set. The existing `TestEveryProcedureIsClassified` covers the classification bookkeeping, including its stale-entry guard. The robot-account integration test seeds through the ent client and covers the `FindByID` semantics the attestation middleware relies on: found, not-found returning `(nil, nil)`, and revoked accounts surfacing `RevokedAt`.

## Compatibility

**Wire-breaking:** a pre-2024 client still calling any `RobotAccountService` RPC gets `Unimplemented` instead of a credential. Nothing in this repo or the platform repo calls it, but it is a public gRPC API, so a self-hosted user's own automation could. The proto deletion is waived in `buf.yaml` (`FILE_NO_DELETE`) for the same reason the earlier `RPC_NO_DELETE` exception was needed — a maintainer decision on the exception stands.

`chainloop integration registered add`, `list` and `describe` become org-level-token-only — the one behaviour change a user is likely to notice. Attaching integrations to workflows is unaffected: the attach flow calls `Attach` directly and never needs to list or read registrations.

**Rotate project-scoped tokens after upgrading.** The robot-account capability is neutralised regardless, but a pre-existing project-scoped token still carries the three `integration_registered` policies in its stored list.

## Follow-up not included here

`IntegrationsService.Register` still performs no project-scope check. More broadly, token minting is delegation with no delegation check — nothing verifies that granted policies are a subset of the caller's own. Closing that would address the class rather than these two instances.


matiasinsaurralde (dismissed):

app/controlplane/pkg/biz/apitoken.go
```diff
@@ -38,12 +38,40 @@ type APITokenJWTConfig struct {
...
+	authz.PolicyArtifactDownload, authz.PolicyReferrerRead,
+	authz.PolicyOrganizationRead,
+
+	// to attach integrations
+	authz.PolicyAvailableIntegrationRead,
+	authz.PolicyAvailableIntegrationList,
+	authz.PolicyRegisteredIntegrationList,
+	authz.PolicyRegisteredIntegrationRead,
+	authz.PolicyAttachedIntegrationList,
+	authz.PolicyAttachedIntegrationAttach,
```
jiparis, line 71: should this one also be removed? Attach can be done at different levels.

migmartri: move registrations to admin ones indeed since they are attached to orgs

matiasinsaurralde (approved):

PR #3338, opened 2026-08-10
feat(cli): add flags to override policy inputs (replace, not append)

javirln:
Adds a way to supply a policy input at `chainloop attestation add` time that replaces the contract-declared value, which the existing append-only `--policy-input-from-file` cannot do.

- `--policy-input [<policy>:]<input>=<value>` sets a policy input to an inline literal value that replaces any contract-declared value for that key. This makes it possible to override a scalar policy input at run time (for example lowering `min_iterations` for a single run) instead of turning it into a multi-value list.
- `--policy-input-from-file` is unchanged: its values are still appended to the contract value.

Both support an optional `<policy>:` prefix to scope an input to a single policy (matched by name or ref); otherwise the input applies to every declaring policy. When several scopes match the same policy, the most specific one wins deterministically.

Also adds `--append`, a flag reserved for a future release: it is accepted but has no effect yet, and will later control whether `--policy-input` and `--policy-input-from-file` append to rather than replace the contract-declared value.

Closes PFM-6906.

AI assistance: this contribution was implemented with Claude Code.

### DIFF
--- a/app/cli/pkg/action/attestation_add.go
+++ b/app/cli/pkg/action/attestation_add.go
@@ -273,6 +273,12 @@ func withSourceArchiveEvidence(opts []crafter.AddOpt) []crafter.AddOpt {
 // contract value instead of appending, keeping a scalar override a scalar. As
 // with contract-declared arguments, individual append values must not embed
 // those delimiters; path globs, the intended use, never do.
+//
+// Precedence for the same input+scope: file inputs are applied first and inline
+// --policy-input values last, so an inline value deterministically wins over a
+// --policy-input-from-file-replace for the same key. (Cobra exposes each
+// repeatable flag as its own slice with no cross-flag ordering, so this fixed
+// precedence — rather than raw CLI argument order — is what we can guarantee.)
 func buildRuntimeInputs(policyInputFiles []*PolicyInputFromFile, policyInputs []*PolicyInput) (*policies.RuntimeInputs, error) {
 	if len(policyInputFiles) == 0 && len(policyInputs) == 0 {
 		return nil, nil
--- a/app/cli/pkg/action/attestation_add.go
+++ b/app/cli/pkg/action/attestation_add.go
@@ -265,20 +265,13 @@ func withSourceArchiveEvidence(opts []crafter.AddOpt) []crafter.AddOpt {
 
 // buildRuntimeInputs reads each policy input file and combines it with the
 // inline --policy-input values, returning them grouped for the policy engine.
-// Append-mode file inputs (--policy-input-from-file) go under Global/Scoped and
-// are newline-joined via policies.MergeRuntimeInputs so repeated inputs merge
-// using the multi-value encoding the engine expects (it splits inputs back on
-// newlines and commas). Replace-mode file inputs (--policy-input-from-file-replace)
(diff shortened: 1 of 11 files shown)


pkg/policies/policies.go
```diff
@@ -400,13 +403,26 @@ func (pv *PolicyVerifier) evaluatePolicyAttachment(ctx context.Context, attachme
 	// Record which runtime inputs actually applied to this policy (i.e. made it
 	// into the computed args because the policy declares them). The values
 	// themselves live in `with`; this only flags the overridden input names.
+	// Skipped entirely on the common path with no runtime inputs to avoid a
+	// per-attachment map allocation.
 	var runtimeInputOverrides []string
-	for k := range effectiveRuntime {
-		if _, ok := args[k]; ok {
+	if len(appendRuntime) > 0 || len(replaceRuntime) > 0 {
```
jiparis, line 411: I think this if block could be replaced by a couple of `slices.Collect(maps.Keys(m))`. There is no need of those for loops if I'm not wrong.

app/cli/pkg/action/attestation_add.go
```diff
@@ -262,43 +263,77 @@ func withSourceArchiveEvidence(opts []crafter.AddOpt) []crafter.AddOpt {
...
+// append — has no meaningful accumulation.
+func addOverrideInput(ri *policies.RuntimeInputs, policy, input, value string) {
+	if policy == "" {
+		ri.GlobalOverride[input] = value
+		return
+	}
+	if ri.ScopedOverride[policy] == nil {
+		ri.ScopedOverride[policy] = map[string]string{}
+	}
+	ri.ScopedOverride[policy][input] = value
```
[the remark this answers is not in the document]
javirln: _Posted by Claude Code (an AI agent) on behalf of @javirln._

Already addressed. The non-determinism was the map range in `forPolicy`; it now resolves matching scopes through `matchingScopes`, which applies them **most-specific-last** (digest/scheme/provider-qualified > bare name, ties broken lexically), so when both a policy name and its ref match the same attachment the merged override value is stable regardless of Go's map iteration order — the more specific scope wins.

Covered by `TestRuntimeInputsForPolicy/most_specific_scoped_override_wins_deterministically` (looped 50x to defeat map randomization) and `TestMatchingScopesOrder`. Fixed in e2ca5965 and e4401c9f.

pkg/policies/policies.go
javirln: _Posted by Claude Code (an AI agent) on behalf of @javirln._

Good call — done in 5dbf491c. Replaced the manual collect-and-sort loop with `slices.Sorted(maps.Keys(overrideNames))` and collapsed the two per-map loops into a single `range []map[string]string{appendRuntime, replaceRuntime}`.

One loop stays, though: the `args[k]` membership check. `ComputeArguments` drops runtime keys the policy doesn't declare (the "argument ... not defined in policy spec, ignoring it" path), and this field is meant to report only inputs that actually applied, so a bare `slices.Collect(maps.Keys(appendRuntime))` would include undeclared keys and could also duplicate a key present in both maps. Net -4 lines with the filter/dedup preserved.

jiparis (approved):

PR #3341, opened 2026-08-11
fix(crafter): bound AccessChk memory use during policy evaluation

javirln:
## Summary

The `SYSINTERNALS_ACCESSCHK` material type could drive the CLI to very high peak memory during client-side policy evaluation, risking an OOM kill of the CI runner for large materials. This bounds that memory use.

- Stream the AccessChk parser line by line instead of building a normalized full-text copy and splitting all lines at once, and stop pinning the whole input in memory via stored substrings.
- Size-gate the verbatim raw-text fallback fields (`raw` and per-object `raw_lines`): retain them for inputs below a threshold and omit them above it, keeping the JSON projection handed to the policy engine from ballooning to multiples of the original file size. These fields are not attested and are unused by current policies, so recorded evidence and evaluations are unchanged.
- Make AccessChk output detection independent of the `raw` field so oversized descriptor-only materials are still recognized.
- Warn at craft time when a material exceeds the retention threshold.

Closes PFM-6922.

## AI assistance

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3341?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/materials/accesschk.go
+++ b/pkg/attestation/crafter/materials/accesschk.go
@@ -62,6 +62,18 @@ func (i *AccessChkCrafter) Craft(ctx context.Context, filePath string) (*api.Att
 		return nil, fmt.Errorf("input does not look like accesschk output: %w", ErrInvalidMaterialType)
 	}
 
+	// The material is attested by digest and stored as-is, but at policy
+	// evaluation time it is projected to JSON client-side on the runner. Above
+	// this size the parser omits the verbatim raw-text fallback to keep peak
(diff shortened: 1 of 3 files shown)


jiparis (approved):

matiasinsaurralde (approved):
LGTM

PR #3343, opened 2026-08-12
fix(artifact-cas): stream uploads to object-store backends to bound CAS memory

javirln:
## Summary

The Artifact CAS bytestream Write RPC buffered the entire artifact in memory before flushing it to the storage backend, so CAS memory grew roughly 1:1 with artifact size and a large upload OOM-killed the CAS pod mid-stream.

Object-store backends (S3, S3 access point, Azure Blob) now advertise a streaming capability and are fed directly from the client stream through an `io.Pipe`, so peak CAS memory stays bounded regardless of artifact size. The OCI backend keeps the fully-buffered path because its push implementation needs the whole layer content up front.

A backend upload failure is always surfaced as an internal error and is never misclassified as a client disconnect. The per-upload size cap is enforced on both the streaming and buffered paths. The download path, including digest verification, is unchanged.

Resolves PFM-6923.

## AI assistance

Claude Code was used in producing this change.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/artifact-cas/internal/service/bytestream.go
+++ b/app/artifact-cas/internal/service/bytestream.go
@@ -115,29 +115,33 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
 	}
 
 	s.log.Infow("msg", "artifact does not exist, uploading", "digest", req.resource.Digest, "name", req.resource.FileName)
-	// Create a buffer that will be filled in the background before sending its content to the backend
-	buffer := newStreamReader(info.MaxBytes)
-	// Add data from first request
-	if err = buffer.Write(req.GetData()); err != nil {
-		if backend.IsUploadSizeExceeded(err) {
-			return status.Error(codes.ResourceExhausted, err.Error())
-		}
-		return sl.LogAndMaskErr(err, s.log)
+
+	// Streaming-capable backends (object stores such as S3/Azure) are fed
+	// directly from the client stream through an io.Pipe, so CAS memory stays
+	// bounded by the chunk/pipe size regardless of artifact size (PFM-6923).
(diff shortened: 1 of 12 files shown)


app/artifact-cas/internal/service/bytestream.go
```diff
@@ -148,22 +152,198 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
...
+// total number of bytes committed to the backend.
+//
+// Tradeoff vs. the buffered path: because bytes are forwarded as they arrive, an
+// over-cap upload has already streamed up to maxBytes to the backend by the time
+// the size guard trips (the client is still correctly rejected with
+// ResourceExhausted). For multipart backends the SDK aborts the in-flight upload
+// best-effort; operators should keep a bucket lifecycle rule to reap any
+// incomplete multipart uploads a failed abort could leave behind. This is
+// inherent to streaming — the alternative (buffer-then-validate) is exactly the
+// OOM this change removes.
```
jiparis: Please remove this comment, it's irrelevant

jiparis (dismissed):
Looks great, please check my comment.
Thanks!

jiparis (approved):

matiasinsaurralde (approved):

PR #3344, opened 2026-08-12
perf(crafter): stream AccessChk material instead of buffering it whole

javirln:
## Summary

Streams the Sysinternals AccessChk material through the crafter instead of reading the whole file into memory.

`accesschk.Parse` now takes an `io.Reader` and parses the input line by line, so the crafter opens the file with `os.Open` rather than `os.ReadFile`, removing the full-file buffer held at craft time. The size-gated raw-text fallback is preserved by tracking the byte count as the input streams and dropping the verbatim `Raw`/`RawLines` fields once `RawRetentionLimit` is crossed; `Report.RawOmitted` reports when that happens so the crafter can warn that string-matching policies will not see those fields for the material. The policy-evaluation projection wraps the in-memory material bytes with a reader, matching the existing text-material paths.

## AI assistance

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
+++ b/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
@@ -239,7 +239,7 @@ func (m *Attestation_Material) ingestMaterialToJSON(rawMaterial []byte, value st
 		// AccessChk emits plain text; project it to JSON so the policy engine,
 		// which only consumes JSON, can evaluate it. The raw text is preserved
 		// in the projection's "raw" field for string-matching fallbacks.
-		report, err := accesschk.Parse(rawMaterial)
+		report, err := accesschk.Parse(bytes.NewReader(rawMaterial))
 		if err != nil {
 			return nil, fmt.Errorf("invalid accesschk material: %w", err)
 		}
--- a/pkg/attestation/crafter/materials/accesschk.go
+++ b/pkg/attestation/crafter/materials/accesschk.go
@@ -44,34 +44,24 @@ func NewAccessChkCrafter(schema *schemaapi.CraftingSchema_Material, backend *cas
 }
 
 func (i *AccessChkCrafter) Craft(ctx context.Context, filePath string) (*api.Attestation_Material, error) {
-	data, err := os.ReadFile(filePath)
+	f, err := os.Open(filePath)
 	if err != nil {
 		return nil, fmt.Errorf("can't open the file: %w", err)
 	}
+	defer func() { _ = f.Close() }()
 
 	// Soft fingerprint: AccessChk emits free-form text, so we only require that
 	// the input is valid text that resembles AccessChk output (a banner, at
-	// least one access entry, or an SDDL/descriptor marker). The raw text is
-	// stored unchanged; it is projected to JSON later for policy evaluation.
-	report, err := accesschk.Parse(data)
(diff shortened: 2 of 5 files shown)


pkg/attestation/crafter/materials/accesschk.go
```diff
@@ -64,14 +66,15 @@ func (i *AccessChkCrafter) Craft(ctx context.Context, filePath string) (*api.Att
...
-			Msg("large AccessChk material: raw-text fallback fields are omitted from the policy input to limit memory use")
+	// RawRetentionLimit the parser omits the verbatim raw-text fallback to keep
+	// peak memory bounded; warn so an operator relying on string-matching
+	// policies is aware the fallback fields will not be present for this material.
+	if report.RawOmitted() {
+		ev := i.logger.Warn().Int("threshold", accesschk.RawRetentionLimit)
+		if fi, statErr := f.Stat(); statErr == nil {
+			ev = ev.Int64("size", fi.Size())
+		}
+		ev.Msg("large AccessChk material: raw-text fallback fields are omitted from the policy input to limit memory use")
```
jiparis: This has nothing to do with policies, does it? policy evaluations use GetEvaluableContent which doesn't call this Craft method.

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -239,7 +239,7 @@ func (m *Attestation_Material) ingestMaterialToJSON(rawMaterial []byte, value st
 		// AccessChk emits plain text; project it to JSON so the policy engine,
 		// which only consumes JSON, can evaluate it. The raw text is preserved
 		// in the projection's "raw" field for string-matching fallbacks.
-		report, err := accesschk.Parse(rawMaterial)
+		report, err := accesschk.Parse(bytes.NewReader(rawMaterial))
```
jiparis, line 242: keep in mind that `rawMaterial` comes from `rawMaterial, err = os.ReadFile(value)` so this is probably not saving anything at evaluation time.

pkg/attestation/crafter/materials/accesschk.go
javirln: You're right — this warning is describing evaluation-time behaviour from the craft path, which doesn't drive policy evaluation (that goes through `GetEvaluableContent`). Removed it here; the operator-facing notice about the raw-text fallback belongs at the evaluation layer and will come with a separate change that bounds evaluation-time memory.

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
javirln: Agreed — `rawMaterial` is already fully read via `os.ReadFile(value)` before this point, so wrapping it in `bytes.NewReader` saves nothing at evaluation time. This change is only adapting the call to the new `io.Reader` signature of `Parse`; the memory win here is craft-time only. Bounding evaluation-time memory (the projection into the OPA input) is a separate change.

jiparis (approved):

PR #3345, opened 2026-08-12
fix(crafter): release go-git packfile descriptors after repository reads

matiasinsaurralde:
Closes PFM-6929

## Summary

Since the go-git v6 alpha.5 bump (#3336), every `PlainOpen*` call builds its own 256-entry descriptor pool holding roughly 3 descriptors per packfile, freed only on LRU eviction — which never fires below 256 entries. alpha.4 released them right after each read.

The crafter opens the repository up to twice per `attestation init` and never closed the storer, so those descriptors stayed open until the process exited. They are now released through `storer.IdleReleaser` once the reads complete; descriptors reopen lazily, so the repository remains usable.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/cioverride.go
+++ b/pkg/attestation/crafter/cioverride.go
@@ -95,6 +95,8 @@ func overrideHeadWithPRCommit(headCommit *HeadCommit, path, actualSHA string, lo
 		return
 	}
 
+	defer releaseGitDescriptors(repo, logger)
+
 	hash := plumbing.NewHash(actualSHA)
 	commit, err := repo.CommitObject(hash)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jiparis (approved):
thanks!

PR #3348, opened 2026-08-13
ci: publish GitHub release as a draft until post-release steps succeed

javirln:
## Summary

The release workflow used to publish the GitHub release immediately via GoReleaser and then keep adding post-release artifacts (SBOMs, vulnerability reports, Chainloop attestation link, source-code attestation, release-note edits). When one of those post-release steps failed, the release was already public but incomplete.

This makes the release atomic from a user's perspective:

- `.goreleaser.yml`: the `release:` block now sets `draft: true`, so GoReleaser creates the release as a draft.
- `.github/workflows/release.yaml`: all existing post-release steps run against the draft (they reference it by tag). A final `success()`-gated **Publish release** step flips the draft to published only after every post-release step has succeeded.

If any post-release step fails, the release stays a draft for a human to inspect or retry, rather than a broken release going public.

Refs PFM-6883.

Assisted-by: Claude Code

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -275,6 +275,16 @@ jobs:
           # Update the release notes and ignore if it fails since we might be lacking permissions to update the release notes
           gh release edit $RELEASE_TAG -n "$modified_notes" || echo -n "Not enough permissions to edit the release notes. Skipping..."
 
+      # GoReleaser publishes the release as a draft (see release.draft in .goreleaser.yml).
+      # Flip it to published only after every post-release step above has succeeded so a
+      # failure leaves the release as a draft for a human to inspect/retry rather than
+      # publishing a broken release.
+      - name: Publish release
+        if: ${{ success() }}
+        env:
+          RELEASE_TAG: ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


javirln:
Good catch — fixed in 69eb523d. The `Include source code on attestation` step ran during the draft window and relied on `gh release download -A tar.gz`, whose auto-generated source asset isn't reliably available for a draft. Switched it to fetch the tarball directly from the tag archive URL (`https://github.com/${GITHUB_REPOSITORY}/archive/refs/tags/${RELEASE_TAG}.tar.gz`), which depends only on the git tag and yields the same auto-generated archive, so it works before publish.

migmartri:
Do we want this? it's in draft

PR #3364, opened 2026-08-24
perf(crafter): de-duplicate AccessChk security descriptors in the policy projection

javirln:
The `SYSINTERNALS_ACCESSCHK` material is projected to JSON client-side for policy evaluation. AccessChk evidence for a registry hive or service database applies a small number of distinct security descriptors to hundreds of thousands of objects through inheritance, and the flat projection repeats each descriptor inline on every object. The policy engine materializes every copy in memory, so a large material drives the CLI to very high peak memory and can OOM-kill the CI runner during evaluation.

This projects the material to a de-duplicated shape: the distinct security descriptors are emitted once in a `descriptors` table, and each object references one by index (`descriptor`). For a large registry material this reduces the projection by roughly 20x (hundreds of thousands of objects collapse to a few hundred distinct descriptors) and brings evaluation peak memory down by an order of magnitude. No object, name, or ACE is dropped — descriptors are shared only when byte-for-byte identical — so recorded evidence and policy findings are unchanged.

Policies read a descriptor via `input.descriptors[obj.descriptor]`.

Closes PFM-6922.

## AI assistance

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
+++ b/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
@@ -239,11 +239,23 @@ func (m *Attestation_Material) ingestMaterialToJSON(rawMaterial []byte, value st
 		// AccessChk emits plain text; project it to JSON so the policy engine,
 		// which only consumes JSON, can evaluate it. The raw text is preserved
 		// in the projection's "raw" field for string-matching fallbacks.
+		//
+		// The projection de-duplicates security descriptors: a registry hive or
+		// service database applies a handful of distinct descriptors to hundreds
+		// of thousands of objects, and repeating each one inline would balloon the
+		// document the policy engine holds in memory (large materials have
+		// OOM-killed CI runners). Objects reference a shared descriptors table by
+		// index instead; no object, name, or ACE is dropped, so policy findings
+		// are unchanged. Policies read a descriptor via input.descriptors[obj.descriptor].
 		report, err := accesschk.Parse(bytes.NewReader(rawMaterial))
 		if err != nil {
 			return nil, fmt.Errorf("invalid accesschk material: %w", err)
 		}
-		return json.Marshal(report)
+		projection, err := report.Project()
+		if err != nil {
+			return nil, fmt.Errorf("failed to project accesschk material: %w", err)
+		}
+		return json.Marshal(projection)
 	case v1.CraftingSchema_Material_CERTCC_DRANZER:
 		// dranzer emits plain text; project it to JSON so the policy engine,
 		// which only consumes JSON, can evaluate it. The raw text is preserved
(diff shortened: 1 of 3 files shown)


pkg/attestation/crafter/materials/accesschk/projection_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// dedup happened: 3 objects, 2 distinct descriptors.
+	assert.Len(t, proj.Objects, 3)
+	assert.Len(t, proj.Descriptors, 2)
+	assert.Equal(t, proj.Objects[0].Descriptor, proj.Objects[1].Descriptor, "identical descriptors must share an index")
+	assert.NotEqual(t, proj.Objects[0].Descriptor, proj.Objects[2].Descriptor)
+
+	// every object reconstructs to its original descriptor, byte-for-byte.
+	for i, po := range proj.Objects {
+		orig := report.Objects[i]
+		d := proj.Descriptors[po.Descriptor]
```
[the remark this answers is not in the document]
javirln: Good call. Added TestProjectJSONShape in 318c577b: it marshals the projection and asserts the wire contract — the descriptors table, objects referencing a descriptor index (and no inline dacl), the retained/omitted raw_lines fallback, and the top-level raw field.

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -239,11 +239,23 @@ func (m *Attestation_Material) ingestMaterialToJSON(rawMaterial []byte, value st
...
+		// document the policy engine holds in memory (large materials have
+		// OOM-killed CI runners). Objects reference a shared descriptors table by
+		// index instead; no object, name, or ACE is dropped, so policy findings
+		// are unchanged. Policies read a descriptor via input.descriptors[obj.descriptor].
 		report, err := accesschk.Parse(bytes.NewReader(rawMaterial))
 		if err != nil {
 			return nil, fmt.Errorf("invalid accesschk material: %w", err)
 		}
-		return json.Marshal(report)
+		projection, err := report.Project()
```
jiparis, line 254: any reason to keep as a separate step from Parse? could we make `Parse` return the "projection" already?

javirln: Separate because the crafter consumes the flat `Report` for validation/annotations; folding the projection into `Parse` would couple parsing with the eval projection

jiparis (approved):

Issue #2518, opened 2025-11-05
chainloop artifact upload seems to fail for S3 buckets


javirln:
We have seen a weird behaviour when trying to upload some artifacts to a S3 bucket. Some artifacts report the following:
```
$ chainloop artifact upload --file ~/Downloads/chainloop-linux-amd64
INF uploading chainloop-linux-amd64 - sha256:44c7fa3a3545c29c8c990699c3209aeaef1d71f3773a0b53d7a5ca9d3910f56b
 ... done! [89.84MB in 7.713s; 11.65MB/s]
ERR server error
```

The error on the CAS side is something like:
```json
{"level":"error","ts":1762342756.7378519,"component":"service","msg":"failed to upload to bucket: upload multipart failed, upload id: THE-ID, cause: operation error S3: CompleteMultipartUpload, https response error StatusCode: 400, RequestID: OTHER-ID, HostID: HOST-ID, api error BadDigest: The sha256 you specified did not match the calculated checksum."...
```

Uploading to OCI seems to work fine:
```
$ chainloop artifact upload --file ~/Downloads/chainloop-linux-amd64
 ... done! [89.84MB in 10.1s; 8.90MB/s]
```

migmartri:
reproduced locally

migmartri:
reverting to a previous commit to the AWS SDK upgrade makes it work, so it was indeed introduced there

```
git revert 7b72e14dc1206e1541cc78785dd6fb1b1ac6709f

mkfile -n 10M temp_10GB_file

go run main.go artifact upload -f temp_10GB_file
WRN API contacted in insecure mode
WRN API contacted in insecure mode
INF uploading temp_10GB_file - sha256:e5b844cc57f57094ea4585e235f36c78c1cd222262bb89d53c94dcb4d6b3e55d
 ... done! [10.49MB in 11.896s; 881.40KB/s]
```

migmartri:
The problem is related to the checksum check 

```

	if b.checksumVerificationEnabled() {
		// Check that the object is uploaded correctly
		input.ChecksumSHA256 = aws.String(hexSha256ToBinaryB64(resource.Digest))
	}

```

and it only happens if the file is uploaded as multipart, meaning it's bigger than 5MB

jiparis:
This seems related https://github.com/aws/aws-sdk-js-v3/issues/6742, using `PutObject` from `client-s3` instead of `Upload` might be a solution

migmartri:
[...]

PR #3365, opened 2026-08-25
feat(artifact-cas): verify uploaded content against the declared digest

javirln:
The Artifact CAS keyed stored artifacts by their client-declared SHA256 without ever
checking that the uploaded bytes hash to that digest. Only the OCI backend verified, via
its own layer-digest check. On S3, S3 access point and Azure Blob any identity holding
`artifact:upload` could store arbitrary content under an arbitrary digest key.

The CAS now streams each upload to a local staging file, hashes it while receiving, and
hands the backend a file only once its digest matches the one the client declared. A
mismatch is rejected with `InvalidArgument` and nothing reaches the backend.

## What changed

- A single spill-then-upload path shared by all backends, replacing the per-backend
  streaming branch and the `StreamingUploader` interface. Verification is no longer
  something each backend implements, and any future backend inherits it.
- Staging files are unlinked as soon as they are created and used through the open
  descriptor, so the kernel reclaims the space whenever the process exits, including when
  it is killed mid-upload. Nothing needs to sweep or reap the staging volume.
- `staging_dir` is required: the CAS refuses to start without one and proves it is
  writable before serving traffic, so a misconfiguration fails at boot rather than
  producing a healthy service that rejects every upload.
- Helm: a dedicated staging emptyDir mounted unconditionally at a fixed path. It sets no
  `sizeLimit`, since exceeding one evicts the Pod rather than failing the upload that
  overran it.
- Staging on disk keeps the service's own memory use independent of artifact size. The
  OCI backend still requires the whole layer in memory, unchanged here.
- The S3 backend relies on this service-layer verification rather than a per-object
  checksum precondition, which cannot be expressed for multipart uploads.

Addresses PFM-6988.

This contribution was produced with the assistance of Claude Code.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3365?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


app/artifact-cas/internal/service/staging.go
```diff
@@ -0,0 +1,61 @@
+//
```
migmartri, line 1: this should probably not be in service package

deployment/chainloop/templates/cas/configmap.yaml
```diff
@@ -14,6 +14,11 @@ metadata:
   {{- end }}
 data:
   server.yaml: |
+    {{- if .Values.cas.staging.enabled }}
```
migmartri, line 17: what should this be configurable? I'd say this is mandatory and the only mechanism

app/artifact-cas/internal/service/bytestream.go
```diff
@@ -165,91 +168,69 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
...
-	}
-	// A backend failure is wrapped so the caller always masks it, never mistaking
-	// a backend-side reset/cancellation for a client disconnect.
-	if uploadErr != nil {
-		return 0, &backendUploadError{uploadErr}
+	s.log.Infow("msg", "artifact verified, uploading now to backend", "name", req.resource.FileName, "digest", req.resource.Digest, "size", size)
+	// IMPORTANT: hand the *os.File to Upload unwrapped. Wrapping it (io.TeeReader,
+	// io.LimitReader, a progress reader) hides io.ReaderAt/io.Seeker and silently
+	// forces the object-store SDK back onto in-memory multipart buffering.
+	if err := storageBackend.Upload(ctx, f, req.resource); err != nil {
```
migmartri, line 223: does this still keep the whole file outside of memory?

app/artifact-cas/internal/service/bytestream.go
```diff
@@ -165,91 +168,69 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
...
+// staging disk while computing its SHA256, verifies the computed digest against
+// the client-declared one, and only then hands the verified file to the backend
+// for storage under the canonical key. Unverified bytes never reach the backend,
+// so the canonical key can never hold content that does not hash to its
+// digest. CAS memory stays bounded because the artifact lives on disk, and
+// the *os.File handed to Upload lets object-store SDKs stream it in bounded parts
+// via io.ReaderAt/io.Seeker rather than buffering it in memory.
+//
+// It returns the number of bytes committed. A digest mismatch is returned as a
+// *digestMismatchError; a backend Upload failure as a *backendUploadError; spill
```
migmartri, line 181: this comment is little bit useless it would be more useful if it explains what it does

app/artifact-cas/cmd/main.go
```diff
@@ -152,6 +159,31 @@ func newProtoValidator() (protovalidate.Validator, error) {
...
+// prepareStagingDir resolves the upload staging directory (falling back to the
+// OS temp dir when unconfigured — dev only; production mounts a dedicated
+// volume), creates it, and removes any leftover staging files from a previous
+// run. It must use the same directory the service is configured with (see
+// serviceOpts / conf.staging_dir).
+func prepareStagingDir(bc *conf.Bootstrap, logger log.Logger) error {
+	dir := bc.GetStagingDir()
+	if dir == "" {
+		dir = os.TempDir()
+		_ = logger.Log(log.LevelWarn, "msg", "staging_dir not configured, falling back to OS temp dir (dev only)", "dir", dir)
```
migmartri, line 171: fail here do not fallback

app/artifact-cas/cmd/main.go
```diff
@@ -152,6 +159,31 @@ func newProtoValidator() (protovalidate.Validator, error) {
...
+	}
+
+	if err := os.MkdirAll(dir, 0o700); err != nil {
```
migmartri, line 174: will this work on scrach container?

app/artifact-cas/cmd/main.go
```diff
@@ -152,6 +159,31 @@ func newProtoValidator() (protovalidate.Validator, error) {
...
+		return err
+	}
+
+	if _, err := service.SweepStagingDir(dir, servicelogger.ScopedHelper(logger, "staging")); err != nil {
```
migmartri, line 178: `staging` is overloaded, what about `staging directory`

migmartri, line 178: why are you sweeping on boot that might cause problems booting, I'd keep startup process as limited as possible

pkg/blobmanager/s3/backend.go
```diff
@@ -46,19 +46,10 @@ type Backend struct {
...
-)
+var _ backend.UploaderDownloader = (*Backend)(nil)
 
 const defaultRegion = "us-east-1"
 
-// SupportsStreaming reports that the s3 backend can upload directly from a
-// streaming reader. The AWS SDK's manager.Uploader consumes the reader in
-// bounded-size parts (multipart upload), so CAS never needs to buffer the whole
-// artifact in memory.
-func (b *Backend) SupportsStreaming() bool { return true }
```
migmartri, line 60: what changed that now this has been removed?

deployment/chainloop/templates/cas/deployment.yaml
```diff
@@ -126,6 +126,12 @@ spec:
               mountPath: "/data/conf"
             - name: jwt-public-key
               mountPath: "/tmp"
+            {{- if .Values.cas.staging.enabled }}
+            # Writable scratch volume for staging + verifying uploads before they
+            # reach the backend (the container root filesystem is read-only).
+            - name: staging
```
migmartri, line 132: can we just make it /tmp?

deployment/chainloop/templates/cas/deployment.yaml
```diff
@@ -158,6 +164,12 @@ spec:
         - name: jwt-public-key
           secret:
             secretName: {{ include "chainloop.cas.fullname" . }}-jwt-public-key
+        {{- if .Values.cas.staging.enabled }}
+        # Node-disk (NOT tmpfs/RAM) scratch space for staging + verifying uploads.
+        - name: staging
+          emptyDir:
+            sizeLimit: {{ .Values.cas.staging.sizeLimit }}
```
migmartri, line 171: why configure this? what happens if it goes over?

migmartri (commented):
In general it looks good to me, see my comments, some thoughts though

- I'd make this mode to be the only mode, not configurable, if it's going to be tmpDir that's ok
- What happened to the streaming capabilities you added in the other PR, why is it no longer needed? can we confirm that during the end to end input->stagingFS->blobFS we never load the whole file in memory?

deployment/chainloop/templates/cas/deployment.yaml
```diff
@@ -126,6 +126,12 @@ spec:
               mountPath: "/data/conf"
             - name: jwt-public-key
               mountPath: "/tmp"
+            {{- if .Values.cas.staging.enabled }}
+            # Writable scratch volume for staging + verifying uploads before they
+            # reach the backend (the container root filesystem is read-only).
+            - name: staging
+              mountPath: {{ .Values.cas.staging.mountPath | quote }}
```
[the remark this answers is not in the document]
migmartri: this is a good mount, we might need another one, i.e /tmp-staging-fs

javirln:
> Posted by Claude Code on behalf of @javirln.

## Verification

**Unit / integration** (`app/artifact-cas`, `pkg/blobmanager`)
- Digest mismatch rejected with `InvalidArgument`, nothing sent to the backend.
- Staging cleanup on every exit path, table-driven: success, digest mismatch, backend
  failure, size cap exceeded, client disconnect.
- `staging_dir` startup contract: unconfigured rejected, missing directory created,
  existing-but-unwritable rejected, probe file removed.
- Upload refused when the staging dir is unset, so `os.CreateTemp` cannot silently fall
  back to the OS temp dir.

**End to end against a real S3 (MinIO, testcontainers, runs in CI)**
- 12 MB multipart upload: the object in the bucket re-hashes to the declared digest.
- Tampered digest: `InvalidArgument`, and `NoSuchKey` at the canonical key.

**End to end with the real CLI against a locally built control plane + CAS + MinIO**
- 25 MB and 1 GB uploads stored with an exact digest match; download round-trip identical.
- Memory stays flat: peak RSS +11.6 MB for a 250 MB artifact, +29 MB for a 1 GB artifact.
- 1 GB upload observed spilling to disk, then reclaimed.

**End to end on a kind cluster, chart-deployed, scratch image built from this branch**
- Pod runs with `readOnlyRootFilesystem: true`, `runAsUser: 1001`, the jwt secret at
  `/tmp` and the staging emptyDir alongside it; 40 MB upload verified in the bucket.
- Removing `staging_dir` from the deployed config crash-loops the pod with
  `staging_dir is required` instead of producing a healthy pod that fails every upload.

**Interrupted uploads**
- Client killed mid-upload: partial staging file released in 12 ms, nothing written to the
  bucket.
- CAS killed mid-upload: before unlinking, a 403 MB partial survived a restart; after,
  the directory holds nothing and the space is reclaimed by the kernel.

app/artifact-cas/internal/service/staging.go
javirln: > Claude Code, on behalf of @javirln.

Gone entirely - the boot sweep is removed, and `staging.go` with it.

deployment/chainloop/templates/cas/configmap.yaml
javirln: > Claude Code, on behalf of @javirln.

Agreed, it is mandatory now. It was worse than an odd option: with `enabled=false` the
fallback was `os.TempDir()` -> `/tmp`, which on this deployment is the read-only
jwt-public-key secret mount. `MkdirAll` succeeded because the directory exists, so the CAS
booted healthy and failed every upload. `cas.staging.enabled` and the configurable
mountPath are both gone; the emptyDir is mounted unconditionally at a fixed path.

app/artifact-cas/internal/service/bytestream.go
javirln: > Claude Code, on behalf of @javirln.

Yes for the object stores, measured rather than argued: peak RSS +11.6 MB for a 250 MB
upload and +29 MB for a 1 GB one. The `*os.File` is passed to `Upload` unwrapped so the
SDKs keep `ReaderAt`/`Seeker` and read it in bounded parts; wrapping it would silently
push them back to in-memory buffering.

Not for OCI: `oci/backend.go` still does `io.ReadAll` and `static.NewLayer` holds the
content again, because go-containerregistry needs the full uncompressed layer up front.
Unchanged by this PR - that is the PFM-6924 work.

app/artifact-cas/internal/service/bytestream.go
javirln: > Claude Code, on behalf of @javirln.

Rewritten - 13 lines down to 9, leading with what the function does and keeping only the
error contract the caller switches on. The SDK/`ReaderAt` rationale moved to the `Upload`
call site where it is actionable.

It also claimed "CAS memory stays bounded because the artifact lives on disk", which is
not true for the OCI backend. It now says the staging keeps this service's memory
independent of artifact size, which is what is actually guaranteed.

app/artifact-cas/cmd/main.go
javirln: > Claude Code, on behalf of @javirln.

Agreed - and there were three fallbacks, not one:

1. this one in `main.go`,
2. `newCommonService`, which defaulted `stagingDir` to `os.TempDir()` independently,
3. `os.CreateTemp` itself, which treats an empty dir as the OS temp dir - so removing (2)
   alone would have quietly reinstated the same behaviour.

All three are gone: an unset `staging_dir` is a startup error, the service keeps no
default, and `spillVerifyUpload` refuses to stage when the directory is unset. Startup
also probes for writability, since `MkdirAll` succeeds on a directory that exists but
cannot be written to. `config.devel.yaml` sets `staging_dir` so local dev still works.

app/artifact-cas/cmd/main.go
javirln: > Claude Code, on behalf of @javirln.

Verified on a kind cluster with a scratch image built from this branch.

With the emptyDir mounted: yes. The pod runs with `readOnlyRootFilesystem: true` and
`runAsUser: 1001`, and reaching Running means the staging setup and its writability probe
both succeeded. A 40 MB upload through the CLI landed in the bucket with a matching digest.

With the old fallback: no. Scratch has no `/tmp` of its own, the root filesystem is
read-only, and the `/tmp` that exists is the read-only secret mount - which is why your
"fail, do not fall back" note was right. Removing `staging_dir` from the deployed config
now crash-loops the pod with `staging_dir is required` rather than booting broken.

app/artifact-cas/cmd/main.go
javirln: > Claude Code, on behalf of @javirln.

Moot now - removing the boot sweep removed the only scoped logger, so there is no
`staging` scope left to name.

javirln: > Claude Code, on behalf of @javirln.

Removed, and there is nothing left for a sweep to do.

It did cover something real, so it was worth checking first. The emptyDir is scoped to the
Pod, not the container - "a container crashing does not remove a Pod from a node. The data
in an emptyDir volume is safe across container crashes" - so a file stranded by a crash
survives an in-place restart. I reproduced it: SIGKILL the CAS mid-upload and a 403 MB
partial sat there across a restart.

Rather than add a reaper, staging files are now unlinked as soon as they are created and
used through the open descriptor. Content stays readable and seekable, so backends are
unaffected, but the kernel frees the space when the descriptor goes away - including on
SIGKILL. Same experiment after the change: nothing in the directory, the artifact visible
only as `cas-upload-NNN (deleted)`, and the space reclaimed on kill. That removes the need
for a sweep rather than deferring it, which keeps startup as limited as you asked.

One caveat: there is a ~4 us window between creating the file and unlinking it (measured
mean 4.3 us, worst 39 us). A SIGKILL inside it strands an inode, but the window closes
before any bytes are written, so what is stranded is empty. `O_TMPFILE` would close it
completely; I skipped it because it needs a fallback for non-Linux dev machines and
filesystems without support, which seemed a poor trade against a zero-byte exposure. Happy
to add it if you disagree.

pkg/blobmanager/s3/backend.go
javirln: > Claude Code, on behalf of @javirln.

Nothing about S3 changed - the capability stopped selecting anything. `SupportsStreaming`
had one caller: the branch in `Write` choosing between the piped-streaming path and the
buffered one. Verification has to happen before any bytes reach a backend, so every upload
spills to the staging file first and there is now exactly one path; the flag had no
remaining caller.

The streaming benefit is kept: `Upload` still gets a seekable `*os.File` and
`manager.Uploader` reads it in bounded parts.

deployment/chainloop/templates/cas/deployment.yaml
javirln: > Claude Code, on behalf of @javirln.

No - `/tmp` is already the jwt-public-key secret mount a few lines up, so a second mount
there collides. Went with your `/tmp-staging-fs` suggestion from the other thread, and the
path is no longer configurable.

deployment/chainloop/templates/cas/deployment.yaml
javirln: > Claude Code, on behalf of @javirln.

It gets the Pod evicted, which is the wrong answer to "this upload is too big", so it is
removed. Exceeding an emptyDir `sizeLimit` makes the kubelet mark the Pod for eviction
rather than failing the request that overran it.

Related gap worth flagging: the bound meant to prevent this is the per-request cap from
the token, and `checkUploadSize` treats `maxSize == 0` as unlimited, so a token issued
without a cap has nothing bounding staging disk. Sizing is now left to the node's ephemeral
storage; we can add `resources.requests.ephemeral-storage` on the CAS container if you want
it attributable.

deployment/chainloop/templates/cas/deployment.yaml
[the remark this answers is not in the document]
javirln: > Claude Code, on behalf of @javirln.

Adopted - the mount path is `/tmp-staging-fs`, and it is fixed rather than configurable,
which also makes the `fail` guard cubic suggested unnecessary.

javirln:
> Claude Code, on behalf of @javirln.

@migmartri - the two points from your review, answered in the threads but summarised here:

**Make staging the only mode.** Done. `cas.staging.enabled` and the configurable mount path
are both gone; the emptyDir is mounted unconditionally at a fixed path, and `staging_dir` is
required rather than falling back. It was worse than a preference: the fallback resolved to
`/tmp`, which is the read-only secret mount, so the CAS booted healthy and failed every
upload.

**What happened to streaming.** `SupportsStreaming` had a single caller - the branch
choosing between the piped path and the buffered one. Verification must happen before any
bytes reach a backend, so there is now one path and the flag had nothing left to select.
Streaming is not lost: the backend still gets a seekable `*os.File` and reads it in bounded
parts.

**End to end memory.** Bounded for the object stores, measured: +11.6 MB peak RSS for a
250 MB artifact, +29 MB for 1 GB. The exception is OCI, which still reads the whole layer
into memory - unchanged here, that is PFM-6924.

app/artifact-cas/internal/service/service.go
```diff
@@ -38,6 +38,10 @@ type commonService struct {
 	backends backend.Providers
 	// best-effort audit events publisher, nil-safe
 	audit *AuditDispatcher
+	// stagingDir is the local directory where uploads are staged on disk while
+	// their SHA256 is verified against the declared digest before reaching the
+	// backend. It must be writable; when unset it defaults to the OS temp dir.
```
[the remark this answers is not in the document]
javirln: > Claude Code, on behalf of @javirln.

Correct - that comment described the default I removed. Fixed: the field now documents that
an unset value is a deployment error and uploads are refused, matching `WithStagingDir` and
`spillVerifyUpload`.

app/artifact-cas/internal/service/bytestream.go
```diff
@@ -116,28 +117,30 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
...
 	//     when it wraps a network reset/cancellation originating backend-side —
 	//     doing so would falsely report success and silently drop the artifact.
-	//   - A stream-read (feed) error is classified: a client disconnect is not a
+	//   - A stream-read (spill) error is classified: a client disconnect is not a
 	//     failure, an exceeded size cap maps to ResourceExhausted, anything else
-	//     is masked.
+	//     (e.g. a staging-disk write failure) is masked.
 	if err != nil {
+		var mismatch *digestMismatchError
+		if errors.As(err, &mismatch) {
```
jiparis: nit comment: note that since go 1.26 these error checks can be replaced by a single line `errors.AsType[*digestMismatchError](err)`

[https://go.dev/doc/go1.26#errorspkgerrors](<https://go.dev/doc/go1.26#errorspkgerrors>)

app/artifact-cas/internal/service/bytestream.go
```diff
@@ -165,91 +168,97 @@ func (s *ByteStreamService) Write(stream bytestream.ByteStream_WriteServer) erro
...
+	// process is killed outright.
+	//
+	// This is what keeps the staging volume bounded. A deferred remove only runs
+	// when the handler returns, so a SIGKILL or an OOM kill mid-upload would
+	// strand a partial artifact on the volume, and nothing would ever reclaim it:
+	// the emptyDir backing it outlives container restarts and is cleared only
+	// when the Pod is removed from the node. Unlinking up front means an
+	// interrupted upload cannot leave anything behind, whatever kills us, so no
+	// sweep or reaper is needed to keep the volume from filling up.
+	if err := os.Remove(f.Name()); err != nil {
```
jiparis, line 205: The comment says it's removing the folder, but here it's removing the temp file itself, right?

pkg/blobmanager/s3/backend.go
```diff
@@ -173,10 +164,12 @@ func (b *Backend) Upload(ctx context.Context, r io.Reader, resource *pb.CASResou
 		},
 	}
 
-	// if b.checksumVerificationEnabled() {
-	// 	// Check that the object is uploaded correctly
-	// 	input.ChecksumSHA256 = aws.String(hexSha256ToBinaryB64(resource.Digest))
-	// }
+	// No per-object ChecksumSHA256 precondition is set here: a whole-object SHA256
```
jiparis: I think this comment is useless here, or at least it doesn't belong here.

jiparis (dismissed):
Looks good to me. Thanks @javirln

app/artifact-cas/internal/service/bytestream.go
javirln: > Claude Code, on behalf of @javirln.

Done, both checks. Note this is the repo's first `errors.AsType` - the other 55 call sites
still use `errors.As`, so happy to revert if you'd rather not set the precedent here.

app/artifact-cas/internal/service/bytestream.go
javirln: > Claude Code, on behalf of @javirln.

Right, it removes the temp file, not the folder. I'd written "directory entry", meaning the
file's name - precise but evidently misleading. Reworded to say the file itself.

pkg/blobmanager/s3/backend.go
javirln: > Claude Code, on behalf of @javirln.

Half agreed. Dropped the part about the CAS service verifying - that's service-layer
knowledge that has no business in a backend.

Kept the S3 constraint itself, trimmed to two lines: a whole-object SHA256 can't be
expressed for a multipart upload, and R2 rejects it. That's backend-local, and setting this
field is what caused #2518, so it's cheap insurance against someone re-adding it. Say the
word and I'll drop it entirely.

jiparis (approved):

deployment/chainloop/values.yaml
```diff
@@ -1478,7 +1478,21 @@ cas:
       drop: ["ALL"]
     seccompProfile:
       type: "RuntimeDefault"
-  
+
+  ## CAS upload staging.
```
migmartri, line 1482: this is outdated no?

PR #3368, opened 2026-08-26
feat(crafter): redact secrets from AI coding sessions before CAS upload

jiparis:
AI coding session evidence carries the raw conversation transcript in `data.raw_session`, which routinely includes whatever the agent read or printed, credentials included. That content was stored in the CAS verbatim.

Secrets are now detected and replaced before the material is stored.

The file on disk is left untouched and remains what policies are evaluated against, so a secret-detection policy still sees what the agent actually captured. The recorded digest therefore describes the redacted artifact rather than the local file, which a `chainloop.material.redacted` annotation makes explicit; the same annotation is what keeps the policy input consistent on inline CAS backends, where the stored content is otherwise also the policy input.

Any scan failure in `attestation add` stop the process rather storing unscanned content. 
Added a CLI option `--skip-secret-redaction` is an explicit opt-out, recorded in the attestation as `chainloop.material.redaction.skipped` so a policy can reject it. What was redacted is published as `chainloop.material.redaction.count` and `chainloop.material.redaction.rules`.

Detection lives behind a `Scanner` interface in a new `internal/redaction` package, which detects against the document rendered as indented JSON and then rewrites the decoded value tree, so the output cannot be malformed JSON. A convergence loop re-scans after each rewrite, making idempotency a property of the algorithm. Every string leaf is eligible except an explicit set of identifiers, timestamps and file paths the platform joins records on.

Detection uses **betterleaks**, which also replaces gitleaks as the library backing the `GITLEAKS_JSON` material parser; the report formats are wire-compatible. `github.com/zricethezav/gitleaks/v8` is dropped, and the net effect is two fewer modules.

Assisted by Claude Code.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/redaction/redaction.go
```diff
@@ -0,0 +1,482 @@
...
+//
+// The size cap and the timeout are related. A successful redaction costs two
+// passes over the document: one that finds the secrets and one that confirms
+// none are left. Both together run at roughly 2.5 MB/s (measured by
+// BenchmarkRedact on an M3 Pro; the cost is regex matching, and rendering the
+// document is negligible beside it), so a document at the cap takes on the order
+// of 20s of real work. The timeout sits well above that deliberately: it is
+// there to stop pathological backtracking, not to bound ordinary work.
+const (
+	DefaultMaxBytes  = 24 << 20
```
migmartri: what happens if the file is too big or if the timeout is reached?

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -53,6 +57,22 @@ var (
 	AnnotationsSBOMMainComponentName    = CreateAnnotation("material.sbom.main_component.name")
 	AnnotationsSBOMMainComponentType    = CreateAnnotation("material.sbom.main_component.type")
 	AnnotationsSBOMMainComponentVersion = CreateAnnotation("material.sbom.main_component.version")
+
+	// AnnotationMaterialRedacted marks a material whose stored content was
+	// transformed by its crafter before upload, to strip secrets out of it. Two
+	// things follow from it: the recorded digest describes the redacted artifact
+	// rather than the file on disk, and policy evaluation must read the untouched
+	// local file instead of the stored content (see GetEvaluableContent).
+	AnnotationMaterialRedacted = CreateAnnotation("material.redacted")
```
migmartri, line 66: nice

pkg/attestation/crafter/materials/chainloop_ai_coding_session.go
```diff
@@ -83,11 +106,18 @@ func (c *ChainloopAICodingSessionCrafter) Craft(ctx context.Context, artifactPat
 		return nil, fmt.Errorf("AI coding session validation failed: %w", err)
 	}
 
-	material, err := uploadAndCraft(ctx, c.input, c.backend, artifactPath, c.logger)
+	craftOpts, report, err := c.redact(ctx, f)
```
migmartri, line 109: has this already happen after the policy evaluations?

migmartri (dismissed):

pkg/attestation/crafter/materials/chainloop_ai_coding_session.go
jiparis: Policy evaluations use the raw file contents (it's read directly from the file). 

What's uploaded to CAS is the redacted version of it (replaced using the `craftOpts.contentOverride` contents)

internal/redaction/redaction.go
jiparis: It fails the `attestation add` operation. We might want to remove these limits for now. AI coding sessions are usually not that big (24MB is a generous limit though)

pkg/attestation/crafter/materials/aicodingsession/redact.go
```diff
@@ -0,0 +1,138 @@
...
+// its original digest. The result is guaranteed to still validate against the AI
+// coding session schema; if it does not, redaction fails rather than uploading
+// either an invalid document or an unredacted one.
+func Redact(ctx context.Context, evidence []byte) ([]byte, *redaction.Report, error) {
+	scanner, err := redaction.DefaultScanner()
+	if err != nil {
+		return nil, nil, fmt.Errorf("initialising the secret scanner: %w", err)
+	}
+
+	redacted, report, err := redaction.New(scanner, redaction.WithPathFilter(eligible)).Redact(ctx, evidence)
```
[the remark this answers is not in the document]
jiparis: Valid, and fixed — in the shared engine rather than here, which is why this file is unchanged and the thread stayed open.

`redaction.decodeObject` now calls `rejectDuplicateKeys` before decoding, so every caller of `Redact` refuses a document whose objects repeat a key. That covers this call path. The bypass was real: the secret in the overwritten value was never scanned, and because nothing was replaced the original bytes were handed back as clean, so failing closed is the right outcome.

Covered at this layer by `TestRedactRejectsDuplicateKeys`, which asserts `Redact` returns `ErrDuplicateKey` and no output for a session whose transcript repeats a `content` key.

migmartri (approved):
thanks

PR #3383, opened 2026-08-28
fix(policies): evaluate redacted materials against the content that was stored

jiparis:
Secret redaction (#3368) only rewrites the copy of a `CHAINLOOP_AI_CODING_SESSION` that leaves the machine. Policies were still evaluated against the untouched original on disk, having access to any potential secret that could have leaked into the AI session. 

The invariant now implemented is that **the policy engine receives exactly the bytes that were stored**.

## Changes

- Crafters that do not store the artifact verbatim implement a `transformCrafter` seam and report the bytes they stored. `materials.Craft` surfaces them as `CraftResult.EvaluableContent`, and the crafter passes them to both policy verifiers via the new `policies.WithMaterialContent` per-call option.
- Content resolution fails closed: a material annotated `chainloop.material.redacted=true` returns `ErrRedactedContentRequired` rather than falling back to the file on disk. This is uniform across CAS backends, inline included.
- `chainloop policy devel eval` threads the same content through, so a policy is developed against the input the real run produces.
- The bundled testdata policy now detects `[REDACTED:<rule>]` placeholders and the redaction annotations, since the raw pattern it looked for is no longer present in its input.

Materials that were not transformed are unaffected and hold nothing extra in memory: their content still resolves from the inline copy or the file on disk. This also closes a gap for `skipUpload` materials, whose sanitized bytes were previously retained nowhere at all.

## Notable consequences

A redacted material's policy input can no longer be reconstructed from persisted crafting state alone, so a future push-time or server-side material evaluation must plumb the bytes through as well. `materials.Craft` now returns `*CraftResult` and `policies.Verifier.VerifyMaterial` takes variadic options — both breaking for external importers of these packages.

## Not covered here

Two gaps remain that are properties of redaction rather than this plumbing: `--skip-secret-redaction` stores the session as captured, and secrets counted as `report.Unlocated` (landing in a protected path, or spanning two JSON leaves) are not redacted in the stored copy either. Detecting placeholders in the upstream `ai-coding-session-no-secrets` policy is the second part of the issue and lives in another repository; that policy should keep its pattern matching alongside the placeholder check, since pattern matching remains the only thing that catches the secrets redaction could not reach.

## AI disclosure

Produced with assistance from Claude Code; every commit carries an `Assisted-by: Claude Code` trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
+++ b/pkg/attestation/crafter/api/attestation/v1/crafting_state.go
@@ -145,34 +145,39 @@ func (m *Attestation_Material) GetEvaluableContent(value string, content []byte)
 		return nil, ErrRedactedContentRequired
 	}
 
-	artifact := m.GetArtifact()
-	if artifact == nil && m.GetSbomArtifact() != nil {
-		artifact = m.GetSbomArtifact().GetArtifact()
-	}
+	if len(content) > 0 {
+		// Chosen ahead of the artifact below so that supplied content wins for
+		// every material kind, including the ones that carry no artifact at all
+		// (a container image, a string) and would otherwise resolve to "{}".
+		//
+		// NOTE: ingestMaterialToJSON re-reads the artifact from `value` for the
+		// kinds it projects from a path (JUNIT_XML, HELM_CHART), so supplied
+		// content is still ignored for those. Unreachable today, since only
+		// CHAINLOOP_AI_CODING_SESSION supplies content and it is JSON-native. A
+		// crafter that starts transforming what it stores for a path-projected
+		// kind must make the projection content-based first, or it will fail open.
+		rawMaterial = content
+	} else {
+		artifact := m.GetArtifact()
+		if artifact == nil && m.GetSbomArtifact() != nil {
+			artifact = m.GetSbomArtifact().GetArtifact()
+		}
 
-	if artifact != nil {
-		switch {
-		case len(content) > 0:
-			// NOTE: ingestMaterialToJSON re-reads the artifact from `value` for
-			// the kinds it projects from a path (JUNIT_XML, HELM_CHART), so
-			// supplied content would be silently ignored for those. Unreachable
-			// today, since only CHAINLOOP_AI_CODING_SESSION supplies content and
-			// it is JSON-native. A crafter that starts transforming what it
-			// stores for a path-projected kind must make the projection
-			// content-based first, or it will fail open.
-			rawMaterial = content
-		case m.InlineCas:
-			rawMaterial = artifact.GetContent()
-		case value == "":
-			return nil, errors.New("artifact path required")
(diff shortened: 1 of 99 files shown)


pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -110,41 +111,73 @@ func (m *Attestation_Material) NormalizedOutput() (*NormalizedMaterialOutput, er
...
+		// every material kind, including the ones that carry no artifact at all
+		// (a container image, a string) and would otherwise resolve to "{}".
+		//
+		// NOTE: ingestMaterialToJSON re-reads the artifact from `value` for the
+		// kinds it projects from a path (JUNIT_XML, HELM_CHART), so supplied
+		// content is still ignored for those. Unreachable today, since only
+		// CHAINLOOP_AI_CODING_SESSION supplies content and it is JSON-native. A
+		// crafter that starts transforming what it stores for a path-projected
+		// kind must make the projection content-based first, or it will fail open.
+		rawMaterial = content
```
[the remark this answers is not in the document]
jiparis: String won't provide redacted content.

pkg/attestation/crafter/api/attestation/v1/crafting_state.go
```diff
@@ -110,41 +111,73 @@ func (m *Attestation_Material) NormalizedOutput() (*NormalizedMaterialOutput, er
...
 }
 
-// GetEvaluableContent returns the content to be sent to policy evaluations
-func (m *Attestation_Material) GetEvaluableContent(value string) ([]byte, error) {
+// ErrRedactedContentRequired is returned when policy evaluation is asked to
+// resolve a material whose stored copy was sanitized, without being given that
+// copy. There is no safe source left: the file on disk is the un-redacted
+// original, and for a CAS-backed material the sanitized bytes were streamed away
+// and are not on the material at all.
+var ErrRedactedContentRequired = errors.New(
```
migmartri, line 119: Why do we need this error?

jiparis: Just to validate that a material with the "redacted" annotation should provide redacted content. Currently, the annotation is only added if the content was changed (it's not added in the case that the redaction didn't have any effect, ie no secrets found)

javirln (approved):

PR #3388, opened 2026-08-31
feat(trace): open source the chainloop trace command

jiparis:
Brings `chainloop trace` — AI coding session capture and attestation — into the OSS CLI. Until now it shipped only in the enterprise CLI, even though the `CHAINLOOP_AI_CODING_SESSION` evidence type, its crafter, its schema and its secret redaction have all been here for a while. This adds the producer side.

See the docs already available at: https://docs.chainloop.dev/guides/chainloop-trace#how-to-trace-ai-coding-sessions

See comment below for the session summary that belongs to this PR and has been attested using a binary built from this branch.

`chainloop trace init` installs git hooks plus agent hooks for Claude Code, Cursor and opencode. From then on, commits are annotated with the sessions that produced them, per-file and per-line AI attribution is recorded under `.git/chainloop-trace/`, and `git push` attests the session as `aicodingsession` evidence. `chainloop trace run -- <cmd>` does the same for a single-shot agent invocation and propagates the wrapped command's exit code. `chainloop trace uninstall` removes everything it installed.

This repo has been dogfooding the command through the enterprise CLI already — see `.chainloop.yml`, `.claude/settings.json` and `devel/sandbox-kit/`.

## Layout

| Path | Contents |
|---|---|
| `app/cli/internal/trace/**` | transcript parsing per agent, line attribution, git state, hook installation, `.chainloop.yml` handling |
| `app/cli/pkg/action/trace_*.go` | `TraceRun`, `RunTracePush`, the agent hook handlers, `AttestationExecutor`, `SubprocessExitError` |
| `app/cli/cmd/trace*.go` | the command tree, plus the hook logger |

Commits are split so the bulk lands as a reviewable move: the internals arrive unchanged apart from the import prefix, then the action layer, then the command layer and wire-up, then license headers and generated docs.

## Notes for review

**`AttestationExecutor` takes its dependencies as parameters.** The enterprise copy read `cmd.ActionOpts` and `cmd.Version` directly, which is an import cycle here since `cmd` imports `pkg/action`. Both are now passed in by the caller, with `ActionOpts`/`CLIVersion` fields on `TraceRunOpts` and `RunTracePushOpts`. Note `cliVersion` must be the bare `cmd.Version` and not `ActionsOpts.CLIVersion` — the latter is `fullVersion()`, which appends the edition and would end up in the attestation predicate.

**The hook logger lives in `cmd`, not `pkg/action`.** Hook commands swap the root logger for a colorless, level-filtered one that also writes to the trace state's `log.txt`, because their output interleaves with git's during a commit or push. `cmd` already owns that logger var, so `trace_hooklog.go` operates on it and adds only a `cmd.Logger()` getter. Putting a mutable exported `Logger` in `pkg/action` would have added a third logger source to a package other projects import.

**`main` gained a `SubprocessExitError` case, first in the switch.** Without it `trace run` always exits 1 regardless of what the wrapped command did.

**Two test adjustments worth a look.** `TestGoGitClient_GeneratedMatcherProjectAttributes` runs the generated-file matcher against the real repository `.gitattributes`, so its expectations were retargeted at this repo's paths — taken from `git check-attr linguist-generated`, and still covering the "last match wins" override (`ent/**` generated, then `ent/migrate/**` and `ent/schema/*` not). And `initGitRepo` now resolves symlinks on its temp dir, without which three attribution tests fail on macOS where `t.TempDir()` returns `/var/folders/...` but the repo root resolves to `/private/var/folders/...`.

**Dependencies:** no new modules. `go mod tidy` promotes `go-git/go-billy/v6` and `sergi/go-diff` from indirect to direct at the versions already in the graph; `go.sum` is unchanged.

## Verification

- `make -C app/cli test` green; `golangci-lint run ./app/cli/...` reports nothing in any added file
- `go mod tidy` and the CLI reference regeneration are both idempotent
- `go vet -tags integration ./app/cli/internal/trace/opencode/` — CI never compiles that file
- Against a built binary in a scratch repo: `trace init` writes executable hooks, a commit fires them, `trace hook claude session-start` runs without contacting the control plane and logs to `log.txt` rather than stderr, `--debug` lowers the stderr floor, `trace run -- sh -c 'exit 42'` exits 42, and `trace uninstall` leaves no residue
- `trace hook` is hidden, so it stays out of `--help` and out of `cli-reference.mdx`

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


migmartri (approved):
🎉

PR #3398, opened 2026-09-04
fix(cli): fail attestation verify when the bundle was never verified

### LINKED ISSUE #3397: attestation verify exits 0 when the bundle carries no verification material
## Problem

`chainloop attestation verify --bundle <file>` prints `WRN attestation couldn't be verified` and exits with code `0` when the attestation could not actually be verified, specifically when:

- the bundle carries no verification material (no signing certificate), or
- the control plane has no trusted root configured (`GetTrustedRoot` returns `Unimplemented`).

CI steps that gate on the exit code therefore treat a bundle that was never verified as good.

When the bundle does carry verification material and verification fails, the command already errors out with a non-zero exit code. The gap is specifically the "nothing to verify against" path.

## Where

- `app/cli/pkg/action/attestation_verify.go`: `verifier.ErrMissingVerificationMaterial` is mapped to `false, nil`, and a missing trusted root also returns `false, nil`.
- `app/cli/cmd/attestation_verify.go`: on a `false` result the command logs a warning and returns `nil`.

## Expected behaviour

`attestation verify` should exit non-zero whenever it cannot confirm that the attestation was verified, and the error message should distinguish "was never verified" from "failed verification".

## Note

The verification helper is shared with `workflow run describe`, which legitimately reports an attestation as not verified without failing the command (it can fall back to key-based envelope verification). The strict, fail-closed behaviour should apply to the `attestation verify` command only.

jiparis:
Fixes #3397

`chainloop attestation verify` exited with code 0 while warning that the attestation couldn't be verified, so CI steps that gate on the exit code accepted bundles that were never verified. This happened when the bundle carries no verification material, and when the control plane has no trusted root configured.

The verify command now fails in both cases, with an error that distinguishes an attestation that was never verified from one that failed verification.

The strict and best-effort verification paths are separated, so `workflow run describe` keeps its current behaviour: an attestation that cannot be verified is reported as not verified without failing the command, and can still fall back to key-based envelope verification.

AI disclosure: this change was produced with the assistance of Claude Code.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


migmartri (approved):
Thanks

PR #3415, opened 2026-09-08
feat(cli): create the trace workflow on init, attached to the AI coding session contract

### LINKED ISSUE #3416: trace does not attach the AI coding session contract to its workflow
`chainloop trace` never binds its workflow to a contract.

`trace init` only installs the git hooks. The workflow is created later, implicitly, by the pre-push hook's `attestation init`, which always gives it an empty default contract — so traced AI coding sessions are never governed by a contract, and nothing tells the user how to change that.

Expected: the trace workflow is created up front and attached to the `chainloop-ai-coding-session` contract when the organization has it, falling back to the default contract when it does not.

migmartri:
## What

`chainloop trace init` and `chainloop trace run` now create the Chainloop workflow the trace attestations target, attached to the `chainloop-ai-coding-session` workflow contract when the organization has it, and to the control plane's default contract otherwise.

Until now the workflow was created implicitly by the pre-push hook, always with an empty per-project contract and with nobody reading the output, so traced AI coding sessions were never governed by a contract.

- New `--contract` flag on `trace init` and `trace run` overrides the contract name. It is used only when the workflow is created and is not persisted to `.chainloop.yml`, so the push path stays unchanged.
- A contract that does not exist is not created: the workflow falls back to the control plane's default contract. A contract named explicitly with `--contract` must exist.
- Existing workflows are never modified, and re-running `trace init` is a no-op for the workflow.
- `trace init` fails before writing hooks or `.chainloop.yml` when the workflow cannot be created. `trace run` only warns and runs the session anyway, since the attestation still creates the workflow.
- Updating `.chainloop.yml` now preserves the user's comments, key order and indentation instead of re-marshalling the file from a map.

No control-plane or API change. One integration test case pins the behaviour trace relies on: an organization-level contract can be attached while the project is still being created.

Closes #3416

## AI assistance

This change was produced with AI assistance (Claude Code); every commit carries an `Assisted-by: Claude Code` trailer.


jiparis (dismissed):
LGTM, thanks!

jiparis (approved):

PR #3421, opened 2026-09-09
feat(cli): point telemetry at the EU PostHog project

javirln:
Points the CLI telemetry at the EU PostHog project through `t.chainloop.dev`, and fixes two problems in the events themselves.

## Telemetry destination

The endpoint and key now target the EU project. Both stay in-source defaults so builds outside the release pipeline (Homebrew, `go install`) keep reporting, and GoReleaser consumes `POSTHOG_API_KEY` / `POSTHOG_ENDPOINT` when they are set, falling back to those defaults otherwise. The release workflow already exported both secrets but nothing read them, so the endpoint was effectively a constant; it is now a release-time input.

The endpoint is compiled in and has no runtime override, so binaries released before this change keep reporting to `crb.chainloop.dev`. That host stays frozen on the previous project rather than being repointed, and the sandbox egress allowlist keeps both hosts so older tags still work in a sandbox.

## Group assignment

Setting the `organization` group replaced the groups map instead of adding to it, so `cp_installation` was dropped from every event that carried an organization. Since only API tokens carry an org id, this split all traffic in two: API-token events had `organization` alone, everything else had `cp_installation` alone, and no event ever carried both. Installation-level aggregation therefore never saw CI traffic, and installations used exclusively from CI with an API token were absent from it entirely.

Both groups now travel together on a single event. The underlying values were always present as event properties, so this restores the group dimension rather than recovering lost data, and it does not backfill history.

`organization` remains API-token-only, because user and federated tokens do not carry an org id at that point in the command lifecycle. This makes the index assignment stable and both groups present when known; giving user-token events an organization group needs the org UUID on the telemetry path and is left as a follow-up.

## Alias volume

An alias between the machine and the account was enqueued on every authenticated command, making `$create_alias` roughly a third of the project's total event volume. It is now limited to interactive user sessions outside CI, which is the only case where stitching a pre-login machine identity onto an account is meaningful. API tokens belong to an organization rather than a person, and federated tokens resolve to the issuer URL, which is identical across installations, so aliasing those merged unrelated CI runners into a single person.

Refs PFM-7213.

---

AI assistance: this change was produced with Claude Code. Affected commits carry an `Assisted-by:` trailer, per AI_POLICY.md.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3421?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/devel/sandbox-kit/spec.yaml
+++ b/devel/sandbox-kit/spec.yaml
@@ -368,9 +368,10 @@ permissions:                                 # v1: `network.allowedDomains:`
       - "crb.chainloop.dev:443"              # CLI usage telemetry (PostHog) - binaries released
                                              # before the endpoint moved. The host is compiled
                                              # in, so an older tag checked out in a sandbox
-                                             # still reaches for it, and a blocked telemetry
-                                             # egress stalls every command for the request
-                                             # deadline instead of failing cleanly.
+                                             # still reaches for it. Commands succeed either
+                                             # way, but a connection that is dropped rather
+                                             # than refused costs each tracked command a
+                                             # couple of seconds waiting on the send deadline.
       - "timestamp.digicert.com:80"          # RFC 3161 timestamp authority (plain HTTP, signed response)
                                              # used during attestation signing; host comes from the
                                              # control plane's signing options
--- a/devel/sandbox-kit/spec.yaml
+++ b/devel/sandbox-kit/spec.yaml
@@ -364,7 +364,13 @@ permissions:                                 # v1: `network.allowedDomains:`
       - "api.cp.chainloop.dev:443"           # control plane - ALWAYS (attest + keyless signing CSR)
       - "api.cas.chainloop.dev:443"          # CAS - only if the org CAS backend is external
       - "api.app.chainloop.dev:443"          # platform backend - the EE CLI also dials this
-      - "crb.chainloop.dev:443"              # CLI usage telemetry (PostHog)
+      - "t.chainloop.dev:443"                # CLI usage telemetry (PostHog) - current builds
+      - "crb.chainloop.dev:443"              # CLI usage telemetry (PostHog) - binaries released
+                                             # before the endpoint moved. The host is compiled
+                                             # in, so an older tag checked out in a sandbox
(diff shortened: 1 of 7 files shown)


devel/sandbox-kit/spec.yaml
```diff
@@ -364,7 +364,13 @@ permissions:                                 # v1: `network.allowedDomains:`
       - "api.cp.chainloop.dev:443"           # control plane - ALWAYS (attest + keyless signing CSR)
       - "api.cas.chainloop.dev:443"          # CAS - only if the org CAS backend is external
       - "api.app.chainloop.dev:443"          # platform backend - the EE CLI also dials this
-      - "crb.chainloop.dev:443"              # CLI usage telemetry (PostHog)
+      - "t.chainloop.dev:443"                # CLI usage telemetry (PostHog) - current builds
+      - "crb.chainloop.dev:443"              # CLI usage telemetry (PostHog) - binaries released
+                                             # before the endpoint moved. The host is compiled
+                                             # in, so an older tag checked out in a sandbox
+                                             # still reaches for it, and a blocked telemetry
```
jiparis: I'm worried about this sentence. This means that commands could fail with no reason in restricted environments

jiparis (dismissed):

devel/sandbox-kit/spec.yaml
migmartri: What do you mean? 

Basically, whoever is using an old sandbox and a new CLI will indeed fail, but I'd say that's not common since this is a new beta feature no?

In the future we should probably pin the CLI too inside the sandbox and link it ot the release process

javirln: It's a false claimed by Claude, I've fixed it. The command is not blocked

jiparis (approved):

PR #3427, opened 2026-09-10
feat(attestation): carry triage/adjudicate data in the AI security context

matiasinsaurralde:
Refs PFM-7229 / PFM-7271

## Summary

Extends the `CHAINLOOP_AI_SECURITY_CONTEXT` material so a single material spans both an un-adjudicated (triage-only) context and a fully adjudicated one, rather than introducing a separate triage-index material kind.

The embedded security-context schema (kept at `security-context-0.1`) gains, as additive optional fields:

- a top-level `survivors` array. The adjudication work queue that Phase-1 triage produces and an adjudicate run drains with a `survivor` definition
- scan_stats counters: survivors_total, adjudicated_commits, triage_input_tokens, triage_output_tokens, and adjudication_complete

A combined-scan context that omits these still validates, and the schema keeps additionalProperties: false throughout. The aisecuritycontext wire struct mirrors the new fields.

### DIFF
--- a/internal/schemavalidators/internal_schemas/aisecuritycontext/ai-security-context-0.1.schema.json
+++ b/internal/schemavalidators/internal_schemas/aisecuritycontext/ai-security-context-0.1.schema.json
@@ -81,6 +81,10 @@
       "items": {
         "$ref": "#/definitions/survivor"
       }
+    },
+    "discarded": {
+      "$ref": "#/definitions/sha_list",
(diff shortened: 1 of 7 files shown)


migmartri (dismissed):
Thanks @matiasinsaurralde I'd like to understand wether since evidenc has value by itself.

migmartri (dismissed):

migmartri:
Thanks @matiasinsaurralde 

So we are going to record all the commits that were triaged as part of the index, not just the ones that failed correct?

migmartri (approved):

Issue #3429, opened 2026-09-10
chainloop trace: print the session link after a successful push


migmartri:
## Summary

After `chainloop trace` attests an AI coding session, the CLI prints `push completed` and nothing else. There is no pointer to the recorded session, so the user has to open the dashboard, find the organization, open the sessions list, and match the row by hand.

`chainloop trace` should print a direct link to each attested session once the push succeeds.

## Proposed behaviour

Both trace push paths, the pre-push git hook and `chainloop trace run`, end in the same shared push routine. After a successful push it should log one line per attested session:

```
AI coding session attested  session=<agent session id>  url=<uiDashboardUrl>/u/<org>/sessions/<agent session id>
```

The link is printed only when the control plane has `ui_dashboard_url` configured. This is the same gate `chainloop attestation push` already applies to the workflow-run link it renders: the value arrives in the attestation `Init` response and is carried in the crafting state, so no additional request is made from inside a git hook. When no dashboard is configured, the line still confirms which sessions were attested, without a URL.

The identifier in the URL is the agent's own session ID, since that is the only identifier the CLI holds at push time. The dashboard resolves a session by that identifier.

## Notes

- The organization in the link comes from the attestation result rather than local configuration, which may be empty when the CLI falls back to the user's default organization.
- A push can attest several sessions, so the output is one line per session.
- Only the session link is in scope. The workflow-run link stays as it is today.

🤖 _Posted by Maximus bot (Claude Code) on behalf of @migmartri_

PR #3434, opened 2026-09-11
feat(cli): surface trace messages to the user in opencode

migmartri:
Extends the user-facing trace messages to opencode. Until now only Claude Code received them: the session-start banner naming where evidence is going, and the link to the attested session after a push.

opencode defines no hook-response protocol of its own, so the hook writes a small JSON contract to stdout and the generated opencode plugin reads it, putting the message on two channels — a TUI toast, which reaches the user directly, and the shell tool's output, which reaches the model so it can repeat the link in its reply.

Two adjustments to the shared layer came with it. The blank lines that frame the banner in Claude Code's transcript move from the shared session-start handler into the Claude provider, since a toast supplies its own frame and any provider now receives the banner unadorned. The instruction that asks the model to repeat a message verbatim is now a single shared constant rather than a copy per provider. The banner's dashboard lookup also runs alongside session tracking instead of after it, so the two waits no longer add up at session start.

Follow-up to #3429 / #3430, which introduced the session link for Claude Code.

This pull request was produced with AI assistance (Claude Code); the commit carries an `Assisted-by` trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3434?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/cli/internal/trace/claude/provider.go
+++ b/app/cli/internal/trace/claude/provider.go
@@ -152,7 +152,13 @@ func (p *Provider) SupportsSystemMessage() bool {
 	return true
 }
 
-// SystemMessage writes a message to stdout for Claude Code to display on session start.
+// SystemMessage writes a message to stdout for Claude Code to display on
+// session start.
+//
+// The blank lines around it are this client's presentation, not the message's:
+// Claude Code prints a systemMessage flush against the surrounding transcript,
+// so without them the banner reads as part of whatever came before. Providers
+// that frame the message themselves add nothing.
 func (p *Provider) SystemMessage(msg string) error {
 	if msg == "" {
 		return nil
@@ -160,7 +166,7 @@ func (p *Provider) SystemMessage(msg string) error {
 
 	resp := struct {
 		SystemMessage string `json:"systemMessage"`
-	}{SystemMessage: msg}
+	}{SystemMessage: "\n\n" + msg + "\n"}
 
 	return json.NewEncoder(os.Stdout).Encode(resp)
 }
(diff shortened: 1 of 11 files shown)


app/cli/internal/trace/claude/provider.go
```diff
@@ -152,15 +152,21 @@ func (p *Provider) SupportsSystemMessage() bool {
...
+// that frame the message themselves add nothing.
 func (p *Provider) SystemMessage(msg string) error {
 	if msg == "" {
 		return nil
 	}
 
 	resp := struct {
 		SystemMessage string `json:"systemMessage"`
-	}{SystemMessage: msg}
+	}{SystemMessage: "\n\n" + msg + "\n"}
```
jiparis, line 169: Is this on purpose? Claude welcome message already looks good with two space lines after the message:
```
SessionStart:startup says:

Chainloop Trace is recording this session.
Evidence will be sent to https://app.chainloop.dev (organization: chainloop, project: chainloop-platform)


```
Also, the session URL message looks good already:
```
Verified: 15 setup tests pass, types, lint, format clean.

Coding Session Available at https://app.chainloop.dev/u/chainloop/sessions/9e55c69d-3005-4b64-b487-397c9fd57372

✻ Brewed for 1m 37s · done 6:05 PM
```

PR #3439, opened 2026-09-14
fix(trace): stop crediting finished sessions on unrelated commits

jiparis:
The `Chainloop-Trace-Sessions` trailer was chosen by intersecting the staged files with every `ai-lines` ledger on disk, with no scoping in time, branch or session liveness. Those ledgers never expire — `WipeTraceDir` preserves them and `GCOrphans` only drops one when no surviving commit references it — so a session pinned itself into the repository after its first commit and kept being stamped onto every later commit that touched any file it had ever edited.

The consequences compounded downstream: a session's evidence file set comes from the diffs of the commits credited to it, so a foreign commit contributed its whole diff, and every file the session had no ranges for was reported as human-written. A single session accumulated commits from unrelated branches and surfaced on pull requests it never touched.

## What changed

Attribution is now consumed when it lands:

- Each `ai-lines` entry carries a `recorded_at`, and the post-commit hook appends a `consumed_by` marker for the files the commit contained.
- Matching runs against the pending set instead of the full range history. Ranges are retained for push-time enrichment, and per-file consumption keeps credit intact for a session committing its work in chunks.
- Entries written before `recorded_at` existed are never pending, so ledgers already on disk stop matching immediately.

Two supporting changes:

- `session-end` persists `active: false` and `session-start` revives it on resume, which lets a push with no commits to go on skip sessions that have ended. `active` was previously written once and never read.
- `GCOrphans` retires sessions older than a week once everything they produced has been attested, since reachability alone never expires them.

`active` remains unsuitable as an attribution signal — `session-end` is not installed for `trace run` and never fires on SIGKILL or a closed terminal.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jiparis:
Thanks — four of the five were real. Addressed in 79a3764.

**Fixed**

- **Session revived from any hook** (`trace_agent_hook.go`). Correct: `ensureSessionTracked` is also called from pre- and post-tool-use, so a late hook resurrected a genuinely ended session. Revival moved to `HandleAgentSessionStart` only, with a test asserting a post-tool-use hook leaves `Active` false.
- **`MarkConsumed` race** (`ailines.go`). Correct, and the sharper failure is that the session loses credit for work still in the working tree. Rather than lock the ledger, each marker now names the edit it retires instead of the time it was written, and replay only clears the file when the pending time is not after the marker's. A concurrent edit's later timestamp keeps it pending. The pending snapshot is an explicit parameter now, since it is the boundary this has to be correct across. Test verified to fail without the change.
- **Clock-dependent ordering tests** (`state_test.go`). Correct — those would tie on a coarse-clock host. Added `RecordLineRangesAt` so the tests set the times, and used the same seam to make the two-session lifecycle test deterministic; it was racing the wall clock too.
- **Legacy entry after a valid one** (`ailines.go`). Folded into the marker change above: the parse-failure path is now only about markers, and an unreadable marker retires the file outright.

**Not changed — deliberate, reasoning now recorded in `staleSessionIDs`**

The GC not sparing a week-old session's uncommitted attribution is the intended trade-off, not an oversight. Keeping it means a session nothing has heard from in a week can still win ownership of a file and pull the whole commit's diff into its evidence — the exact bug this PR exists to fix. By that point its transcript is usually gone (`raw/` is wiped on every push), so `ParseSession` fails and the attestation it wins yields no evidence anyway, leaving a commit record pointing at nothing. Attributing that edit to nobody under-credits the AI; the alternative credits the wrong session. Note also that a session with no commits at all was already dropped by the pre-existing reachability rule, independent of this change.

jiparis:
Good catch, and valid — fixed in 5f2d38dc.

You were right that a timestamp cannot bound the marker: a racing edit landing in the same clock tick compared equal, so the marker swept it up and the session lost credit for work still in the working tree. That is the same coarse-clock tie that forced the ordering tests off the wall clock, so keying retirement to a time was never going to hold.

Rather than add a separate ordering token, the bound is now the entry's position in the ledger. `PendingEdit` carries `{At, Seq}`, and a marker records the `consumed_seq` it retires; replay clears the file only when the pending position is at or below it. The ledger is append-only, so positions are unique and strictly increasing whatever the clock does, and an edit appended after the snapshot sits past the bound and stays pending. A marker with no position retires the file outright, matching the markers written before the field existed.

Ownership between sessions still goes by `At` — positions from different ledgers are not comparable, so only the time can order two sessions' edits to the same file.

Two tests added, both verified to fail against the timestamp-bounded version: a racing edit that shares the snapshot's exact timestamp survives, and a marker with no position retires the file.

matiasinsaurralde (approved):

PR #3443, opened 2026-09-15
Allow hyphens in annotation names

### LINKED ISSUE #3405: Allow hyphens in annotation names
### What

Annotation names currently only accept word characters, so `my_annotation` is valid but `my-annotation` is rejected. Contract names and material names both allow hyphens, so this is an inconsistency rather than a deliberate restriction — users coming from either of those naturally reach for a hyphen and get a validation error.

### Where

`app/controlplane/api/workflowcontract/v1/crafting_schema.proto`:

```proto
message Annotation {
  string name = 1 [(buf.validate.field).string.pattern = "^[\\w]+$"]; // Single word optionally separated with _
  // This value can be set in the contract or provided during the attestation
  string value = 2;
}
```

`^[\w]+$` matches letters, digits and underscore. Hyphen is not in the class.

### Suggested change

Widen the pattern to allow hyphens, keeping it consistent with how other names are validated, and update the trailing comment so it no longer says underscores only.

Things to check while making the change:

- Regenerate the protobuf code after editing the `.proto` (the generated Go must be committed alongside it).
- Annotation names flow into the integrations engine for filtering and interpolation — confirm a hyphenated name survives that path.
- Add a test case covering a hyphenated name, and one confirming genuinely invalid names are still rejected.
- Existing underscore names must keep working; this is purely additive.

### Why it is a good first issue

The change itself is one line in a `.proto`. The value is in doing it properly: regenerating, testing both the new and existing cases, and checking nothing downstream assumed the narrower character set.

wangyusheng1985:
Annotation names were limited to word characters, so values like `my-annotation` were rejected even though contract and material names already allow hyphens.

This widens the Annotation.name validation pattern to accept hyphens while remaining additive for existing underscore and alphanumeric names. Generated protobuf, TypeScript, and JSON Schema artifacts are updated to match, and hyphenated names continue to work through dependency-track interpolation and filter matching.

Assisted-by: Claude Code

Validation observed for this change:
- `gofmt -l app/controlplane/api/workflowcontract/v1/crafting_schema_test.go app/controlplane/plugins/core/dependency-track/v1/extension_test.go`
- `go test ./app/controlplane/api/workflowcontract/v1/ ./app/controlplane/plugins/core/dependency-track/v1/ -count=1 -run 'TestValidateAnnotations|TestAnnotationNameConstraintAllowsHyphen|TestResolveProjectName|TestVerifyAllFilters'`

Fixes #3405

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3443?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


migmartri:
Cam you please sign the commit? Thanks!

wangyusheng1985:
Do you mean a cryptographic commit signature or a DCO Signed-off-by trailer? Please point me to the applicable contribution policy.

migmartri:
> Do you mean a cryptographic commit signature or a DCO Signed-off-by trailer? Please point me to the applicable contribution policy.

Both, https://github.com/chainloop-dev/chainloop/blob/main/CONTRIBUTING.md#commit-format

Thanks

PR #3445, opened 2026-09-16
perf(controlplane): make referrer discovery independent of the caller's organization count

javirln:
Referrer discovery is the query behind the Trust Hub graph. Its cost grew with how many organizations the caller belongs to rather than with the data being asked about.

## The project visibility filter is a constant size

The filter was built by OR-ing one `EXISTS` subquery per organization the caller belongs to, so its shape grew with the caller's membership list. At 23 organizations the planner stops walking the root's edges and evaluates the whole disjunction as a row filter over the referrer table.

It now emits at most two branches for any number of organizations: one organization list matched on the project's own organization column, and one list of `(organization, project)` pairs for individually granted projects. Pairs rather than two independent sets, so a project granted in one organization cannot be reached through another the caller happens to share.

A project-scoped discover goes from 34.3 ms to 6.8 ms, returning identical rows.

## Also here

The referrer visibility filter is expressed as a correlated `EXISTS` rather than through Ent's generated `IN (subquery)`, and the references query no longer asks for a `DISTINCT` it cannot need — the traversal's join table is keyed on `(referrer_id, referred_by_id)` and the projection carries the primary key. Neither is a measured speed-up; both forms plan the same way and run within noise of each other at every node degree measured.

## Tests

The hand-written visibility filter is required to return exactly what the Ent-generated one does, across every state it can meet. Project visibility is covered against the database: a caller granted a project reaches it, a caller restricted to none does not, and a grant naming one organization does not reach a project living in another.

## Compatibility

No schema change, no migration and no new index. No API change, so existing clients — including the CLI — see the same responses.

Assisted-by: Claude Code


migmartri:
Measured this against production data, by reconstructing the SQL ent emits for the project-scoped `DiscoverPrivate` path and running `EXPLAIN (ANALYZE, BUFFERS)` on each shape. Root was a `GIT_HEAD_COMMIT` with 2 references, caller in 37 orgs.

**Nothing needs to change in the code.** Both changes are correct, and they compound:

| Shape | Execution time |
| --- | ---: |
| `main` today | 3,610 ms |
| + correlated `EXISTS` only | 594 ms |
| + both changes | 178 ms |

One correction to the description, though:

> Neither is a measured speed-up; both forms plan the same way and run within noise of each other at every node degree measured.

Against `main`, the correlated `EXISTS` is a **6.1× speed-up on its own**. On current code the uncorrelated `IN (subquery)` is 3,585 ms of the 3,610 ms: Postgres materialises every referrer visible to the caller — seq scan of 1.79M `referrer_workflows` rows, then an external merge sort of 1.76M rows spilling 34 MB to disk — in order to filter 2 candidate rows. The correlated form turns that into 2 index probes on `referrer_workflows_pkey`, 0.09 ms.

I suspect the benchmark missed it because it swept node degree. This cost is independent of node degree — it scales with the size of the caller's visible-referrer set and the row count of `referrer_workflows`.

The constant-size project filter is doing real work too. With only the `EXISTS` fix applied, the planner stops walking the root's edges and seq-scans `referrers` applying the disjunction as a row filter (556 ms) — precisely the failure mode the description predicts. Collapsing the disjunction is what removes it. So the two changes are complementary, and which one looks dominant depends on which is already in place.

One caveat on the 178 ms: my reconstruction inlines the root referrer id as a scalar subquery, which introduces an InitPlan and pushes the planner into a parallel seq scan of `referrer_references` — 157 of those 178 ms. The real code binds the id directly, so the true figure is lower; everything else in that plan sums to about 3 ms.

Unrelated to performance, and only a suggestion: `referrerVisibleInProjects` now enforces the RBAC project rule in SQL, while `isReferrerVisible` still enforces it in Go. Two implementations of one rule that must not drift — might be worth a test asserting they agree, rather than each being verified separately.

migmartri (approved):

migmartri:
## Security review — no RBAC or authorization regression

Reviewed the head (`957265cf`) against `main`, focusing on the authorization paths this PR rewrites. The final head touches only `app/controlplane/pkg/data/referrer.go` plus tests — the `DiscoverEdges` RPC added mid-PR (with the project-scoping issue cubic flagged: a project-scoped caller omitting `project_name` seeing cross-project edges) was reverted, so no new externally reachable surface ships. Each substitution is semantics-preserving:

### 1. `referrerVisibleToOrgs` (correlated EXISTS) vs `referrer.HasWorkflowsWith(...)` — equivalent

- Same conditions: `workflows.deleted_at IS NULL`, org matched via `IN`. The old form joined the `organizations` table for ID equality; the new one matches the FK column `workflows.organization_id` — identical under the FK, and neither form filters soft-deleted orgs, so no widening.
- **Empty org list**: ent v0.14.6 renders `sql.In(col)` with zero args as `FALSE` (`dialect/sql/builder.go:1054`) — deny, same as the generated form. Deny-by-default is preserved.
- The strongest guard in the PR is here: `referrer_visibility_integration_test.go` runs the hand-written filter and the Ent-generated one against the **same database fixtures** across every state they can meet (no workflow, deleted-only, mixed-orgs, duplicate org IDs, `nil` orgs, traversal shape under `Unique(false)`) and requires identical ID sets, with expectations written out so a change in either form must be justified here.

### 2. `projectVisibilityPredicate` (constant two-branch form) vs the per-org disjunction — equivalent, and one real trap avoided

- The per-org semantics map exactly: no map entry → org column match; entry with empty slice → contributes nothing; entry with grants → `org == orgID AND project.id IN (grants)`. The `(organization_id, id) IN ((org, proj)…)` pair form is the same conjunction. A `nil` predicate still becomes `sql.False()` upstream — nothing visible — deny, unchanged.
- The trap: the obvious "constant-shape" rewrite (two independent `IN` sets) **would** have been a genuine cross-org grant escalation — a project granted in one restricted org becoming reachable through another. The pair form prevents exactly that, and it is pinned three ways: `TestProjectVisibilityPredicateDoesNotCrossOrgs` (pair shape, argument pairing, absence of a bare `"projects"."id" IN (`), `TestProjectVisibilityPredicateSkipsOrgsWithoutGrants`, and the DB-backed biz test where a grant recorded under org2 for a project living in org1 is rejected.

### 3. `Unique(false)` (DISTINCT removal) — not an authz control, invariant holds

`referrer_references` has composite PK `(referrer_id, referred_by_id)` (`ent/migrate/schema.go:915`), so a fixed root matches each reference at most once, and every predicate on the query is a semi-join or scalar — join-free, no row multiplication. DISTINCT never filtered access, only multiplicity. The invariant is documented in-code with a warning that future predicates must stay join-free, and tested ("returned once, without DISTINCT"). Pagination cursor unaffected.

### Injection surface

None added. `projectGrantsPredicate` binds UUIDs via the builder's `Args` (parameterized); project name, version, and digest go through ent predicates. The hand-written builder only writes identifiers and the literal ` IN ` keyword — no user input is interpolated.

### Non-blocking observations

- **The reverted endpoint must not re-land as-was.** If `DiscoverEdges`/`EdgesAmong` returns, the project-scoping rule must apply to every edge endpoint *unconditionally*, not only when the caller supplies `project_name`. Worth a note wherever the reverted work is tracked so those commits aren't cherry-picked back.
- **Pre-existing, unchanged here**: `isReferrerVisible` hydrates via `WithWorkflows()` without a `deleted_at` filter, so its Go-side org/project sets can include soft-deleted workflows while the SQL filter excludes them. SQL is the stricter one, so no exposure — but it is the drift risk behind my earlier suggestion that a test assert the two implementations agree.
- Chainloop SAST/secrets/OWASP checks pass; only controlplane data-layer files changed, so the vulnerability, IaC, and workflow scans were correctly out of scope.

PR #3450, opened 2026-09-17
fix(trace): restore gRPC server spans and add HTTP server spans

### LINKED ISSUE #3448: OTel: server spans missing for HTTP and (conditionally) gRPC — traces root at middleware spans
## Problem

With OpenTelemetry tracing enabled (`observability.tracing`), traces in the backend (Tempo/Jaeger) root at middleware spans such as `WithCurrentMembershipsMiddleware` or `WithCurrentAPITokenAndOrgMiddleware` instead of at a transport-level server span. This affects **both** HTTP and gRPC entrypoints, so there is no per-request server span carrying the RPC method / HTTP route as the trace root.

### 1. gRPC server span is silently dropped when `max_recv_msg_size` is set

`app/controlplane/internal/server/grpc.go` registers the OTel stats handler via Kratos `grpc.Options`:

```go
var serverOpts = []grpc.ServerOption{
    grpc.Middleware(craftMiddleware(opts)...),
    grpc.UnaryInterceptor(protovalidateMiddleware.UnaryServerInterceptor(opts.Validator)),
    grpc.Options(grpcLib.StatsHandler(otelgrpc.NewServerHandler())),
}
...
if v := opts.ServerConfig.Grpc.GetMaxRecvMsgSize(); v > 0 {
    serverOpts = append(serverOpts, grpc.Options(grpcLib.MaxRecvMsgSize(int(v))))
}
```

But Kratos `grpc.Options` **assigns**, it does not append (`kratos/v2@v2.9.2/transport/grpc/server.go:124`: `s.grpcOpts = opts`). So whenever `server.grpc.max_recv_msg_size` is configured — which the Helm chart exposes as `controlplane.grpcMaxRecvMsgSize` and is the documented knob for allowing larger attestations — the second `grpc.Options(...)` overwrites the first and the `otelgrpc` StatsHandler is never installed. Result: no gRPC server spans at all.

Fix: merge both raw options into a single `grpc.Options(grpcLib.StatsHandler(...), grpcLib.MaxRecvMsgSize(...))` call (Kratos appends `srv.grpcOpts` to its own option list, so one combined call is enough).

### 2. HTTP server has no transport instrumentation at all

`NewHTTPServer` (`app/controlplane/internal/server/http.go`) builds the Kratos `http.Server` with only the application middleware chain — there is no Kratos `tracing.Server()` middleware anywhere in the repo, and `otelhttp` is only an indirect dependency. Requests arriving via REST transcoding therefore produce no server span and no `traceparent` extraction; the first span of the trace is whatever middleware runs first.

Fix: add Kratos `tracing.Server()` as the first middleware on the HTTP server only (adding it to the shared chain would duplicate the gRPC stats-handler span), or wrap the composed handler with `otelhttp.NewHandler`.

## Expected behavior

Every request trace rooted at a server span: `controlplane.v1.<Service>/<Method>` for gRPC and `HTTP <METHOD> <route>` for transcoded HTTP, with middleware/biz spans nested underneath.

## Notes

- grpc-web traffic (`grpcweb.WrapServer`) re-dispatches through `grpc.Server.ServeHTTP`, which does wire the stats handler (grpc-go v1.83.2), so fixing cause 1 covers grpc-web too.
- The sampler is `TraceIDRatioBased` rather than `ParentBased` (`app/controlplane/internal/server/otel.go`), so incoming parent sampling decisions are ignored; worth revisiting separately.

migmartri:
## Problem

With tracing enabled, traces rooted at middleware spans (`WithCurrentMembershipsMiddleware`, `WithCurrentAPITokenAndOrgMiddleware`, ...) instead of transport-level server spans — for both gRPC and HTTP entrypoints. See #3448.

## Root causes

1. **gRPC**: kratos' `grpc.Options` *assigns* `s.grpcOpts` instead of appending (`kratos/v2@v2.9.2/transport/grpc/server.go:124`). `NewGRPCServer` passed `grpc.Options(grpcLib.StatsHandler(otelgrpc.NewServerHandler()))` and then, when `max_recv_msg_size` was configured (Helm `controlplane.grpcMaxRecvMsgSize`), a second `grpc.Options(grpcLib.MaxRecvMsgSize(...))` — silently wiping the stats handler and with it all gRPC server spans.
2. **HTTP**: the Kratos HTTP server had no server-side span instrumentation at all (no `tracing.Server()` middleware, no `otelhttp`), so transcoded REST requests started their trace at the first middleware span, with no `traceparent` extraction.

## Fix

- `grpc.go`: pass all raw grpc-go server options in a single `grpc.Options(...)` call, with a comment explaining the assignment semantics so it doesn't regress.
- `http.go`: prepend kratos `tracing.Server()` as the outermost HTTP middleware. It uses the global tracer provider and extracts W3C tracecontext by default. gRPC is untouched there (already covered by the stats handler, so no double spans).

## Verification

New regression tests in `app/controlplane/internal/server/otel_test.go` spin up real Kratos servers with an in-memory exporter:

- `TestGRPCServerSpanWithMaxRecvMsgSize` — gRPC server span `grpc.health.v1.Health/Check` is produced with `MaxRecvMsgSize` set. During development the split-`grpc.Options` variant was confirmed to produce **zero** spans, reproducing the bug.
- `TestHTTPServerSpanFromTracingMiddleware` — HTTP request produces a server span named after the route (`/ping`).

`go build ./app/controlplane/... ./app/artifact-cas/...`, `go vet`, and `go test ./app/controlplane/internal/server/` all pass. artifact-cas checked: only one `grpc.Options` call, unaffected.

Fixes #3448

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


javirln (dismissed):

migmartri:
Addressed the review feedback in d75c1897:

**P2 — gRPC test now exercises `NewGRPCServer`**: `TestGRPCServerSpanWithMaxRecvMsgSize` builds the server through the real `NewGRPCServer` constructor (minimal `Opts`: server config with `MaxRecvMsgSize` set, validator, logger, stub `AuthService` for the status-service registration) and asserts the `grpc.health.v1.Health/Check` server span is exported. A regression that splits the raw options into a second `grpc.Options` call makes this test fail (verified during development: split calls → zero spans).

**P2 — HTTP test now exercises the production middleware list**: the chain construction was extracted into `httpMiddlewares(opts)`, which `NewHTTPServer` now consumes; `TestHTTPServerSpanFromTracingMiddleware` builds a Kratos HTTP server with that exact chain and asserts the `/ping` server span. Building full `NewHTTPServer` would require a working `AuthService` (OIDC config, login/callback handlers) — the extracted function covers the changed wiring without that coupling.

**P3 — comment accuracy**: the comment on `httpMiddlewares` now states that handlers registered via `httpSrv.Handle` (auth login/callback, Prometheus metrics) and `/openapi.yaml` bypass the middleware chain and still produce no server span. Instrumenting those raw handlers would need `otelhttp` wrapping — left out of this PR's scope.

jiparis (dismissed):

migmartri:
Assessment of the review round, plus a follow-up fix in cc25bad3:

- **P2 (gRPC test path) — fully valid.** The original test hand-wrote the combined options and asserted they work; that tests kratos, not our wiring, and the real regression mode (a second `grpc.Options` call in `NewGRPCServer`) would have slipped through. Fixed by building through the real constructor — cheaper than expected since only `AuthSvc.AuthURLs.Login` is dereferenced at construction, and the stats handler records the span even though the auth middleware rejects the health check.
- **P2 (HTTP test path) — directionally right, impractical as stated.** Full `NewHTTPServer` construction needs a working `AuthService` (OIDC config, login/callback handlers invoked at construction). Compromise: extracted `httpMiddlewares(opts)` (now consumed by `NewHTTPServer`) and test through that chain. Covers dropping/reordering `tracing.Server()`; the remaining gap (rewiring `NewHTTPServer` away from `httpMiddlewares`) is accepted.
- **P3 (comment accuracy) — correct, and now fully fixed rather than documented.** The auth login/callback and Prometheus metrics handlers registered via `httpSrv.Handle` bypass the middleware chain; they are now wrapped with `otelhttp.NewHandler` individually (`go.mod`: `otelhttp` promoted from indirect to direct dependency). The remaining untraced endpoints are `/openapi.yaml` and `/download/{digest}` — kratos Context handlers that can't be wrapped without adapters, and low-value static/redirect endpoints anyway. The comment on `httpMiddlewares` states this precisely.

No false positives in this round; each finding pointed at a real weakness.

javirln (approved):

PR #3454, opened 2026-09-18
feat(cli): record command duration in telemetry

### LINKED ISSUE #3453: feat(cli): record command duration in telemetry without blocking the command
## Problem

The CLI reports a `command_executed` telemetry event per command, but it is sent from the root command's pre-run hook, before the command body runs. That placement is deliberate: the network round trip overlaps the command, so telemetry never delays the user. It also means the event cannot carry how long the command took, so there is no way to see per-command latency, spot regressions in `attestation push`, or compare CI and interactive runs.

Reporting the duration from the same process would put a network round trip (roughly 100 ms on a cold connection) on the exit path of every command, which is exactly what the current design avoids. Writing the duration to disk and shipping it with the next invocation loses the last command run on a machine, which in CI is usually `attestation push`.

## Proposed approach

Emit the event at the end of the command and hand delivery to a detached child process running the same binary with a hidden `telemetry flush` subcommand:

- `cmd.Execute` measures the command, builds a small JSON payload (command path, `duration_ms`, `success`, a coarse `error_kind` such as the gRPC status code name, control plane URL, org name, and the parsed token identity, never the token itself) and spawns `chainloop telemetry flush` detached with the payload on stdin. The parent releases the child and returns immediately.
- The child rebuilds the event with the existing tracker code, derives OS, arch, CI runner and machine ID itself, and delivers it under the existing 2 s flush deadline.
- `duration_ms` is sent as a number so PostHog can aggregate it (p50, p90, p95 per command).
- Everything stays behind the existing `DO_NOT_TRACK` and dev-build gates. The `telemetry flush` command is excluded from telemetry so it does not report itself.

Known limitations: interrupted commands (Ctrl-C, CI timeout) are not reported, and the child is lost when the CLI is PID 1 of a container that exits right after.

migmartri:
## What

The CLI reports a `command_executed` telemetry event per command, but it was sent from the root command's pre-run hook, before the command body ran. That placement was deliberate, since the network round trip overlapped the command and so never delayed the user, but it meant the event could not carry how long the command took.

The event now goes out when the command finishes and carries:

- `duration_ms` as a number, so averages and percentiles are expressible in an insight
- `success` as a boolean
- `error_kind` on failure, holding the gRPC status code name or `other`, never the error message

To keep the command from ever waiting on the network, delivery moves to a detached child process running the same binary as a hidden `chainloop telemetry flush`, receiving the payload on an anonymous pipe. The parent releases the child and exits.

`telemetry.Tags` now holds typed values rather than strings, which is what lets the duration arrive as a number.

Shell completion is no longer reported. The shell runs it on every press of the Tab key, so those events were counting keystrokes as commands.

## Notes

The auth token never crosses the process boundary. The child receives only the parsed identity that telemetry has always reported. Everything stays behind the existing `DO_NOT_TRACK` and dev-build gates, and the flush command is excluded from telemetry so it cannot report itself.

Two cases are not reported: commands interrupted before they return, and the case where the CLI is PID 1 of a container that exits immediately afterwards.

The public telemetry documentation lives in the docs repository and needs a follow-up for the new properties.

Closes #3453

AI assistance: this change was produced with Claude Code. Every commit carries an `Assisted-by: Claude Code` trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


migmartri:
The **Chainloop AI Policies** check reports one `ai-config-no-secrets` violation. It is a false positive and there is nothing in the diff to fix.

All five hits are the same line of existing code, `app/cli/cmd/root.go`, which holds the PostHog project key:

```go
posthogAPIKey = "phc_…" // gitleaks:allow
```

That value is already committed, is annotated `gitleaks:allow`, and the comment above it states it is not sensitive and is overridable at build time. It sits a few lines from the telemetry code this PR changes, so reading the file put the literal into the session transcript as tool output. The policy scans the transcript and does not honour the `gitleaks:allow` directive there, only in a repository scan.

No credential was printed by a command, and nothing new was exposed. Flagging for a human to waive rather than amending the branch to chase it.

migmartri:
Thanks. Four of the five are fixed in 448b6088; one does not reproduce.

**Fixed**

- **P1, the child not checking the opt-out.** Correct, and worth closing even though the parent already gates before spawning: the only way to reach it was invoking the hidden command by hand, but `DO_NOT_TRACK` should hold for the binary rather than for one path through it. Verified by piping a payload into `chainloop telemetry flush` with the variable set and confirming nothing reaches the endpoint.
- **P2, the re-exec dropping config context.** Correct, and worse than described. The child reads nothing from the config, but `initConfigFile` still ran, created the config directory, wrote a default file, and called `panic` when the directory could not be created. Reproduced with an unwritable config home: a normal command dies with `panic: creating config file ... permission denied`, while the child now exits 0 and still delivers.
- **P2, unresolved commands reported with an empty name.** Correct. Bare `chainloop` and unknown subcommands both hand back the root command, and both were reported under an empty string. Now skipped.
- **P2, no coverage for the detached spawn.** Fair, and it was the one part with only manual verification behind it. Added tests that the payload arrives on the child's stdin, that the parent returns without waiting, and that a child which cannot start or exits early neither hangs nor fails the command. The no-waiting test was mutation-checked: adding a `Wait()` to the parent makes it fail.

**Not reproduced**

The P2 about the unknown-field test payload being malformed JSON. The payload is a Go raw string literal containing no backslashes, and the case does reach unknown-field validation. Confirmed by mutation: commenting out `DisallowUnknownFields()` makes `TestRunTelemetryFlushRejectsBadPayloads/unknown_field_from_a_newer_parent` fail, which is exactly the regression the case is there to catch. Leaving it as is.

migmartri:
Both correct, fixed in 25095e32. Good catch, and it was the more valuable of the two rounds.

`Version` is `dev` in a test binary, and `isTelemetryDisabled` treats a dev build as telemetry off, so `reportCommand` returned at its first line in both tests. They asserted that no child process was started, which was true for a reason that had nothing to do with the code under test.

The tests now make the binary look like a released one for their duration. I also added the positive case they were missing, asserting that a resolved command with telemetry on really does start a child. Without it, a `reportCommand` that returned early for every input would have satisfied both of the negative tests.

Verified by mutation, which is what I should have done when I wrote them:

| Mutation | Before | After |
|---|---|---|
| Remove the `DO_NOT_TRACK` gate | passed | fails |
| Remove the empty-command filter | passed | fails |

migmartri:
Correcting my earlier note on the `ai-config-no-secrets` failure. I said five hits, all the same line. The final count is nine, and the extra four were self-inflicted while investigating, which is worth recording for whoever waives this.

The breakdown:

- **Five** are the same existing line in `app/cli/cmd/root.go`, the PostHog project key, which is already committed beside a scanner-suppression directive and a comment stating it is not sensitive and is overridable at build time. Any change to CLI telemetry has to read that file.
- **One** came from querying the violation detail itself. That response quotes each offending line, so reading the report re-entered the value into the transcript.
- **Three** came from opening and updating an internal note about this exact class of false positive. The note stored an example module digest and the word patterns the scanner matches.

So the policy is partly reporting on the act of looking at its own output. Nothing new was exposed in any of the nine, and no credential was printed by a command.

Worth weighing against it: the scan of the actual diff is a separate check, and it passed. `secret-scan` / `secrets-detection` is green, as are `sast` and `owasp-top10-2025`. The finding is a property of the session transcript, not of the code being merged.

Still asking for a waiver rather than amending the branch.

jiparis (approved):

javirln (approved):

PR #3455, opened 2026-09-18
feat(sandbox-kit): version and publish the Docker Sandboxes kit (PFM-7373)

migmartri:
Gives the `chainloop-trace-claude` Docker Sandboxes kit a release lifecycle, and narrows it to a single supported mode.

### Publishing

The kit declares a `version:` in its spec that tracks the Chainloop release whose CLI it ships. `bump-chart-and-dagger-version.sh` rewrites it on release alongside the chart's `appVersion`, and merging the resulting bump PR triggers a new `Package Sandbox Kit` workflow that publishes to `docker.io/chainloop/sbx-kit-claude` under both the version tag and `latest`, with a signature and SLSA provenance attached — mirroring how `package_chart.yaml` reacts to changes under `deployment/chainloop/`.

Docker Hub rather than ghcr: sbx's default `kit.allowedSources` is already `["docker.io/"]`, so consumers pull the kit without changing their host settings. Authentication is via a Docker Hub OIDC connection, so no long-lived registry credential is stored in the repository; access is governed by a ruleset matching the subject claim `repo:chainloop-dev/chainloop:ref:refs/heads/main`.

`v1.109.0` and `latest` have been published manually to bootstrap the repository.

### Persistent tracing only

The kit now requires a repository already initialized with `chainloop trace init` and refuses to start otherwise, pointing the user at that command. Identity always comes from the committed `.chainloop.yml`, so the `traceMode`, `chainloopOrg`, `chainloopProject` and `chainloopWorkflow` arguments are gone.

`chainloop trace run` is dropped deliberately: it ignores `.chainloop.yml` by design, and its teardown wipes `.git/chainloop-trace/` and strips the committed hooks — destructive on exactly the repositories this kit accepts. The trade-off is that a session whose work is never pushed attests nothing.

### Layout and docs

The spec moves to `devel/sandbox-kit/claude/` to make room for further kits. A `discover` job derives the publish matrix from the directories holding a `spec.yaml` and the bump script globs the same tree, so a new kit is published with no workflow edit. The trigger is scoped to `devel/sandbox-kit/*/spec.yaml` so a README edit cannot republish a released tag, and a version already present in the registry is skipped rather than overwritten. `sbxenv.yaml` keeps pointing at the in-repo copy so the local environment runs the kit as it exists on the current branch.

The spec gains a usage header pointing at the guide, and sheds its long-form rationale in favour of short notes referring to `devel/sandbox-kit/README.md`, which already carries the same material. The README's `sbx run` invocations are corrected: the kit is `kind: sandbox`, so it is the agent and belongs in the positional slot rather than behind `--kit`, which accepts mixins only.

AI assistance: this change was produced with Claude Code.


matiasinsaurralde (dismissed):

devel/sandbox-kit/claude/spec.yaml
```diff
@@ -0,0 +1,201 @@
+# Claude Code in a Docker Sandbox, traced with Chainloop.
+#
+# GUIDE: https://docs.chainloop.dev/guides/docker-sandboxes  (publishing shortly)
```
jiparis, line 3: remove "publishing shortly", it's already live.

jiparis (approved):

PR #557, opened 2024-03-03
chore(db): update workflow_run revision used

migmartri:
This patch adds a migration that updates `workflowrun.contract_revision_used` to its contract  `revision`. 

Before

```sql
select "contract_revision_used","contract_revision_latest" from workflow_runs;
 contract_revision_used | contract_revision_latest 
------------------------+--------------------------
                      0 |                        2
                      0 |                        2
                      0 |                        2
                      0 |                        1
                      0 |                        2
                      0 |                        2
                      0 |                        2
                      0 |                        2
                      0 |                        2
                      0 |                        2
                      0 |                        1
                      2 |                        2

```

after

```
--migration ran
UPDATE 11
select "contract_revision_used","contract_revision_latest" from workflow_runs;
 contract_revision_used | contract_revision_latest 
------------------------+--------------------------
                      2 |                        2
                      2 |                        2
                      1 |                        2
                      1 |                        1
                      2 |                        2
                      1 |                        2
                      2 |                        2
                      2 |                        2
                      2 |                        2
                      2 |                        2
                      2 |                        2
                      1 |                        1

```

refs https://github.com/chainloop-dev/frontend/issues/206

PR #3456, opened 2026-09-19
docs(controlplane): fix stale schema change instructions

alucarded:
The control plane README tells contributors to create the versioned migration for an Ent schema change with `make migration_new`. That target stopped diffing the Ent schema in ef06a7ca (#557, March 2024), when it became `atlas migrate new` and the diff behaviour moved to the new `migration_sync` target, which `generate` depends on. Following the step as written produces an empty migration file. The sentence has also been truncated mid-way since #220, with an unclosed backtick.

This documents the current procedure: `make -C app/controlplane generate` writes the migration via `migration_sync`, and the root `make generate` does not. It also notes that a data backfill is a separate, hand-written migration, as in #2454.

AI disclosure: this change was produced with AI assistance (Claude Code); the affected commit carries an `Assisted-by` trailer.

<a href="https://cubic.dev/pr/chainloop-dev/chainloop/pull/3456?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


migmartri (approved):

migmartri:
Thanks for your contribution!

